PR #43, opened 2016-11-26
WIP: create the libp2p base class

diasdavid:
>  This makes creating libp2p builds so nice


diasdavid:
Going to merge this one in as a 'first version of the base class'

src/index.js
```diff
@@ -1,19 +1,252 @@
...
+    this.swarm.handle(protocol, handlerFunc, matchFunc)
+  }
+
+  unhandle (protocol) {
+    this.swarm.unhandle(protocol)
+  }
+}
+
+module.exports = {
+  Node: Node
```
dignifiedquire: Why not `module.exports = Node`? simpler to import and there is only one thing exported

PR #48, opened 2016-12-01
feat: add ping to the mix

diasdavid:


diasdavid:
@dignifiedquire @VictorBjelkholm quick thumbs up? :)

dignifiedquire:
what about tests?

diasdavid:
@dignifiedquire all of these get tested in their builds, until we have an `interface-libp2p` (and quite honestly, once we figure out a way to create a batch of tests that is agnostic to a bundle)

PR #64, opened 2017-03-24
Formalising the API

diasdavid:


diasdavid:
Finished polishing and documenting the libp2p API. Feel welcome to post comments and review.

Now:

- [ ] Update the codebase
- [ ] Update libp2p-ipfs-nodejs
- [ ] Update libp2p-ipfs-browser

README.md
```diff
@@ -56,23 +56,144 @@ npm install --save libp2p
...
+
+`callback` is a function with the following `function (err) {}` signature, where `err` is an Error in case starting the node fails.
+
+#### `libp2p.stop(callback)`
+
+> Stop the libp2p Node.
+
+`callback` is a function with the following `function (err) {}` signature, where `err` is an Error in case stopping the node fails.
+
+#### `libp2p.dial(peer [, protocol, callback])`
```
dignifiedquire: I would like to split this into `dial` and `connect`.

- `dial(peer, protocol, callback)` all args required does what this does currently
- `connect(peer, callback)` only ensures a connection is made to the peer, does not return a conn in the callback

README.md
```diff
@@ -56,23 +56,144 @@ npm install --save libp2p
...
+
+---------------------
+
+`SOON™`
+
+#### `libp2p.findPeers`
+
+#### `libp2p.findProviders`
+
+#### `libp2p.record.put`
```
dignifiedquire: I think these should be called `dht.put` and `dht.get` as they allow you to put arbitrary values not only records

README.md
diasdavid: Interesting, I'm thinking about it, not 100% convinced yet, but I can definitely see the potential of having a 'more clean way' for people to migrate from that swarm.connect that go-ipfs kind of standardised.

README.md
diasdavid: go-ipfs only allows (at least per js-ipfs-api tests) to put/get records, provider or IPNS.

It might make sense to just expose it as dht, so that we have a clean upgrade path for when we have the routing and record interfaces well established.

dignifiedquire: Not sure why that is, the underlying method do not have this restriction: https://github.com/libp2p/go-libp2p-kad-dht/blob/master/routing.go#L35

diasdavid: @dignifiedquire because there is no current value for IPFS in having people using the DHT as a generic key-value store.

diasdavid: That being said, don't feel the need to shoehorn this API methods as you develop the DHT, they aren't solidified yet.

README.md
diasdavid: Thinking it twice, I like to keep it simple in libp2p land.

dignifiedquire: Why does this make it simpler? It's just overloading a single method with even more functionality

src/index.js
```diff
@@ -142,149 +137,79 @@ class Node {
...
-  pingById (id, callback) {
-    assert(this.isOnline, OFFLINE_ERROR_MESSAGE)
-    callback(new Error('not implemented yet'))
-  }
-
-  // TODO
-  pingByMultiaddr (maddr, callback) {
-    assert(this.isOnline, OFFLINE_ERROR_MESSAGE)
-    callback(new Error('not implemented yet'))
+  isOn () {
```
dignifiedquire: Hmm why `isOn` and `isOnline` both?

src/index.js
```diff
@@ -142,149 +137,79 @@ class Node {
...
+    return this.isOnline
   }
 
-  pingByPeerInfo (peerInfo, callback) {
-    assert(this.isOnline, OFFLINE_ERROR_MESSAGE)
+  ping (peer, callback) {
+    assert(this.isOn, OFFLINE_ERROR_MESSAGE)
```
dignifiedquire: `isOn` is a method

src/index.js
```diff
@@ -142,149 +137,79 @@ class Node {
...
-  }
-
-  hangUpByPeerInfo (peer, callback) {
+  hangUp (peer, callback) {
     assert(this.isOnline, OFFLINE_ERROR_MESSAGE)
+    const peerInfo = this._getPeerInfo(peer)
 
-    this.peerBook.removeByB58String(peer.id.toB58String())
-    this.swarm.hangUp(peer, callback)
+    this.peerBook.removeByB58String(peerInfo.id.toB58String())
```
dignifiedquire: This is not good, just because we hang up does not mean we want to forget about the peer details

README.md
diasdavid: Simpler as in "start fresh" and not have to take in a legacy 'connect'. I kind of like your idea, just not convinced about the dial and connect keyword: is it intuitive for users?

src/index.js
```diff
@@ -142,149 +137,79 @@ class Node {
...
+      } catch (err) {
+        p = new PeerInfo(PeerId.createFromB58String(peerIdB58Str))
+      }
+      p.multiaddr.add(peer)
+    } else if (PeerId.isPeerId(peer)) {
+      const peerIdB58Str = peer.toB58String()
+      try {
+        p = this.peerBook.getByB58String(peerIdB58Str)
+      } catch (err) {
+        // TODO this is where PeerRouting comes into place
```
dignifiedquire: Actually it doesn't. As far as I understand additions to the peerbook are made directly from e.g. the dht while peers are discovered. If you want to dial a peerid you do not know you make a manual request on the libp2p level to the peer routing mechanism to get multiaddrs. The automatic check should happen at the ipfs level.

README.md
dignifiedquire: Maybe, on the other hand as long as dial creates a connection if it's missing I am not sure this is actually needed currently. Lets leave it as is for now, need to work through more code until  am sure

src/index.js
diasdavid: Here, what would happen would be something like a this.findPeers, to see if we can find the Peer we are looking for. If we then retrieve it from the PeerBook, that is ok.

dignifiedquire: But we already do a lookup in the peerbook, we can not start a dht query here, as this method gets called by the dht to dial, so it could get stuck in a loop

src/index.js
```diff
@@ -142,149 +137,79 @@ class Node {
...
-  // Disconnecting (hangUp) methods
-  //
+  hangUp (peer, callback) {
+    assert(this.isOn(), OFFLINE_ERROR_MESSAGE)
+    const peerInfo = this._getPeerInfo(peer)
 
-  hangUpById (id, callback) {
-    // TODO
-    callback(new Error('not implemented yet'))
+    this.peerBook.removeByB58String(peerInfo.id.toB58String())
```
dignifiedquire: @diasdavid THIS IS GOING TO BREAK THINGS when we have a dht :(

diasdavid: Yeah, and we actually don't need it.

diasdavid: I was trying to remember why this was added here and I just figured out. pre-dht era, I treated the PeerBook as our source of truth for peers we are connected, and so: https://github.com/ipfs/js-ipfs/blob/master/src/core/components/swarm.js#L12-L47 relies on that.

We could do a match with swarm.muxedConns, but a better solution is to actually have a little bit more state on the PeerInfo object I believe. Same goes for the address that gets picked, so that when we do log out the peers we are connected, we only print the addr we are actually using (like go-ipfs does).

dignifiedquire: Maybe we should separate these two like go does, so that we have one storage for all details about peers we have ever seen and one storage where we just store the currently connected peers with the address they are connected with.

diasdavid: swarm.muxedConns is kind of that place where "just store the currently connected peers with the address they are connected with.".

I feel it is more reasonable to keep all that state in just one place.

dignifiedquire: sounds good, lets keep the discussion in #68

PR #66, opened 2017-03-27
feat: implement dialById

dignifiedquire:
Need this in preparation for dht


diasdavid:
see https://github.com/libp2p/js-libp2p/pull/64 instead

diasdavid:
Done in master now :)

PR #93, opened 2017-06-30
node emits "online" event once online

pgte:
To support https://github.com/ipfs/js-ipfs/pull/891

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -187,6 +187,8 @@ class Node extends EventEmitter {
           return each(this.modules.discovery, (d, cb) => d.start(cb), cb)
         }
         cb()
+
+        this.emit('online')
       },
       (cb) => {
         if (this._dht) {


src/index.js
```diff
@@ -187,6 +187,8 @@ class Node extends EventEmitter {
           return each(this.modules.discovery, (d, cb) => d.start(cb), cb)
         }
         cb()
+
+        this.emit('online')
```
diasdavid: I'm not convinced that `online`/`offline` is the right wording here. In js-ipfs we have:

image

start/stop sounds better because `online` has a lot of assumptions today. To me, start/stop say "hey, the thing you set up, is running". 

Thoughts?

pgte: Agree, "online" is somewhat ambiguous, and not the goal. I'd say started and stopped are what are the really important events for the user (at the libp2p and ifs levels).

Following that trail, I think this leads to replacing `.isOnline()`  with `.isStarted()`, no?

diasdavid: @pgte I like that :) (maybe `isRunning`?)

pgte: Perhaps we should use the same verbs, otherwise it may get confusing to users, no?
The other alternative is to be explicit (in the docs) about the state machine (states) and the transitions (events). Example:

starting event -> starting state -> started event -> running state -> stopping event -> stopped state

Makes sense?

diasdavid: 👍🏽

pgte:
As discussed, switched to "start" and "stop" events. Also, replaced `.isOn()` with `.isStarted()`.

PR #101, opened 2017-07-07
Create CODE_OF_CONDUCT.md

diasdavid:


VictorBjelkholm:
Hm, we should not include a code of conduct document in every repository, better to just have the line "This repository falls under the IPFS Code of Conduct." with a link to https://github.com/ipfs/community/blob/master/code-of-conduct.md since otherwise changes will be harder to make.

VictorBjelkholm:
Similar to go-ipfs: https://github.com/ipfs/go-ipfs/#contributing

diasdavid:
Fair point. The same way we handle CONTRIBUTING.md here

PR #116, opened 2017-08-03
WIP - feat: Circuit Relay

dryajov:
Tests are failing because `libp2p-circuit` is not published to npm.


src/index.js
```diff
@@ -53,6 +54,8 @@ class Node extends EventEmitter {
       this.swarm.on('peer-mux-closed', (peerInfo) => {
         this.emit('peer:disconnect', peerInfo)
       })
+
+      this.modules.transport.push(new Circuit(this.swarm, _options.relay))
```
dryajov: This is probably not the best place to instantiate `Circuit`, I would much rather move it to IPFS where the `libp2p` node is set up and instantiate it there, but since it depends on `libp2p-swarm` it would require moving swarm instantiation there too.

diasdavid: This is not what we are looking for. Relay should not be added as a transport (although it uses the same interface). It should be activated in libp2p-swarm.

src/index.js
```diff
@@ -14,6 +14,7 @@ const PeerInfo = require('peer-info')
 const PeerBook = require('peer-book')
 const mafmt = require('mafmt')
 const multiaddr = require('multiaddr')
+const Circuit = require('libp2p-circuit')
```
diasdavid: This should be added in swarm, a feature that gets enabled, just like `connection.reuse`

src/index.js
```diff
@@ -168,7 +171,7 @@ class Node extends EventEmitter {
         this.swarm.transport.add(
           transport.tag || transport.constructor.name, transport)
       } else if (transport.constructor &&
-                 transport.constructor.name === 'WebSockets') {
+        transport.constructor.name === 'WebSockets') {
```
diasdavid: unnecessary style change

src/index.js
```diff
@@ -298,16 +301,24 @@ class Node extends EventEmitter {
...
     if (PeerInfo.isPeerInfo(peer)) {
       p = peer
-    // Multiaddr instance (not string)
+      // Multiaddr instance (not string)
     } else if (multiaddr.isMultiaddr(peer)) {
-      const peerIdB58Str = peer.getPeerId()
+      let peerId
+      if (peer.toString().includes('p2p-circuit')) {
+        // last segment is always the peer ipfs segment
+        peerId = multiaddr(peer.toString().split('p2p-circuit').pop())
```
diasdavid: What's happening here?

test/node.js
```diff
@@ -8,3 +8,4 @@ require('./nodejs-bundle/stream-muxing')
 require('./nodejs-bundle/discovery')
 require('./nodejs-bundle/peer-routing')
 require('./nodejs-bundle/content-routing')
+require('./nodejs-bundle/circuit/circuit')
```
diasdavid: why not just `./nodejs-bundle/circuit`?

test/nodejs-bundle/circuit/dial-over-any-relay.js
```diff
@@ -0,0 +1,99 @@
...
+const waterfall = require('async/waterfall')
+const utils = require('./helpers/utils')
+
+const chai = require('chai')
+chai.use(require('dirty-chai'))
+
+const expect = chai.expect
+
+describe('test connecting over any relay', function () {
+  this.timeout(500000)
```
diasdavid: Does it really need this timeout?

test/nodejs-bundle/circuit/dial-over-any-relay.js
```diff
@@ -0,0 +1,99 @@
...
+
+  let portBase = 9010 // TODO: randomize or mock sockets
```
diasdavid: Let's make sure to get this done before merging

test/nodejs-bundle/circuit/dial-over-any-relay.js
```diff
@@ -0,0 +1,99 @@
...
+      },
+      (err, nodes) => {
+        if (err) {
+          return done(err)
+        }
+
+        testNodes = nodes
+        done()
+      })
+  }
```
diasdavid: Use createNode https://github.com/libp2p/js-libp2p/blob/master/test/nodejs-bundle/utils.js#L12-L34, this is adding a lot of code duplication.

test/nodejs-bundle/circuit/dial-over-any-relay.js
```diff
@@ -0,0 +1,99 @@
...
+  beforeEach(function (done) {
+    setUpNodes(multiplex, () => {
+      let nodeA = testNodes['nodeA']
+      let nodeB = testNodes['nodeB']
+      let relayNode = testNodes['relayNode']
+
+      waterfall([
+        (cb) => nodeA.dial(relayNode.peerInfo, cb),
+        (conn, cb) => nodeB.dial(relayNode.peerInfo, cb)
+      ], () => setTimeout(done, 1000)) // WS needs some time to initialize
```
diasdavid: Why does WS need more time? Sounds like a race condition, ming elaborate ?

test/nodejs-bundle/circuit/dial-over-any-relay.js
```diff
@@ -0,0 +1,99 @@
...
+    utils.dialAndReverse(
+      testNodes.nodeB,
+      testNodes.nodeA,
+      ['hello', 'hello1', 'hello2', 'hello3'],
+      (err, result) => {
+        expect(err).to.be.null()
+        expect(result[0]).to.equal('hello'.split('').reverse('').join(''))
+        expect(result[1]).to.equal('hello1'.split('').reverse('').join(''))
+        expect(result[2]).to.equal('hello2'.split('').reverse('').join(''))
+        expect(result[3]).to.equal('hello3'.split('').reverse('').join(''))
```
diasdavid: What's happening here?

test/nodejs-bundle/circuit/dial-over-specific-relay-transport.js
```diff
@@ -0,0 +1,140 @@
...
+          return done(err)
+        }
+
+        testNodes = nodes
+        relaySpy1 = sinon.spy(testNodes['relayNode1'].swarm.transports[Circuit.name].listeners[0].hopHandler, 'handle')
+        relaySpy2 = sinon.spy(testNodes['relayNode2'].swarm.transports[Circuit.name].listeners[0].hopHandler, 'handle')
+
+        done()
+      })
+  }
```
diasdavid: Same, use https://github.com/libp2p/js-libp2p/blob/master/test/nodejs-bundle/utils.js#L12-L34

test/nodejs-bundle/circuit/dial-over-specific-relay-transport.js
```diff
@@ -0,0 +1,140 @@
...
+      expect(relaySpy1.args.some((a) => {
+        return a[0] &&
+          a[0].dstPeer &&
+          a[0].dstPeer.addrs[0].toString() === `/ipfs/${testNodes['nodeB'].peerInfo.id.toB58String()}`
+      })).to.be.ok()
+
+      expect(result[0]).to.equal('hello'.split('').reverse('').join(''))
+      done(err)
+    })
+  })
```
diasdavid: Also needs some fail test cases.

test/nodejs-bundle/circuit/dial-over-specific-relay.js
```diff
@@ -0,0 +1,205 @@
...
+const chai = require('chai')
+chai.use(require('dirty-chai'))
+
+const expect = chai.expect
+
+const nodeKeys = require('./fixtures/nodes')
+const Circuit = require('libp2p-circuit')
+
+describe(`listen on an explicit relay addr`, function () {
+  this.timeout(500000)
```
diasdavid: This timeout is huge, something is wrong here, why is this required?

test/nodejs-bundle/circuit/dial-over-specific-relay.js
```diff
@@ -0,0 +1,205 @@
...
+        if (err) {
+          return done(err)
+        }
+
+        testNodes = nodes
+        relaySpy1 = sinon.spy(testNodes['relayNode1'].swarm.transports[Circuit.name].listeners[0].hopHandler, 'handle')
+        relaySpy2 = sinon.spy(testNodes['relayNode2'].swarm.transports[Circuit.name].listeners[0].hopHandler, 'handle')
+
+        done()
+      })
```
diasdavid: You have repeated this coded multiple times in every single test, please use the DRY principle and try to reuse what exists already, such as: https://github.com/libp2p/js-libp2p/blob/master/test/nodejs-bundle/utils.js#L12-L34

test/nodejs-bundle/circuit/dial-over-specific-relay.js
```diff
@@ -0,0 +1,205 @@
...
+      ], done)
+    })
+
+    afterEach(function circuitTests (done) {
+      utils.stopNodes(testNodes, done)
+    })
+
+    it('dial over the correct relay', function (done) {
+      utils.dialAndReverse(testNodes['nodeA'], testNodes['nodeB'], ['hello'], (err, result) => {
+        expect(err).to.be.null()
```
diasdavid: switch for `expect(err).to.not.exist()`

test/nodejs-bundle/circuit/dial-over-specific-relay.js
```diff
@@ -0,0 +1,205 @@
...
+
+        expect(relaySpy1.args.some((a) => {
+          return a[0].dstPeer && a[0].dstPeer.addrs[0].toString() === `/ipfs/${testNodes['nodeB'].peerInfo.id.toB58String()}`
+        })).to.be.ok()
```
diasdavid: Make this assertion simpler. We have to always strive to make things understandable, the code in js-libp2p and js-ipfs gets reviewed, used and maintained by a very large community.

test/nodejs-bundle/circuit/dial-over-specific-relay.js
```diff
@@ -0,0 +1,205 @@
...
+
+        expect(result[0]).to.equal('hello'.split('').reverse('').join(''))
```
diasdavid: ??

test/nodejs-bundle/circuit/dial-over-specific-relay.js
```diff
@@ -0,0 +1,205 @@
...
+  //       expect(relaySpy2.args[0][1].toString())
+  //         .to
+  //         .equal((`/ipfs/${testNodes['nodeA'].peerInfo.id.toB58String()}`))
+  //
+  //       expect(result[0]).to.equal('hello'.split('').reverse('').join(''))
+  //
+  //       done(err)
+  //     })
+  //   })
+  // })
```
diasdavid: If not used, delete.

test/nodejs-bundle/circuit/fixtures/nodes.js
```diff
@@ -0,0 +1,25 @@
...
+  pubKey: 'CAASpgIwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQC6pg6LYWbY+49SOYdYap6RPqKqZxg80IXeo3hiTUbiGtTruxVYZpnz3UbernL9J9mwlXJGRUQJUKmXmi1yePTQiyclpH0KyPefaWLbpxJQdCBI1TPZpDWo2hutWSPqhKBU1QyH2FLKQPWLdxuIX1cNFPtIlSl5gCxN6oiDIwh7++kxNM1G+d5XgJX6iHLlLoNv3Wn6XYX+VtYdyZRFk8gYyT2BrISbxmsrSjSOodwUUzF8TYTjsqW6ksL2x0mrRm2cMM9evULktqwU+I8l9ulASDbFWBXUToXaZSL9M+Oq5JvZO0WIjPeYVpAgWladtayhdxg5dBv8aTbDaM5DZvyRAgMBAAE='
+}
```
diasdavid: Is this really needed?

test/nodejs-bundle/circuit/helpers/test-node.js
```diff
@@ -0,0 +1,21 @@
...
+        muxer: [muxer],
+        crypto: options.isCrypto ? [secio] : null
+      },
+      discovery: []
+    }
+    super(modules, peerInfo, null, options)
+  }
+}
+
+module.exports = TestNode
```
diasdavid: Why another bundle if you use it inside the nodejs bundle?

test/nodejs-bundle/circuit/helpers/utils.js
```diff
@@ -0,0 +1,110 @@
...
+    (key, cb) => {
+      nodes[key].stop(cb)
+    },
+    (err) => {
+      if (err) {
+        return callback(err)
+      }
+      callback()
+    })
+}
```
diasdavid: This is again more repeated code. See https://github.com/libp2p/js-libp2p/blob/master/test/nodejs-bundle/utils.js#L12-L34

test/nodejs-bundle/circuit/helpers/utils.js
```diff
@@ -0,0 +1,110 @@
...
+  eachAsync(Object.keys(nodes),
+    (key, cb) => {
+      nodes[key].stop(cb)
+    },
+    (err) => {
+      if (err) {
+        return callback(err)
+      }
+      callback()
+    })
```
diasdavid: This whole callback can be just:

```
eachAsync(Object.keys(nodes), (key, cb) => nodes[key].stop(cb), callback)
```

diasdavid:
Also, from IRC

```
19:14 <daviddias> dryajov: not quite
19:14 <daviddias> I remember discussing how circuit gets integrated in swarm
19:14 <daviddias> and why we didn't need the priority 
19:14 <daviddias> in transports
19:14 <daviddias> right now as it stands, swarm is just trying to burst and open all the connections through all transports and multiaddrs
19:14 <daviddias> including relay
19:15 <daviddias> which should not be the case
19:15 <daviddias> relay should be used as a last resource if any of the direct dials don't work
19:15 <daviddias> that is why relay is enabled as a special option in swarm
```

dryajov:
@diasdavid we can handle `circuit` as a special case, when transports are being selected, i.e. skip circuit and only use it if all other transports failed...

test/nodejs-bundle/circuit/dial-over-any-relay.js
dryajov: Without it, I would get intermittent timeouts without it, specially with WS which seem to take a while for the connection to get established... Definitely something to investigate once we iron everything else out.

test/nodejs-bundle/circuit/dial-over-any-relay.js
dryajov: This is something I still need to investigate, but in general WS seems to need some time to initialize. Definitely something to look into.

test/nodejs-bundle/circuit/dial-over-any-relay.js
dryajov: Just a few quick tests to make sure more than one result gets sent, not the cleanest - will clean up.

test/nodejs-bundle/circuit/dial-over-any-relay.js
dryajov: @diasdavid I think not randomizing is actually fine, this tests are not ran in parallel so this ports shouldn't conflict with anything, but in rare cases. I think removing the TODO makes sense.

test/nodejs-bundle/circuit/helpers/test-node.js
dryajov: The reason I'm _for now_ using a custom node and utilities is because circuit tests are a bit more involved and require controlling which transports are enabled per node. I'll refactor both the utility as well as my tests to allow for more control over how the node is setup.

dryajov:
@diasdavid I think this is ready for another pass - tests are still not perfect but better than what I had before, IMO :)

diasdavid:
@dryajov were all the things we chatted Thursday done? Where is the tracking list?

dryajov:
@diasdavid Still in progress. I'm working on cleaning things up and adding interop tests right now - been going a little slower than I expected, but should wrap up this week. Also, I will update https://github.com/ipfs/js-ipfs/issues/833 with all the latest progress and outstanding issues.

dryajov:
This needs libp2p-circuit and libp2p-swarm released.

test/nodejs-bundle/circuit/dial-over-any-relay.js
diasdavid: Why does this need to be so complicated? https://github.com/libp2p/js-libp2p/blob/master/test/nodejs-bundle/tcp%2Bwebsockets%2Bwebrtc-star.js boots a lot of nodes without requiring so much customization and. code duplication

- Why do you need to set up the transports explicitly? Libp2p only boots transports it has listeners to
- Why do you need to setup muxer and isCrypto?

test/nodejs-bundle/circuit/dial-over-any-relay.js
```diff
@@ -0,0 +1,99 @@
...
+          isCrypto: true,
+          muxer: muxer,
+          addrs: [
+            `/ip4/0.0.0.0/tcp/${portBase++}/ws`
+          ]
+        }
+      },
+      (err, nodes) => {
+        if (err) {
+          return done(err)
```
diasdavid: use `expect`

test/nodejs-bundle/circuit/dial-over-any-relay.js
diasdavid: Have you done any further investigation? Where is the issue to track this?

test/nodejs-bundle/circuit/dial-over-any-relay.js
diasdavid: Where is the clean up happening?

test/nodejs-bundle/circuit/dial-over-specific-relay-transport.js
```diff
@@ -0,0 +1,141 @@
...
+const chai = require('chai')
+chai.use(require('dirty-chai'))
+
+const expect = chai.expect
+
+const nodeKeys = require('./fixtures/nodes')
+const Circuit = require('libp2p-circuit')
+
+describe(`listen on an explicit chained relay addr`, function () {
+  this.timeout(500000)
```
diasdavid: This timeout is not ok. @dryajov please figure out why it needs such a timeout and fix or document why it needs to be here

test/nodejs-bundle/circuit/dial-over-specific-relay-transport.js
```diff
@@ -0,0 +1,141 @@
...
+
+  let portBase = 9030 // TODO: randomize or mock sockets
```
diasdavid: Please randomize

test/nodejs-bundle/circuit/helpers/utils.js
diasdavid: @dryajov please apply CR

test/nodejs-bundle/circuit/helpers/utils.js
```diff
@@ -0,0 +1,110 @@
+'use strict'
```
diasdavid: This file should be part of the utils at the top `test/utils` It is also duplicating code that is already there.

test/nodejs-bundle/utils.js
```diff
@@ -10,6 +10,7 @@ const waterfall = require('async/waterfall')
 const pull = require('pull-stream')
 
 function createNode (multiaddrs, options, callback) {
+  options = options || {}
```
diasdavid: This should come after the if condition

src/index.js
```diff
@@ -43,6 +42,9 @@ class Node extends EventEmitter {
       // If muxer exists, we can use Identify
       this.swarm.connection.reuse()
 
+      // If muxer exists, we can use Relay for listening/dialing
+      this.swarm.connection.enableCircuitRelay(_options.relay)
```
diasdavid: This should be enabled by an option and not by default

package.json
```diff
@@ -1,19 +1,19 @@
 {
   "name": "libp2p",
-  "version": "0.12.4",
+  "version": "0.12.0",
```
diasdavid: why drop the version?

dryajov:
@diasdavid I discovered an issue with circuit while working on getting this tests to pass, here is a PR that adresses them - https://github.com/libp2p/js-libp2p-circuit/pull/17.

src/index.js
dryajov: this is being handled inside the `enableCircuitRelay` - https://github.com/libp2p/js-libp2p-swarm/blob/master/src/connection.js#L100.

dryajov:
@diasdavid this needs the latest circuit master release.

diasdavid:
Already did that, right?

diasdavid:
@dryajov aegir failed me and I published manually

dryajov:
@diasdavid ah.. didn't realize.

dryajov:
I'm updating dependencies, and pushing shortly - it should build with the latest circuit.

dryajov:
Tests are failing in node 8, but I can't see the specific error because of an issue with a karma dep fixed here - https://github.com/ipfs/aegir/pull/164

Issue #117, opened 2017-08-29
libp2p-websocket-star


mkg20001:
I have created a module called [libp2p-websocket-star](https://github.com/mkg20001/js-libp2p-websocket-star) that works similar to libp2p-webrtc-star just without webrtc.

What do you guys think?

mkg20001:
I've now added it to [mkg-tool](https://github.com/mkg20001/mkg-tool) and created a public signaling server at `/libp2p-webrtc-star/ip4/148.251.206.162/tcp/9090/ws/`. It works great

diasdavid:
That's pretty cool, you probably want to add a specific multiaddr for it to not be confused with the webrtc-star transport :)

diasdavid:
This is going to enable a lot of people that want to use js-ipfs P2P like but are having troubles with WebRTC.

@lgierth can we get one of the servers running hosted by us?

lgierth:
What does it does?

Can we just make it part of the regular websockets transport?

mkg20001:
It's a rewritten version of the libp2p-webrtc-star transport that creates the streams over the signaling server (socket.io) directly instead of over webrtc.
I doubt it's a good idea to merge the websocket transport with that.

diasdavid:
I agree with @mkg20001. It has a different mechanism.

lgierth:
I'm all for it -- but let's make sure this multiaddr scheme is correct, @diasdavid you mentioned in the past the webrtc-star scheme needs an overhaul because it doesn't correctly represent encapsulation.

diasdavid:
Yeah, here https://github.com/libp2p/js-libp2p-webrtc-star/issues/110. Let's figure out this tomorrow over coffee ☕️ as proposed :)

mkg20001:
I've written a small solution (currently not on github) to how to make the listening process more secure:
Before listening a ~4kb pseudo random string is generated and a hash is created using `b58encode(sha5(sha5(key)))` that is then added to the address as another ipfs id with `.encapsulate("ipfs/"+hash)`. (This should later be changed to use it's own multiaddr like `/instance/<hash>` or `/id/<hash>`)

For the client to listen on the multiaddr it has to provide the key (the random string). This way there can be multiple instances of the same peer-id online without having conflicts.

[...]

PR #122, opened 2017-09-06
Bring libp2p-websocket-star to the Transports family! 🌟

mkg20001:
Adds tests for libp2p-websocket-star
Ref libp2p/js-libp2p-websocket-star#3


circle.yml
```diff
@@ -6,8 +6,12 @@ dependencies:
   pre:
     - google-chrome --version
     - curl -L -o google-chrome.deb https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+    - for v in $(curl http://archive.ubuntu.com/ubuntu/pool/main/n/nss/ | grep "href=" | grep "libnss3.*deb\"" -o | grep -o "libnss3.*deb" | grep "3.28" | grep "14.04"); do curl -L -o $v http://archive.ubuntu.com/ubuntu/pool/main/n/nss/$v; done && rm libnss3-tools*_i386.deb libnss3-dev*_i386.deb
     - sudo dpkg -i google-chrome.deb || true
+    - sudo dpkg -i libnss3*.deb || true
     - sudo apt-get update
+    - sudo apt-get install -f || true
+    - sudo dpkg -i libnss3*.deb
```
diasdavid: @mkg20001 is this the fix for all the Chrome errors we started seeing in Circle?? Awesome!!!

//cc @VictorBjelkholm

mkg20001: @diasdavid Yes.

diasdavid:
Ok, I've gone through and give a big CR (with refactoring in place) to websocket-star. Overall there is a ton of work there and I'm really happy to see it coming together, thank you @mkg20001 👏🏽👏🏽👏🏽👏🏽!

The CR+Refactor can be found in:
- https://github.com/libp2p/js-libp2p-websocket-star/pull/16
- https://github.com/libp2p/js-libp2p-websocket-star-rendezvous/pull/1

To release and integrate websocket-star with confidence, recommending our users to try it out, I still want to see some issues solved and/or improved. I've opened issues in both repos so that it is clear what is missing, lists can be found in:

- https://github.com/libp2p/js-libp2p-websocket-star-rendezvous/issues
- https://github.com/libp2p/js-libp2p-websocket-star/issues

@mkg20001 is this something you would like to finish? I can continue helping by providing CR and coding time, just not sure how much time I can allocate that soon :) 

Thanks!

mkg20001:
> @mkg20001 is this something you would like to finish? I can continue helping by providing CR and coding time, just not sure how much time I can allocate that soon :)

Yes but I'm also working on https://github.com/ZeroNetJS/zeronet-js/ and my last year in school just began so I won't be working on this project that much

diasdavid:
I didn't know about zeronet-js with libp2p! That's rad! :D

mkg20001:
> Actually, this is more than rad! it is super-duper rad! Wanna show it working today at the IPFS All Hands?

After thinking about that for a while and talking with my parents I decided it would be a good idea. I'm usually a bit shy and this would be the first every meeting/conference/etc where I would participate. But I would like to participate. How can I join? How does that work?

diasdavid:
@dryajov @VictorBjelkholm can I get your review here as well?

package.json
```diff
@@ -60,6 +60,8 @@
     "libp2p-tcp": "~0.11.0",
     "libp2p-webrtc-star": "~0.13.0",
     "libp2p-websockets": "~0.10.1",
+    "libp2p-websocket-star": "0.3.0-rc4",
+    "libp2p-websocket-star-signal": "0.3.0-rc4",
```
VictorBjelkholm: Haven't seen that we are doing RC versions anywhere else, can we release stable versions of these?

mkg20001: Fixed

mkg20001:
Rebased all changes against current master and released stable versions of the libp2p-websocket-star* modules.
Merge?

.aegir.js
```diff
@@ -15,14 +17,28 @@ const before = (done) => {
...
+    port: 14444,
+    refreshPeerListIntervalMS: 1000,
+    strictMultiaddr: false
+  }, (err, _server) => {
+    if (err) {
+      throw err
+    }
+    server2 = _server
+    ready()
+  })
```
diasdavid: @mkg20001 mind putting the two tasks, `sigServer.start` and `wsRendezvous.start` in a [`async.parallel`](http://caolan.github.io/async/docs.html#parallel) flow?

.aegir.js
```diff
@@ -38,12 +54,7 @@ const before = (done) => {
 }
 
 const after = (done) => {
-  setTimeout(() => node.stop((err) => {
-    if (err) {
-      return done(err)
-    }
-    server.stop(done)
-  }), 2000)
+  setTimeout(() => require('async/each')([node, server, server2], (s, n) => s.stop(n), done), 2000)
```
diasdavid: a) Please do not require modules throughout the codebase, always require it at the top.
b) Preserve the `{}` when nesting multiple calls.
c) Use async/parallel for this

package.json
```diff
@@ -62,6 +62,8 @@
     "libp2p-tcp": "~0.11.1",
     "libp2p-webrtc-star": "~0.13.2",
     "libp2p-websockets": "~0.10.4",
+    "libp2p-websocket-star": "~0.5.0",
+    "libp2p-websocket-star-rendezvous": "~0.2.0",
```
diasdavid: Thank you for keeping the `~` for <1.0.0 :)

test/browser-bundle/browser-bundle.js
```diff
@@ -36,11 +37,13 @@ class Node extends libp2p {
   constructor (peerInfo, peerBook, options) {
     options = options || {}
     const webRTCStar = new WebRTCStar()
+    const wsStar = new WebSocketStar({ id: peerInfo.id })
```
diasdavid: Does it still require that the PeerInfo gets passed on the constructor? Why?

test/nodejs-bundle/tcp+websockets+websocket-star.js
```diff
@@ -0,0 +1,247 @@
...
+            expect(Object.keys(nodeAll.swarm.muxedConns)).to.have.length(1)
+            cb()
+          },
+          (cb) => {
+            const peers = nodeWS.peerBook.getAll()
+            expect(Object.keys(peers)).to.have.length(1)
+            expect(Object.keys(nodeWS.swarm.muxedConns)).to.have.length(1)
+            cb()
+          }
+        ], done)
```
diasdavid: All of the checks happening are synchronous, there is no reason to use async.parallel

test/nodejs-bundle/tcp+websockets+websocket-star.js
```diff
@@ -0,0 +1,247 @@
...
+            expect(Object.keys(peers)).to.have.length(2)
+            expect(Object.keys(nodeAll.swarm.muxedConns)).to.have.length(0)
+            cb()
+          },
+          (cb) => {
+            const peers = nodeWS.peerBook.getAll()
+            expect(Object.keys(peers)).to.have.length(1)
+            expect(Object.keys(nodeWS.swarm.muxedConns)).to.have.length(0)
+            cb()
+          }
```
diasdavid: Same from?

test/nodejs-bundle/tcp+websockets+websocket-star.js
```diff
@@ -0,0 +1,247 @@
...
+            expect(Object.keys(nodeAll.swarm.muxedConns)).to.have.length(1)
+            cb()
+          },
+          (cb) => {
+            const peers = nodeWStar.peerBook.getAll()
+            expect(Object.keys(peers)).to.have.length(1)
+            expect(Object.keys(nodeAll.swarm.muxedConns)).to.have.length(1)
+            cb()
+          }
+        ], done)
```
diasdavid: Same here

test/nodejs-bundle/tcp+websockets+websocket-star.js
```diff
@@ -0,0 +1,247 @@
...
+            expect(Object.keys(nodeAll.swarm.muxedConns)).to.have.length(0)
+            cb()
+          },
+          (cb) => {
+            const peers = nodeWStar.peerBook.getAll()
+            expect(Object.keys(peers)).to.have.length(1)
+            expect(Object.keys(nodeWStar.swarm.muxedConns)).to.have.length(0)
+            cb()
+          }
+        ], done)
```
diasdavid: Same here

test/nodejs-bundle/tcp+websockets+websocket-star.js
mkg20001: I copied this from test/nodejs-bundle/tcp+websockets+webrtc-star.js ( https://github.com/libp2p/js-libp2p/blob/master/test/nodejs-bundle/tcp%2Bwebsockets%2Bwebrtc-star.js#L107-L118 )

test/browser-bundle/browser-bundle.js
patrickdet: in the README of ws-star it says: `const ws = new WSStar({ id: id }) // the id is required for the crypto challenge`

patrickdet: this requierment to have the `{ id: peerInfo.id }` makes it impossible to configure the `WebSocketStar` through the IPFS node constructor like:

```
libp2p: { // add custom modules to the libp2p stack of your node
    modules: {}
  }
```

because the peer id is not available at that point

mkg20001: How else can the transport get the id?

diasdavid: The first question should be "why does this Id exist in the first place", if:

- a) is to auth with the rendezvous point, the reality is that the rendezvous point doesn't have any list of pre-shared ids, so using the PeerId or something else at random is the same
- b) is to auth with the other peers, then that is a double crypto challenge because secio already does that.

test/nodejs-bundle/tcp+websockets+websocket-star.js
diasdavid: Wanna improve both?

test/browser-bundle/browser-bundle.js
mkg20001: It is to prevent https://github.com/libp2p/js-libp2p-webrtc-star/issues/116

.aegir.js
mkg20001: They are already parallel...

test/nodejs-bundle/tcp+websockets+websocket-star.js
mkg20001: Sure

test/browser-bundle/browser-bundle.js
diasdavid: @mkg20001 I understand but note that to fix it you don't need the node peer-id, it could be some other id of the client.

mkg20001: @diasdavid I already had implemented that some time ago and you rejected it. https://github.com/libp2p/js-libp2p/issues/117#issuecomment-325929963

diasdavid: There are two things here:

- a) Authentication between peers (end to end) - assured by SECIO ✔️
- b) Authentication with the rendezvous point  - what you are trying to achieve with the PeerId

b) is a bit more complicated because you want to peers to prove who they are with the rendezvous point. 

I'm torn that one the best solution is through PeerId because that really breaks the transport interface expectation. We also want to move away from centralized rendezvous points and just make any node be able to do that, which kind of means that a transport will need access to libp2p itself. Once we end up doing this, it might lead to changing how modules are loaded to a DI scheme.

mkg20001: There are now 4 possible solutions:
 - Use `wsstar.lazySetId()` somewhere later in the code
 - Don't add the peerId and run the server with `--disableCryptoChallenge` (*insecure!*)
 - https://github.com/libp2p/js-libp2p/issues/117#issuecomment-325929963
 - DI scheme as mentioned above (isn't implemented so this isn't a good short-term solution)
 - Something else that works short-term at least

Which one would be the best? (I prefer the first one)

Beanow:
Can you provide a configuration you've tested this with in `js-ipfs`?

In my tests to implement this I found some odd issues with the multi addresses.
Using a test server,
```js
config: {
  Addresses: {
    Swarm: [
      '/dns4/ws-star-signal-2.servep2p.com/wss/p2p-websocket-star/'
    ]
  }
}
```

My logs output:
> Swarm listening on /dns4/ws-star-signal-2.servep2p.com/wss/p2p-websocket-star/ipfs/QmcdoJPn4CStGjJd58XQE3GfLZQHSshMovhoGMNFoHd9px/ipfs/QmcdoJPn4CStGjJd58XQE3GfLZQHSshMovhoGMNFoHd9px

And it does not successfully connect with any peers over websocket star.

When I add this to [`src/core/runtime/libp2p-browser.js`](https://github.com/ipfs/js-ipfs/blob/master/src/core/runtime/libp2p-browser.js)

```js
peerInfo.multiaddrs.clear()
peerInfo.multiaddrs.add(multiaddr('/dns4/ws-star-signal-1.servep2p.com/wss/p2p-websocket-star/'))
const wsstar = new WSStar({id: peerInfo.id})
```

It will do peer discovery, bitswam and pubsub correctly.

test/browser-bundle/browser-bundle.js
diasdavid: @mkg20001 what about deploying a first version without the cryptoChallenge so that js-ipfs users can try it out today and we work on the DI scheme (which @pgte and me will start using more of his time towards that) as a separate endeavor so that then we can add the crypto challenge?

mkg20001:
Try it with https://github.com/mkg20001/js-ipfs/tree/feat-websocketstar or apply [this](https://github.com/ipfs/js-ipfs/compare/master...mkg20001:master.diff) patch and see if it works

test/browser-bundle/browser-bundle.js
mkg20001: @diasdavid The peerId IS accesible at the time of transport creation. No need to use the insecure version. https://github.com/ipfs/js-ipfs/compare/master...mkg20001:master.diff

mkg20001: @diasdavid Insecure server running at `/dns/ws-star-signal-4.servep2p.com/wss/p2p-websocket-star/`

patrickdet:
The only annoying thing is that now it is not possible to configure the `libp2p` transport with the `WebSocketStar` from the configuration like:

```
  const ipfs = new Ipfs({
    libp2p: {
      modules: {
        transport: [new WebSocketStar({id: peerInfo.id})],
        discovery: [new WebSocketStar({id: peerInfo.id})]
      }
    }
  })
```

because in that scenario the `peerInfo` is not available yet. Not sure if it is critical to have that though.

mkg20001:
You can use it without peer id and use this signalling server instead: `/dns/ws-star-signal-4.servep2p.com/wss/p2p-websocket-star/`. note that it's not as secure as the other servers (no crypto challenge)

diasdavid:
Let's focus on figuring out #130 to get that PeerId on the transport. Until then let's move with the _not as secure, also not yet audited version_.

.aegir.js
```diff
@@ -4,9 +4,12 @@ const Node = require('./test/nodejs-bundle/nodejs-bundle.js')
 const PeerInfo = require('peer-info')
 const PeerId = require('peer-id')
 const pull = require('pull-stream')
+const {parallel} = require('async')
```
diasdavid: Please use standard `require('async/parallel')`

.aegir.js
```diff
@@ -4,9 +4,12 @@ const Node = require('./test/nodejs-bundle/nodejs-bundle.js')
...
 
 const sigServer = require('libp2p-webrtc-star/src/sig-server')
+const wsRendezvous = require('libp2p-websocket-star-rendezvous')
 let server
+let server2
```
diasdavid: Since we now have two "servers", let's use qualifying names:

- wrtcRendezvous
- wsRendezvous

.aegir.js
```diff
@@ -15,14 +18,28 @@ const before = (done) => {
   let count = 0
   const ready = () => ++count === 2 ? done() : null
 
-  sigServer.start({ port: 15555 }, (err, _server) => {
+  sigServer.start({
+    port: 15555
+  }, (err, _server) => {
```
diasdavid: Unnecessary style change.

.aegir.js
diasdavid: I mean, to make it more structured with Parallel. I also see Ready being called 3 times and only checked for 2

test/nodejs-bundle/tcp+websockets+webrtc-star.js
```diff
@@ -91,57 +91,32 @@ describe('TCP + WebSockets + WebRTCStar', () => {
       (cb) => nodeTCP.stop(cb),
       (cb) => nodeWS.stop(cb),
       (cb) => nodeWStar.stop(cb),
-      (cb) => ss.stop(done)
+      (cb) => ss.stop(cb)
```
diasdavid: good catch!

test/nodejs-bundle/tcp+websockets+webrtc-star.js
```diff
@@ -91,57 +91,32 @@ describe('TCP + WebSockets + WebRTCStar', () => {
...
-          },
-          (cb) => {
-            const peers = nodeTCP.peerBook.getAll()
-            expect(Object.keys(peers)).to.have.length(1)
-            expect(Object.keys(nodeTCP.swarm.muxedConns)).to.have.length(1)
-            cb()
-          }
-        ], done)
-      }
+      setTimeout(check, 500, nodeTCP, done, 1, 1)
```
diasdavid: Please use instead:

`setTimeout(() => check(nodeTCP, 1, 1, done), 500`

And update the function arguments respectively (rule: callback always comes last)

test/nodejs-bundle/tcp+websockets+websocket-star.js
```diff
@@ -0,0 +1,154 @@
+/* eslint-env mocha */
```
diasdavid: Instead of creating a copy of `tcp+websockets+webrtc-star` please do try to augment the other test suite

diasdavid:
@mkg20001 thanks for pushing this :) I've made some comments. I am happy to finish this one tomorrow early morning if you don't have time until then. Thank you!

mitra42:
I see a number of configurations posted as not working above,  is there a conclusion of a configuration that works ? Could it be posted here.

PR #123, opened 2017-09-12
Remove non-prepared link in document

djKooks:
It seems development part is not exist in this document, so removed link and commented as WIP. I'm not sure this is working in progress so please let me know if I'm wrong.
Also, `package` part seems it is under `usage` menu in current document.

Please look on. Thanks.


diasdavid:
thanks for catching this @djKooks. Just fixed the README on https://github.com/libp2p/js-libp2p/pull/128

Issue #126, opened 2017-10-19
DHT routing example code does not discover peers and exits irregularly


pvh:
- **Version**: 2a81fc75a8e6ff106e872f6a392b0d04cd488439
- **Platform**: ubuntu 17.10 / node v8.7.0
- **Subsystem**: js-libp2p-dht-kad

#### Steps to reproduce the error:
1. checkout js-libp2p
2. `copy examples/peer-and-content-routing/{12}.js` somewhere
3. `npm install` all the things
4. `node {1,2}.js`

#### Result
The scripts fail to discover other peers and exit irregularly.

This is the DEBUG=* from routing/1.js: https://gist.github.com/anonymous/fc0b56120e94b054d0006f35fe8a5620

dryajov:
thanks for reporting this, @pvh !

nunofmn:
It seems that `100 ms` is not enough time to peers establish connections.
By increasing to `300 ms` was possible to discover other peers
```js
// Set up of the cons might take time
(cb) => setTimeout(cb, 100)
```
I'm going to open a PR to increase the delay.

pvh:
On a different machine here the examples work with a 100ms timeout but not a 10ms timeout, so it appears you've found the right cause, Nuno. I would suggest a more descriptive error message is worth doing. Also, if you're fixing the examples the link from the README.md is broken.

diasdavid:
Thanks for fixing this @nunofmn and @pvh for catching :)

PR #127, opened 2017-10-20
docs(examples): fix peer-content routing example scripts

nunofmn:
Related to #126.

Increase the added delay to let the peer establish connections from `100 ms` to `300 ms`.


diasdavid:
thanks @nunofmn :)

PR #131, opened 2017-11-20
remove innactive multiaddrs

pgte:
Removes inactive multiaddresses from peerInfo. Should fix https://github.com/ipfs/js-ipfs/issues/213


src/index.js
```diff
@@ -177,6 +177,14 @@ class Node extends EventEmitter {
       }
     })
 
+    // detect which multiaddrs we don't have a transport for and remove them
+    multiaddrs.forEach((multiaddr) => {
+      if (!multiaddr.toString().match(/\/p2p-circuit($|\/)/) &&
```
pgte: @diasdavid I'm not particularly proud of this line of code, but I learned that if I remove these addresses (although they apparently have no corresponding transport), tests start failing.. Any insight?

diasdavid: Why not use https://github.com/whyrusleeping/js-mafmt ?

test/nodejs-bundle/multiaddr-trim.js
```diff
@@ -0,0 +1,32 @@
...
+chai.use(require('dirty-chai'))
+const expect = chai.expect
+const createNode = require('./utils').createNode
+
+describe('multiaddr trim', () => {
+  let node
+  it('can create a test node with an irrelevant multiaddr', (done) => {
+    createNode(
+      [
+        '/ip4/0.0.0.0/tcp/0/p2p-webrtc-direct'
```
diasdavid: Can we have some tests with the transports that are currently used by IPFS?

src/index.js
pgte: @diasdavid this is not validation, the goal of this is to remove the multiaddrs for which the node has no transport for..

pgte: (see next LOC)

diasdavid: Can't you use the filter from each transport to then identify the subset of multiaddrs to which no transport has grabbed?

pgte: I'd have to create a separate set named "used multiaddresses" to which I would add the selected multiaddresses for each transport. After that I would have to compare the "used multiaddresses" set to the current multiaddress set and remove all those that are not present.

(and the line we're discussing is not about the filtering , that's the next line. This line is about not removing multiaddresses that have `/p2p-circuit` in them..)

diasdavid: Yes, and that should happen after the swarm.listen step so that multiaddrs such that have things like `0.0.0.0` and port `0` get updated first. This would be a good test to add as well

diasdavid: @pgte did you get this last comment?

pgte: Yup, will be working on this 👍

pgte: @diasdavid please check 67db4e2060a8081be29f6b5e38511b40802ef177, where the final set of multiaddrs is being computed *after* start.

If it's really important I could use sets to do this math, but I'd have to do it very explicitly as currently MultiAddrSet (https://github.com/libp2p/js-peer-info/blob/master/src/multiaddr-set.js) does not seam to support the necessary set operations..

diasdavid:
Just realized that I missed this comment https://github.com/libp2p/js-libp2p/pull/131#discussion_r153232034

> MultiAddrSet (https://github.com/libp2p/js-peer-info/blob/master/src/multiaddr-set.js) does not seam to support the necessary set operations..

You can always expand it.

Please add a test(even if failing now) of a `/ip4/0.0.0.0/tcp/0` to make sure that the final set is the correct one.

pgte:
@diasdavid done in 59362291c0d0e00c68495355df76eb0fe375a25a

pgte:
resolved to test with non-conflatable addresses, making sure in the test that, not only the transport-less address is removed, also the wildcard multiaddr is replaced with the actual address resulting from the `swarm.listen()`.

diasdavid:
@pgte it is not entirely clear to me that this work is finished from your last comment. Could you confirm?

diasdavid:
@pgte thank you so much for pushing this fix! \o/

pgte:
@diasdavid yes, should be finished. No problem! 👍

PR #135, opened 2017-12-01
use that crypto puzzle!

diasdavid:


diasdavid:
Next https://github.com/libp2p/js-libp2p-webrtc-star/issues/128

PR #138, opened 2017-12-14
fix: prevent "The libp2p node is not started yet" when stopping

pgte:
When issuing a `node.stop()`, some asynchronous actions may still be happening, which sometimes originates random "The libp2p node is not started yet" errors.

This prevents that by assuming the node is only stopped after the stopping process is complete.


diasdavid:
Nice! good catch!

PR #147, opened 2018-01-05
tcp/443 on webrtc-star example.

fiatjaf:
According to https://github.com/libp2p/js-libp2p-webrtc-star/issues/134 and my own fiddling with this example.


diasdavid:
thanks! :)

PR #151, opened 2018-01-12
Looks like a typo in DHT setup

Gozala:
I think there was a typo that would have ignored passed options and fallback to defaults for DHT datastore.


lgierth:
:')

diasdavid:
thanks @Gozala :)

PR #161, opened 2018-02-07
Update to libp2p-switch

diasdavid:


package.json
```diff
@@ -39,7 +39,7 @@
   "dependencies": {
     "async": "^2.6.0",
     "libp2p-ping": "~0.6.0",
-    "libp2p-swarm": "~0.35.1",
+    "libp2p-switch": "~0.36.0",
```
dryajov: shouldn't this reset versioning?

diasdavid: why?

dryajov: Thats a good question - I was thinking along the lines of once the name of the module changes, it might be worth resetting the version, but thinking some more I realize that there is no benefit in doing that. nvm.

PR #164, opened 2018-02-14
feat: add pubsub to libp2p

diasdavid:


diasdavid:
image

CI is green (we need to support Windows at some point, wink wink @richardschneider :))

Issue #46, opened 2016-11-27
Improve libp2p config


diasdavid:
Right now, it is impossible to signal through the config if we want to use a transport just for dialing and not for listening, in fact, go-ipfs makes the hard decision of disabling any dialing from a transport from where we are not listening. 

The issue is that with this design, we have to make something hacky/hard coded to enable the WebSockets transport within libp2p-swarm for dialing in the browser.

https://github.com/libp2p/js-libp2p/blob/master/src/index.js#L97-L102

Ideas:
- Have a `dialer` multiaddr addr to signal that no listener should be started, but it should be considered to dial anyway
- Have a more config option within the config.json that informs swarm to do that.

diasdavid:
Please review proposal here https://github.com/libp2p/js-libp2p/pull/166

jacobheun:
With #166 merged I think we could add transport configuration for something like `isDialer` and then push all dialer only transports into the switch transports after startup. This could be enabled by default, as that is already a set expectation with web sockets.

diasdavid:
@jacobheun related, see https://github.com/libp2p/js-libp2p/issues/201

diasdavid:
DONE \o/ :D

Issue #159, opened 2018-02-05
libp2p.next()


diasdavid:
This is the tracking issue for the next generation of the JavaScript implementation of js-libp2p. In this coming version we want:

- [x] Expose PubSub as a core primitive
  - https://github.com/libp2p/js-libp2p/issues/140
- [x] Make mplex a real thing
  - https://github.com/libp2p/js-libp2p-mplex/issues/68 and opened pull-stream only use for take https://github.com/libp2p/js-libp2p-mplex/issues/73
- [x] Improve the config management
  - https://github.com/libp2p/js-libp2p/issues/46
  - https://github.com/libp2p/js-libp2p/pull/166
- [x] Formalize Peer Routing and Content Routing Interfaces for internal modules
  -  [x] Update libp2p-kad-dht to follow the interface specs https://github.com/libp2p/js-libp2p-kad-dht/pull/25
  - [x] Create https://github.com/libp2p/js-libp2p-delegated-peer-routing
  - [x] Create https://github.com/libp2p/js-libp2p-delegated-content-routing
- [ ] Ensure we can set requirements for certain protocols
  - https://github.com/libp2p/js-libp2p-swarm/issues/239
  - https://github.com/libp2p/js-libp2p/pull/166 (also being solved here)
- [ ] Make libp2p a proper state machine
- [ ] Refactor and add DI support so that libp2p has a common backbone for all the things that plug to it.
  - https://github.com/libp2p/js-libp2p/issues/130
- [ ] Replace the external rendezvous points from *-star transports
  - https://github.com/libp2p/js-libp2p/issues/134
- [x] Add Statistics Tracking (@pgte)
  - https://github.com/libp2p/js-libp2p/issues/157 
- [x] Implement the Connection Manager (@pgte)
  - https://github.com/libp2p/js-libp2p-connection-manager

Issue #130, opened 2017-11-08
dependency injection during boot time


pgte:
Some libp2p modules require or would benefit from knowing the `peerId` and other values. It happens that, at the time the declarative options are created, the peerId is not yet known. As an example, libp2p-websocket-star [would require the `PeerId` object to prevent spoofing](https://github.com/libp2p/js-libp2p/pull/122#discussion_r147589693). It happens that this value is not known before boot.

Here I propose a modules definition form that is backwards compatible and can be summed up to:

If a module is a function, libp2p assumes it is a constructor and executes that function, which should return a the module instance.

This would allow to declare config like this:

```js
{
  modules: {
    transport: [
      wrtcStar // instance,
      ({peerId}) => new WSStar({id: peerId})
    ]
  }
}
```

pgte:
One problem with this setting is the discovery. Usually, a discovery module is extracted from the transport module, something like this:

```js
const wstar = new WRTCStar()
{
  modules: {
    transport: [
      wstar
    ],
    discovery: [ wstar.discovery ]
  }
}
```

One option to mitigate this would be to implicitely add the discovery module if the transport instance contains a `.discovery` attribute.

Another workaround would be to cache it like this:

```js
let wsstar // caches the WSStar object instance
const createWSStar = ({peerId}) => (wsstar ? wsstar : wsstar = new WSStar({id: peerId}))

const libp2pOptions = {
  modules: {
    transport: [
      createWSStar
    ],
    discovery: [
      (options) => createWSStar(options).discovery
    ]
  }
}
```

Thoughts?

pgte:
(Probably should replace `peerId` with `peerInfo`, but not really relevant for the overall discussion of the DI method here).

dryajov:
Another approach might be passing an instance of swarm to the transport during initialization, the peerId/peerInfo can be extracted from it. We already do that for DHT and Circuit.

diasdavid:
Let's continue evaluating this, I need to think more about it but I am in the same page that we need DI.

dryajov:
overall, 👍 for DI, should make things a lot more testable and such!

mkg20001:
Currently I'm using a `.setSwarm(swarm)` function after transport creation to hack around this limitation in https://github.com/libp2p/js-libp2p-websocket-star/pull/43
Something similar could be implemented. Like `if (typeof transport._setSwarm === 'function') transport._setSwarm(swarm)` after transport creation in libp2p.
[...]

PR #166, opened 2018-02-20
new config

diasdavid:
For https://github.com/libp2p/js-libp2p/issues/46. Takes into account:
- https://github.com/ipfs/notes/issues/209
- https://github.com/libp2p/js-libp2p-switch/issues/239

Would love to get a ton of feedback from libp2p users.

-------------

- [x] Add support to receive transport instances or Classes 
- Needs:
  - [x] https://github.com/libp2p/js-libp2p-railing/pull/75
  - [x] https://github.com/libp2p/js-libp2p-mdns/pull/74


README.md
```diff
@@ -101,35 +101,76 @@ const libp2p = require('libp2p')
...
+      },
+      EXPERIMENTAL: {                 // Experimental features ("behind a flag")
+        pubsub: true,
+        dht: true
+      }
     }
 
-    super(modules, peerInfo, peerBook, options)
+    // overload any defaults of your bundle
+    Object.assign(options, _options)
```
alanshaw: Double checking you want an `assign`? You've gone to a lot of trouble to establish defaults which are easily removed. For example, if I pass an options object like `{ config: { addrs: { listen: ['/ip4/0.0.0.0/tcp/5555'] } } }` I blow away all the other defaults that have been set for `peerDiscovery`, `protocolMuxing` etc.

Do you need a merge instead?

diasdavid: You are right, it need to be a deepMerge.

README.md
```diff
@@ -101,35 +101,76 @@ const libp2p = require('libp2p')
...
+        contentRouting: {}            // this will change once we factor that into two modules, for now do the following line:
+        dht: DHT                      // DHT enables PeerRouting, ContentRouting and DHT itself components
       },
-      discovery: [
-        new MulticastDNS(peerInfo)
-      ],
-      // DHT is passed as its own enabling PeerRouting, ContentRouting and DHT itself components
-      dht: DHT
+      config: {                       // The config object is the part of the config that can go into a file, config.json.
+        addrs: {                      // Multiaddrs
```
mkg20001: The `peerInfo` already contains multiaddrs. What happens with that?

README.md
```diff
@@ -101,35 +101,76 @@ const libp2p = require('libp2p')
...
-        new WS()
-      ],
-      connection: {
-        muxer: [
+  constructor (_peerInfo, _peerBook, _options) {
+    const options = {
+      peerInfo: _peerInfo             // The Identity of your Peer
+      peerBook: _peerBook             // Where peers get tracked, if undefined libp2p will create one instance
+      modules: {
+        transport: [
```
mkg20001: I think some kind of first and second class transport system is needed.
For example: the new ws-star needs the libp2p swarm to dial the server via WS but WS is independent from the swarm.
So any listen/dial stuff for first class transports would be setup before the second class transports are setup (so dialing does not fail)

README.md
```diff
@@ -101,35 +101,76 @@ const libp2p = require('libp2p')
...
+          }
+          // .. other discovery module options.
+        },
+        peerRouting: {},
+        contentRouting: {},
+        protocolMuxing: {             // Protocol muxing sequence. libp2p uses sane defaults (below). You can override these with:
+          transport: {
+            secio: {
+              spdy: {
+                '*': '*'
```
mkg20001: What does this exactly mean? Is the thing on the left the multiprotocol for secio and the thing on the right the protocol that is actually being dialed?

README.md
```diff
@@ -101,35 +101,76 @@ const libp2p = require('libp2p')
...
-      ],
-      // DHT is passed as its own enabling PeerRouting, ContentRouting and DHT itself components
-      dht: DHT
+      config: {                       // The config object is the part of the config that can go into a file, config.json.
+        addrs: {                      // Multiaddrs
+          listen: []                  // To listen
+          ann: []                     // To announce/share with the network (i.e DNS addr)
+          notAnn: []                  // To keep private
+        },
+        peerDiscovery: {
```
mkg20001: How will the options be passed to the module here?

README.md
```diff
@@ -101,35 +101,76 @@ const libp2p = require('libp2p')
...
+          mdns: {                     // mdns options
+            interval: 1000            // ms
+            enabled: true
+          },
+          webrtc-star: {              // webrtc-star options
```
VictorBjelkholm: Not consistent, should be `webrtcStar`

README.md
```diff
@@ -101,35 +101,76 @@ const libp2p = require('libp2p')
...
       },
-      discovery: [
-        new MulticastDNS(peerInfo)
-      ],
-      // DHT is passed as its own enabling PeerRouting, ContentRouting and DHT itself components
-      dht: DHT
+      config: {                       // The config object is the part of the config that can go into a file, config.json.
+        addrs: {                      // Multiaddrs
+          listen: []                  // To listen
+          ann: []                     // To announce/share with the network (i.e DNS addr)
```
VictorBjelkholm: Not a fan of these abbraviations, makes it difficult to understand what it really is unless you have context. 

Also, taking inspiration from go-ipfs, it would turn out something like this:

```
{
  addresses: {
    swarm: [] // to listen
    announce: [] // for sure to announce
    noAnnounce: [] // stop these from announcing at all cost
    api: ''
    gateway: ''
  }
}
```

README.md
diasdavid: Both transports will be first class with #130 occurring during the #159. Not the concern of this proposal/PR

README.md
diasdavid: They get added/merged. This is to enable the config to be a single JSON file that has the addrs without any complex setup.

README.md
diasdavid: Good questions. 

1) The nested'ness here is to represent order/options. This says "After the transport, always do SECIO, after SECIO you can do SPDY or MPLEX, then do whatever" 

2) re: `'*': '*'` - My intention is to mean "Now protocol mux whatever protocol you want".

README.md
diasdavid: They will, will present these details after this first pass on the looks of the config.

README.md
diasdavid: good catch 👍

README.md
diasdavid: Ack.

README.md
dryajov: we had a proposal for transport priorities at some point, but it was rejected.

README.md
dryajov: look into lodash defaults/deepDefaults instead, they do what you need.

README.md
diasdavid: @dryajov link it please

dryajov: @diasdavid https://lodash.com/docs/4.17.5#defaults and https://lodash.com/docs/4.17.5#defaultsDeep

diasdavid: image

I'm confused, wrong c&p?

dryajov: my bad, this is the correct one - https://github.com/libp2p/js-libp2p-tcp/pull/78

diasdavid: @dryajov that was a proposal for a pseudo priority based on what the module says. It needs to be a user defined.

diasdavid:
I've continue my work on this PR. Made some progress and overall I like where this is headed (which is eventually enabling simpler configuration and #130). 

I tried passing the responsibility completely to libp2p to instantiate all of its modules (which was not the case for Transports and Discovery services) but I hit the 🐓&🥚 of transports such as WebRTC that might require special binding modules such as `wrtc`. 

I do want to make sure that everything that goes on `config` is serializable to JSON and I do want libp2p to be able to control the instantiation of modules, however, I believe I can not escape the need to be able to pass already instantiated modules too. It is not hard, just annoying because now I have to add a hint to the module to tells me if the reference is to an instance or to a class.

If this just sounded like alien 👽 mumbo jumbo to you, do not worry, I'll carry on with the coding and once it is fully ready, it will provide more clarity. If you are interested in understanding, let me know :)

dryajov:
> I have to add a hint to the module to tells me if the reference is to an instance or to a class. 

This sounds like you're recreating IoC (inversion of control) containers all over :) There are some out there that do that already.

README.md
```diff
@@ -101,47 +101,91 @@ libp2p becomes very simple and basically acts as a glue for every module that co
...
-        new MulticastDNS(peerInfo)
-      ],
-      // DHT is passed as its own enabling PeerRouting, ContentRouting and DHT itself components
-      dht: DHT
+
+      // Enable/Disable Experimental features
+      EXPERIMENTAL: {                 // Experimental features ("behind a flag")
+        pubsub: true,
+        dht: true
+      }
```
diasdavid: Make `EXPERIMENTAL` part of the config.

README.md
```diff
@@ -101,47 +101,91 @@ libp2p becomes very simple and basically acts as a glue for every module that co
...
+        peerRouting: {},
+        contentRouting: {},
+        protocolMuxing: {             // Protocol muxing sequence. libp2p uses sane defaults (below). You can override these with:
+          transport: {
+            secio: {
+              spdy: '*',              // '*' means, from this point forward, accept any protocol
+              mplex: '*'
+            }
+          }
+        }
```
diasdavid: This feature will have to be implemented at the Switch Level

diasdavid: Before merging, I'll remove these options from the README and continue working on this feature design at https://github.com/libp2p/js-libp2p/issues/201

README.md
```diff
@@ -101,47 +101,90 @@ libp2p becomes very simple and basically acts as a glue for every module that co
...
 //   transport: websockets + tcp
-//   stream-muxing: SPDY
+//   stream-muxing: spdy & mplex
 //   crypto-channel: secio
 //   discovery: multicast-dns
 
 const libp2p = require('libp2p')
 const TCP = require('libp2p-tcp')
 const WS = require('libp2p-websockets')
 const spdy = require('libp2p-spdy')
```
diasdavid: uppercase

README.md
```diff
@@ -101,47 +101,90 @@ libp2p becomes very simple and basically acts as a glue for every module that co
...
+const mplex = require('libp2p-mplex')
```
diasdavid: uppercase + needs to be included in the modules

README.md
```diff
@@ -101,47 +101,90 @@ libp2p becomes very simple and basically acts as a glue for every module that co
...
 const secio = require('libp2p-secio')
```
diasdavid: uppercase

README.md
```diff
@@ -101,47 +101,90 @@ libp2p becomes very simple and basically acts as a glue for every module that co
...
+        dht: DHT                      // DHT enables PeerRouting, ContentRouting and DHT itself components
+      },
+
+      // libp2p config options (typically found on a config.json)
+      config: {                       // The config object is the part of the config that can go into a file, config.json.
+        addresses: {                  // Multiaddrs
+          listen: []                  // To start transport listener
+          announce: []                // To announce/share with the network (i.e DNS addr)
+          noAnnounce: []              // To keep private
+        },
```
diasdavid: These are extra address to the PeerInfo.

- [ ] Impl this feature
- [ ] Add tests
  - [ ] listen
  - [ ] announce
  - [ ] noAnnounce

diasdavid: Will break this into a separate issue to not delay this refactor any longer

diasdavid: https://github.com/libp2p/js-libp2p/issues/202

test/utils/bundle-browser.js
```diff
@@ -0,0 +1,85 @@
...
+          wsStar
+        ],
+        streamMuxer: getMuxers(_options.muxer),
+        connEncryption: [
+          SECIO
+        ],
+        peerDiscovery: [
+          // TODO add tags to all Discovery Mechanisms
+          // wrtcStar.discovery,
+          // wsStar.discovery,
```
diasdavid: Need to add a tag to both of these discovery mechanisms (following Bootstrap and mdns) and then these tests will work like a charm 👌🏽

mkg20001: ws-star: https://github.com/libp2p/js-libp2p-websocket-star/pull/51

mkg20001: wrtc-star: https://github.com/libp2p/js-libp2p-webrtc-star/pull/147

src/index.js
```diff
@@ -151,24 +151,33 @@ class Node extends EventEmitter {
...
               if (typeof D === 'function') {
+                this._config.peerDiscovery[D.tag].peerInfo = this.peerInfo
                 d = new D(this._config.peerDiscovery[D.tag])
               } else {
                 d = D
+                console.log(d)
               }
 
+              console.log('-> Waiting for peers')
+              d.on('peer', () => console.log('found peer'))
```
diasdavid: testing note: this never fires.

test/utils/bundle-browser.js
```diff
@@ -78,6 +78,10 @@ class Node extends libp2p {
 
     defaultsDeep(_options, defaults)
 
+    // TODO: is there a defaultsDeep that doesn't copy the refs?
+    _options.modules.peerDiscovery.push(wrtcStar.discovery)
+    _options.modules.peerDiscovery.push(wsStar.discovery)
```
diasdavid: Is there a defaultsDeep that doesn't clone objects that it doesn't need to clone/merge?

diasdavid: @alanshaw, @pgte, @satazor, @jacobheun et al, calling out for your JS wisdom to help on this one :)

satazor: You can use `https://lodash.com/docs/4.17.10#mergeWith` with a customizer function that does the "defaults" logic by checking for undefineds.

jacobheun: I think mergeWith will still clone the objects. You could use something like [deepmerge](https://github.com/KyleAMathews/deepmerge) and turn cloning off. The array merging would need to get updated for it though. For example, the following would combine options and defaults. Any arrays that exist in both, the options array would win out. This should be all object references, no clones.

```js
_options = deepmerge({_options, defaults, { clone: false, arrayMerge: (destinationArray, sourceArray, options) => sourceArray })
```

satazor: @jacobheun I just tried it and it doesn't:

screen shot 2018-06-19 at 11 36 59

The lodash documentation says the following:

> Array and plain object properties are merged recursively

jacobheun: Ah yep, you're right. Also just did a test with mergeWith for custom merging and it looks solid.

diasdavid: Ideally, we want to replace the defaultsDeep + the later array.push by something simpler. Currently, I don't see that using another package with multiple options and custom merge functions is better.

jacobheun: I agree that abstracting that out to a simple method call is the way to go, so we can use that elsewhere and have consistent defaults behavior, if there are no existing modules that do it the way we need it to behave. I think the issue with doing array.push later is that if users want to specify their own discovery array and not use those two defaults, they will have to remove them after creating the libp2p node. 

As a user, I would expect that if I pass options for `modules.peerDiscovery`, I wouldn't get the defaults. If I want to add additional peerDiscovery options, I'd expect to have to add them after node creation. If those rules are intended to be different I think they need to be explicitly called out.

README.md
```diff
@@ -101,47 +101,85 @@ libp2p becomes very simple and basically acts as a glue for every module that co
...
+        relay: {                      // Circuit Relay options
+          enabled: false,
+          hop: {
+            enabled: false,
+            active: false
+          }
+        }
+        // Enable/Disable Experimental features
+        EXPERIMENTAL: {               // Experimental features ("behind a flag")
+          pubsub: false,
```
diasdavid: - [ ] Needs to get tested

diasdavid:
@jacobheun wanna power duo this one with me? The final steps that I see are on https://github.com/libp2p/js-libp2p/pull/166#issuecomment-394722032, but I would really appreciate a fine grained CR to make sure nothing gets missed.

jacobheun:
@diasdavid let's do it! I'll start running through the code in more detail.

src/index.js
```diff
@@ -126,30 +127,58 @@ class Node extends EventEmitter {
...
           // always add dialing on websockets
-          this.switch.transport.add(ws.tag || ws.constructor.name, ws)
+          this._switch.transport.add(ws.tag || ws.constructor.name, ws)
         }
 
         // all transports need to be setup before discover starts
-        if (this.modules.discovery) {
-          return each(this.modules.discovery, (d, cb) => d.start(cb), cb)
+        if (this._modules.peerDiscovery && this._config.peerDiscovery) {
+          each(this._modules.peerDiscovery, (D, cb) => {
```
jacobheun: It would be nice for the each callback to be named something else since we're touching this code. Being `cb` for `each` and for `series` makes this harder to follow than it needs to be.

src/index.js
```diff
@@ -175,10 +204,10 @@ class Node extends EventEmitter {
         // detect which multiaddrs we don't have a transport for and remove them
         const multiaddrs = this.peerInfo.multiaddrs.toArray()
 
-        transports.forEach((transport) => {
+        this._transport.forEach((transport) => {
```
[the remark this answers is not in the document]
jacobheun: @diasdavid I've made some changes locally to this branch for testing, that include these changes, the rebase of master, as well as a switch to use @nodeutils/defaults-deep, which wraps lodash.mergeWith, to fix the push issue with defaults.

Would you like me to push those changes to this branch or up to a separate branch so they're easier to review?

I can also help with the examples (maybe from the bottom check boxes up so we dont collide) if you'd like.

src/index.js
```diff
@@ -126,30 +127,58 @@ class Node extends EventEmitter {
...
+          this._switch.transport.add(ws.tag || ws.constructor.name, ws)
         }
 
         // all transports need to be setup before discover starts
-        if (this.modules.discovery) {
-          return each(this.modules.discovery, (d, cb) => d.start(cb), cb)
+        if (this._modules.peerDiscovery && this._config.peerDiscovery) {
+          each(this._modules.peerDiscovery, (D, cb) => {
+            // If enabled then start it
+            if (this._config.peerDiscovery[D.tag].enabled) {
```
jacobheun: We need to also verify the existence of `this._config.peerDiscovery[D.tag]`. This will break when that is inevitably missing from the config. 

This brings up a question around defaults if the config is not present. Does the config **have** to exist and be **enabled** in order for the discovery service to be added?

I think the logic should be:
* If `_modules.peerDiscovery` contains an instance, use that and ignore info in `_config.peerDiscovery`. If I pass an instance I'd expect that to be enabled and used as is.
* If `_modules.peerDiscovery` contains a function, its configuration exists in `_config.peerDiscovery` and isn't explicitly `enabled: false`, it is added. Otherwise, it will be ignored.

I think this helps make configuration easier. If I supply something to modules I want it enabled unless I explicitly disable it.

src/index.js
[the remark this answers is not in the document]
diasdavid: Perhaps do a PR first (I want to see the mergeWidth) and after the first PR we can work on the same branch.

👍 and ❤️ for helping with the examples!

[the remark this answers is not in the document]
jacobheun: https://github.com/libp2p/js-libp2p/pull/206 I removed the rebase from master to make that potential merge easier/cleaner.

src/index.js
diasdavid: Let's rather use a strategy such as https://github.com/ipfs/js-ipfs/blob/master/src/core/config.js#L5-L41 to ensure that the config exists, letting it represent the truth about _the configuration_. There are so many options that it is better to be explicit rather than creating too many ternaries.

alanshaw:
Can we merge and release this as is and finish the examples in separate PRs?

jacobheun:
@alanshaw I am working on this today. I think we need to resolve, https://github.com/libp2p/js-libp2p/pull/166#discussion_r196830767 first, which I will work on. With the changes to the config, I feel it's important to make sure we have proper validation there so users aren't confused why their config is not working.

I should have this resolved today or tomorrow.

alanshaw:
🚀 awesome @jacobheun! Apologies I hadn't clocked https://github.com/libp2p/js-libp2p/pull/166#discussion_r196830767

jacobheun:
Okay, so I think I have everything checked off. I've rebased from master so everything is up to date (@diasdavid fyi on the rebase in case you pull). 

I've coupled the latest changes into two commits, hopefully this will help keep things clear.
1. The first commit is the updated examples. I ran through all of them a final time after my last changes and everything runs. Woo!
2. I added Joi to validate the configuration. The second commit is pretty much about that, but also includes some of the misc changes @alanshaw mentioned to add to future PRs. This resulted in some code clean up, as we're able to rely on thrown errors from config validation to avoid some unnecessary property checks.

I added tests for the new config validation. There is some custom logic there for peerDiscovery in modules and config, as well as experimental dht.

**Pubsub**: pubsub should now only be initialized if it's enabled via EXPERIMENTAL. Calling any methods on pubsub when it's not enabled will result in an error.

Provided there are no changes requested from those 2 latest commits, I think this should be good to go.

ci/Jenkinsfile
```diff
@@ -1,2 +1,2 @@
 // Warning: This file is automatically synced from https://github.com/ipfs/ci-sync so if you want to change it, please change it there and ask someone to sync all repositories.
-javascript()
+javascript(['nodejs_versions': ['8.11.3']])
```
diasdavid: @jacobheun @alanshaw switched to Node.js 8 for now until wrtc has builds for node 10

examples/chat/README.md
```diff
@@ -1 +1,13 @@
...
+
+## Setup
+1. Install the modules, `npm install`.
+2. Open 2 terminal windows in the `./src` directory.
+
+## Running
+1. Run the listener in window 1, `node listener.js`
+2. Run the dialer in window 2, `node dialer.js`
+3. Type a message in either window and hit _enter_
+4. Tell youself secrets to your hearts content!
```
diasdavid: Thank you for writing the tutorial. Helps for #186

diasdavid:
Only test that failed is Jenkins related

image

PR #174, opened 2018-03-04
update echo and chat examples to use libp2p-switch

zcstarr:
Fixes the current echo example


hacdias:
Thanks @zcstarr. Could you also update it on [examples/chat/src/listener.js](https://github.com/libp2p/js-libp2p/blob/master/examples/chat/src/listener.js)?

zcstarr:
np, sure!

Issue #65, opened 2017-03-26
Checking for WebSockets constructor name fails under uglification


jackkleeman:
Currently to add the WebSockets transport index.js does a check for the constructor.name being 'WebSockets'. This fails under uglification unless 'WebSockets' is made a reserved word in the uglify config, as constructor names are one of the things uglify removes. I recommend that either this is documented, or a different solution is devised; there is already a TODO in the code suggesting that this is found. For example, we could standardise some kind of boolean flag in a transport class which tells libp2p to include it even if there are no listening addresses. That wouldn't be websocket-specific, and wouldn't require checking the name against a string.

dignifiedquire:
This is unfortunate, we use this pattern in a couple of places. I guess we need to find a new way to do these kinds of detections.

diasdavid:
Trying to come up with an elegant solution and I always end up in making something that looks too complicated. The crux of this problem is the fact that we might handle several versions of the same module that are compatible with each other, but because of the dependency tree, they coexist.

Seems that we need to go 🦆 typing all the way, which isn't the best, but will suffice for the majority (if not all) the cases.

mqzry:
_Warning: I did not look at the code yet._
Can't you just specify a regex to the mangle-regex option of uglify that matches functions starting with a capital? You do get a dependency on how the transpiler picks the prototype.name for some constructor.name.

Can you explain the crux of the problem a bit more? I might try to solve this issue

dryajov:
Actually `libp2p` (as well as `libp2p-swarm`) uses the `.tag` property as the primary way to determine the name of the transport, it uses `.constructor.name` only as a fallback, so it should be possible to just set the `.tag` as the name of the transport. I like this solution and I explicitly set `.tag` to `Circtuit` in the `libp2p-circuit`, for example.

diasdavid:
@fsdiogo is handling this.

fsdiogo:
PR https://github.com/libp2p/js-libp2p/pull/182 fixes this.

fsdiogo:
Check https://github.com/ipfs/js-ipfs/issues/1321 for more info.

PR #177, opened 2018-03-08
avoid constructor.name . fix for #65

ya7ya:
this avoids using `constructor.name` so uglify can mangle and compress `js-ipfs` properly.

this should close https://github.com/libp2p/js-libp2p/issues/65

cc @diasdavid


src/index.js
```diff
@@ -129,7 +130,7 @@ class Node extends EventEmitter {
         this.switch.transport.add(
           transport.tag || transport.constructor.name, transport)
       } else if (transport.constructor &&
-                 transport.constructor.name === 'WebSockets') {
+                 transport instanceof WebSockets) {
```
diasdavid: Same feedback as https://github.com/multiformats/js-multiaddr/pull/55#discussion_r173337049

ya7ya: @diasdavid hey, 

**solution 1** : What if we add a `name` property to the `WebSockets` class which isn't gonna be mangled by uglify like so
```javascript
class WebSockets {
 constructor (opts) {
   this.name = 'WebSockets'
 }
}
```

**Solution 2** : use both `instanceOf` and `constructor,name` like so
` if (transport.constructor.name === 'WebSockets' || transport instanceOf WebSockets)`

What do you think.??

diasdavid: solution 1 is closer to a real solution. Even better would be just to have a `isWebSockets` method

diasdavid:
See @pgte proposal https://github.com/ipfs/js-ipfs/issues/1131#issuecomment-373299739

fsdiogo:
PR https://github.com/libp2p/js-libp2p/pull/182 fixes this.

diasdavid:
This has been fixed in all the js-ipfs deps thanks to @fsdiogo. @ya7ya thanks for helping us diagnose and provide solutions. Check https://github.com/ipfs/js-ipfs/issues/1321 for more info.

Issue #191, opened 2018-05-08
Tests are failing against master


dryajov:
- **Version**: master
- **Platform**: All (at least it appears so)
- **Subsystem**: libp2p, libp2p-secio

#### Description:
Master is failing to build with:
- https://travis-ci.org/libp2p/js-libp2p
- https://circleci.com/gh/libp2p/js-libp2p/540#tests/containers/0

<img width="689" alt="screen shot 2018-05-08 at 1 02 42 pm" src="https://user-images.githubusercontent.com/1094341/39777390-b0b94fb2-52c0-11e8-8ce3-4cbe6f6535c6.png">

The `Bad signature` error message seems to be coming from `libp2p-secio` - https://github.com/libp2p/js-libp2p-secio/blob/6d09b472ec8d96040c54fbee630dc0b117a628bb/src/handshake/crypto.js#L151

I'm not sure about the second failure.

#### Steps to reproduce the error:

- Clone this repo and run `npm run test`.

PR #192, opened 2018-05-09
[WIP] feat: use pull stream based mplex

dryajov:
NOTE: Im using libp2p-switch as a git dep, so that we can test it on CI. This is temporary and should be changed once libp2p-switch is released.


dryajov:
Tests are also failing against master - https://github.com/libp2p/js-libp2p/issues/191, it doesn't seem to be related to the mplex changes.

diasdavid:
Work on https://github.com/libp2p/js-libp2p-mplex/pull/76

PR #193, opened 2018-05-12
Latest wrtc is breaking support in Node.js

diasdavid:
ref #191

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: node_js
 
 matrix:
   include:
-  - node_js: 'stable'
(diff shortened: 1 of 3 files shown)


diasdavid:
Seems that `wrtc` is causing issues 

image

diasdavid:
simple-peer also fails with latest wrtc https://github.com/diasdavid/simple-peer/pull/1

diasdavid:
locking v to 0.1.1 for now. Reported at #194 to be debugged in async

diasdavid:
image

no mac os x workers :(

PR #206, opened 2018-06-20
refactor: use different defaultsDeep and clean up some code

jacobheun:
## defaultsDeep
This includes updates to the already changed examples.

This leverages `@nodeutils/defaults-deep` instead of `lodash.defaultsdeep`. This ensures that cloning of objects does not happen. This will also make sure that an array passed within options, will not be stomped on by defaults. Passed arrays win over defaults.

Example:
```js
const options = {
  modules: {
    peerDiscovery: [Bootstrap, wsStar.discovery]
  }
}

const defaults = {
  modules: {
    peerDiscovery: [MulticastDNS]
  }
}

return defaultsDeep(options, defaults)
```
will result in a returned object deeply equivalent to the options.

## code updates
* Clarifies some nested callback naming
* Removes redundant looping over transports

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ const MPLEX = require('libp2p-mplex')
 const SECIO = require('libp2p-secio')
 const MulticastDNS = require('libp2p-mdns')
 const DHT = require('libp2p-kad-dht')
-const defaultsDeep = require('lodash.defaultsdeep')
+const defaultsDeep = require('@nodeutils/defaults-deep')
 
 class Node extends libp2p {
   constructor (_peerInfo, _peerBook, _options) {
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -113,7 +113,7 @@ const MPLEX = require('libp2p-mplex')
 const SECIO = require('libp2p-secio')
 const MulticastDNS = require('libp2p-mdns')
 const DHT = require('libp2p-kad-dht')
-const defaultsDeep = require('lodash.defaultsdeep')
+const defaultsDeep = require('@nodeutils/defaults-deep')
```
diasdavid: great find!

Issue #98, opened 2017-07-05
Be a state machine. Just like js-ipfs :) 


diasdavid:

diasdavid:
@pgte is this still something you would like to pursue?

pgte:
Yes.

lgierth:
What does this entail? Is it something that other libp2p impls should be doing too?

diasdavid:
@lgierth not at all. It is just how the libp2p Node in JS changes its state.

diasdavid:
@pgte you have a ton of experience on this and have pointed out multiple issues with the current state-machine in js-ipfs. Do you have any recommendations? Modules to use, examples to take inspiration from so that we don't fall in the same thing? I want to implement it asap for #159

pgte:
@diasdavid I think it can be boiled down to this:

* define a FSM (states and transitions) for a libp2p node
* let the user commands drive transitions of the state machine
* act on state changes instead of user commands

Some states transitions are initiated by the machine itself. Apply the same principle: do not act directly, only start transitions, decoupling state transition from handling the state transition. This only should allow you to escape some nasty bugs and require less verification and synchronisation.

As for the library, I like https://www.npmjs.com/package/fsm-event 's API.

Any more pointers you would like?

pgte:
Ha, also another think, to your consideration:
(This doesn't apply directly to libp2p, I think, but can perhaps be translated to IPFS more directly.)

Allow operations while offline. If the operation requires the node to be online, wait until it is inline instead of failing.
For instance, pubsub publish can perhaps fail silently if the node is offline (since the API offers no delivery guarantees).

While, OTOH, the `file.get(cid)` can be postponed to when the node is online.

This would allow a more friendly programming style, similar to many database drivers out there, where we don't need to wait until we connect before issuing commands or queries.

diasdavid:
Thank you! We use fsm-event in js-ipfs https://github.com/ipfs/js-ipfs/blob/master/src/core/state.js#L7 but I know you had some issues with how it was implemented. Could you elaborate on those?

pgte:
Sure. As an example, let's see `ipfs.start()`: https://github.com/ipfs/js-ipfs/blob/master/src/core/components/start.js

It's confounding [issuing the state transition](https://github.com/ipfs/js-ipfs/blob/master/src/core/components/start.js#L28) and [implementing the state transition](https://github.com/ipfs/js-ipfs/blob/master/src/core/components/start.js#L30-L51).

[...]

PR #207, opened 2018-06-21
docs: 2018 Q3 OKRs

diasdavid:
**libp2p:**
- [Spreadsheet for 2018 2018 Q3](https://docs.google.com/spreadsheets/d/1HTXfgR5FyPTFhsTkFPRThkeMvHvCgJOaAs7BSl_vQ_0/edit#gid=1241853194)
- [PR for 2018 Q2 for reference](https://github.com/ipfs/js-ipfs/pull/1281)

### DIFF
--- a/OKR.md
+++ b/OKR.md
@@ -15,7 +15,7 @@ We try to frame our ongoing work using a process based on quarterly Objectives a
 **libp2p debugging is not painful**
 
 - Error handling is done correctly and error codes are part of the API
-- libp2p and libp2p-switch become full state machines
+- libp2p and libp2p-switch become full state machines - @jacobheun
 - libp2p has a network visualization tool 
 
 **libp2p is seen as the go to p2p networking stack**
--- a/OKR.md
+++ b/OKR.md
@@ -12,11 +12,14 @@ We try to frame our ongoing work using a process based on quarterly Objectives a
 - IPNS works over PubSub (+ go-libp2p interop) - @vascosantos
 - IPNS works over DHT (+ go-libp2p interop) - @vascosantos
 
-**libp2p debugging is not painful**
+**libp2p debugging and interop testing is not painful**
 
 - Error handling is done correctly and error codes are part of the API - @jacobheun
 - libp2p and libp2p-switch become full state machines - @jacobheun
 - libp2p has a network visualization tool - @vascosantos
+- js-libp2p has a daemon with a RPC API and a client library
+- We get the RPC API standardised so that go-libp2p and rust-libp2p can implement it as well
+- Write the first set of interop tests that tests
 
 **libp2p is seen as the go to p2p networking stack**
 
--- a/OKR.md
+++ b/OKR.md
@@ -9,8 +9,8 @@ We try to frame our ongoing work using a process based on quarterly Objectives a
 - Delegated Peer Routing is completed 
 - Delegated Content Routing is completed  
 - DHT interops with go-libp2p
-- IPNS works over PubSub (+ go-libp2p interop)
-- IPNS works over DHT (+ go-libp2p interop)
+- IPNS works over PubSub (+ go-libp2p interop) - @vascosantos
+- IPNS works over DHT (+ go-libp2p interop) - @vascosantos
 
 **libp2p debugging is not painful**
 
--- a/OKR.md
+++ b/OKR.md
@@ -4,31 +4,7 @@ We try to frame our ongoing work using a process based on quarterly Objectives a
 
 ## 2018 Q3
 
-**All pieces are there. Ship Peer & Content Routing + IPNS**
-
-- Delegated Peer Routing is completed - @jacobheun
-- Delegated Content Routing is completed - @jacobheun
-- DHT interops with go-libp2p - @vascosantos
-- NAT Traversal is implemented in js-libp2p (for at least TCP) - @jacobheun
-- IPNS works over PubSub (+ go-libp2p interop) - @vasco-santos
-- IPNS works over DHT (+ go-libp2p interop) - @vasco-santos
-
-**libp2p debugging and interop testing is not painful**
-
-- Error handling is done correctly and error codes are part of the API - @jacobheun
-- libp2p and libp2p-switch become full state machines - @jacobheun
-- libp2p has a network visualization tool - @vasco-santos
-- js-libp2p has a daemon with a RPC API and a client library - @jacobheun
-- We get the RPC API standardised so that go-libp2p and rust-libp2p can implement it as well - @jacobheun
-- Write the first set of interop tests that tests - @jacobheun
-
-**libp2p is seen as the go to p2p networking stack**
-
-- js-libp2p releases get proper fanfare - @diasdavid
-- kick off the js.libp2p.io Welcome page - @diasdavid
-- Onboard 3 new full time contributors - @diasdavid
-
-Once these OKRs are finished, you will be able to find them on the [2018 Q3 libp2p OKRs Spreadsheet](https://docs.google.com/spreadsheets/d/1HTXfgR5FyPTFhsTkFPRThkeMvHvCgJOaAs7BSl_vQ_0/edit#gid=1241853194)
+Find the js-libp2p OKRs for 2018 Q2 at the [2018 Q3 libp2p OKRs Spreadsheet](https://docs.google.com/spreadsheets/d/1HTXfgR5FyPTFhsTkFPRThkeMvHvCgJOaAs7BSl_vQ_0/edit#gid=1241853194)
 
 ## Previous Quarters
 
--- a/OKR.md
+++ b/OKR.md
@@ -9,6 +9,7 @@ We try to frame our ongoing work using a process based on quarterly Objectives a
 - Delegated Peer Routing is completed - @jacobheun
 - Delegated Content Routing is completed - @jacobheun
 - DHT interops with go-libp2p - @vascosantos
+- NAT Traversal is implemented in js-libp2p (for at least TCP) - @jacobheun
 - IPNS works over PubSub (+ go-libp2p interop) - @vasco-santos
 - IPNS works over DHT (+ go-libp2p interop) - @vasco-santos
 
@@ -17,9 +18,9 @@ We try to frame our ongoing work using a process based on quarterly Objectives a
 - Error handling is done correctly and error codes are part of the API - @jacobheun
 - libp2p and libp2p-switch become full state machines - @jacobheun
 - libp2p has a network visualization tool - @vasco-santos
-- js-libp2p has a daemon with a RPC API and a client library
-- We get the RPC API standardised so that go-libp2p and rust-libp2p can implement it as well
-- Write the first set of interop tests that tests
+- js-libp2p has a daemon with a RPC API and a client library - @jacobheun
+- We get the RPC API standardised so that go-libp2p and rust-libp2p can implement it as well - @jacobheun
+- Write the first set of interop tests that tests - @jacobheun
 
 **libp2p is seen as the go to p2p networking stack**
 
--- a/OKR.md
+++ b/OKR.md
@@ -1,0 +1,19 @@
+# Quarterly Objectives and Key Results
+
+We try to frame our ongoing work using a process based on quarterly Objectives and Key Results (OKRs). Objectives reflect outcomes that are challenging, but realistic. Results are tangible and measurable.
+
+## 2018 Q3
+
+**Objective 1**
+- KR 1
+- KR 2 
+
+**Objective 2**
+- KR 1
+- KR 2 
+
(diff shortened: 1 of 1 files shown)


mgoelzer:
/ cc @jacobheun @vasco-santos @steverichmond

I want to try to connect [this tab for js-libp2p OKRs](https://docs.google.com/spreadsheets/d/1HTXfgR5FyPTFhsTkFPRThkeMvHvCgJOaAs7BSl_vQ_0/edit#gid=1241853194) with [this tab for overall libp2p project OKRs](https://docs.google.com/spreadsheets/d/1HTXfgR5FyPTFhsTkFPRThkeMvHvCgJOaAs7BSl_vQ_0/edit#gid=0).

By connect, I really just mean that generally the two should be aiming at the same goals.  We can do this in a bottoms-up way, where we start with the ongoing work by the JS libp2p team and better align the overall goals to fit those plans.  Or we can do it in a top-down way, i.e., start by getting to consensus on the overall libp2p goals and use that to set goals for js-libp2p.  Thoughts?

jacobheun:
@mgoelzer I think we are likely going to need to leverage both approaches. I see the direction of libp2p being heavily influenced by the community and the projects it integrates with. For example, we know that right now libp2p can be extremely difficult to troubleshoot errors. It takes up a lot of time doing so, not just for libp2p developers, but for its consumers as well. Making that easier to do as well as making the system more fault tolerant in general should be high priorities for us, because it will improve user experience and allow us to spend subsequent quarters focused more on feature development. 

Where I see the more top-down approach coming into play is through event connections, market research, PL goals and general community outreach. What types of things are libp2p/ipfs users or potential users looking for that we can put into place? What support does libp2p need to continue to be successful, and are we growing at the rate we need to?

I think a lot of these can tie together pretty well. (More demos of work being completed that can be shared with the community, interesting proof of concept work with new features, etc.)

jacobheun:
Here is a draft of a fairly large initiative I'd like to see for Q3. I think it's reasonable for this to be completed within 3 months.

## Q3 OKR Proposal: Libp2p stability
One of the recurring themes of Q2 has been issues with error handling and troubleshooting.

There is an issue in js-ipfs, [#1406](https://github.com/ipfs/js-ipfs/issues/1406), that discusses a roadmap for standardizing error codes across the ipfs ecosystem. I think this is going to be a huge improvement for diagnosing errors when they arise. In order to better leverage standardized errors and improve our ability to handle and diagnose them in the libp2p workflow, I think we should focus on making the following updates to libp2p modules.

### State Machines

#### Make libp2p a state machine, [#98](https://github.com/libp2p/js-libp2p/issues/98)
Along with allowing users to initiate libp2p transitions, rather than change state directly, this would also help to better diagnose issues. If we know what state/transition libp2p was in when an error occured, it should help us better understand how to replicate that issue. Right now we have to rely on stack traces (which are can currently be very unhelpful), or view the logs, provided someone was able to have them running when they hit an issue.

A clearly documented state machine, also has the benefit of better informing users of the underlying flows in libp2p. They'd be able to see all the available states, their transitions and what they transition to. As the libp2p connection workflow is a topic for the upcoming summit in Berlin, it's clear that this could be extremely helpful. I think the [connection proposal](https://github.com/libp2p/js-libp2p-switch/issues/24#issuecomment-386331717) in libp2p-switch is an example of how this could help clarify workflows, especially with additional comments.

#### Make libp2p-switch a state machine, including its connections, [#24](https://github.com/libp2p/js-libp2p-switch/issues/24)
There are two major things this lets us do.
1. It makes it easy to understand internal state and the state of connections. This allows us to better handle individual errors that occur for a given connection and translate its state accordingly. 
2. It makes it easier for libp2p and the connection manager to understand what's going on with switch, so they can make more informed decisions. Notification of state transitions, and connection transitions, along with logging via debug of those transitions should also make it easier for users to diagnose issues.

### Standard Errors

#### Standardize connection errors and add them to the ipfs initiative, [#1406](https://github.com/ipfs/js-ipfs/issues/1406)
This must be done to make the implementation successful. 

#### Implement in [interface-connection](https://github.com/libp2p/interface-connection) modules
By leveraging standard errors for all modules implementing the connection interface, the libp2p-switch connection state machines will be better able to respond to the errors appropriately. 
Fatal codes could terminate connections outright and not cause libp2p or switch to crash the node. Unstable connections could be reported up to libp2p and the connection manager to handle. The connection manager could make decisions on which connections to keep, and/or potentially add dialing backoffs for connections (network outage, a restarting peer, etc).

diasdavid:
All in for improving debuggability! The other P0 objective should be supporting js-ipfs getting Connectivity Magic, namely interopable and reliable DHT + IPNS.

vasco-santos:
@mgoelzer I agree that we should be synced between those 2 tabs. I agree with @jacobheun as the OKRs tab should represent the intended direction of the project as a whole, and GO and JS should follow them. Moreover, JS and GO implementations will probably have other aspects that are inherent to the current state of each project, and if those aspects are critical they should also be tackled. 

Accordingly, I would like to know if there is an [IMPLEMENTATION_STATUS.md](https://github.com/ipfs/ipfs/blob/master/IMPLEMENTATION_STATUS.md), as we have for ipfs.  I have looked for it, but I couldn’t find it and I think that this document could also be important for OKRs consideration.

Meanwhile, I have some suggestions for OKRs in the JS side that I consider to be important for the following quarter. I will describe them later in here.

diasdavid:
Hi all, thank you so much for writing these. I've a feel that they are really close to what we will be writing in the spreadsheet at the end.

As food for thought, I wrote up -- https://github.com/ipfs/pm/issues/654 -- to remind everyone that if we miss to understand our user and use case needs, we will only be able to see the needs that we see during the development.

If I were to name the most important objective is to nail down the DHT + IPNS implementations and respective interop. That will enable libp2p and IPFS in JS to achieve the connectivity magic that everyone expects from the project.

diasdavid:
@mgoelzer linking [my comment to the libp2p 2018 OKRs here for reference -- comment](https://docs.google.com/spreadsheets/d/1HTXfgR5FyPTFhsTkFPRThkeMvHvCgJOaAs7BSl_vQ_0/edit?pli=1&disco=AAAACAlhLC0)

OKR.md
```diff
@@ -0,0 +1,21 @@
+# Quarterly Objectives and Key Results
+
+We try to frame our ongoing work using a process based on quarterly Objectives and Key Results (OKRs). Objectives reflect outcomes that are challenging, but realistic. Results are tangible and measurable.
+
+## 2018 Q3
```
diasdavid: Let's continue discussing the OKRs over the markdown, taking inspiration from the notes already shared in this thread.

I've defined 2 main objectives so far (rewording is welcome if you see the need) for this quarter which capture the most requested and needed features and bug fixes.

Let's shoot to get the final list of KRs by Wednesday, final review on Thursday and ship it on Friday.

vasco-santos: @diasdavid  should we add `IPNS parity and interoperability with go`?

diasdavid: Absolutely @vasco-santos, good call!

I've updated the OKRs list and started assigning some based on previous work. Please review:

- Check if wording makes sense.
- Check if they are achievable this quarter.
- Check the ones you will want to take the lead.

vasco-santos: I believe that we can achieve this list!! 

I would like to take the lead on `DHT interops with go-libp2p`. 
By the way, is there any issue for the network visualization tool? I think I would like to go with it as well.

OKR.md
```diff
@@ -0,0 +1,31 @@
...
+
+**All pieces are there. Ship Peer & Content Routing + IPNS**
+
+- Delegated Peer Routing is completed 
+- Delegated Content Routing is completed  
+- DHT interops with go-libp2p
+- IPNS works over PubSub (+ go-libp2p interop) - @vascosantos
+- IPNS works over DHT (+ go-libp2p interop) - @vascosantos
```
vasco-santos: IPNS is currently implemented in `go-ipfs`, so I think we should have `go-ipfs interop`

OKR.md
achingbrain: Do you think it's worth having a 'technical debt' section?  There was something about speeding the tests up here a minute ago but it seems to have been removed.

Feature parity with the go API would also be a good thing to shoot for.

OKR.md
diasdavid: image

You mean anything different from what is already there?

OKR.md
diasdavid: @achingbrain you are confusing js-ipfs OKRs with js-libp2p OKRS. See https://github.com/ipfs/js-ipfs/pull/1409/files

OKR.md
vasco-santos: I mean:

```
IPNS works over PubSub (+ go-ipfs interop) - @vascosantos
IPNS works over DHT (+ go-ipfs interop) - @vascosantos
```

Since IPNS is implemented inside the scope of `js-ipfs` / `go-ipfs`

OKR.md
diasdavid: @vasco-santos awesome! Created an issue for you https://github.com/libp2p/js-libp2p/issues/215

diasdavid: With this, I assigned the remaining to @jacobheun, hope you agree. I do think that we need to find more stuff to keep @jacobheun busy, otherwise he will be done in the first couple of weeks :D

vasco-santos: cool, thanks @diasdavid ! Then we sync about the tool 🙂

OKR.md
```diff
@@ -12,11 +12,14 @@ We try to frame our ongoing work using a process based on quarterly Objectives a
...
 
-**libp2p debugging is not painful**
+**libp2p debugging and interop testing is not painful**
 
 - Error handling is done correctly and error codes are part of the API - @jacobheun
 - libp2p and libp2p-switch become full state machines - @jacobheun
 - libp2p has a network visualization tool - @vascosantos
+- js-libp2p has a daemon with a RPC API and a client library
+- We get the RPC API standardised so that go-libp2p and rust-libp2p can implement it as well
+- Write the first set of interop tests that tests
```
diasdavid: @jacobheun how do you feel about adding and taking these. It would be pretty sweet and we would have the chance to model a lot of the things with regards to interop testing :)

OKR.md
achingbrain: > you are confusing js-ipfs OKRs with js-libp2p OKRS

Oops, yes.  I've added the comment there.

OKR.md
```diff
@@ -0,0 +1,34 @@
...
+- Delegated Content Routing is completed - @jacobheun
+- DHT interops with go-libp2p - @vascosantos
+- IPNS works over PubSub (+ go-libp2p interop) - @vascosantos
+- IPNS works over DHT (+ go-libp2p interop) - @vascosantos
+
+**libp2p debugging and interop testing is not painful**
+
+- Error handling is done correctly and error codes are part of the API - @jacobheun
+- libp2p and libp2p-switch become full state machines - @jacobheun
+- libp2p has a network visualization tool - @vascosantos
```
vasco-santos: @diasdavid you missed a hyphen on my handle 😛

OKR.md
jacobheun: @diasdavid haha, yes I'm good with taking the rest. I know NAT traversal was part of q2 but I'd like to make sure that gets done, as I think it's important to making libp2p the painless go to in p2p.

diasdavid: rad!! And you are right, let's add NAT back in.

diasdavid:
All right, I think we are done here! \o/ @jacobheun, @vasco-santos, if you agree, I'll move these to the spreadsheet, assign priorities and wrap up this PR. 👍 this comment for YES!

diasdavid:
image

[OKRs for 2018 Q3 are set!](https://docs.google.com/spreadsheets/d/1HTXfgR5FyPTFhsTkFPRThkeMvHvCgJOaAs7BSl_vQ_0/edit#gid=1241853194) Godspeed team! And may the muxers be with you ⚡️

mgoelzer:
> Accordingly, I would like to know if there is an IMPLEMENTATION_STATUS.md

Yes, we're working on something similar for the project over here https://github.com/libp2p/libp2p/pull/39.  The idea would be to have our main KPI (meaning the thing we track long term, beyond just any single quarter) be %age completion of that list, which could grow over time.  The list itself needs more cleanup -- it was copied out of https://github.com/ipfs/ipfs/blob/master/ROADMAP-TO-1.0.0.md (just the libp2p parts) and needs some further cleanup.  You can see some comments from @diasdavid about parts that I've left out, and I also need to improve it's formatting.

diasdavid:
@mgoelzer `IMPLEMENTATION_STATUS.md` is a different doc from ROADMAP, check https://github.com/ipfs/ipfs/blob/master/IMPLEMENTATION_STATUS.md

mgoelzer:
@diasdavid Yes, right, I'm just pointing out that it's a similar type of document (list of things with emojis to indicate completion status).

PR #209, opened 2018-06-28
feat: enable peer discovery modules by default

alanshaw:
This PR will enable any provided peer discovery modules by default if no configuration for the module is supplied/needed.

As before, modules can be explicitly disabled or enabled by passing config.

**This enables pre-configured modules (instances) to be passed and enabled without them having to have a `tag` and an unused config section.**

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -2,40 +2,19 @@
 
 const Joi = require('joi')
 
-const schema = Joi.object({
+const ModuleSchema = Joi.alternatives().try(Joi.func(), Joi.object())
+
+const OptionsSchema = Joi.object({
   // TODO: create proper validators for the generics
   connectionManager: Joi.object(),
   peerInfo: Joi.object().required(),
   peerBook: Joi.object(),
   modules: Joi.object().keys({
-    transport: Joi.array().items(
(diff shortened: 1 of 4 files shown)


alanshaw:
@jacobheun I don't think that peer discovery module validation code is working properly 😟 . I just found problems with the test for it and fixed them and it fails now:

https://github.com/libp2p/js-libp2p/pull/209/commits/f753d61c59661ef339f96c750bca7724a6749d64

I'm not changing any defaults, just allowing modules that don't require any config or have already been configured to still be registered and started. IMHO it's unnecessary to require a developer to pass a preconfigured module instance and also have to set a config value with `{ enabled: true }` somewhere else in the options. I think it is a reasonable expectation that if I pass a module and say nothing about whether it is enabled or disabled then I want it to be enabled.

I don't think we should validate config for each of the discovery modules - that's outside of the scope of _this_ module. We can't know what all the config options are and keep that info up to date. All we could meaningfully do would be to validate a boolean `enabled` property, which I think should just be optional anyway - hence this PR.

alanshaw:
@jacobheun how about now? Is better 😛 ?

PR #210, opened 2018-06-28
Add ability to pass a connection protector for private networking

jacobheun:
This adds the ability to specify a connection protector (private networking) via the new libp2p configuration.  Once https://github.com/libp2p/js-libp2p-pnet/pull/1 is released, we will be able to add that as a protector, such as through js-ipfs.

This will require an update to the libp2p-switch version once [js-libp2p-switch#266](https://github.com/libp2p/js-libp2p-switch/pull/266) is released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ const SECIO = require('libp2p-secio')
 const MulticastDNS = require('libp2p-mdns')
 const DHT = require('libp2p-kad-dht')
 const defaultsDeep = require('@nodeutils/defaults-deep')
+const Protector = require('libp2p-pnet')
 
 class Node extends libp2p {
   constructor (_peerInfo, _peerBook, _options) {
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -135,6 +136,7 @@ class Node extends libp2p {
         connEncryption: [
           SECIO
         ],
+        connProtector: new Protector(/*protector specific opts*/),
```
diasdavid: Do we want to have multiple Protectors or just just have it inside and enable it as a config value?

jacobheun: The implementation is based off of go which just supports a single, configurable protector. Users will be able to supply any protector they want, so it's extendable in the future as more protectors are created. In theory we could turn it into a list of protectors but I can't think of a use case where you'd want to do that as multiple levels of encryption on top of secio is really expensive. 

If we produce more "intelligent" protectors in the future, those would be easy to swap out.

PR #221, opened 2018-07-13
chore: update deps

diasdavid:


diasdavid:
The new libp2p-switch is causing (or revealed):

```
  1) .contentRouting le ring nodeA.contentRouting.provide:

     Error: stream ended with:0 but wanted:1

      at drain (node_modules/pull-reader/index.js:43:26)

      at node_modules/pull-reader/index.js:63:18

      at node_modules/pull-reader/index.js:114:13

      at node_modules/pull-reader/index.js:114:13

      at onNext (node_modules/pull-catch/index.js:17:17)

      at drain (node_modules/stream-to-pull-stream/index.js:141:18)

      at Channel.<anonymous> (node_modules/stream-to-pull-stream/index.js:158:5)

      at endReadableNT (node_modules/readable-stream/lib/_stream_readable.js:1010:12)

      at process._tickCallback (internal/process/next_tick.js:63:19)
```

jacobheun:
The pull-reader update exposed a bug in the dht where were were reading on write only responses. The dht is still working properly, but since we're getting no data when add providers, which is expected, we pull-reader errors because we tried to read a byte. 

https://github.com/libp2p/js-libp2p-kad-dht/pull/30

PR #225, opened 2018-07-19
fix: start and stop connection manager with libp2p

### LINKED ISSUE #224: Connection manager is not started
I can't see where the connection manager is started! It is instantiated but `start` is never called 😟 

https://github.com/libp2p/js-libp2p/blob/d9059dbad9b33e2c1434b5c13ff500ed34aa348f/src/index.js#L45

refs https://github.com/ipfs/js-ipfs/issues/1435#issuecomment-406113817

jacobheun:
Resolves #224 

This will start the connection manager on `libp2p.start` and stop it on `libp2p.stop`. Previously users would need to start and stop it themselves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,9 @@ class Node extends libp2p {
             active: false
           }
         },
+        dht: {
+          kBucketSize: 20
+        },
         // Enable/Disable Experimental features
         EXPERIMENTAL: {               // Experimental features ("behind a flag")
           pubsub: false,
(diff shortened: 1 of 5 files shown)


jacobheun:
@diasdavid added a test to verify subcomponents are being started and stopped along with libp2p.

jacobheun:
npm is having a bad day, I will check back tomorrow and re-run the tests when npm is feeling better.

Issue #120, opened 2017-09-04
Delegated Routing


daviddias:
> Following https://github.com/ipfs/pm/pull/492

Tasks:
- [x] Expose findPeers and findProovs in the Gateway
- [x] Create module that exposes:
  - [x] Peer Routing - https://github.com/libp2p/js-libp2p-delegated-peer-routing
  - [x] Content Routing - https://github.com/libp2p/js-libp2p-delegated-content-routing

diasdavid:
@lgierth any update on the first task of the list?

diasdavid:
@lgierth mind updating this issue with what we discussed and the latest developments? Thank you!

lgierth:
Here are my notes:

- expose dht/findpeer, dht/findprovs, swarm/connect, refs commands
  - https://github.com/ipfs/go-ipfs/pull/4595
- implement dht/putpeer command
- issue for "awesome delegated routing" endeavour
- update libp2p interfaces: peer-routing and content-routing
- comment on js-ipfs relay plans
  - announce /p2p-circuit addrs
  - configure static list of ~20 well-known relay nodes, pick 2
- debug /wss connection death bug
  - check out js-ipfs bootstrapper tests

diasdavid:
> A couple more notes.

The Peer Routing flows will use the findPeer call directly. While the Content Routing flow will be:
- for providing:
  - Ask the delegated peer to fetch the content
  - Ask the delegated peer to become a provider (which happens by fetching the content in the first place)
- for finding providers
  - Ask the delegated peer to tell us who is providing
  - Ask the delegated peer to fetch that data
  - Fetch the data from the delegated peer.

RangerMauve:
Where are the specs for this functionality?

dignifiedquire:
Implementations are being done here:

- https://github.com/libp2p/js-libp2p-delegated-peer-routing/pull/1
- https://github.com/libp2p/js-libp2p-delegated-content-routing/pull/1

jacobheun:
Both libp2p/js-libp2p-delegated-peer-routing and libp2p/js-libp2p-delegated-content-routing are about ready for release. I've got a local branch I am working on to integrate those into libp2p. The biggest issue right now is that https://ipfs.io is not working for those two.

I will make sure the local tests aren't hitting external endpoints, but the default peer and content routing won't work until ipfs.io is fixed.

[...]

PR #242, opened 2018-08-22
Delegated Peer and Content Routing

jacobheun:
This is ready!

This PR adds the ability to configure delegate routing modules, https://github.com/libp2p/js-libp2p/issues/120. 

I've included an example of the delegated routing working.  

**Note**: until https://github.com/ipfs/go-ipfs/pull/4595 is released, running a compatible local node for the example requires building go-ipfs from the `feat/delegated-routing` branch and running that locally.  This will also not work in production until a Gateway node is deployed with delegated routing support: https://github.com/ipfs/infrastructure/issues/426.

This also provides libp2p users with the ability to create and supply their own peer and content routing modules.


src/index.js
```diff
@@ -99,8 +99,9 @@ class Node extends EventEmitter {
     }
 
     // Attach remaining APIs
-    this.peerRouting = peerRouting(this)
-    this.contentRouting = contentRouting(this)
+    // If peer or content routing modules have been provided, use those, otherwise use the dht
+    this.peerRouting = this._modules.peerRouting || peerRouting(this)
+    this.contentRouting = this._modules.contentRouting || contentRouting(this)
```
jacobheun: @diasdavid my original thought was to either use the provided peer and content routing modules, or the dht. After finishing up the core of the work and thinking more about the example and usage I feel like it would be better to use both if they are enabled. The thought being that leveraging a delegate is a good way to bootstrap a node into the network while it grows its own network, but that overtime its network would have a lot of the information. We could also change peer and content routing to allow for multiple modules and in the future potentially introduce some priority logic for ordering the calls to prevent flooding the network, if that becomes an issue.

diasdavid: Think of it as a Stream Muxer or a Transport, you can use one, many, all the same time, disable some, you name it :)

jacobheun:
Delegated routing is ready for review! The IPFS updates aren't finished or deployed yet, but this is ready for when it is!

cc: @lgierth

src/content-routing.js
```diff
@@ -1,20 +1,83 @@
...
+     */
+    findProviders: (key, options, callback) => {
+      if (routers.length === 0) {
+        return callback(new Error('No content routers available'))
+      }
+
+      if (typeof options === 'function') {
+        callback = options
+        options = {}
+      } else if (typeof options === 'number') {
```
vasco-santos: Can you just add a comment here for future deprecation?

src/content-routing.js
```diff
@@ -1,20 +1,83 @@
...
-      node._dht.findProviders(key, timeout, callback)
+      const tasks = routers.map((router) => {
+        return (cb) => router.findProviders(key, options, (err, results) => {
+          if (err) {
+            return cb(err)
+          }
+
+          // If we don't have any results, we need to provide an error to keep trying
+          if (!results || Object.keys(results).length === 0) {
+            return cb(Object.assign(new Error('not found'), {
```
vasco-santos: use `err-code` module here

src/content-routing.js
```diff
@@ -1,20 +1,83 @@
...
+     *
+     * @param {CID} key The CID key of the content to find
+     * @param {object} options
+     * @param {number} options.maxTimeout How long the query should run
+     * @param {function(Error, Result<Array>)} callback
+     * @returns {void}
+     */
+    findProviders: (key, options, callback) => {
+      if (routers.length === 0) {
+        return callback(new Error('No content routers available'))
```
vasco-santos: add `err-code` here?

src/content-routing.js
```diff
@@ -1,20 +1,83 @@
...
+     * Once a content router succeeds, iteration will stop.
+     *
+     * @param {CID} key The CID key of the content to find
+     * @param {object} options
+     * @param {number} options.maxTimeout How long the query should run
+     * @param {function(Error, Result<Array>)} callback
+     * @returns {void}
+     */
+    findProviders: (key, options, callback) => {
+      if (routers.length === 0) {
```
vasco-santos: I would simply use 

```js
if (!routers.length)
```

But it is a detail. You have this in several places if you agree please change them all, otherwise leave like this

src/content-routing.js
```diff
@@ -1,20 +1,83 @@
...
+     *
+     * @param {CID} key The CID key of the content to find
+     * @param {function(Error)} callback
+     * @returns {void}
+     */
     provide: (key, callback) => {
-      if (!node._dht) {
-        return callback(new Error('DHT is not available'))
+      if (routers.length === 0) {
+        return callback(new Error('No content routers available'))
```
vasco-santos: `err-code`

src/peer-routing.js
```diff
@@ -1,13 +1,59 @@
...
+     *
+     * @param {String} id The id of the peer to find
+     * @param {object} options
+     * @param {number} options.maxTimeout How long the query should run
+     * @param {function(Error, Result<Array>)} callback
+     * @returns {void}
+     */
+    findPeer: (id, options, callback) => {
+      if (routers.length === 0) {
+        return callback(new Error('No peer routers available'))
```
vasco-santos: `err-code`

src/peer-routing.js
```diff
@@ -1,13 +1,59 @@
...
+
+      const tasks = routers.map((router) => {
+        return (cb) => router.findPeer(id, options, (err, result) => {
+          if (err) {
+            return cb(err)
+          }
+
+          // If we don't have a result, we need to provide an error to keep trying
+          if (!result || Object.keys(result).length === 0) {
+            return cb(Object.assign(new Error('not found'), {
```
vasco-santos: `err-code`

src/peer-routing.js
```diff
@@ -1,13 +1,59 @@
...
+
+          cb(null, result)
+        })
+      })
+
+      tryEach(tasks, (err, results) => {
+        if (err && err.code !== 'NOT_FOUND') {
+          return callback(err)
+        }
+        results = results || null
```
vasco-santos: shouldn't this be?

```js
results = results || []
```

test/content-routing.node.js
```diff
@@ -7,87 +7,363 @@ const chai = require('chai')
...
+      it('let kbucket get filled', (done) => {
+        setTimeout(() => done(), 250)
+      })
+
+      it('nodeA.contentRouting.provide', (done) => {
+        nodeA.contentRouting.provide(cid, done)
+      })
+
+      it('nodeE.contentRouting.findProviders for existing record', (done) => {
+        nodeE.contentRouting.findProviders(cid, 5000, (err, providers) => {
```
vasco-santos: Use the`maxTimeout` instead of `5000`, as we intend to deprecate it in the future.

test/content-routing.node.js
```diff
@@ -7,87 +7,363 @@ const chai = require('chai')
...
+          expect(err).to.not.exist()
+          expect(providers).to.have.length.above(0)
+          done()
+        })
+      })
+
+      it('nodeC.contentRouting.findProviders for non existing record (timeout)', (done) => {
+        const cid = new CID('QmTp9VkYvnHyrqKQuFPiuZkiX9gPcqj6x5LJ1rmWuSnnnn')
+
+        nodeE.contentRouting.findProviders(cid, 5000, (err, providers) => {
```
vasco-santos: Use the`maxTimeout` instead of `5000`, as we intend to deprecate it in the future.

test/content-routing.node.js
```diff
@@ -7,87 +7,363 @@ const chai = require('chai')
...
+        const dhtStub = sinon.stub(nodeA._dht, 'findProviders').callsArgWith(2, null, results)
+        const delegateStub = sinon.stub(delegate, 'findProviders').throws(() => {
+          return new Error('the delegate should not have been called')
+        })
 
-      nodeE.contentRouting.findProviders(cid, 5000, (err, providers) => {
-        expect(err).to.not.exist()
-        expect(providers).to.have.length(0)
-        done()
+        nodeA.contentRouting.findProviders('a cid', 5000, (err, results) => {
```
vasco-santos: Use the`maxTimeout` instead of `5000`, as we intend to deprecate it in the future.

test/content-routing.node.js
```diff
@@ -7,87 +7,363 @@ const chai = require('chai')
...
+          done()
+        })
+      })
+
+      it('should use the delegate if the dht fails to find providers', (done) => {
+        const results = [true]
+        const dhtStub = sinon.stub(nodeA._dht, 'findProviders').callsArgWith(2, null, [])
+        const delegateStub = sinon.stub(delegate, 'findProviders').callsArgWith(2, null, results)
+
+        nodeA.contentRouting.findProviders('a cid', 5000, (err, results) => {
```
vasco-santos: Use the`maxTimeout` instead of `5000`, as we intend to deprecate it in the future.

src/content-routing.js
jacobheun: Good catch, I completely forgot to switch that over!

jacobheun:
@vasco-santos I pushed updates from your feedback. I also needed to re-work some of the code in the docs to fix linting there. The logic is unchanged though.

vasco-santos:
Looks good to me!

PR #248, opened 2018-09-19
fix: add maxtimeout to dht get

vasco-santos:
In the context of having the `DHT` enabled by default in `js-ipfs`,  several problems were identified on [js-ipfs/856#issuecomment-422495198](https://github.com/ipfs/js-ipfs/pull/856#issuecomment-422495198). In this case, `js-libp2p` does not allow the definition of a `maxTimeout`.

- [js-ipfs](https://github.com/ipfs/js-ipfs/pull/856/files#diff-e57a164358205eaa436f8fbf8430ed39R36) sends it
- [js-libp2p/dht.js#L12](https://github.com/libp2p/js-libp2p/blob/master/src/dht.js#L12) receives the `maxTimeout` and the `callback` is lost.
- [js-libp2p-kad-dhtindex.js#L186](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/index.js#L186) receives it
- This will result in loosing the callback on the way.

NOTE: There are no tests for the DHT in `js-ipfs`, is there any reason?


vasco-santos:
hey @jacobheun 

I updated the PR with your feedback 🙂

jacobheun:
Woot, looks good!

src/dht.js
```diff
@@ -9,19 +9,29 @@ module.exports = (node) => {
 
       node._dht.put(key, value, callback)
     },
-    get: (key, callback) => {
+    get: (key, maxTimeout, callback) => {
```
diasdavid: things like timeouts should be in a options object rather than a arg otherwise you will fall into a future of having a huge array of options. @vasco-santos @jacobheun

vasco-santos: You are right! I will create a new PR for changing it here, as well as for `js-libp2p-kad-dht` and will change `js-ipfs` too.

PR #251, opened 2018-09-24
fix: start kad dht random walk

vasco-santos:
Currently the DHT's `randomWalk` was not being started. Accordingly, a peer did not find other peers in the network unless we connect them manually.

Needs:

- [x] [js-libp2p-kad-dht#42](https://github.com/libp2p/js-libp2p-kad-dht/pull/42)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -265,7 +265,9 @@ class Node extends EventEmitter {
       },
       (cb) => {
         if (this._dht) {
-          return this._dht.stop(cb)
+          return this._dht.randomWalk.stop(() => {
+            this._dht.stop(cb)
+          })
         }
         cb()
       },
--- a/src/index.js
+++ b/src/index.js
@@ -85,8 +85,11 @@ class Node extends EventEmitter {
     // dht provided components (peerRouting, contentRouting, dht)
     if (this._config.EXPERIMENTAL.dht) {
       const DHT = this._modules.dht
+      const enabledDiscovery = this._config.dht.enabledDiscovery !== false
+
       this._dht = new DHT(this._switch, {
         kBucketSize: this._config.dht.kBucketSize || 20,
+        enabledDiscovery,
         // TODO make datastore an option of libp2p itself so
         // that other things can use it as well
         datastore: dht.datastore
@@ -203,10 +206,7 @@ class Node extends EventEmitter {
         // have to set started here because DHT requires libp2p is already started
         this._isStarted = true
         if (this._dht) {
-          this._dht.start(() => {
-            this._dht.randomWalk.start()
-            cb()
-          })
+          this._dht.start(cb)
         } else {
           cb()
         }
@@ -265,9 +265,7 @@ class Node extends EventEmitter {
       },
       (cb) => {
         if (this._dht) {
-          return this._dht.randomWalk.stop(() => {
-            this._dht.stop(cb)
-          })
+          return this._dht.stop(cb)
         }
         cb()
       },
--- a/src/index.js
+++ b/src/index.js
@@ -203,7 +203,10 @@ class Node extends EventEmitter {
         // have to set started here because DHT requires libp2p is already started
         this._isStarted = true
         if (this._dht) {
-          this._dht.start(cb)
+          this._dht.start(() => {
+            this._dht.randomWalk.start()
+            cb()
+          })
         } else {
           cb()
         }
--- a/README.md
+++ b/README.md
@@ -162,7 +162,8 @@ class Node extends libp2p {
           }
         },
         dht: {
-          kBucketSize: 20
+          kBucketSize: 20,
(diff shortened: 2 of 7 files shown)


src/index.js
```diff
@@ -203,7 +203,10 @@ class Node extends EventEmitter {
         // have to set started here because DHT requires libp2p is already started
         this._isStarted = true
         if (this._dht) {
-          this._dht.start(cb)
+          this._dht.start(() => {
+            this._dht.randomWalk.start()
+            cb()
+          })
```
diasdavid: Also make sure to stop it. Add tests for this.

pgte: Probably not the place for this, but:
I find it weird that we have to call something that is so kademlia-specific. Other DHT solutions may not have this concept of a random walk. IMO, _dht.start() should be enough for the DHT to do all the necessary things for it to work..

vasco-santos: @pgte at first I was starting it into the `libp2p-kad-dht`. However, after @diasdavid 's feedback and reading the [docs](https://github.com/libp2p/js-libp2p-kad-dht#peer-discovery), I added it here, as it should probably be started with the other `Peer Discovery Service`.

vasco-santos: More context also in the [js-libp2p-kad-dht#42](https://github.com/libp2p/js-libp2p-kad-dht/pull/42)

vasco-santos: But I agree in the part of loosing the plugability of using a completelly different DHT! Let's wait for @diasdavid feedback

jacobheun: If the intention is for random walk to behave as a peer discovery service, so that users can choose to add it or not, it should be passed in via the peerDiscovery array configuration for libp2p. That way it can be started/stopped with all other discovery services. 

If it's core to the dht working properly, it should be started when the dht is (inside the dht).

vasco-santos: I am more inclined to have this started in the `DHT` after @pgte and @jacobheun options. What about yout @diasdavid ?

diasdavid: It should be opt in. Users need to have a way to decide if they want their node to issue random queries to the DHT.

A DHT is a:
- Peer Routing
- Content Routing
- Peer Discovery

And we have to enable opt-in by config.

jacobheun: The DHT logic will likely need to get refactored so that, like transports, users can provide an instance or a function. This way if they want to enable the discovery service they'd create the DHT instance and pass `dht.discovery`, or similar, to the `peerDiscovery` array.

jacobheun: I think we should wait until that is configurable to add it in. Having it autostart now and then change in the future to needing to be added via peerDiscovery configuration could end up being confusing to users. It would be great for users to start being able to see an increase in connected peers, but releasing an update later that breaks this would probably result in some frustration. Since it's not currently running by default, I would recommend we take the extra time to get if configurable first.

vasco-santos: I understand your point @jacobheun . But I am still not convinced that in the `libp2p` perspective we will want to have the `random-walk` disabled by default. Otherwise, we would need to connect manually through a chain of peers, in order to be able to query the all network, which seems strange to me. 

However, I can start looking at its modification in the meanwhile.

vasco-santos: After discussing this offline with @jacobheun we believe that the best way is to allow the configuration of the `randomWalk`, but turn it on by default. That way restricted nodes will be able to turn it off.

At first, we tried to understand how we could have `randomWalk` the same way as the other `peer-discovery` work from the interface perspective, but the other discovery services are more targeted. In the `randomWalk`, we will not find a specific peer, but we will ask everyone we know, and everyone they now, if they know a random peer, which is a way of meeting the others, but it is really tight with the `DHT` behavior. 

We ended up thinking, how can we understand that a new peer we just met was a result of the `randomWalk`, or any other event, since the base codebase for this is shared between `randomWalk` and the `DHT` itself. All things considered, a considerable effort would need to be done, in order to get an interface that we can be sure to have an exactly equal behavior to what we expect, as the other `peer-discovery` modules. Accordingly, should we announce every single peer that we find during the walk? or should we announce only the random one, if we find it? If we go through the second one, we will end up having a different behaviour compared to the other `peer-discovery` modules.

Considering [js-libp2p#242/files#diff-b0f6b211af5c98be53893311b36c8d1fR9](https://github.com/libp2p/js-libp2p/pull/242/files#diff-b0f6b211af5c98be53893311b36c8d1fR9), the peer/content routing are separated from the DHT by design,  the `randomWalk` in `peerDiscovery` do not seem that necessary.

Thus, we think it is better to go by simply enabling by default the `randomWalk` and allow an easy way through `libp2p` to disable it.

vasco-santos:
I am having problems with the tests for `websockets stress tests`.

The same tests are failing in master.

jacobheun:
@vasco-santos I am working on that now, there's a resiliency issue with the latest libp2p-mplex. I'll have a PR shortly

PR #256, opened 2018-10-02
chore: change dependency name from libp2p-railing to libp2p-bootstrap

vasco-santos:
The package `libp2p-railing` was [previously renamed](https://github.com/libp2p/js-libp2p-bootstrap/pull/79) to `libp2p-bootstrap`, but this change was not propagated to here.


vasco-santos:
Updated! Did not add those files to commit 😓

hubertokf:
Should update it on examples: 
https://github.com/libp2p/js-libp2p/tree/master/examples/discovery-mechanisms

PR #257, opened 2018-10-05
Make libp2p a state machine

### LINKED ISSUE #98: Be a state machine. Just like js-ipfs :)

jacobheun:
Exposes the new `dialFSM` calls from [switch#278](https://github.com/libp2p/js-libp2p-switch/pull/278), giving libp2p users the option to listen to more detailed information from dials via returned ConnectionFSM's.

This also adds state machine logic to libp2p and ensures restarts are idempotent.

Ideally `dialFSM` would return the state machines instead of via callback, but checking the peer data may require an async call to the dht, so the callback logic stays the same.

TODO:
- [x] Update readme with new methods
- [x] Update switch dependency once it's released

Depends on https://github.com/libp2p/js-libp2p-switch/pull/278
Resolves #98


src/index.js
```diff
@@ -110,15 +114,183 @@ class Node extends EventEmitter {
...
+      log.error(err)
+      this.emit('error', err)
+    })
+  }
+
+  /**
+   * Starts the libp2p node and all sub services
+   *
+   * @param {function(Error)} callback
+   * @fires Node#start
```
vasco-santos: improper usage of jsdocs `@fires`

src/index.js
```diff
@@ -110,15 +114,183 @@ class Node extends EventEmitter {
...
+  start (callback = () => {}) {
+    this.once('start', callback)
+    this.state('start')
+  }
+
+  /**
+   * Stop the libp2p node by closing its listeners and open connections
+   *
+   * @param {function(Error)} callback
+   * @fires Node#stop
```
vasco-santos: improper usage of jsdocs `@fires`

jacobheun:
@vasco-santos I updated callback language in the readme everywhere to reflect the change. I also moved all `@fired` jsdocs to the class level and made sure all events are documented there. Thanks for catching those!

PR #273, opened 2018-11-02
[WIP] feat: allow to add a value with provide

vasco-santos:
Aiming to allow the possibility to add a value while notifying the network that we provide it.

In the context of enabling the following test [interface-ipfs-core/js/src/dht/get.js#L48-L64](https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/dht/get.js#L48-L64)

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,10 +262,10 @@ Required keys in the `options` object:
 - `options`: object of options
 - `options.maxTimeout`: Number milliseconds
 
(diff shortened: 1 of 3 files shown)


vasco-santos:
Closed in favour of [js-ipfs#856#issuecomment-435507070](https://github.com/ipfs/js-ipfs/pull/856#issuecomment-435507070)

PR #283, opened 2018-11-12
chore: add maxNumProviders to findprovs

vasco-santos:
In the context of [libp2p/js-libp2p-kad-dht#54](https://github.com/libp2p/js-libp2p-kad-dht/issues/54). 

Needs:

- [x] [libp2p/js-libp2p-kad-dht#55](https://github.com/libp2p/js-libp2p-kad-dht/pull/55)


vasco-santos:
Thanks, I feel that feat is better in fact!

PR #286, opened 2018-11-23
fix: validate dht config

vasco-santos:
In the context of the issue reported [ipfs/js-ipfs#1657#issuecomment-440758206](https://github.com/ipfs/js-ipfs/issues/1657#issuecomment-440758206)


vasco-santos:
Yes, your solution sounds way better. Can you create the PR with it?

PR #288, opened 2018-11-29
feat: allow configurable validators and selectors to the dht

vasco-santos:
Allow configurable validators and selectors. Users of `libp2p` with `libp2p-kad-dht` must be able to set their own validators and selectors. 

For instance, this is important for `js-ipfs` to use `/ipns/` namespaces in the DHT

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -105,8 +105,8 @@ class Node extends EventEmitter {
         kBucketSize: this._config.dht.kBucketSize,
         enabledDiscovery,
         datastore: this.datastore,
-        validators: this._config.dht.validators || {},
-        selectors: this._config.dht.selectors || {}
+        validators: this._config.dht.validators,
+        selectors: this._config.dht.selectors
       })
     }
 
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -102,9 +102,11 @@ class Node extends EventEmitter {
       const enabledDiscovery = this._config.dht.enabledDiscovery !== false
 
       this._dht = new DHT(this._switch, {
-        kBucketSize: this._config.dht.kBucketSize || 20,
+        kBucketSize: this._config.dht.kBucketSize,
         enabledDiscovery,
-        datastore: this.datastore
+        datastore: this.datastore,
+        validators: this._config.dht.validators || {},
+        selectors: this._config.dht.selectors || {}
```
jacobheun: Is there any reason these can't just be null, instead of creating an empty object?

vasco-santos: No, I am also doing this in `kad-dht`, I will change it now! Thanks :)

PR #289, opened 2018-11-29
Fix/dht config default

jacobheun:
Replaces https://github.com/libp2p/js-libp2p/pull/286


jacobheun:
This is being done in #288 already, closing.

Issue #300, opened 2018-12-18
pubsub.unsubscribe should comply with interface-ipfs-core spec, since it links to it in the docs


noot:
- **Version**: 0.23.1
- **Platform**: Ubuntu 18.04
- **Subsystem**:

Type: Bug
Severity: high
Description:
Cannot unsubscribe from a pubsub topic after subscribing. Following the IPFS core API here: https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/PUBSUB.md

Steps to reproduce the error:
```
const libp2p = require("libp2p")
const TCP = require("libp2p-tcp")
const PeerInfo = require("peer-info")
const waterfall = require("async/waterfall")
const defaultsDeep = require("@nodeutils/defaults-deep")
const Mplex = require("libp2p-mplex")
const SECIO = require('libp2p-secio')
const Bootstrap = require('libp2p-bootstrap')

const bootstrappers = [
	'/ip4/127.0.0.1/tcp/43107/ipfs/Qmbm3gwdCZM9GQJg1fpsooZDJ6ma56AFPxqYup53NhvstB'
]

class MyBundle extends libp2p {
	constructor(_options) {
		const defaults = {
			modules: {
				transport: [TCP],
				streamMuxer: [Mplex],
				connEncryption: [SECIO],
				peerDiscovery: [Bootstrap]
			},
			config: {
				peerDiscovery: {
					bootstrap: {
						interval: 2000,
						enabled: true,
						list: bootstrappers
					},
				},
				EXPERIMENTAL: {
					pubsub: true
				}
			}
		}

		super(defaultsDeep(_options, defaults))
	}
}

function createNode(callback) {
	waterfall([
		(cb) => PeerInfo.create(cb),
		(peerInfo, cb) => {
			peerInfo.multiaddrs.add('/ip4/0.0.0.0/tcp/0')
			node = new MyBundle({peerInfo: peerInfo})
			node.start(cb)
		}
	], (err) => callback(err, node))
}

createNode((err, node) => {
	const topic = 'fruit-of-the-day'
	const receiveMsg = (msg) => console.log(msg.toString())

	node.pubsub.subscribe(topic, receiveMsg, (err) => {
	  if (err) {
	    return console.error(`failed to subscribe to ${topic}`, err)
	  }

	  console.log(`subscribed to ${topic}`)

	  // unsubscribe a second later
	  setTimeout(() => {
[...]

PR #302, opened 2018-12-20
add callback to pubsub.unsubscribe and test (#300)

noot:
As per the issue I opened: https://github.com/libp2p/js-libp2p/issues/300

Upon `pubsub.unsubscribe`, the node was successfully unsubscribing, however the callback was never being called, so there was no explicit way to confirm success.

I have changed `unsubscribe` as to call the callback, so it now complies with the interface-ipfs-core spec. I have also added a test.

Let me know if any changes are needed, thanks!

### DIFF
--- a/src/pubsub.js
+++ b/src/pubsub.js
@@ -44,7 +44,9 @@ module.exports = (node) => {
         floodSub.unsubscribe(topic)
       }
 
-      setImmediate(() => callback())
+      if (typeof callback === 'function') {
+        setImmediate(() => callback())
+      }
     },
 
     publish: (topic, data, callback) => {
--- a/src/pubsub.js
+++ b/src/pubsub.js
@@ -33,7 +33,7 @@ module.exports = (node) => {
       subscribe(callback)
     },
 
-    unsubscribe: (topic, handler) => {
+    unsubscribe: (topic, handler, callback) => {
       if (!node.isStarted() && !floodSub.started) {
         throw new Error(NOT_STARTED_YET)
       }
@@ -43,6 +43,8 @@ module.exports = (node) => {
       if (floodSub.listenerCount(topic) === 0) {
         floodSub.unsubscribe(topic)
       }
+
+      setImmediate(() => callback())
     },
 
     publish: (topic, data, callback) => {
--- a/test/pubsub.node.js
+++ b/test/pubsub.node.js
@@ -5,6 +5,7 @@
 
 const chai = require('chai')
 chai.use(require('dirty-chai'))
+chai.use(require('chai-checkmark'))
 const expect = chai.expect
 const parallel = require('async/parallel')
 const waterfall = require('async/waterfall')
@@ -53,6 +54,7 @@ function stopTwo (nodes, callback) {
 describe('.pubsub', () => {
   describe('.pubsub on (default)', (done) => {
     it('start two nodes and send one message, then unsubscribe', (done) => {
+      expect(4).checks(done)
       const data = Buffer.from('test')
       const handler = (msg, nodes, cb) => {
         expect(msg.data).to.eql(data)
@@ -67,9 +69,8 @@ describe('.pubsub', () => {
               expect(err).to.not.exist()
             })
             nodes[0].pubsub.unsubscribe('pubsub', handler, (err) => {
-              expect(err).to.not.exist()
-              console.log('\tunsubscribed!')
               done()
+              expect(err).to.not.exist()
             })
           })
         },
--- a/test/pubsub.node.js
+++ b/test/pubsub.node.js
@@ -52,21 +52,25 @@ function stopTwo (nodes, callback) {
 // TODO: consider if all or some of those should come here
 describe('.pubsub', () => {
   describe('.pubsub on (default)', (done) => {
-    it('start two nodes and send one message', (done) => {
+    it('start two nodes and send one message, then unsubscribe', (done) => {
+      const data = Buffer.from('test')
+      const handler = (msg, nodes, cb) => {
+        expect(msg.data).to.eql(data)
+        cb(null, nodes)
+      }
       waterfall([
         (cb) => startTwo(cb),
         (nodes, cb) => {
-          const data = Buffer.from('test')
-          nodes[0].pubsub.subscribe('pubsub',
-            (msg) => {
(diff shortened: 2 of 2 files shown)


noot:
Not sure how to fix the remaining failing tests, sorry.

src/pubsub.js
```diff
@@ -43,6 +43,8 @@ module.exports = (node) => {
       if (floodSub.listenerCount(topic) === 0) {
         floodSub.unsubscribe(topic)
       }
+
+      setImmediate(() => callback())
```
jacobheun: Since the callback is being added, even though it should have already exists, we should check that it is actually a function before invoking it.

test/pubsub.node.js
```diff
@@ -52,23 +52,26 @@ function stopTwo (nodes, callback) {
...
-                expect(err).to.not.exist()
-              }), 500)
-            }
-          )
+            }), 500)
+            setTimeout(() => nodes[0].pubsub.unsubscribe('pubsub', handler, (err) => {
+              expect(err).to.not.exist()
+              console.log('tunsubscribed!')
+              done()
+            }), 600)
```
jacobheun: We should get rid of the timeouts in this test. Subscribing -> publishing -> unsubscribing should be able to be executed in serial successfully without a timeout.

noot: done!

src/pubsub.js
noot: makes sense, I added a check

test/pubsub.node.js
```diff
@@ -52,23 +52,26 @@ function stopTwo (nodes, callback) {
...
-              setTimeout(() => nodes[1].pubsub.publish('pubsub', data, (err) => {
-                expect(err).to.not.exist()
-              }), 500)
-            }
-          )
+            })
+            nodes[0].pubsub.unsubscribe('pubsub', handler, (err) => {
+              expect(err).to.not.exist()
+              console.log('\tunsubscribed!')
+              done()
```
jacobheun: It's possible done could be called before we run the expect in the publish handler. It would be good to include `chai-checkmark` in this file and mark each of the expects.  That way we can do `expect(4).checks(done)` so we can ensure everything is verified before finishing the test.

Remove the console.log as well and I think this is good to go. Thank you for the PR!

noot: Sounds good! I've added chai-checkmark and the check to make sure all the expects happen, as well as moved done and got rid of the console.log.

noot: When I remove `done()` at line 72, I get `Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.` If we place `done()` at line 77, at the end of the waterfall, it works. I'm not sure if this is corrrectly allows for the expect check to happen though.

jacobheun:
Sorry for the delay, I've been away on holiday but am back now. The test done logic is still not happening correctly. Done is being called and is passed as a callback to waterfall, this isn't correct. The timeout you were hitting is probably because things aren't firing in the right order, or are being triggered early. I can take a deeper look at what the issue is later today.

noot:
@jacobheun No worries! Sounds good, thanks. Let me know what issues you find.

jacobheun:
There were some errors in the test, primarily some of the function arguments and `.mark()` not being added for chai-checkmark. The delay after the subscribe is also important, otherwise publish will likely be called before the peer is aware of the subscriber.

I think it was a bit hard to follow so I made some changes along with the fixes that hopefully make things a bit easier to read.

```js
    it('start two nodes and send one message, then unsubscribe', (done) => {
      // Check the final series error, and the publish handler
      expect(2).checks(done)

      let nodes
      const data = Buffer.from('test')
      const handler = (msg) => {
        // verify the data is correct and mark the expect
        expect(msg.data).to.eql(data).mark()
      }

      series([
        // Start the nodes
        (cb) => startTwo((err, _nodes) => {
          nodes = _nodes
          cb(err)
        }),
        // subscribe on the first
        (cb) => nodes[0].pubsub.subscribe('pubsub', handler, cb),
        // Wait a moment before publishing
        (cb) => setTimeout(cb, 500),
        // publish on the second
        (cb) => nodes[1].pubsub.publish('pubsub', data, cb),
        // unsubscribe on the first
        (cb) => nodes[0].pubsub.unsubscribe('pubsub', handler, cb),
        // Stop both nodes
        (cb) => stopTwo(nodes, cb)
      ], (err) => {
        // Verify there was no error, and mark the expect
        expect(err).to.not.exist().mark()
      })
    })
```

If that looks good and makes sense to you can you push the update?

noot:
@jacobheun That's a lot easier to follow, I think it makes sense.  I'll push the changes now

jacobheun:
I'm going to look into the windows ci failure. It's possible it's just flakey, but it needs to get fixed either way.

noot:
okay cool, let me know if I need to make any changes.

noot:
@jacobheun hey, any updates or changes needed?

jacobheun:
@noot not yet, I was away last week, hoping to take a look at it this week.

test/pubsub.node.js
```diff
@@ -52,26 +53,37 @@ function stopTwo (nodes, callback) {
...
+        (cb) => startTwo((err, _nodes) => {
+          nodes = _nodes
+          cb(err)
+        }),
+        // subscribe on the first
+        (cb) => nodes[0].pubsub.subscribe('pubsub', handler, cb),
+        // Wait a moment before publishing
+        (cb) => setTimeout(cb, 500),
+        // publish on the second
+        (cb) => nodes[1].pubsub.publish('pubsub', data, cb),
```
jacobheun: @noot sorry for the delay, it looks like the tests were just failing because windows needs some time before unsubscribing. This should fix it:
```suggestion
        (cb) => nodes[1].pubsub.publish('pubsub', data, cb),
        // Wait a moment before unsubscribing
        (cb) => setTimeout(cb, 500),
```

noot: I've made the changes!

jacobheun:
@noot thank you for the PR! I've squashed and merged this to master via 679d446daa0b43c5a1d77c8f300b34c32c678022 to fix the code and commitlint.

PR #315, opened 2019-02-01
Change from ipfs to p2p protocol

thomaseizinger:
https://github.com/multiformats/js-multiaddr/pull/76 changed the default protocol from ipfs to p2p.

js-multiaddr is a transitive dependency of peer-info, so in order to get this change, we had to bump the version of peer-info.

### DIFF
--- a/examples/libp2p-in-the-browser/1/package.json
+++ b/examples/libp2p-in-the-browser/1/package.json
@@ -23,7 +23,7 @@
     "libp2p-spdy": "~0.12.1",
     "libp2p-webrtc-star": "~0.15.3",
     "libp2p-websocket-star": "~0.8.1",
-    "libp2p-websockets": "^0.12.0",
+    "libp2p-websockets": "~0.12.0",
     "peer-info": "~0.15.1"
   }
 }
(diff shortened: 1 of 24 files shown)


thomaseizinger:
The same change should probably be applied to all the other examples as well. However, I realized that some of them don't have a `package.json`. Adding that in order to test every example would kind of out of scope :D 

Should I change the other examples as part of this PR as well even if I cannot fully test them?

examples/libp2p-in-the-browser/1/package.json
```diff
@@ -23,7 +23,7 @@
     "libp2p-spdy": "~0.12.1",
     "libp2p-webrtc-star": "~0.15.3",
     "libp2p-websocket-star": "~0.8.1",
-    "libp2p-websockets": "~0.12.0",
-    "peer-info": "~0.14.1"
+    "libp2p-websockets": "^0.12.0",
```
jacobheun: semver version match for libp2p-websockets should stay the same.
```suggestion
    "libp2p-websockets": "~0.12.0",
```

thomaseizinger:
I've addresses the points you mentioned, let's see if I broke anything :see_no_evil:

PR #317, opened 2019-02-04
Prepare config validation for new randomWalk config parameters

thomaseizinger:
As per: https://github.com/libp2p/js-libp2p-kad-dht/issues/76

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -34,7 +34,12 @@ const OptionsSchema = Joi.object({
     dht: Joi.object().keys({
       kBucketSize: Joi.number().default(20),
       enabled: Joi.boolean().default(true),
-      enabledDiscovery: Joi.boolean().default(true),
+      randomWalk: Joi.object().keys({
+        enabled: Joi.boolean().default(true),
+        queriesPerPeriod: Joi.number().default(1),
+        interval: Joi.number().default(30000),
+        timeout: Joi.number().default(10000),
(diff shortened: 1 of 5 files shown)


jacobheun:
randomWalk isn't getting picked up by the dht config. We should probably just update dht creation to take the whole config object. I'm thinking we should just do this:

```diff
diff --git a/src/index.js b/src/index.js
index 0350b69..b7b7adf 100644
--- a/src/index.js
+++ b/src/index.js
@@ -105,14 +105,10 @@ class Node extends EventEmitter {
     // dht provided components (peerRouting, contentRouting, dht)
     if (this._config.dht.enabled) {
       const DHT = this._modules.dht
-      const enabledDiscovery = this._config.dht.enabledDiscovery !== false

       this._dht = new DHT(this._switch, {
-        kBucketSize: this._config.dht.kBucketSize,
-        enabledDiscovery,
-        datastore: this.datastore,
-        validators: this._config.dht.validators,
-        selectors: this._config.dht.selectors
+        ...{ datastore: this.datastore },
+        ...this._config.dht
       })
     }
```

thomaseizinger:
Good find!

Your suggestion makes a lot of sense, will implement it that way :)

jacobheun:
I went ahead and fixed the linting and merged this via 91e60d4. Thanks for submitting this!

PR #321, opened 2019-02-06
PUBSUB: unsubscribe without handler reference

Isan-Rivkin:
Hi :-)  
following up on an issue in [interface-js-ipfs-core](https://github.com/ipfs/interface-js-ipfs-core/issues/436) I'm creating this PR. 
Also created PR in [js-ipfs-http-client](https://github.com/ipfs/js-ipfs-http-client/pull/940) accordingly. 
This would help in cases like mine where my implementation is not able to support references to a handler function.
I will add the final url in the README to the method documentation once I merge the final PR `to interface-js-ipfs-core` since its pointing nowhere now. 

**Edit:** 

Opened a [PR](https://github.com/ipfs/interface-js-ipfs-core/pull/437) on `interface-js-ipfs-core`

### DIFF
--- a/src/pubsub.js
+++ b/src/pubsub.js
@@ -8,7 +8,27 @@ module.exports = (node) => {
   const floodSub = new FloodSub(node)
 
   node._floodSub = floodSub
+  const _internalUnsubscribe = (input)=>{
+    let {topic,handler,callback} = input
 
+    if (!node.isStarted() && !floodSub.started) {
+      throw new Error(NOT_STARTED_YET)
+    }
+
+    if(handler){
+      floodSub.removeListener(topic, handler)
+    }else{
+      floodSub.removeAllListeners(topic)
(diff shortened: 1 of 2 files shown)


jacobheun:
Per https://github.com/ipfs/interface-js-ipfs-core/pull/437#issuecomment-461403871, I think we should just overload `unsubscribe`.

cc @vasco-santos

jacobheun:
Also, when you update, can you fix linting `npm run lint`?

Isan-Rivkin:
@jacobheun thanks, I commented in [interface-js-ipfs](https://github.com/ipfs/interface-js-ipfs-core/pull/437).
Will do the linting, thanks for the hint I forgot.

vasco-santos:
> I think we should just overload `unsubscribe`.

Subscribe this point

PR #322, opened 2019-02-13
chore: move to travis

hugomrdias:


jacobheun:
Windows still needs to be resolved, it's having issues finding the wrtc module.

daviddias:
@jacobheun did it work on Windows before?

jacobheun:
@daviddias ci has been green for quite a while on master. Did we spin jenkins down already? I was going to look at the specifics of the old windows tests, but it looks like those are gone now.

daviddias:
I advise that Windows tests are disabled and a new PR to enable them is open right away so that the libp2p team can benefit from Travis running the tests for now.

jacobheun:
@daviddias, yeah @hugomrdias and I were talking about this earlier. With jenkins inactive that's our best path forward. I'm doing some checking today and will do that if I cannot find a different solution in the next few hours.

hugomrdias:
you can do this instead
```yaml
matrix:
  fast_finish: true
  allow_failures:
    - os: windows
```

daviddias:
@jacobheun see ^^

jacobheun:
If we're trying to _actively_ fix the windows build it seems wasteful to have windows running at all as it's guaranteed to fail right now. I'd rather see a single, focused branch only running windows.

If we end up putting off windows until Travis becomes more robust there, then it think it makes sense because then you get "auto discovery" of newly passing windows builds anytime you push a branch or PR.

PR #326, opened 2019-02-18
test: add circuit browser test

jacobheun:
Browser tests weren't being run for circuit relay. This adds a basic echo test for two browser nodes over a nodejs node to give us better visibility to potential issues.

### DIFF
--- a/test/circuit-relay.browser.js
+++ b/test/circuit-relay.browser.js
@@ -1,0 +1,98 @@
+/* eslint-env mocha */
+'use strict'
+
+const chai = require('chai')
+chai.use(require('dirty-chai'))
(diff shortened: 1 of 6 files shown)


test/circuit-relay.browser.js
```diff
@@ -0,0 +1,97 @@
...
+const {
+  getPeerRelay
+} = require('./utils/constants')
+
+function setupNodeWithRelay (addrs, options) {
+  options = {
+    config: {
+      relay: {
+        enabled: true
+      }
```
vasco-santos: spread `options.config`here?

jacobheun:
Okay, that's been fixed.

PR #328, opened 2019-02-19
chore: add error codes to dht and pubsub errors

vasco-santos:
Added error codes to the dht and pubsub errors, as well as replaced `setImmediate` by `nextTick`.

In the context of [ipfs/js-ipfs#1879](https://github.com/ipfs/js-ipfs/pull/1879)

### DIFF
--- a/src/dht.js
+++ b/src/dht.js
@@ -3,13 +3,13 @@
 const nextTick = require('async/nextTick')
 const errCode = require('err-code')
 
+const { messages, codes } = require('./errors')
+
 module.exports = (node) => {
   return {
     put: (key, value, callback) => {
       if (!node._dht) {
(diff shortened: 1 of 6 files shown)


jacobheun:
It would be good to have the codes exported from a constants file so we don't risk miss typing them. Also, most of the nextTick logic could probably just have the anonymous function removed since it doesn't need any local references.
```js
nextTick(callback, errCode(new Error('DHT is not available'), 'ERR_DHT_DISABLED'))
```

vasco-santos:
Thanks @jacobheun 

That should be fixed

vasco-santos:
I added tests for all of those!

jacobheun:
@vasco-santos it looks like linting needs to be fixed in the tests.

PR #330, opened 2019-02-25
enhancement: allow configuration for transport modules

robertkiel:
Idea:

- Especially when using stateful transport modules, i. e. those based on techniques like WebRTC, it might be reasonable to give them the opportunity to take configuration parameters.
- Transport modules like WebRTC need to tunnel signalling messages through other peers. For that reason, they need to be able to dial peers by using other transport modules.

Implementation:

- instantiate transport modules with given configuration object
- pass internal switch instance to the newly created transport module instance

What's missing:

- let transport modules specify their own configuration parameters and extend `src/config.js` at startup appropriately


jacobheun:
Switch should really just get passed in with the options and [interface-transport](https://github.com/libp2p/interface-transport/tree/v0.3.7#creating-a-transport-instance) docs should get updated to document that that's the case, rather than being a secondary param.

```js
const options = { ...this._config[Transport.name], switch: this._switch }
new Transport(options)
```

Also, `Transport.name` is not safe with minification, so that will need to change.

robertkiel:
I think this is solved by b06ca1b 👍 

Thanks :-)

PR #348, opened 2019-04-04
fix: remove unneeded peerbook puts

jacobheun:
With https://github.com/libp2p/js-libp2p-kad-dht/pull/92 the dht test for peer discovery was out of date. I fixed it and updated the other tests to use lower intervals so we dont have to wait 10seconds for discovery to happen.

The circuit tests were also inaccurate so those were fixed.

Switch had a significant number of code added recently, so the bundle size threshold needed to be increased slightly.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -243,10 +243,7 @@ class Node extends EventEmitter {
     this._getPeerInfo(peer, (err, peerInfo) => {
       if (err) { return callback(err) }
 
-      this._switch.dial(peerInfo, protocol, (err, conn) => {
-        if (err) { return callback(err) }
-        callback(null, conn)
-      })
+      this._switch.dial(peerInfo, protocol, callback)
     })
   }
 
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -246,7 +245,6 @@ class Node extends EventEmitter {
 
       this._switch.dial(peerInfo, protocol, (err, conn) => {
```
vasco-santos: We can just pass the callback now. Lines below have to be removed

```suggestion
      this._switch.dial(peerInfo, protocol, callback)
```

src/index.js
```diff
@@ -275,9 +273,6 @@ class Node extends EventEmitter {
       if (err) { return callback(err) }
 
       this._switch.dialFSM(peerInfo, protocol, (err, connFSM) => {
```
```suggestion
      this._switch.dialFSM(peerInfo, protocol, callback)
```

src/index.js
jacobheun: Ah yeah, good call! I've updated those.

PR #349, opened 2019-04-04
feat: auto dial discovered peers

jacobheun:
This PR leverages the strategies detailed at https://github.com/libp2p/js-libp2p/pull/349#issuecomment-480256157 to implement auto dial in libp2p. This should improve our ability to maintain an appropriate level of connections based on the min and max peers configured for the `ConnectionManager`. 

Applications leveraging libp2p should ensure they are not dialing discovered peers when using this update. Applications that do wish to handle all dialing can disable `autoDial` in the config.

Depends on https://github.com/libp2p/js-libp2p-switch/pull/325

### DIFF
--- a/PEER_DISCOVERY.md
+++ b/PEER_DISCOVERY.md
@@ -1,0 +1,59 @@
+# Peer Discovery and Auto Dial
+
+**Synopsis**:
+* All peers discovered are emitted via `peer:discovery` so applications can take any desired action.
+* Libp2p defaults to automatically connecting to new peers, when under the [ConnectionManager](https://github.com/libp2p/js-libp2p-connection-manager) low watermark (minimum peers).
+  * Applications can disable this via the `peerDiscovery.autoDial` config property, and handle connections themselves.
+  * Applications who have not disabled this should **never** connect on peer discovery, unless performing a specific action.
+
+## Scenarios
+In any scenario, if a peer is discovered it should be added to the PeerBook. This ensures that
+even if we don't dial to a node when we discover it, we know about it in the event that it
+becomes known as a provider for something we need.
+
+### 1. Joining the network
+The node is new and needs to join the network. It currently has 0 peers.
+**Discovery Mechanisms**: [Ambient Discovery](#ambient-discovery)
+
+### Action to take
+Connect to discovered peers. This should have some degree of concurrency limiting. While the case should be low, if we immediately discover more peers than our high watermark we should avoid dialing them all.
+
+### 2. Connected to some
+The node is connected to other nodes. The current number of connections is less than the desired low watermark.
+**Discovery Mechanisms**: [Ambient Discovery](#ambient-discovery) and [Active Discovery](#active-discovery)
+
+### Action to take
+Connect to discovered peers. This should have some degree of concurrency limiting. The concurrency may need to be modified to reflect the current number of peers connected. The more peers we have, the lower the concurrency may need to be.
+
+### 3. Connected to enough
+**Discovery Mechanisms**: [Ambient Discovery](#ambient-discovery) and [Active Discovery](#active-discovery)
+
+### Action to take
+None. If we are connected to enough peers, the low watermark, we should not connect to discovered peers. As other peers discover us, they may connect to us based on their current scenario.
+
+For example, a long running node with adequate peers is on an MDNS network. A new peer joins the network and both become aware of each other. The new peer should be the peer that dials, as it has too few peers. The existing node has no reason to dial the new peer, but should keep a record of it in case it later becomes an important node due to its contents/capabilities.
+
+Avoiding dials above the low watermark also allows for a pool of connections to be reserved for application specific actions, such as connecting to a specific content provider via a DHT query to find that content (ipfs-bitswap).
+
+### 4. Connected to too many
+The node has more connections than it wants. The current number of connections is greater than the high watermark.
+
+[WIP Connection Manager v2 spec](https://github.com/libp2p/specs/pull/161)
+**Discovery Mechanisms**: [Ambient Discovery](#ambient-discovery) and [Active Discovery](#active-discovery)
+
+## Discovery Mechanisms
+Means of which a libp2p node discovers other peers.
+
+### Active Discovery
+Through active use of the libp2p network, a node may discovery peers.
+
+* Content/Peer routing (DHT, delegated, etc) provider and peer queries
+* DHT random walk
+* Rendezvous servers
+
+### Ambient Discovery
+Leveraging known addresses, or network discovery mechanisms, a node may discover peers outside of the bounds of the libp2p network.
+
+* Bootstrap
+* MDNS
+* proximity based (bluetooth, sound, etc)
--- a/PEER_DISCOVERY.md
+++ b/PEER_DISCOVERY.md
@@ -4,7 +4,7 @@
 * All peers discovered are emitted via `peer:discovery` so applications can take any desired action.
 * Libp2p defaults to automatically connecting to new peers, when under the [ConnectionManager](https://github.com/libp2p/js-libp2p-connection-manager) low watermark (minimum peers).
   * Applications can disable this via the `peerDiscovery.autoDial` config property, and handle connections themselves.
-  * Applications who have not disabled this should **never** connect on peer discovery, unless performing a specific action.
+  * Applications who have not disabled this should **never** connect on peer discovery. Applications should use the `peer:connect` event if they wish to take a specific action on new peers.
 
 ## Scenarios
 In any scenario, if a peer is discovered it should be added to the PeerBook. This ensures that even if we don't dial to a node when we discover it, we know about it in the event that it becomes known as a provider for something we need. The scenarios listed below detail what actions the auto dialer will take when peers are discovered.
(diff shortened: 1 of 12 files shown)


alanshaw:
That's interesting - can you foresee any other strategies other than low and all?

jacobheun:
After thinking through this a bit more I put together some notes. The main take away would be to change the behavior of libp2p to:
1. auto dial discovery peers under the low watermark
2. allow this to be disabled
3. applications should not dial on discovery unless they have disabled this.

## Scenarios
In any scenario, if a peer is discovered it should be added to the PeerBook. This ensures that
even if we don't dial to a node when we discover it, we know about it in the event that it
becomes known as a provider for something we need.

### 1. Joining the network
The node is new and needs to join the network. It currently has 0 peers.
**Discovery Mechanisms**: [Ambient Discovery](#ambient-discovery)

### Action to take
Connect to discovered peers. This should have some degree of concurrency limiting. While the case should be low, if we immediately discover more peers than our high watermark we should avoid dialing them all.

### 2. Connected to some
The node is connected to other nodes. The current number of connections is less than the desired low watermark.
**Discovery Mechanisms**: [Ambient Discovery](#ambient-discovery) and [Active Discovery](#active-discovery)

### Action to take
Connect to discovered peers. This should have some degree of concurrency limiting. The concurrency may need to be modified to reflect the current number of peers connected. The more peers we have, the lower the concurrency may need to be.

### 3. Connected to enough
**Discovery Mechanisms**: [Ambient Discovery](#ambient-discovery) and [Active Discovery](#active-discovery)

### Action to take
None. If we are connected to enough peers, the low watermark, we should not connect to discovered peers. As other peers discover us, they may connect to us based on their current scenario.

For example, a long running node with adequate peers is on an MDNS network. A new peer joins the network and both become aware of each other. The new peer should be the peer that dials, as it has too few peers. The existing node has no reason to dial the new peer, but should keep a record of it in case it later becomes an important node due to its contents/capabilities.

Avoiding dials above the low watermark also allows for a pool of connections to be reserved for application specific actions, such as connecting to a specific content provider via a DHT query to find that content (ipfs-bitswap).

### 4. Connected to too many
The node has more connections than it wants. The current number of connections is greater than the high watermark.

[WIP Connection Manager v2 spec](https://github.com/libp2p/specs/pull/161)
**Discovery Mechanisms**: [Ambient Discovery](#ambient-discovery) and [Active Discovery](#active-discovery)

## Discovery Mechanisms
Means of which a libp2p node discovers other peers.

### Active Discovery
Through active use of the libp2p network, a node may discovery peers.

* Content/Peer routing (DHT, delegated, etc) provider and peer queries
* DHT random walk
* Rendezvous servers

### Ambient Discovery
Leveraging known addresses, or network discovery mechanisms, a node may discover peers outside of the bounds of the libp2p network.

* Bootstrap
* MDNS
* proximity based (bluetooth, sound, etc)


## Thoughts
* All peers discovered should be emitted via `peer:discovery` as applications may wish to do something with that information.
* Libp2p should default to automatically connecting to new peers, when under the low watermark.
  * Applications should be able to disable this, and handle connections themselves.
  * Applications who have not disabled this should **never** connect on peer discovery.

dirkmc:
This sounds like a good approach 👍 

A few questions:
1. Should we make any distinction between inbound / outbound / duplex connections?
2. Do the LOW / ALL discovery strategies still apply in this second set of concepts?
3. Should the dial queue have a priority assigned to dials? (eg low for discovered peers, high for peers that are actively dialed by libp2p clients)

jacobheun:
> Should we make any distinction between inbound / outbound / duplex connections?

There may be value in doing so in the future. I think the first step here will be improving how inbound connections are being handled in the switch and adding the appropriate tags to the connections, which is something the connection manager needs.

> Do the LOW / ALL discovery strategies still apply in this second set of concepts?

No, we shouldn't need the discovery strategies anymore. I pushed up the latest code I have (tests need to be fixed). This changes things over to a configurable `autoDial` approach.

> Should the dial queue have a priority assigned to dials? (eg low for discovered peers, high for peers that are actively dialed by libp2p clients)

Yes, I think they should. Doing auto dialing makes this a lot easier to know where things are coming from so deprioritizing these calls becomes a lot easier.

vasco-santos:
Thanks for clearly exposing this problem and your ideas in this regard @jacobheun ! 🙌 

Totally agree with the following reasoning 👍 :

1) Automatic dial when below the low watermark
2) Don't dial when higher than the low watermark. We need to keep dial queues ready for application specific actions  and the applications should be able to connect if they want to.
3) Behaviors configurable from outside for applications being able overwrite

alanshaw:
👍 thanks for putting so much thought into this @jacobheun - this sounds like a good way forward to limit the number of dials we're doing.

alanshaw:
Can we add your documentation here to the README?

jacobheun:
Since there's a good amount of info there I created a separate readme with the peer discovery / auto dial information. I also updated some language in the main readme and linked to the new readme.

src/index.js
```diff
@@ -468,6 +440,94 @@ class Node extends EventEmitter {
...
+   *
+   * @private
+   * @param {PeerInfo} peerInfo
+   */
+  _peerDiscovered (peerInfo) {
+    peerInfo = this.peerBook.put(peerInfo)
+
+    if (!this.isStarted()) return
+
+    this.emit('peer:discovery', peerInfo)
```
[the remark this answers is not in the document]
jacobheun: It keeps a record of those peers and will emit them on startup. This avoids the scenario where you disable autoDial and get a bootstrap or mdns peer discovered result before libp2p starts. If you try to dial and libp2p hasn't finished starting it will fail. This just makes libp2p take care of the logic internally.

src/index.js
```diff
@@ -468,6 +440,94 @@ class Node extends EventEmitter {
...
+   * @param {function(Error)} callback
+   */
+  _setupPeerDiscovery (callback) {
+    // Once we start, emit and dial any peers we may have already discovered
+    this.state.on('STARTED', () => {
+      this.peerBook.getAllArray().forEach((peerInfo) => {
+        this.emit('peer:discovery', peerInfo)
+        this._maybeConnect(peerInfo)
+      })
+    })
```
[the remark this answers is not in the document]
jacobheun: This should really run anytime the node starts. It will get run multiple times if the node is started again, though. I can just register this in the constructor to avoid the need to remove and re-add it on a restart.

PEER_DISCOVERY.md
```diff
@@ -0,0 +1,59 @@
+# Peer Discovery and Auto Dial
+
+**Synopsis**:
+* All peers discovered are emitted via `peer:discovery` so applications can take any desired action.
+* Libp2p defaults to automatically connecting to new peers, when under the [ConnectionManager](https://github.com/libp2p/js-libp2p-connection-manager) low watermark (minimum peers).
+  * Applications can disable this via the `peerDiscovery.autoDial` config property, and handle connections themselves.
+  * Applications who have not disabled this should **never** connect on peer discovery, unless performing a specific action.
```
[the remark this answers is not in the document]
jacobheun: Thinking through this more, I don't think there's any reason for a connect to ever happen on peer discovery with auto dial on. `peer:connect` would be a more appropriate event to take action on.

src/index.js
```diff
@@ -428,16 +410,14 @@ class Node extends EventEmitter {
...
-            }),
-            cb
-          )
-        }
-        cb()
+        // stop all discoveries before continuing with shutdown
+        parallel(
+          this._discovery.map((d) => {
+            d.removeListener('peer', this._peerDiscovered)
+            return (_cb) => d.stop(() => { _cb() })
```
dirkmc: Can stop() return an error? If so, should we log it here? If not this can just be `return (_cb) => d.stop(_cb)`

src/index.js
```diff
@@ -468,6 +449,86 @@ class Node extends EventEmitter {
...
+   * @param {PeerInfo} peerInfo
+   */
+  _maybeConnect (peerInfo) {
+    // If auto dialing is on, check if we should dial
+    if (this._config.peerDiscovery.autoDial === true && !peerInfo.isConnected()) {
+      const minPeers = this._options.connectionManager.minPeers || 0
+      if (minPeers > Object.keys(this._switch.connection.connections).length) {
+        log('connecting to discovered peer')
+        this._switch.dialer.connect(peerInfo, (err) => {
+          log.error('could not connect to discovered peer', err)
```
dirkmc:
```suggestion
          err && log.error('could not connect to discovered peer', err)
```

src/index.js
```diff
@@ -468,6 +449,86 @@ class Node extends EventEmitter {
...
+   * connected peers is less than the configured `ConnectionManager`
+   * minPeers.
+   * @private
+   * @param {PeerInfo} peerInfo
+   */
+  _maybeConnect (peerInfo) {
+    // If auto dialing is on, check if we should dial
+    if (this._config.peerDiscovery.autoDial === true && !peerInfo.isConnected()) {
+      const minPeers = this._options.connectionManager.minPeers || 0
+      if (minPeers > Object.keys(this._switch.connection.connections).length) {
```
dirkmc: Is an entry added to this._switch.connection.connections as soon as dial is called? If not, then when there are a bunch of peers discovered (eg on startup) this may dial more peers than `minPeers`

jacobheun: It's added as soon as the queue for that peer starts. So in order for this to dial more peers than `minPeers` we would have to dial more peers than the max parallel limit of the queue (currently 100 by default).

src/index.js
jacobheun: Yes they can, good call. logs added.

Issue #356, opened 2019-04-15
Dont discover self


jacobheun:
Original issue: https://github.com/libp2p/js-libp2p-switch/issues/331.

On peer discovery, if the peer happens to be us, we should avoid adding ourselves to the PeerBook and emitting a `peer:discovery` event.

mkg20001:
Additionally maybe log a warning like "WARN: Discovery module $NAME caused this peer to discover itself. This is a bug and should be reported at $URL" (url could be read from package.json, as npmjs.com currently does)
The warning should be shown just once. possibly also print a line to debug log everytime this happens so user knows how frequently this is occurring

jacobheun:
> Additionally maybe log a warning like "WARN: Discovery module $NAME caused this peer to discover itself.

While I think this is valuable I think having the warning code always be present might be unncessary. I think if we log an error we'll at least have the stack trace of where it came from and the frequency it happens.  Additionally I think the right thing to do would be to add self discovery tests to https://github.com/libp2p/interface-peer-discovery (along with the other tests it should have), and get that added to each of the discovery service test suites. @mkg20001 what do you think about that approach?

mkg20001:
@jacobheun Tests sound great, should also be mandatory. About log: Put it in debug log. That should be enough, so developers know it's happening and can report bugs

jacobheun:
> About log: Put it in debug log. 

I added the error to debug for `libp2p:error` in https://github.com/libp2p/js-libp2p/pull/357.

> Tests sound great, should also be mandatory.

I'll create an issue over at interface-discovery to track this.

PR #357, opened 2019-04-15
fix: bail when discovering self

jacobheun:
fixes https://github.com/libp2p/js-libp2p/issues/356

If a discovery service causes us to discover ourselves, dont do anything. This also logs out an error so that we can trace the discovery back to the module that did it.

Note, since not all modules currently have access to `self.peerInfo` they can't avoid emitting it, so we need to prevent that here.

ref https://github.com/libp2p/js-libp2p-webrtc-star/issues/171#issuecomment-483199746


src/index.js
```diff
@@ -476,6 +477,10 @@ class Node extends EventEmitter {
    * @param {PeerInfo} peerInfo
    */
   _peerDiscovered (peerInfo) {
+    if (peerInfo.id.toB58String() === this.peerInfo.id.toB58String()) {
```
dirkmc: Is it more efficient to compare the underlying buffers? ie peerInfo.id.id.equals(this.peerInfo.id.id)

jacobheun: It's actually not. The stringified version is already stored when the instance is created, so there's no conversion actually happening, it's a straight in-memory check. For the buffer check, if the buffers aren't strictly equivalent (the majority of cases) `.compare` is run to check each byte, which is slower.

dirkmc: Got it 👍

PR #360, opened 2019-04-28
Added .js in front of the require

ShivamRawat0l:
May make the new folks confuse with the other file


jacobheun:
Fixed in master

PR #362, opened 2019-05-06
feat: sign pubsub messages

jacobheun:
This upgrades floodsub so that all messages published are signed

Requires https://github.com/libp2p/js-libp2p-floodsub/pull/78

I ran libp2p/interop against this branch to verify it's passing, https://github.com/libp2p/interop/runs/119819194

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "err-code": "^1.1.2",
     "fsm-event": "^2.1.0",
     "libp2p-connection-manager": "~0.1.0",
-    "libp2p-floodsub": "~0.15.8",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "err-code": "^1.1.2",
     "fsm-event": "^2.1.0",
     "libp2p-connection-manager": "~0.1.0",
-    "libp2p-floodsub": "~0.15.8",
+    "libp2p-floodsub": "libp2p/js-libp2p-floodsub#feat/signing",
```
vasco-santos:
```suggestion
    "libp2p-floodsub": "~0.16.0",
```

Issue #361, opened 2019-04-28
Suggestion: circuit-relay-like module, but for negotiating WebRTC


guysv:
So, the best part of circuit-relay, is the fact that it allows to connect nodes where neither can listen for incoming connections.

Unfortunately, running a circuit-relay HOP node is expensive, and by default HOP is turned off.

**tl;dr**: I hacked js-libp2p-circuit into negotiating WebRTC signals using newly relayed connections, and establish a direct connection between unconnected peers using a helping 3rd party.

Such connection scheme is more considerate of "relay" nodes, because they only need to relay the signaling data, and not the entire connection.
Another plus, is that now browsers should be able to connect to one another **directly**.

I uploaded the [modded js-libp2p-circuit](https://github.com/guysv/js-libp2p-webrtc-poc/tree/webrtc-poc) repo and a [demo](https://github.com/guysv/js-libp2p-webrtc-poc-demo) to github:
You can view the diff to the original circuit-relay module in this PR: https://github.com/guysv/js-libp2p-webrtc-poc/pull/1

To run the demo:
1. `npm install` (note: I included a custom package-lock.json in the demo repo so that js-libp2p-switch will use my modded circuit module, if you feel extra paranoid, regenerate the package-lock.json and fill in `guysv/js-libp2p-webrtc-poc#webrtc-poc`
2. run the hop node `DEBUG=libp2p:circuit:* node src/hop.js`
3. run listener node `DEBUG=libp2p:circuit:* node src/listener.js`
4. run dialer node `DEBUG=libp2p:circuit:* node src/dialer.js`
5. look for those logs I added, which indicate that WebRTC connection was established with listener node
```
# dialer node
  libp2p:circuit:dialer Writing signal to circuit connection +21ms
  libp2p:circuit:dialer Writing signal to circuit connection +1ms
  libp2p:circuit:dialer Read signal from circuit connection +46ms
  libp2p:circuit:dialer Connected! +21ms
```
```
# listener node
  libp2p:circuit:stop Read signal from circuit connection +0ms
  libp2p:circuit:stop Read signal from circuit connection +3ms
  libp2p:circuit:stop Writing signal to circuit connection +11ms
  libp2p:circuit:stop Writing signal to circuit connection +2ms
  libp2p:circuit:stop Connected! +22ms
```
The connection is not really usable, but those logs indicate that a WebRTC channel was established, good enough for POC :)

[...]

PR #363, opened 2019-05-16
Instantiate transports with a reference to node

guysv:
As discussed [here](https://github.com/libp2p/js-libp2p/issues/361#issuecomment-491229478), transports could benefit from having a reference to the libp2p node on creation (in case the transport depends on perior connections, e.g. webrtc, relay).

As far as I can tell, this change is backwards compatible.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -345,7 +345,7 @@ class Node extends EventEmitter {
       let t
 
       if (typeof Transport === 'function') {
-        t = new Transport({libp2p: this})
+        t = new Transport({ libp2p: this })
       } else {
         t = Transport
       }
--- a/src/index.js
+++ b/src/index.js
@@ -345,7 +345,7 @@ class Node extends EventEmitter {
       let t
 
       if (typeof Transport === 'function') {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -345,7 +345,7 @@ class Node extends EventEmitter {
       let t
 
       if (typeof Transport === 'function') {
-        t = new Transport()
+        t = new Transport({libp2p: this})
```
jacobheun: Linting fix:
```suggestion
        t = new Transport({ libp2p: this })
```

jacobheun: Other than the linting issue this looks fine. Travis is failing due to this and commitlint. I went ahead and disabled commitlint on master so it doesn't block tests going forward. If you rebase master and fix the linting issue I suggested above, travis tests should properly run.

I'll update the commit message on squash/merge.

guysv:
```diff
+bundlesize: { maxSize: '222kB' },
-bundlesize: { maxSize: '218kB' },
```
?

PR #365, opened 2019-05-24
feat: integrate gossipsub by default

vasco-santos:
BREAKING CHANGE: new configuration for deciding the implementation of pubsub to be used.
In this context, the experimental flags were also removed.

This PR modified the configuration of a libp2p node, in order to allow the selection of the pubsub implementation to use. Also Gossipsub was added as the default pubsub implementation in the tests

Needs: 

- [x] [ChainSafe/gossipsub-js/pull/21](https://github.com/ChainSafe/gossipsub-js/pull/21)
- [x] Fallback to floodsub [ChainSafe/gossipsub-js#34](https://github.com/ChainSafe/gossipsub-js/pull/34)
- [x] Release `gossipsub-js`

Unblocks:

- [ ] [libp2p/js-libp2p-daemon#19](https://github.com/libp2p/js-libp2p-daemon/pull/19)
- [ ] [libp2p/interop#17](https://github.com/libp2p/interop/pull/17)


vasco-santos:
cc @wemeetagain @Mikerah

test/pubsub.node.js
```diff
@@ -145,9 +148,133 @@ describe('.pubsub', () => {
...
+          })
+        },
+        // Stop both nodes
+        (cb) => stopTwo(nodes, cb)
+      ], (err) => {
+        // Verify there was no error, and mark the expect
+        expect(err).to.not.exist().mark()
+      })
+    })
+    it('publish should fail if data is not a buffer', (done) => {
```
jacobheun: I know this is not in scope, and we can move this discussion to an issue, but should we remove the buffer requirement and allow any valid `Buffer.from(data)` here? Being able to just pass a string to publish feels like a quick and easy change for users, especially since we're moving this out of experimental.

test/pubsub.node.js
```diff
@@ -145,9 +148,133 @@ describe('.pubsub', () => {
...
+        // subscribe on the first
+        (cb) => nodes[0].pubsub.subscribe('pubsub', handler, cb),
+        // Wait a moment before publishing
+        (cb) => setTimeout(cb, 500),
+        // publish on the second
+        (cb) => nodes[1].pubsub.publish('pubsub', data, cb),
+        // ls subscripts
+        (cb) => nodes[1].pubsub.ls(cb),
+        // get subscribed peers
+        (cb) => nodes[1].pubsub.peers('pubsub', cb),
```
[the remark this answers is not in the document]
vasco-santos: Yes, I am removing that. The implementations should be responsible for those tests

test/pubsub.node.js
vasco-santos: I agree with you!

I will change that now, as we are moving out of experimental.

src/pubsub.js
```diff
@@ -55,29 +52,33 @@ module.exports = (node) => {
...
       }
 
+      if (!Buffer.isBuffer(data) && typeof data !== 'string') {
+        return nextTick(callback, errCode(new Error('data must be a Buffer or a String'), 'ERR_DATA_IS_NOT_VALID'))
+      }
+
       if (!Buffer.isBuffer(data)) {
-        return nextTick(callback, errCode(new Error('data must be a Buffer'), 'ERR_DATA_IS_NOT_A_BUFFER'))
+        data = Buffer.from(data)
       }
```
jacobheun: The buffer check could be simplified by just trying to cast it immediately. So instead of doing both of these checks just do:

```js
try {
  data = Buffer.from(data)
} catch (err) {
  return nextTick(callback, errCode(new Error('data must be convertible to a Buffer'), 'ERR_DATA_IS_NOT_VALID'))
}
```

vasco-santos: cool! way better

src/index.js
```diff
@@ -121,8 +121,11 @@ class Libp2p extends EventEmitter {
...
     }
 
-    // enable/disable pubsub
-    if (this._config.EXPERIMENTAL.pubsub) {
+    // start pubsub
+    if (this._config.pubsub.enabled) {
+      const Pubsub = this._modules.pubsub
+
+      this._pubsub = new Pubsub(this)
       this.pubsub = pubsub(this)
```
jacobheun: What do you think about just doing `this.pubsub = pubsub(this, this._modules.pubsub)` and letting `./src/pubsub.js` take care of keeping track of pubsub? I think it gets confusing to have both _pubsub and pubsub. I was having a hard time keeping track of what was what.

vasco-santos: Seems better your suggestion. Will refactor it now

src/index.js
```diff
@@ -121,9 +121,9 @@ class Libp2p extends EventEmitter {
       })
     }
 
-    // enable/disable pubsub
-    if (this._config.EXPERIMENTAL.pubsub) {
-      this.pubsub = pubsub(this)
+    // start pubsub
+    if (this._config.pubsub.enabled) {
```
jacobheun: @vasco-santos what do you think about changing this logic a bit. I think if `this._modules.pubsub` is set, and `this._config.pubsub.enabled !== false` we should use pubsub. This would avoid people needing to add **and** enable it. We do this for discovery. If you've configured it, there's a **really** good chance you want to use it :)

vasco-santos: I agree, this is a better dev ex

vasco-santos:
@jacobheun gossipsub was released 🎉 

Just rebased and updated the PR

jacobheun:
0.26.0-rc.0 is out!

PR #367, opened 2019-06-06
feat: add createLibp2p to generate a PeerInfo instance

jacobheun:
`createLibp2p` is a new exported helper function that allows users to create a libp2p instance without worrying about creating a PeerInfo instance first.

Now instead of needing to do this:
```js
const PeerInfo = require('peer-info')
const Libp2p = require('libp2p')
PeerInfo.create((err, peerInfo) => {
  const libp2p = new Libp2p({
    ...options,
    peerInfo
  }) 
  // Use libp2p
})
```

you can just do:
```js
const { createLibp2p } = require('libp2p')
createLibp2p(options, (err, libp2p) => {
    // Use libp2p
}) 
```

While the code change is not significant, it removes the need to know about `PeerInfo` creation prior to creating a libp2p instance. This should hopefully help reduce the barrier to entry by a small degree.

I also changed `Node` to `Libp2p` because the naming is unnecessarily confusing and nondescript.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,8 +10,10 @@ const errCode = require('err-code')
 const each = require('async/each')
 const series = require('async/series')
 const parallel = require('async/parallel')
+const nextTick = require('async/nextTick')
 
 const PeerBook = require('peer-book')
+const PeerInfo = require('peer-info')
 const Switch = require('libp2p-switch')
 const Ping = require('libp2p-ping')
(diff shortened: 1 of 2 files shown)


dirkmc:
Nice 👍

Was there a docs change too?

jacobheun:
> Was there a docs change too?

No but good call, I am updated deps so I will include a docs update with that.

PR #368, opened 2019-06-06
docs: add createLibp2p to readme

jacobheun:
**Also**:
* update dependencies
* Fix issue with wrtc rendezvous in tests due to it moving to a promise based api (due to the recent hapi upgrade)


dirkmc:
Would it make sense for `createLibp2p()` to be the standard way to create a Libp2p node in the docs? It might be simpler for newcomers

jacobheun:
> Would it make sense for createLibp2p() to be the standard way to create a Libp2p node in the docs? It might be simpler for newcomers

Yeah I think that makes sense. I'll swap those two and figure out what's up with the tests from the dep update.

README.md
```diff
@@ -204,13 +204,28 @@ class Node extends libp2p {
...
-#### Create a Node - `new libp2p.Node(options)`
+#### Create a Node - `Libp2p.createLibp2p(options, callback)`
 
-> Creates an instance of the libp2p.Node.
+> Behaves exactly like `new Libp2p(options)`, but doesn't require a PeerInfo. One will be generated instead
+
+```js
+const { createLibp2p } = require('libp2p')
+createLibp2p(options, (err, libp2p) => {
+  libp2p.start(() => {})
```
dirkmc: It may be worth adding a line to check the error here in case people blindly copy and paste this code

jacobheun: Yeah I agree, the start callback should also probably due that. I'm thinking just a simple throw if the `err` exists to try and be overtly clear if there's an issue.

dirkmc: Sounds good 👍

jacobheun:
All the signaling server references have been updated, and I just consolidated to use only the server spun up by `.aegir.js`

PR #370, opened 2019-06-12
chore: update keywords and description

jacobheun:
I noticed the only keyword being used in package.json was IPFS, so I added some more libp2p specific keywords and also updated the description. The homepage also now points to libp2p.io.

I also reordered a couple other fields in package.json to group them a bit better.


vasco-santos:
CI failures are unrelated do these changes

PR #381, opened 2019-07-18
feat: promisify all api methods that accept callbacks

achingbrain:
This is a stop-gap until the full async/await migration can be completed.  It means we can refactor tests of other modules that depend on this module without having to mix async flow control strategies.

N.b. some methods that were previously callable without callbacks (e.g. `node.start()`, `node.stop()`, etc) now require callbacks otherwise a promise is returned which, if rejected, can cause `unhandledPromiseRejection` events and lead to memory leaks where exceptions are thrown and resources are not cleaned up.

### DIFF
--- a/src/pubsub.js
+++ b/src/pubsub.js
@@ -3,6 +3,7 @@
 const nextTick = require('async/nextTick')
 const { messages, codes } = require('./errors')
 const FloodSub = require('libp2p-floodsub')
+const promisify = require('promisify-es6')
 
 const errCode = require('err-code')
 
@@ -12,7 +13,7 @@ module.exports = (node) => {
   node._floodSub = floodSub
 
   return {
-    subscribe: (topic, options, handler, callback) => {
+    subscribe: promisify((topic, options, handler, callback) => {
       if (typeof options === 'function') {
         callback = handler
         handler = options
@@ -33,13 +34,19 @@ module.exports = (node) => {
       }
 
       subscribe(callback)
-    },
+    }),
 
-    unsubscribe: (topic, handler, callback) => {
+    unsubscribe: promisify((topic, handler, callback) => {
       if (!node.isStarted() && !floodSub.started) {
         return nextTick(callback, errCode(new Error(messages.NOT_STARTED_YET), codes.PUBSUB_NOT_STARTED))
       }
-      if (!handler && !callback) {
+      if (!callback) {
+        // we will always be passed a callback because of promisify. If we weren't passed one
+        // really it means no handler was passed..
+        callback = handler
+        handler = null
+      }
+      if (!handler) {
         floodSub.removeAllListeners(topic)
       } else {
         floodSub.removeListener(topic, handler)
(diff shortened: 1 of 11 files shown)


jacobheun:
We need to update the readme for this. Since @achingbrain is out, I'll go ahead and update that.

jacobheun:
I went ahead and added a generic note at the top of the API section about promisify support and the move to async/await. I think it makes sense to update the API code once we do the actual switch over here, in case there are any changes to the way the api behaves.

src/pubsub.js
```diff
@@ -33,13 +34,19 @@ module.exports = (node) => {
...
-    },
+    }),
 
-    unsubscribe: (topic, handler, callback) => {
+    unsubscribe: promisify((topic, handler, callback) => {
       if (!node.isStarted() && !floodSub.started) {
         return nextTick(callback, errCode(new Error(messages.NOT_STARTED_YET), codes.PUBSUB_NOT_STARTED))
       }
-      if (!handler && !callback) {
+      if (!callback) {
```
jacobheun: This is going to be a breaking change. Currently js-ipfs calls `self.libp2p.pubsub.unsubscribe(topic, handler)`, which will end up resulting in this calling callback as a handler. We'll need to do this as a minor release and then update js-ipfs.

vasco-santos: Yeah, I was aiming to have this and gossipsub added for a new minor release

jacobheun: Yes, that would be ideal to release them together. I think I could also do https://github.com/libp2p/js-libp2p/issues/384 (pulling in the core modules) as part of the same minor.

The unfortunate thing about this logic is that if you want to unsubscribe a single handler, you have to use callbacks, otherwise we assume the handler is a callback.

vasco-santos: For #384 we will also have to get the tests for those modules here, as well as probably refactor some stuff, as I believe we will end up with a lot of replicated stuff. So, unless you feel that #384 will not take a long time, I think we should get those 2, as they are kind of ready to go. And we can unblock gossipsub on `js-libp2p` and `js-ipfs`, and more async migration work.

---

yeah, I see it. This is a tricky case because of the 2 last arguments with a function

jacobheun: I'll open a PR for 384 and if it's not ready by then, we can release without it. I don't want to hold anything up with.

----

A different route we could go is to return `Promise.resolve()` if callback is not defined. The caveat of this approach is that if you want to unsubscribe from everything, you have to provide a `null` or `undefined` handler, but you don't have to use callbacks ever.

vasco-santos: hmm, it is more coherent with the API to do what you suggest. But, we need to have proper docs for help on that

jacobheun: I think we can note the difference in the readme, and maybe add an example of the calls to the jsdocs for the method to help. The docs currently point to the ipfs interface for the api, which describes the calls as one of:
```js
await libp2p.unsubscribe(topic)
await libp2p.unsubscribe(topic, handler)
libp2p.unsubscribe(topic, handler, callback)
```
as the code currently is, the first two calls are the same through promisify. We'd temporarily have the following until we go full async/await:
```js
await libp2p.unsubscribe(topic, null)
await libp2p.unsubscribe(topic, handler)
libp2p.unsubscribe(topic, handler, callback)
```

vasco-santos: Yeah, I think we can go with that approach

jacobheun: @vasco-santos I am thinking of opening an additional PR to fix the subscribe method since we're doing a breaking change. Right now, the order of params (topic, options, handler, callback) does not match the interface api, it would be nice to correct that with this release. I can add jsdocs for it and the rest of the pubsub methods in that PR. What do you think?

jacobheun:
@vasco-santos this is ready for one last review pass and should be good to merge. I had to fix some tests due to the pubsub change and a recent release of the delegate content router.

PR #392, opened 2019-07-31
chore: promisify pubsub start and stop

vasco-santos:
With #381 and #365 , the gossipsub PR got in after and we did not promisify `start` and `stop`. This PR solves that


jacobheun:
v0.26.0-rc.1 is out with this

PR #393, opened 2019-07-31
fix: pubsub default config

jacobheun:
* Also fixes examples for the latest release

The pubsub config was defaulting to `enabled=false`. This makes things a bit annoying since you now have to provide the desired pubsub module in configuration. If we provide the module we shouldn't need to enable it also, it should be enabled by default. If users for some reason want to provide the module but disable it, that should be the override behavior.


jacobheun:
v0.26.0-rc.2 is out.

Issue #266, opened 2018-10-26
Awesome Endeavour: Async Iterators


daviddias:
@libp2p/javascript-team see thread at [Awesome Endeavour: Async Iterators](https://github.com/ipfs/js-ipfs/issues/1670)

**Migration Notice: As libp2p modules migrate to async/await there will be some level of incompatibility when trying to include the latest release of modules. You can see the latest support minor versions of each module below.**

```
    "libp2p-bootstrap": "^0.9.x",
    "libp2p-circuit": "^0.3.x",
    "libp2p-delegated-content-routing": "^0.2.x",
    "libp2p-delegated-peer-routing": "^0.2.x",
    "libp2p-floodsub": "^0.17.x",
    "libp2p-gossipsub": "^0.0.x",
    "libp2p-kad-dht": "^0.15.x",
    "libp2p-mdns": "^0.12.x",
    "libp2p-mplex": "^0.8.x",
    "libp2p-secio": "^0.11.x",
    "libp2p-spdy": "^0.13.x",
    "libp2p-tcp": "^0.13.x",
    "libp2p-webrtc-star": "^0.16.x",
    "libp2p-websockets": "^0.12.x",
    "libp2p-websocket-star": "^0.10.x",
    "libp2p-websocket-star-rendezvous": "^0.4.x"
```

## JS Libp2p Refactor

As more of the dependencies for libp2p are finishing their migrations, we are starting on the internals of js-libp2p. Due to the fact that there are a LOT of things that need to be updated internally, we will be cutting the work into smaller, isolated chunks. This will allow us to iterate, avoiding a much larger PR.

### Working Branch

All async PRs to js-libp2p internals that break compatibility should be opened against the [refactor/async-await](https://github.com/libp2p/js-libp2p/tree/refactor/async-await) branch.

### Submitting a PR

If there is a subsystem that you wish to take on, please add a comment to the internals refactor [Trello card](https://trello.com/c/OseiuYno/22-refactor-js-libp2p-internals), or open a DRAFT PR, so that others know it's being worked on.

- The existing Libp2p test suite will be removed from the refactor branch. This will allow each subsystem refactor to add new tests and avoid duplication. When refactoring a subsystem, you may pull pertinent tests from master and refactor those to be async, if you are not sure what should be tested.
- A PR should have a full test suite for that system. Whenever possible, we should test the internals of a system, without relying on **all** of libp2p. This ensures we're able to test in better isolation, and will help with being able to iterate on the refactor.

[...]

PR #397, opened 2019-08-02
docs: async migration

vasco-santos:
In the context of `js-libp2p` being a skeleton project, when a new user tries to use this there is a probability of hitting problems, as a consequence of the on going efforts to migrate the codebase to `async`. For instance, if we try to install `libp2p-bootstrap` to use in the bundle, it will use the newest version by default, which is already migrated, taking problems to `js-libp2p`, which still expects callbacks.

This PR adds a note to new users in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ npm install --save libp2p
 
 ## Usage
 
+**IMPORTANT NOTE**: We are currently on the way of migrating all our `libp2p` modules to use `async await` and `async iterators`, instead of callbacks and `pull-streams`. As a consequence, when you start a new libp2p project, we must check which versions of the modules you should use. For now, it is required to use the modules using callbacks with `libp2p`, while we are working on getting the remaining modules ready for a full migration. For more details, you can have a look at [libp2p/js-libp2p#266](https://github.com/libp2p/js-libp2p/issues/266).
+
 ### [Tutorials and Examples](/examples)
 
 You can find multiple examples on the [examples folder](/examples) that will guide you through using libp2p for several scenarios.


didlie:
I think the issue is that code migration to satisfy syntax is destructive to branch developers and almost irrelevant. Whatever code style was chosen and used to create a working and usable project, that is the code that should continue to be maintained and working. if obfuscation and annoyance and wasted time is the goal, then migrate the code bit-by-bit and ruin working branch applications and invalidate ports of the project... by all means.

vasco-santos:
@didlie I agree with avoiding immediate breaking changes, especially if they are really impactful. However, in this case, we have been talking about this since last year #266 

The main goals with this change are to improve the efficiency of the codebase, decrease the bundle size of `js-libp2p` and provide an improved developer experience to the libp2p users. Supporting and maintaining the current implementations would not allow anything of this.

With this PR we are essentially pointing this information to new contributors, who would start a new project and try to use the latest versions of each libp2p modules, which is still not supported. For ongoing projects, each module that got upgraded was released as a breaking change, and users will be able to see if they should update to the newest version of a module, or wait until all the code migration is complete (until getting into `js-libp2p`).

didlie:
it will make the code more condense and less readable and less adaptable...why not just let it be compiled? compilation size should be identical.

mikeal:
@didlie compilation won’t fix the pull-stream issues. this isn’t just a style change, the actually semantics of how data is processed change and are cleaned up quite a bit by migrating to async iterators.

didlie:
@mikeal there is nothing unethical about mixing code versions in a single repo.  It can actually help developers distinguish.

Issue #401, opened 2019-08-10
Make pubsub option emitSelf configurable


a1300:
First and foremost: Thanks for this great package and the hard work! :+1: 

<br/>

- **Version**: 0.26.0
- **Platform**: Linux Kubuntu-18 4.18.0-25-generic
- **Subsystem**:

#### Description:
Currently is the option **emitSelf** for `pubsub` packages ([libp2p/js-libp2p-floodsub](https://github.com/libp2p/js-libp2p-floodsub) and [ChainSafe/gossipsub-js](https://github.com/ChainSafe/gossipsub-js)) hardcoded in the code.

https://github.com/libp2p/js-libp2p/blob/fd738f9d51e2166861ce8327c8ef2f4eb52a27be/src/pubsub.js#L10

Both `pubsub` packages support the **emitSelf** option. It would be nice to add the option to specify if self-published messages should be received.

Example: 
```diff
export class Bundle extends libp2p {
  constructor(_options) {
    const defaults = {
      modules: {
        transport: [TCP],
        streamMuxer: [Mplex],
        connEncryption: [SECIO],
        peerDiscovery: [Bootstrap],
        dht: DHT,
        pubsub: GossipSub,
      },
      config: {
        pubsub: {
          enabled: true,
+         emitSelf: false,
        },
      },
    };
    super(defaultsDeep(_options, defaults));
  }
```

a1300:
If this feature request is granted I would like to have this issue assigned to me :slightly_smiling_face:

jacobheun:
Yes, we should pass the configuration into the Pubsub initialization. We should account for other Pubsub options as well when doing this, such as the strict signing and verification params https://github.com/libp2p/js-libp2p-pubsub/blob/v0.2.0/src/index.js#L30-L31. I'd expect the whole pubsub config object to get passed.

jacobheun:
Fixed in #404

PR #404, opened 2019-08-15
fix: pubsub configuration

a1300:
This PR fixes issue #401 and adds the following new `pubsub` configuration options:
- `emitSelf` (boolean)
- `signMessages` (boolean)
- `strictSigning` (boolean)

@jacobheun I wasn't sure if I should add tests for this bug fix because it seemed to me that this repository isn't testing if configuration is correctly passed to the Modules. Should I add some tests?

Thanks!

### DIFF
--- a/examples/pubsub/README.md
+++ b/examples/pubsub/README.md
@@ -63,7 +63,7 @@ node2 received: Bird bird bird, bird is the word!
 node1 received: Bird bird bird, bird is the word!
 ```
 
-You can change the pubsub `emitSelf` option if you don't want that the publishing node also receives its own messages.
+You can change the pubsub `emitSelf` option if you don't want the publishing node to receive its own messages.
 
 ```JavaScript
 const defaults = {
--- a/examples/pubsub/README.md
+++ b/examples/pubsub/README.md
@@ -1,6 +1,6 @@
 # Publish Subscribe
 
-Publish Subscribe is also included on the stack. Currently, we have on PubSub implementation which we ship by default [libp2p-floodsub](https://github.com/libp2p/js-libp2p-floodsub), with many more being researched at [research-pubsub](https://github.com/libp2p/research-pubsub).
+Publish Subscribe is also included on the stack. Currently, we have two PubSub implementation available [libp2p-floodsub](https://github.com/libp2p/js-libp2p-floodsub) and [libp2p-gossipsub](https://github.com/ChainSafe/gossipsub-js), with many more being researched at [research-pubsub](https://github.com/libp2p/research-pubsub).
 
 We've seen many interesting use cases appear with this, here are some highlights:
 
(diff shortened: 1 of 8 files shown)


examples/pubsub/README.md
```diff
@@ -40,11 +57,29 @@ The output of the program should look like:
...
-QmWpvkKm6qHLhoxpWrTswY6UMNWDyn8hN265Qp9ZYvgS82 Bird bird bird, bird is the word!
-QmWpvkKm6qHLhoxpWrTswY6UMNWDyn8hN265Qp9ZYvgS82 Bird bird bird, bird is the word!
-QmWpvkKm6qHLhoxpWrTswY6UMNWDyn8hN265Qp9ZYvgS82 Bird bird bird, bird is the word!
+node2 received: Bird bird bird, bird is the word!
+node1 received: Bird bird bird, bird is the word!
+node2 received: Bird bird bird, bird is the word!
+node1 received: Bird bird bird, bird is the word!
+```
+
+You can change the pubsub `emitSelf` option if you don't want that the publishing node also receives its own messages.
```
jacobheun:
```suggestion
You can change the pubsub `emitSelf` option if you don't want the publishing node to receive its own messages.
```

a1300:
@jacobheun Ok, I will soon add some tests!

a1300:
@jacobheun quick question: Do you want that I test the toggling of `emitSelf` with both `libp2p-gossipsub` and `libp2p-floodsub`?

jacobheun:
@a1300 1 implementation is fine, ultimately we're just verifying that it's getting the configuration. Those modules should handle more in depth testing for themselves.

jacobheun:
Even just testing that a mock Pubsub implementation is getting the right options would be sufficient.

a1300:
@jacobheun I am currently writing the tests, and I wondered if the following is not a bug on its own:

The following code is the config that defines the default properties of the `PubSub` config:
```JavaScript
  // Pubsub config
  pubsub: s('object?', {
    // Pubsub defaults
    enabled: true,
    emitSelf: true,
    signMessages: true,
    strictSigning: true
  })
```

If I now create a `Bundle` with only a subset of the options above then I would expect that the default values would be set, but they aren't.
```JavaScript
const bundle = new libp2p({
  modules: {
    transport: [ TCP ],
    streamMuxer: [ Mplex ],
    connEncryption: [ SECIO ],
    peerDiscovery: [ MulticastDNS ],
    pubsub: Gossipsub
  },
  config: {
    peerDiscovery: {
      mdns: {
        interval: 2000,
        enabled: true
      }
   },
   pubsub: {
      enabled: true,
      emitSelf: true
    }
  }
}
```

The `PubSub` implementation is getting called only with:
```JavaScript
{
  enabled: true,
  emitSelf: true
}
```

But I would expect that it would be called with:
```JavaScript
{
  enabled: true,
  emitSelf: true,
  signMessages: true,
  strictSigning: true
}
```

Is this a bug or expected behavior? Thanks!

jacobheun:
It's expected, but not ideal. The config needs some improvements. Right now it's validating an optional object and not their sub properties. If any object is supplied it uses that and doesn't do any  copying of the default properties. Ideally we'd get the defaults populated, while avoiding making the config too strict. The reason for this is that if we add a new config option in the future to pubsub or another module, we need to avoid libp2p config validation from throwing an error, so that it doesn't have to be updated to use the latest pubsub.

I'll create a ticket to track improving the config, don't worry about it for this, you may just need to be verbose with configuration for some of the tests. (Note: Both existing pubsub modules do properly add defaults for the missing values, so we are catching it there as we should be).

a1300:
@jacobheun please review the tests :slightly_smiling_face:

jacobheun:
@a1300 looks good! I ran the example as well and everything looks good there too. Thanks for the PR!

PR #449, opened 2019-08-25
Generate Typescript Typings via JsDoc

samparsky:
Generates Typescript Typings via JsDoc

* Adds typings to the js-libp2p module
* Add typings generation to travis.yml
* Include typings test


mpetrunic:
I don't think this is gonna work, you are using types like `PeerInfo` and `PeerId` that are not linked here (should be created in their own repos first)

- https://github.com/libp2p/js-peer-info
- https://github.com/libp2p/js-peer-info

vasco-santos:
FYI: we are moving on this [way](https://github.com/libp2p/js-libp2p/issues/659) for the support for typescript users. Thanks to `aegir` that now supports typescript definitions from the jsdocs

vasco-santos:
Closing this as we shipped `0.30` with typedefs generated on release. Thanks anyway for the PR

Issue #399, opened 2019-08-08
README still links to Waffle


marten-seemann:
And since Waffle doesn't exist anymore, it's a dead link.

jacobheun:
Fixed in #450

PR #450, opened 2019-08-27
Remove dead link from readme

geastwood:
PR for #399 

Simply remove the link, if there is a new link, just let me know, will update it.


README.md
```diff
@@ -29,7 +29,7 @@
 
 We've come a long way, but this project is still in Alpha, lots of development is happening, API might change, beware of the Dragons 🐉..
 
-**Want to get started?** Check our [examples folder](/examples). You can check the development status at the [Waffle Board](https://waffle.io/libp2p/js-libp2p).
+**Want to get started?** Check our [examples folder](/examples).
 
 [![Throughput Graph](https://graphs.waffle.io/libp2p/js-libp2p/throughput.svg)](https://waffle.io/libp2p/js-libp2p/metrics/throughput)
```
jacobheun: We should remove the throughput graph below as that is also referencing waffle.
```suggestion
```

geastwood: Sure, should have grabbed README :D

PR #459, opened 2019-09-30
Add streaming iterables guide

jacobheun:
The usage section is currently blank, but I think getting this in sooner is best so that anyone can iterate on this. I am aiming to get some examples added in the next couple days.

I also moved the peer discovery readme since there is now a sub directory for docs.

**Note: This is going into the async refactor branch**


doc/STREAMING_ITERABLES.md
```diff
@@ -0,0 +1,32 @@
+# Iterable Streams
+
+> This document is a guide on how to use Iterable Streams in Libp2p. As a part of the refactor away from callbacks, we have also moved to using Iterable Streams instead of [pull-streams](https://pull-stream.github.io/). If there are missing usage guides you feel should be added, please submit a PR!
```
vasco-santos: Should we point the Issue here for context?

jacobheun: 👍 done!

PR #461, opened 2019-10-01
refactor(async): update transports subsystem

jacobheun:
This PR creates a single transportManager that is attached to libp2p, `libp2p.transportManager`. 

The TransportManager is responsible for keeping track of transports and listeners. It also handles routing single multiaddr dials to their respective transport.

- Per https://github.com/libp2p/js-libp2p/issues/266#issue-374385232, I have removed all the tests. Only the TransportManager tests currently exist.
- I removed transports that have not yet been converted from tests, only tcp and websockets are currently being tested.
- I removed some of the start, stop, and config logic of libp2p, as each of the subsystems being refactored in the future should update those functions as needed. This should make it easier for each subsystem refactor to avoid clashing with one another.


src/transport-manager.js
```diff
@@ -0,0 +1,177 @@
+'use strict'
+
+const pSettle = require('p-settle')
+const { codes } = require('./errors')
+const errCode = require('err-code')
+const debug = require('debug')
+const log = debug('libp2p:transports')
+log.error = debug('error:libp2p:transports')
```
vasco-santos: We have been using the `:error` in the end on the transports. Is this intended?

src/transport-manager.js
```diff
@@ -0,0 +1,177 @@
...
+    }
+    return null
+  }
+
+  /**
+   * Starts listeners for each given Multiaddr.
+   * @async
+   * @param {Multiaddr[]} addrs
+   */
+  async listen (addrs) {
```
vasco-santos: We can add new transports anytime during runtime. We should then be able to listen on transports added after a `listen`.

jacobheun: It's totally doable in the current state, and you could also just call listen at any point if you have new addresses for existing transports.

```js
libp2p.transportManager.add(transportKey, TransportConstructor)
await libp2p.transportManager.listen(newTransportAddrs)
```

Issue #451, opened 2019-09-03
Dialing a multiaddr should dial it first


jacobheun:
## The Problem

Currently in libp2p if you dial a `Multiaddr`, or its string version, the `PeerInfo` record for that peer will be retrieved and then the `Multiaddr` will be added to its multiaddr list before dialing. If the peer already has known addresses, the `Multiaddr` originally provided might not be dialed first.

For example, let's say I am trying to manually connect to an IPFS preload node, `/dns4/node0.preload.ipfs.io/tcp/443/wss/ipfs/QmZMxNdpMkewiVZLMRxaNxUeZpDUb34pWjZ1kZvsd16Zic`. If I have no information about this node, this address will be dialed, as its the only address.
Now, if I have been on the network for a while, or have previously connected to this node, I will likely have some additional addresses, such as:
```log
/dns4/node0.preload.ipfs.io/tcp/443/wss
/ip4/127.0.0.1/tcp/4001
/ip4/108.61.156.24/tcp/4001
/ip4/45.77.150.82/tcp/4001
/ip6/::1/tcp/4001
/ip6/2001:19f0:5:66d1::/tcp/4001
/ip6/2001:19f0:5:fc8:5400:1ff:feeb:77b3/tcp/4001
/ip4/127.0.0.1/tcp/8081/ws
```

The problem here, is that the dialing logic causes us to iterate over our available transports and the target peers Multiaddrs. The order of this dial is currently determined by the order our transports were added, and the order of their Multiaddrs. Most likely, this will be the first `TCP` address in the list, assuming we support TCP as it is often added to [configuration first](https://github.com/ipfs/js-ipfs/blob/v0.37.1/src/core/runtime/libp2p-nodejs.js#L33-L35). Instead of dialing `/dns4/node0.preload.ipfs.io/tcp/443/wss` we would first dial `/ip4/127.0.0.1/tcp/4001`. This may fail and we will continuing trying other addresses, **however**, if we happen to have a node running locally on this address, such as go-ipfs, we would connect and then secio would terminate the connection with the error [dialed to the wrong peer, Ids do not match](https://github.com/libp2p/js-libp2p-secio/blob/eee15dded19d5faae47a91d9c1f0ed74b221bf13/src/handshake/crypto.js#L80) as the Peer dialed is not the target Peer. Since the connection has already been "successfully" dialed, no other addresses are currently being tried. Ideally, we should try all addresses until we have a full, encrypted connection.

## Solutions
[...]

PR #462, opened 2019-10-02
refactor(async): add dialer and upgrader

jacobheun:
This PR adds Dialer and [Upgrader](https://github.com/libp2p/interface-transport#upgrader) components. This moves away from the convoluted connection upgrade process the Switch goes through now, by leveraging a Dialer and Upgrader. An Upgrader takes a [MultiaddrConnection](https://github.com/libp2p/interface-transport#multiaddrconnection) and goes through the process of upgrading it into a [libp2p connection](https://github.com/libp2p/interface-connection#connection-1). This upgrade process will include things like crypto and muxer selection.

The Dialer will:
- Allow dialing a single address (**new!**) https://github.com/libp2p/js-libp2p/issues/451
- Allow dialing to a peer on any of its addresses
- ~~Allow for parallel dials with abort support (**new!**)~~ *This has the potential to introduce the simultaneous dial problem that go libp2p currently has if nodes don't abort the same connections, so I'm punting on this for a future PR*


src/upgrader.js
```diff
@@ -0,0 +1,359 @@
...
+    } catch (err) {
+      log.error('Failed to upgrade outbound connection', err)
+      await maConn.close(err)
+      throw err
+    }
+
+    log('Successfully upgraded outbound connection')
+
+    // Create the muxer
+    const muxer = new Muxer({
```
vasco-santos: what do you think of creating a function to create the muxer for outbound and inbound? It seems that only the log is different, so it would be easier to maintain less code

src/upgrader.js
```diff
@@ -0,0 +1,359 @@
...
+        connection.addStream(stream, protocol)
+        this._onStream({ stream, protocol })
+      },
+      // Run anytime a stream closes
+      onStreamEnd: muxedStream => {
+        connection.removeStream(muxedStream.id)
+      }
+    })
+
+    const newStream = async protocols => {
```
vasco-santos: same as muxer

src/upgrader.js
```diff
@@ -0,0 +1,359 @@
...
+    const protocols = Array.from(cryptos.keys())
+    log('selecting inbound crypto protocol', protocols)
+
+    try {
+      const { stream, protocol } = await mss.handle(protocols)
+      const crypto = cryptos.get(protocol)
+      log('encrypting inbound connection...')
+      const cryptoResponse = await crypto.secureInbound(localPeer, stream)
+
+      if (cryptoResponse) {
```
vasco-santos: What happens if we do not return?

src/upgrader.js
```diff
@@ -0,0 +1,359 @@
...
+  }
+
+  /**
+   * Attempts to encrypt the incoming `connection` with the provided `cryptos`.
+   * @private
+   * @async
+   * @param {PeerId} localPeer The initiators PeerInfo
+   * @param {*} connection
+   * @param {Map<string, Crypto>} cryptos
+   * @returns {[connection, string]} An encrypted connection and the tag of the `Crypto` used
```
vasco-santos: Not returning an array

src/upgrader.js
```diff
@@ -0,0 +1,359 @@
...
+  /**
+   * Attempts to encrypt the given `connection` with the provided `cryptos`.
+   * The first `Crypto` module to succeed will be used
+   * @private
+   * @async
+   * @param {PeerId} localPeer The initiators PeerInfo
+   * @param {*} connection
+   * @param {PeerId} remotePeerId
+   * @param {Map<string, Crypto>} cryptos
+   * @returns {[connection, string]} An encrypted connection and the tag of the `Crypto` used
```
vasco-santos: No returning an array

src/dialer.js
```diff
@@ -0,0 +1,100 @@
...
+   *
+   * @async
+   * @param {PeerInfo} peerInfo The address to dial
+   * @param {object} [options]
+   * @param {AbortSignal} [options.signal] An AbortController signal
+   * @returns {Promise<Connection>}
+   */
+  async connectToPeer (peerInfo, options = {}) {
+    const addrs = peerInfo.multiaddrs.toArray()
+    // TODO: Allow for parallel dials to a peer
```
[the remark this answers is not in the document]
jacobheun: Some of these are existing issues, so I will make sure we have issues for them. Things that need to be done before this releases I will add Trello cards for since we're micro tracking there.

src/upgrader.js
jacobheun: multistream usage is different also, listener vs dialer and `select` vs `handle`

src/upgrader.js
jacobheun: This should just throw, there's no need to check the result, fixing this.

jacobheun:
Since `libp2p-interfaces` has been released, I also switch out `interface-connection` for that.

src/upgrader.js
vasco-santos: From what I am seeing, I don't see those differences:

```js
// UpgradeInbound

  const muxer = new Muxer({
    // Run anytime a remote stream is opened
    onStream: async muxedStream => {
      const mss = new Multistream.Listener(muxedStream)
      log('inbound: incoming stream opened on %s', protocol)
      connection.addStream(stream, protocol)
      this._onStream({ stream, protocol })
    },
    // Run anytime a stream ends
    onStreamEnd: muxedStream => {
      connection.removeStream(muxedStream.id)
    }
  })

```

```js
// UpgradeOutbound
  const muxer = new Muxer({
    // Run anytime a remote stream is created
    onStream: async muxedStream => {
      const mss = new Multistream.Listener(muxedStream)
      log('outbound: incoming stream opened on %s', protocol)
      connection.addStream(stream, protocol)
      this._onStream({ stream, protocol })
    },
    // Run anytime a stream closes
    onStreamEnd: muxedStream => {
      connection.removeStream(muxedStream.id)
    }
  })

```

Same for `const newStream`

jacobheun: Ah yes, you're right, I was looking at the wrong thing.

jacobheun: The main issue with moving the muxer creation into a new function is that it's heavily reliant on the scope it's currently in. For example, it needs access to the connection, which can't be created until the muxer is made.

However, I do think we could move the whole of connection creation into its own function, which would keep the scope we need and do the deduplication.

PR #463, opened 2019-10-13
fix: performance bottleneck in stat.js

betamos:
Array.shift seems to be very slow, perhaps linear, on some engines, resulting in  _update consuming a lot of CPU.

I transferred a large payload of 2.4GB between two machines running Node 12 on a local network and attached a Chrome profiler which showed a hotspot on the target machine of about 15s in this single line of code (the shift operation). The total transfer time was about 45s.

The CPU consumption was so high that the target machine was not able to consume the stream as new data was received, resulting in the TCP window filling up and causing network congestion. This was at speeds around 60MB/s on a high end MBP.

After this change, the time for the loop was down to about 1s. This might sound like a big improvement but 1s is still a lot for stats that are not important to all clients; perhaps consider allowing an option for turning stats off completely.

### DIFF
--- a/src/switch/stats/stat.js
+++ b/src/switch/stats/stat.js
@@ -139,10 +139,10 @@ class Stats extends EventEmitter {
     this._timeout = null
     if (this._queue.length) {
       let last
-      while (this._queue.length) {
-        const op = last = this._queue.shift()
-        this._applyOp(op)
+      for (last of this._queue) {
+        this._applyOp(last)
       }
+      this._queue = []
 
       this._updateFrequency(last[2]) // contains timestamp of last op
 


betamos:
Thanks for the swift response and merge. I didn't realize Stats was load bearing, then perhaps it is wiser to make some of the heavier features configurable at least.

Btw, yesterday it looked like the Windows tests were failing due to this change, but now they look fine. Any idea what happened?

jacobheun:
> Thanks for the swift response and merge. I didn't realize Stats was load bearing, then perhaps it is wiser to make some of the heavier features configurable at least.

Yeah, Stats definitely needs some love. We're going to be working on the Connection Manager soon and will hopefully work out a minimum set of needed stats (if any) that are required for it to manage connections properly. From there we can shut everything else off by default I think.

> Btw, yesterday it looked like the Windows tests were failing due to this change, but now they look fine. Any idea what happened?

I should have mentioned I reran them. Due to some timing issues on the windows CI environment the tests can be flakey for stats with when they are checked. I'm planning to get that resolved when I go through it as part of the async refactor in the next couple of weeks.

PR #466, opened 2019-10-20
docs: add stream wrapping example

jacobheun:
docs: add iterable types from alanshaw's gist

This adds a new, and the first, example for streaming iterables. I was working on the crypto interface and needed to do this, so sharing it here.

I also pulled over snippets of the iterables types from @alanshaw's gist to avoid being reliant on that always existing, but I made sure to link to it to credit his work.

### DIFF
--- a/doc/STREAMING_ITERABLES.md
+++ b/doc/STREAMING_ITERABLES.md
@@ -7,7 +7,7 @@
 - [Iterable Streams](#iterable-streams)
   - [Table of Contents](#table-of-contents)
   - [Usage Guide](#usage-guide)
-    - [Wrapping Duplex Iterables](#wrapping-duplex-iterables)
+    - [Transforming Bidirectional Data](#transforming-bidirectional-data)
   - [Iterable Stream Types](#iterable-stream-types)
     - [Source](#source)
     - [Sink](#sink)
@@ -17,9 +17,9 @@
 
 ## Usage Guide
 
-### Wrapping Duplex Iterables
+### Transforming Bidirectional Data
 
-Sometimes you may need to wrap an existing duplex stream in order to perform incoming and outgoing [transforms](#transform) on data. This type of wrapping is commonly used in encryption streams. Using [it-pair][it-pair] and [it-pipe], we can do this rather easily, given an existing [duplex iterable](#duplex).
+Sometimes you may need to wrap an existing duplex stream in order to perform incoming and outgoing [transforms](#transform) on data. This type of wrapping is commonly used in stream encryption/decryption. Using [it-pair][it-pair] and [it-pipe][it-pipe], we can do this rather easily, given an existing [duplex iterable](#duplex).
 
 ```js
 const duplexPair = require('it-pair/duplex')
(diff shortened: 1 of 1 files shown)


doc/STREAMING_ITERABLES.md
```diff
@@ -7,11 +7,144 @@
...
+    - [Transform](#transform)
+    - [Duplex](#duplex)
   - [Iterable Modules](#iterable-modules)
 
 ## Usage Guide
 
-**Coming Soon!**
+### Wrapping Duplex Iterables
+
+Sometimes you may need to wrap an existing duplex stream in order to perform incoming and outgoing [transforms](#transform) on data. This type of wrapping is commonly used in encryption streams. Using [it-pait][it-pair] and [it-pipe], we can do this rather easily, given an existing [duplex iterable](#duplex).
```
vasco-santos:
```suggestion
Sometimes you may need to wrap an existing duplex stream in order to perform incoming and outgoing [transforms](#transform) on data. This type of wrapping is commonly used in encryption streams. Using [it-pair][it-pair] and [it-pipe], we can do this rather easily, given an existing [duplex iterable](#duplex).
```

doc/STREAMING_ITERABLES.md
```diff
@@ -7,11 +7,144 @@
...
+
+Sometimes you may need to wrap an existing duplex stream in order to perform incoming and outgoing [transforms](#transform) on data. This type of wrapping is commonly used in encryption streams. Using [it-pair][it-pair] and [it-pipe], we can do this rather easily, given an existing [duplex iterable](#duplex).
+
+```js
+const duplexPair = require('it-pair/duplex')
+const pipe = require('it-pipe')
+
+// Wrapper is what we will write and read from
+// This gives us two duplex iterables that are internally connected
+const wrapper = duplexPair()
```
alanshaw: You might want to name these pairs to make it easier to understand in the example.

```suggestion
const [client, server] = duplexPair()
```

doc/STREAMING_ITERABLES.md
```diff
@@ -7,11 +7,144 @@
...
+)
+
+// We can now read and write from the other half of our pair
+pipe(
+  ['some data'],
+  wrapper[1],
+  async (source) => {
+    for await (const chunk of source) {
+      // Chunk will be a BufferList, we can slice it to get the underlying buffer
+      console.log('Data: %s', chunk.slice())
```
alanshaw: You could just call `toString()`. In the example there's nothing to say what `existingDuplex` is yielding so personally I wouldn't complicate it.

https://github.com/rvagg/bl#toString

```suggestion
      console.log('Data: %s', chunk.toString())
```

doc/STREAMING_ITERABLES.md
```diff
@@ -7,11 +7,144 @@
...
+    - [Source](#source)
+    - [Sink](#sink)
+    - [Transform](#transform)
+    - [Duplex](#duplex)
   - [Iterable Modules](#iterable-modules)
 
 ## Usage Guide
 
-**Coming Soon!**
+### Wrapping Duplex Iterables
```
alanshaw: I'd express this in terms of what it enables you to do. Typically people know what they want to do but don't know how to do it.

doc/STREAMING_ITERABLES.md
jacobheun: I went with `internal` and `external`. I don't feel great about these names, but the pairing is used to have 1 half integrate with the external duplex connection and the other half is what we would use internally to read/write data to.

Issue #453, opened 2019-09-19
Merge peer-id & peer-info into one, Merge peer-book into js-libp2p


daviddias:
We've talked this in a PR but creating a issue so that we don't miss it.

- [ ] Merge peer-id & peer-info into one
- [x] Merge peer-book into js-libp2p

daviddias:
While at it, rename the module name to be `libp2p-peer`

jacobheun:
I think what I'd really like to see is for peer-info to go away altogether. What it does could and should, imo, be managed by the peer-book (PeerStore, which would be great to rename as part of pulling this into libp2p to avoid confusion with go and the specs). Everything ultimately gets housed in the PeerStore anyway, so fully centralizing it there would be ideal, and it would remove a layer of complexity.

cc: @vasco-santos

vasco-santos:
I agree with jacob here! Peerbook should own the responsibilities of `peer-info`.

+1 on renaming this to `PeerStore` and I want to get this work done as a groundwork to get to [libp2p/specs#164](https://github.com/libp2p/specs/issues/164)

daviddias:
> I think what I'd really like to see is for peer-info to go away altogether. What it does could and should, imo, be managed by the peer-book

Just note that there are multiple modules that require the concept of a `Peer` but they don't need to know about a `PeerStore`. That said, I suggest to give it a try and see how viable it is on the refactor to remove the _Class_ `Peer` (prev peer-info) altogether, I might be missing some nuances that might make it nicer.

> (PeerStore, which would be great to rename as part of pulling this into libp2p to avoid confusion with go and the specs).

👍🏽

daviddias:
Now that PeerStore is becoming part of js-libp2p with the async await refactor. Can we also get peer-id and peer-info merged as well?

vasco-santos:
Yes, I have on my list to tackle that soon :)

PR #467, opened 2019-10-23
refactor: pubsub

vasco-santos:
This PR aims to add the pubsub subsystem into `js-libp2p`.

In order to make it possible, the `peer-store` was created, as well as the `registrar`. 

The `peer-store` will replace `peer-book`, using the same name as `go-libp2p`, being this the initial step for [libp2p/js-libp2p#453](https://github.com/libp2p/js-libp2p/issues/453), as well as for the new version of `peer-store`, which we will be working on after the refactor.

The `registrar` aims to be a bridge between all subsystems and `js-libp2p` (`peer-store` + `connection-manager`). Before this refactor, `js-libp2p` subsystems like `pubsub` and `dht` were responsible for tracking new peers that appear in the network and try to dial them using their respective protocols. Other than that, each subsystem was responsible for managing the connections and dials they were doing. This way, we aim to centralize this logic and enable subsystems to only care about their work, and leave the dialing and connection management for who should take care of them. It is also important pointing out that this is the first step to enable us to enhance the `connection-manager`. More context on [libp2p/notes#13](https://github.com/libp2p/notes/issues/13)

Needs:

- [ ] https://github.com/ChainSafe/gossipsub-js/pull/49


src/registrar.js
```diff
@@ -0,0 +1,121 @@
...
+    // }
+
+    // open new stream
+    // await Promise.all()
+
+    // TODO: return list?
+    return []
+  }
+
+  unregister (multicodecs) {
```
jacobheun: Per the above I think this should take a registration id.

src/peer-store.js
```diff
@@ -0,0 +1,154 @@
...
+   * @param {PeerInfo} peerInfo
+   */
+  update (peerInfo) {
+    const recorded = this.peers.get(peerInfo.id.toB58String())
+
+    // pass active connection state
+    const ma = peerInfo.isConnected()
+    if (ma) {
+      recorded.connect(ma)
+    }
```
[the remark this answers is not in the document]
vasco-santos: The `connect` does the `multiaddr` validation, if this is what you are saying.

And yes, I agree that we need to take into account those cases, should I add a `TODO:` comment?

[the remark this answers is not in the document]
jacobheun: We don't need to take any action on this, just getting my thoughts down while reviewing. We'll need to create a larger issue to work on multiaddr confidence.

src/registrar.js
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  /**
+   * Register handlers for a set of multicodecs given
+   * @param {Array<string>|string} multicodecs
+   * @param {object} handlers
+   * @param {function} handlers.onConnect
+   * @param {function} handlers.onDisconnect
+   * @param {object} topology
+   * @return {Array<object>}
```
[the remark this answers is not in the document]
vasco-santos: My idea was to return the connected peers that we already knew that have the protocol. But, we can only return the id and call the `onConnect` with them.

src/pubsub.js
```diff
@@ -105,61 +44,46 @@ module.exports = (node, Pubsub, config) => {
...
 
-      const subscriptions = Array.from(pubsub.subscriptions)
-
-      nextTick(() => callback(null, subscriptions))
-    }),
+      return pubsub.getTopics()
+    },
 
-    peers: promisify((topic, callback) => {
+    getPeersSubscribed: (topic) => {
```
jacobheun: Would we ever have pubsub peers that aren't subscribed, if so, are those different? Just trying to get a better understanding of what this is expected to return. Might be good to add some jsdocs to the methods missing them.

test/pubsub/implementations.node.js
```diff
@@ -0,0 +1,101 @@
...
+
+describe('Pubsub subsystem is able to use different implementations', () => {
+  let peerInfo, remotePeerInfo
+  let libp2p, remoteLibp2p
+  let remAddr
+
+  beforeEach(async () => {
+    [peerInfo, remotePeerInfo] = await Promise.all([
+      PeerInfo.create(),
+      PeerInfo.create()
```
jacobheun: Since the peers fixture util you added exists now, it would be good to use that. Creating default PeerInfo's can add up in the test suite due to the key generation delay.

test/pubsub/implementations.node.js
```diff
@@ -0,0 +1,101 @@
...
+      libp2p.start(),
+      remoteLibp2p.start()
+    ])
+
+    const libp2pId = libp2p.peerInfo.id.toB58String()
+    remAddr = remoteLibp2p.transportManager.getAddrs()[0]
+
+    const connection = await libp2p.dialProtocol(remAddr, multicodec)
+
+    await new Promise((resolve) => setTimeout(resolve, 1000))
```
jacobheun: What is the delay here for, a comment would be helpful? We have the `delay` module installed for dev, so you could use that instead.

test/pubsub/implementations.node.js
```diff
@@ -0,0 +1,101 @@
...
+      expect(msg.data.toString()).to.equal(data)
+      defer.resolve()
+    })
+
+    // wait for remoteLibp2p to know about libp2p subscription
+    await pWaitFor(() => {
+      const subscribedPeers = remoteLibp2p.pubsub.getPeersSubscribed(topic)
+      return subscribedPeers.includes(libp2pId)
+    })
+    await new Promise((resolve) => setTimeout(resolve, 1000))
```
jacobheun: If we're waiting for the subscribe to show up do we still need this delay?

test/pubsub/operation.node.js
```diff
@@ -0,0 +1,174 @@
...
+
+describe('Pubsub subsystem operates correctly', () => {
+  let peerInfo, remotePeerInfo
+  let libp2p, remoteLibp2p
+  let remAddr
+
+  beforeEach(async () => {
+    [peerInfo, remotePeerInfo] = await Promise.all([
+      PeerInfo.create(),
+      PeerInfo.create()
```
jacobheun: Same thing here, use fixture utils.

test/pubsub/operation.node.js
```diff
@@ -0,0 +1,174 @@
...
+      libp2p = await create(mergeOptions(subsystemOptions, {
+        peerInfo
+      }))
+
+      remoteLibp2p = await create(mergeOptions(subsystemOptions, {
+        peerInfo: remotePeerInfo
+      }))
+
+      await libp2p.start()
+      await remoteLibp2p.start()
```
jacobheun: These should probably get started in parallel

test/pubsub/operation.node.js
```diff
@@ -0,0 +1,174 @@
...
+
+      expect(connection).to.exist()
+      expect(libp2p.pubsub._pubsub.peers.size).to.be.eql(1)
+      expect(remoteLibp2p.pubsub._pubsub.peers.size).to.be.eql(1)
+
+      expect(libp2p.registrar.onConnect.callCount).to.equal(1)
+      expect(remoteLibp2p.registrar.onConnect.callCount).to.equal(1)
+    })
+
+    it('should receive pubsub messages', async () => {
```
jacobheun: Is this test redundant with the implementations test? Looks to be validating the same things.

src/pubsub.js
vasco-santos: We can have peers connected with pubsub, but that did not subscribe the topic provided.

I will add jsdocs here and more concrete stuff like examples in the docs PR

test/pubsub/implementations.node.js
vasco-santos: it was just for debugging and I forgot to remove it. Will do it now

test/pubsub/operation.node.js
```diff
@@ -0,0 +1,172 @@
...
+        return subscribedPeers.includes(libp2pId)
+      })
+      remoteLibp2p.pubsub.publish(topic, data)
+
+      await defer.promise
+    })
+  })
+
+  // TODO: Needs identify push
+  describe.skip('pubsub started after connect', () => {
```
jacobheun: This suite should be able to be run now.

test/registrar/registrar.spec.js
```diff
@@ -32,53 +33,7 @@ describe('registrar', () => {
...
-
-      try {
-        registrar.register(topologyProps)
-      } catch (err) {
-        expect(err).to.exist()
-        return
-      }
-      throw new Error('should fail to register a protocol if the onDisconnect handler is not provided')
-    })
+    // TODO: not valid topology
```
jacobheun: Did you want to add something for this?

src/upgrader.js
```diff
@@ -186,7 +186,7 @@ class Upgrader {
           const { stream, protocol } = await mss.handle(Array.from(this.protocols.keys()))
           log('%s: incoming stream opened on %s', direction, protocol)
           connection.addStream(stream, protocol)
-          this._onStream({ connection, stream, protocol })
+          this._onStream({ connection, stream, protocol, remotePeer })
```
jacobheun: There shouldn't be a need to pass the `remotePeer`, as it's on the connection, `connection.remotePeer`. Same with the other two changes below

src/registrar.js
```diff
@@ -109,6 +111,11 @@ class Registrar {
    * @return {string} registrar identifier
    */
   register (topology) {
+    assert(
+      Topology.isTopology(topology) ||
+      MulticodecTopology.isMulticodecTopology(topology),
+      'topology must be an instance of interfaces/topology')
```
jacobheun: `is-class` should handle the base class extension check, so we should just need to check that it is a Topology, `Topology.isTopology(topology)`

vasco-santos: Cool, that works :)

PR #469, opened 2019-10-25
refactor: crypto and pnet

jacobheun:
* Updates crypto usage to async/await
* Implements the [plaintext 2 protocol](https://github.com/libp2p/specs/tree/master/plaintext) for testing
* Updates pnet to use async/await


src/insecure/plaintext.js
```diff
@@ -0,0 +1,88 @@
+'use strict'
+
+const protobuf = require('protons')
+const protocol = '/plaintext/2.0.0'
```
vasco-santos: Can you put this const after all `require`? I prefer to have all the `require` first and then constants

src/insecure/plaintext.js
```diff
@@ -0,0 +1,88 @@
...
+const PeerId = require('peer-id')
+const debug = require('debug')
+const log = debug('libp2p:plaintext')
+log.error = debug('libp2p:plaintext:error')
+const { UnexpectedPeerError, InvalidCryptoExchangeError } = require('libp2p-interfaces/src/crypto/errors')
+
+const {
+  Exchange,
+  KeyType
+} = protobuf(`
```
vasco-santos: should we isolate the protobuf in its own file? I think it is easier to have it because we can easily find the proto file and it does not add noise to the source code

src/insecure/plaintext.js
```diff
@@ -0,0 +1,88 @@
...
+  const response = (await lp.decodeFromReader(shake.reader).next()).value
+  const id = Exchange.decode(response.slice())
+  log('read pubkey exchange from peer %j', remoteId)
+
+  let peerId
+  try {
+    peerId = await PeerId.createFromPubKey(id.pubkey.Data)
+  } catch (err) {
+    log.error(err)
+    throw new InvalidCryptoExchangeError('Remote did not provide their public key')
```
vasco-santos:
```suggestion
    throw new InvalidCryptoExchangeError('Remote did not provide its public key')
```

src/pnet/crypto.js
```diff
@@ -1,52 +1,46 @@
...
-  )
+  return (source) => (async function * () {
+    for await (const chunk of source) {
+      yield Buffer.from(xor.update(chunk.slice()))
+    }
+  })()
 }
 
 /**
  * Creates a pull stream to decrypt messages in a private network
```
vasco-santos:
```suggestion
  * Creates a stream iterable to decrypt messages in a private network
```

src/pnet/crypto.js
```diff
@@ -1,52 +1,46 @@
...
 const xsalsa20 = require('xsalsa20')
 const KEY_LENGTH = require('./key-generator').KEY_LENGTH
 
 const log = debug('libp2p:pnet')
 log.trace = debug('libp2p:pnet:trace')
-log.err = debug('libp2p:pnet:err')
+log.error = debug('libp2p:pnet:err')
 
 /**
  * Creates a pull stream to encrypt messages in a private network
```
vasco-santos:
```suggestion
  * Creates a stream iterable to encrypt messages in a private network
```

src/pnet/index.js
```diff
@@ -1,12 +1,17 @@
...
 const Errors = require('./errors')
-const State = require('./state')
-const decodeV1PSK = require('./crypto').decodeV1PSK
+const {
+  createBoxStream,
+  createUnboxStream,
+  decodeV1PSK
+} = require('./crypto')
+const handshake = require('it-handshake')
+const NONCE_LENGTH = require('./key-generator').NONCE_LENGTH
```
vasco-santos:
```suggestion
const { NONCE_LENGTH } = require('./key-generator')
```

test/dialing/direct.node.js
```diff
@@ -171,66 +173,90 @@ describe('Dialing (direct, TCP)', () => {
...
-      }
+  describe('libp2p.dialer', () => {
+    let peerInfo
+    let remotePeerInfo
+    let libp2p
+    let remoteLibp2p
+    let remoteAddr
+
+    before(async () => {
+      const [peerId, remotePeerId] = await Promise.all([
```
vasco-santos: Should we have an util `createPeerInfoFromFixtures`? We could provide `n` and receive `n` `PeerInfos`. This is being replicated in a lot of tests

test/dialing/direct.node.js
```diff
@@ -171,66 +173,90 @@ describe('Dialing (direct, TCP)', () => {
...
-  it('should use the dialer for connecting', async () => {
-    libp2p = new Libp2p({
-      peerInfo,
-      modules: {
-        transport: [Transport],
-        streamMuxer: [Muxer],
-        connEncryption: [mockCrypto]
-      }
+    after(async () => {
+      await remoteLibp2p.stop()
```
vasco-santos: Just do:

```js
after(() => remoteLibp2p.stop())
```

.aegir.js
```diff
@@ -1,23 +1,38 @@
...
-    upgrader: mockUpgrader,
-    onConnection: () => {}
+  // Use the last peer
+  const peerId = await PeerId.createFromJSON(Peers[Peers.length - 1])
+  const peerInfo = new PeerInfo(peerId)
+  peerInfo.multiaddrs.add(MULTIADDRS_WEBSOCKETS[0])
+
+  libp2p = new Libp2p({
+    peerInfo,
+    modules: {
```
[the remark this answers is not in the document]
vasco-santos: I created this on libp2p/js-libp2p#470](https://github.com/libp2p/js-libp2p/pull/470)

test/dialing/direct.node.js
vasco-santos: I created this on [libp2p/js-libp2p#470](https://github.com/libp2p/js-libp2p/pull/470)

.travis.yml
```diff
@@ -20,7 +20,7 @@ jobs:
   include:
     - stage: check
       script:
-        - npx aegir build --bundlesize
+        # - npx aegir build --bundlesize
```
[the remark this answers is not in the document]
jacobheun: That might be good, it will just keep failing until we can remove all the non async modules. I originally just commented this out to avoid the 500 failures I was getting at the time from the bundlesize site.

test/dialing/direct.node.js
jacobheun: Okay, once we get through these next few merges I can go through and do some deduplication.

PR #470, opened 2019-11-01
feat: peer store

vasco-santos:
Initial implementation of `peer-store`, which will replace [libp2p/js-peer-book](https://github.com/libp2p/js-peer-book), using the same name as `go-libp2p`, being this the initial step for [libp2p/js-libp2p#453](https://github.com/libp2p/js-libp2p/issues/453), as well as for the new version of `peer-store`, which we will be working on after the refactor.

Work extracted from #467


test/peer-store/peer-store.spec.js
```diff
@@ -0,0 +1,161 @@
...
+    const [peerInfo] = await peerUtils.createPeerInfo(1)
+    const id = peerInfo.id.toB58String()
+
+    let removed = peerStore.remove(id)
+    expect(removed).to.eql(false)
+
+    // Put the peer in the store
+    peerStore.put(peerInfo)
+
+    const peers = peerStore.getAllArray()
```
jacobheun: I'd just access peers directly. As mentioned I would prefer to not have `getAllArray`, no need to create an array when we have the Map.
```suggestion
    const peers = peerStore.peers.size
```

test/peer-store/peer-store.spec.js
```diff
@@ -0,0 +1,161 @@
...
+    expect(peers).to.have.lengthOf(1)
+
+    removed = peerStore.remove(id)
+    expect(removed).to.eql(true)
```
jacobheun: Need to validate it's actually been removed.
```suggestion
    expect(removed).to.eql(true)
    expect(peerStore.peers.size).to.equal(0)
```

src/peer-store/index.js
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    throw errCode(new Error('PeerInfo was not found'), 'ERR_NO_PEER_INFO')
+  }
+
+  /**
+   * Get an array with all peers known.
+   * @returns {Array<PeerInfo>}
+   */
+  getAllArray () {
```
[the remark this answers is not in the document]
vasco-santos: sounds good

PR #471, opened 2019-11-02
feat: registrar

vasco-santos:
The `registrar` aims to be a bridge between all subsystems and `js-libp2p` (`peer-store` + `connection-manager`). Before this refactor, `js-libp2p` subsystems like `pubsub` and `dht` were responsible for tracking new peers that appear in the network and try to dial them using their respective protocols. Other than that, each subsystem was responsible for managing the connections and dials they were doing. This way, we aim to centralize this logic and enable subsystems to only care about their work, and leave the dialing and connection management for who should take care of them. It is also important pointing out that this is the first step to enable us to enhance the `connection-manager`. 

More context on [libp2p/notes#13](https://github.com/libp2p/notes/issues/13) and [cryptpad](https://cryptpad.fr/code/#/2/code/edit/lakY6PqQ+tf2E+Ezds0PDErK/)

This PR is on top of #470 branch and was also extracted from #467


src/index.js
```diff
@@ -54,25 +55,30 @@ class Libp2p extends EventEmitter {
...
-    this._switch = new Switch(this.peerInfo, this.peerBook, this._options.switch)
+    this._switch = new Switch(this.peerInfo, this.peerStore, this._options.switch)
 
     // Setup the Upgrader
     this.upgrader = new Upgrader({
       localPeer: this.peerInfo.id,
       onConnection: (connection) => {
         const peerInfo = getPeerInfo(connection.remotePeer)
+
+        this.peerStore.put(peerInfo)
```
jacobheun: The put here doesn't do anything. `connection.remotePeer` is a PeerId, so `getPeerInfo` is just going to check the peerStore anyway. I would delete the `getPeerInfo` usage and just do: 
```js
const peerInfo = this.peerStore.get(connection.remotePeer.toB58String())
```

src/index.js
```diff
@@ -54,25 +55,30 @@ class Libp2p extends EventEmitter {
...
+        this.registrar.onConnect(peerInfo, connection)
         this.emit('peer:connect', peerInfo)
       },
       onConnectionEnd: (connection) => {
         const peerInfo = getPeerInfo(connection.remotePeer)
```
jacobheun: Same thing here, I think we can get rid of getPeerInfo eventually in favor of direct peerStore access. Especially when we know we're going to have a PeerId. `getPeerInfo` is really only useful when we have no idea what the type is.

src/registrar.js
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Responsible for notifying registered protocols of events in the network.
+ */
+class Registrar {
+  /**
+   * @param {PeerStore} peerStore
+   * @constructor
+   */
+  constructor (peerStore) {
```
jacobheun: Let's pass an object, we may need to expand this in the future.
```suggestion
  constructor ({ peerStore }) {
```

src/registrar.js
```diff
@@ -0,0 +1,165 @@
...
+     * @type {Map<string, conn>}
+     */
+    this.connections = new Map()
+
+    /**
+     * Map of topologies per multicodec
+     *
+     * @type {Map<string, object>}
+     */
+    this.multicodecTopologies = new Map()
```
jacobheun: This should just be `topologies`, the registrar shouldn't care about multicodecs, it really shouldn't have any knowledge of the internals of a Topology. It should just care about its desired connection min/max values and the current peers the topology has identified as being in their desired topology.

src/registrar.js
```diff
@@ -0,0 +1,165 @@
...
+   * TODO: this should live in the ConnectionManager
+   * @param {PeerInfo} peerInfo
+   * @param {Connection} conn
+   * @returns {void}
+   */
+  onConnect (peerInfo, conn) {
+    assert(PeerInfo.isPeerInfo(peerInfo), 'peerInfo must be an instance of peer-info')
+    assert(Connection.isConnection(conn), 'conn must be an instance of interface-connection')
+
+    this.connections.set(peerInfo.id.toB58String(), conn)
```
jacobheun: This could be a problem, there may be more than 1 connection to a peer. It's not the desired behavior but it can happen and we don't want to lose tracking of this. The connection Map should probably take an Array of connections for now.

src/registrar.js
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  /**
+   * Remove a disconnected peer from the record
+   * TODO: this should live in the ConnectionManager
+   * @param {PeerInfo} peerInfo
+   * @param {Error} [error]
+   * @returns {void}
+   */
+  onDisconnect (peerInfo, error) {
```
jacobheun: This needs to account for multiple connections to a peer. We should only trigger disconnect behaviors if we actually have 0 open connections to the peer. It might be helpful to also pass the connection to this method, as removing that specific connection would be easier.

src/registrar.js
```diff
@@ -0,0 +1,165 @@
...
+
+    this.connections.delete(peerInfo.id.toB58String())
+  }
+
+  /**
+   * Get a connection with a peer.
+   * @param {PeerInfo} peerInfo
+   * @returns {Connection}
+   */
+  getPeerConnection (peerInfo) {
```
jacobheun: Maybe just `getConnection`? Adding `peer` feels unnecessary.

src/registrar.js
```diff
@@ -0,0 +1,165 @@
...
+  /**
+   * Register handlers for a set of multicodecs given
+   * @param {Array<string>|string} multicodecs
+   * @param {object} handlers
+   * @param {function} handlers.onConnect
+   * @param {function} handlers.onDisconnect
+   * @param {object} topology properties for topology
+   * @return {string} registrar identifier
+   */
+  register (multicodecs, handlers, topology = {}) {
```
jacobheun: `topology` should be required. The registrar shouldn't create them, it should just track and inform them. With that the `handlers` shouldn't need to be passed, because they're part of the topology.

`multicodecs` shouldn't be needed and should exist solely in the `topology`

src/registrar.js
```diff
@@ -0,0 +1,165 @@
...
+  /**
+   * Unregister topology.
+   * @param {string} id registrar identifier
+   * @return {void}
+   */
+  unregister (id) {
+    const topology = this.multicodecTopologies.get(id)
+
+    if (!topology) {
+      throw errCode(new Error('no registrar found for the provided id'), 'ERR_NO_REGISTRAR')
```
jacobheun: I don't think we need to throw here. If it's not being tracked, that's technically a successful unregistration imo.

test/registrar/registrar.node.js
```diff
@@ -0,0 +1,67 @@
...
+
+  it('should inform registrar of a new connection', async () => {
+    libp2p = new Libp2p(mergeOptions(baseOptions, {
+      peerInfo
+    }))
+
+    sinon.spy(libp2p.registrar, 'onConnect')
+    sinon.spy(libp2p.registrar, 'onDisconnect')
+
+    const connection = await libp2p.dial(remoteAddr)
```
jacobheun: Mock the internal dial as we don't actually need to connect to another peer for these tests.

src/connection-manager/topology.js
```diff
@@ -0,0 +1,102 @@
...
+    this.peers = new Map()
+
+    // Handlers
+    this._onConnect = onConnect
+    this._onDisconnect = onDisconnect
+
+    this._onProtocolChange = this._onProtocolChange.bind(this)
+
+    // Set by the registrar
+    this._peerStore = null
```
jacobheun: After thinking about this more we should just reference the peerstore from the registrar, `this.registrar.peerStore`.

src/connection-manager/topology.js
```diff
@@ -0,0 +1,102 @@
...
+    if (existingPeer && protocols.filter(protocol => this.multicodecs.includes(protocol)).length === 0) {
+      this._onDisconnect({
+        peerInfo
+      })
+    }
+
+    // New to protocol support
+    for (const protocol of protocols) {
+      if (this.multicodecs.includes(protocol)) {
+        this.tryToConnect(peerInfo, this.registrar.getPeerConnection(peerInfo))
```
jacobheun: What's the point of `tryToConnect`? Ultimately I think this line should just be an add to `this.peers`. When registration happens we could use `Proxy` to have the registrar watch for changes on each Toplogies peer set. When that changes it can take action on prioritizing peers for the connection manager to connect/disconnect to.

The flow I am thinking of is:
1. A new peer is discovered and we are under the min threshold for a topology (multicodec, bootstrap, root, etc)
1. The connection manager takes the highest priority peers and dials to them (for a typical node on startup, this would be the bootstrap peers, and any other priority peers like ipfs preload nodes)
1. When the connection is made, the registrar is notified and in turn notifies the topologies, so they can take action if desired
  a. A multicodec topology could just ignore this, because it cares about protocol updates
1. When identify runs, the protocol change is triggered by the peer store, and the multicodec does a check for its multicodecs
  a. If the multicodec is present, it adds the peer to its list of peers
  b. The peer set change, causes the registrar to update the priority of that peer
1. The reprioritized peer will now be connected/disconnected accordingly by the connection manager

src/index.js
vasco-santos: When a `peerB` dials `peerA` the `peerStore` for `peerA` does not have the `peerB` on it. That way, we would have `peerInfo` = `undefined`
`getPeerInfo` puts in the `peerStore` if it is provided. I changed for the following:

```js
const peerInfo = getPeerInfo(connection.remotePeer, this.peerStore)

this.registrar.onConnect(peerInfo, connection)
this.emit('peer:connect', peerInfo)
```

src/index.js
vasco-santos: `getPeerInfo` created the `peer-info` if it does not exist, while `peerStore.get` returns undefined if it does not exist in the store

src/index.js
vasco-santos: Oh wait, if I do that `getPeerInfo` tries to `put` invalid `PeerInfo` to the store in some scenarios from other tests.

src/registrar.js
vasco-santos: In this scenario, I am also using `multicodecs` to try to connect to the already connected peers using that multicodec. If I don't use them here, I will need to try `onConnect` to all the peers, which will fail the `newStream` inside Pubsub (as an example). Other than that, we will need to also try to connect with peers from the peerStore that are not connected

src/connection-manager/topology.js
vasco-santos: The point for `tryToConnect` is basically the following use case:

When a peer starts a subsystem (`pubsub` as an example), the peer may already be connected to several peers. Take into account that someone enables the subsystem in runtime. By that time, the peer will know and be connected to several peers (some of them may support the protocol of the subsystem). Once the subsystem is started and the registrar is notified of it, the registrar must inform the topology that it is already connected with some peers (`tryToConnect`), and the topology will do `onConnect` if it is inside the thresholds. This way, the `tryToConnect` is basically a `subsystem.onConnect` with a validation before, in order to guarantee that we are inside the thresholds.

The flow that you described looks good, except for the initial part that I just described. However, I am not sure if we are getting too complex with the `Proxy`.

test/registrar/registrar.node.js
vasco-santos: The test needs the remotePeer to receive the dial. I just changed it accordingly

src/registrar.js
jacobheun: This is only needed right now because Identify hasn't been refactored yet, right? The Registrar shouldn't need to care about multicodecs once identify is running. I think eventually we could have a peer discovery mechanism that checked the multicodecs our node supports (those added via `libp2p.handle`) and querying the network or rendezvous servers for nodes that also support those. This still wouldn't require getting multicodecs from the topology as we'll know them from `handle`/`unhandle` calls.

src/connection-manager/topology.js
jacobheun: >  the registrar must inform the topology that it is already connected with some peers

I would expect that when a topology is registered, the registrar would simply trigger `onConnect` for any existing connections. It would basically be a replay mechanism for lazily started systems.

> However, I am not sure if we are getting too complex with the Proxy

It's not really complex though, it's just another way of creating an observer. We could do it with hooks or events, but the Proxy is cleaner IMO. The registrar should be the only thing observing `topology.peers`.

vasco-santos: > I would expect that when a topology is registered, the registrar would simply trigger onConnect for any existing connections. It would basically be a replay mechanism for lazily started systems.

I see 2 problems with that. The topology needs to update its `peers` set and if we are already connected with a lot of peers, we may immediately exceed the threshold. That's why I would rather have a `tryToConnect` method.

> It's not really complex though, it's just another way of creating an observer. We could do it with hooks or events, but the Proxy is cleaner IMO. The registrar should be the only thing observing topology.peers.

I think we can iterate on that later, as we will not need this until getting to connectionManager

src/registrar.js
vasco-santos: I think this [conversation](https://github.com/libp2p/js-libp2p/pull/471#discussion_r342507203) has more insights. 

Basically, I am using them for getting the already connected peers.

I have a doubt though: Once pubsub starts, a `identify-push` will occur. If we have already connected and identified other peers (who support pubsub), we will already have that information right? This way, `protocols` will not change and we need to try to connect using pubsub on those peers.

src/connection-manager/topology.js
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Check if a new peer support the multicodecs for this topology.
+   * @param {Object} props
+   * @param {PeerInfo} props.peerInfo
+   * @param {Array<string>} props.protocols
+   */
+  _onProtocolChange ({ peerInfo, protocols }) {
+    const existingPeer = this.peers.get(peerInfo.id.toB58String())
+
+    protocols.filter(protocol => this.multicodecs.includes(protocol))
```
jacobheun: This should get set to some `hasProtocol` or similar variable, it's not doing anything right now and is being called again in the if statement below.

src/peer-store/index.js
```diff
@@ -0,0 +1,190 @@
...
+            peerInfo: obj,
+            multiaddrs: value.toArray()
+          })
+        } else if (prop === 'protocols') {
+          this.emit('change:protocols', {
+            peerInfo: obj,
+            protocols: Array.from(value)
+          })
+        }
+        return true
```
jacobheun: I believe you will need to call Reflect.set here for the changes to occur.
```suggestion
        return Reflect.set(obj, prop, value)
```

src/peer-store/index.js
```diff
@@ -0,0 +1,190 @@
...
+      }
+    })
+
+    this.peers.set(peerInfo.id.toB58String(), peerProxy)
```
jacobheun: A problem with this approach is that the original `PeerInfo` that is being added isn't tracked, the new instance is. So if the original instance has any updates occur to it after the add, the changes will never trigger. I think it would be safer to require changes to a peer happen through the peer store (the deprecation of PeerInfo) rather than trying to track the metadata like this, as we don't have complete control of the instances being used.

src/registrar.js
```diff
@@ -0,0 +1,139 @@
...
+    assert(Connection.isConnection(conn), 'conn must be an instance of interface-connection')
+
+    const id = peerInfo.id.toB58String()
+    const storedConn = this.connections.get(id)
+
+    if (storedConn) {
+      storedConn.push(conn)
+    } else {
+      this.connections.set(id, [conn])
+    }
```
jacobheun: The topologies are never informed of new connections here as they are with onDisconnect.

src/registrar.js
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  /**
+   * Add a new connected peer to the record
+   * TODO: this should live in the ConnectionManager
+   * @param {PeerInfo} peerInfo
+   * @param {Connection} conn
+   * @returns {void}
+   */
+  onConnect (peerInfo, conn) {
```
jacobheun: What do you think about changing the connection tracking to just use the connection id, since we have those now? `onConnect` could just take the connection, since we can determine the PeerInfo using its `remotePeer` and the PeerStore. This would also simplify having to track multiple connections per peer.

src/connection-manager/topology.js
```diff
@@ -0,0 +1,128 @@
...
+
+    this._onProtocolChange = this._onProtocolChange.bind(this)
+  }
+
+  set registrar (registrar) {
+    this._registrar = registrar
+    this._registrar.peerStore.on('change:protocols', this._onProtocolChange)
+
+    // Add connected peers to the topology
+    this._addConnectedPeers()
```
jacobheun: I think we could simply the peer/connection add logic a bit. Instead of `_addConnectedPeers` and `tryToConnect` we can just have an `_updatePeers` function that takes an Iterable of PeerInfo objects. This would allow us to call it with the peerStore peer set values, and with `[peerInfo]` from protocolChange.

```js
this._updatePeers(this._registrar.peerStore.peers.values())

_updatePeers (peerInfoIterable) {
  for (const peerInfo of peerInfoIterable) {
    if (this.multicodecs.filter(multicodec => peerInfo.protocols.has(multicodec))) {
      // Add the peer regardless of whether or not there is currently a connection
      this.peers.set(peerInfo.id.toB58String(), peerInfo)
      // If there is a connection, call _onConnect
      const connection = this._registrar.getConnection(peerInfo)
      connection && this._onConnect(connection)
    } else {
      // Remove any peers we might be tracking that are no longer of value to us
      this.peers.delete(peerInfo.id.toB58String())
    }
  }
}
```

src/connection-manager/topology.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  /**
+   * Notify protocol of peer disconnected.
+   * @param {PeerInfo} peerInfo
+   * @param {Error} [error]
+   * @returns {void}
+   */
+  disconnect (peerInfo, error) {
+    if (this.peers.delete(peerInfo.id.toB58String())) {
```
jacobheun: We shouldn't remove the peer from tracking here. The only time a peer should get removed is if they don't support our protocol, or we've decided they shouldn't be included in our Topology, such as with certain mesh overlays. For multicodec, the peer still has value to us, we just don't have a connection anymore. By keeping it in our peer set we are telling the registrar this peer is useful to us, so if we go below our connection threshold, Connection Manager knows that specific peer has value. If no Topology has it in their peer sets, it won't get dialed (unless it is randomly dialed when we are below our global threshold of minimum connections).

src/peer-store/index.js
vasco-santos: > I think it would be safer to require changes to a peer happen through the peer store

Yes, that is why I made this change now, as a first step to deprecate `PeerInfo`.
With the new instance of `peer-info`, we cannot change the added `PeerInfo` from outside (same as when we destruct it).

Then, I added the proxy so that when people do the `get` the changes are triggered for now. Once we get rid of the `peer-info` we just need to also remove the `Proxy`. 

I can change this to not have the `peer-info` if you prefer to make it now, but I would rather do it in a subsequent PR instead of tracking code changes in the 3 current PRs. I think that with the current code we have guarantees that what is in the `peer-store` is always correct according to the events we emit.

src/registrar.js
vasco-santos: When a new connection happens, I was expecting that the `peer-store` will trigger the `protocol-change` after identify and we do not need to inform the topologies with the connection.

src/connection-manager/topology.js
vasco-santos: I was deleting this from a perspective of when a peer gets offline and not on a perspective of we intentionally disconnect to it. I agree with you in a perspective of intentionally disconnecting, but what are your thoughts on a peer getting offline?

src/registrar.js
vasco-santos: If we move in that direction, when we want to get a connection from a peer, we will need to iterate the collection and get the `remotePeer` from the connection and then check if it matches. This simplifies the multiple connections tracking but makes the `getConnection` more complex. I am not sure if we should change it or not, but we can do it if you feel this is a more viable approach.

jacobheun: Yeah, that's a good point, I think we can leave this for now. We could potentially create an id dictionary for quicker lookups both ways, but that might be overkill at the moment.

src/registrar.js
jacobheun: We can leave this for now. There may be Topologies in the future that need to known when connections are created but that don't care about the multicodecs, but I don't think we have any of those planned right now. A static peer set Topology, such as a Bootstrap or Priority Node topology won't care about the multicodecs, but they also won't care about connections, just expressing their desire to be connected to certain peers.

src/peer-store/index.js
vasco-santos: I was expecting the following flow to be required for updates:

```js
libp2p.peerStore.add(peerInfo)
peerInfo.protocols.add('/new/protocol')
libp2p.peerStore.add(peerInfo)
```

To update the record

jacobheun: 👍 , should the second call be `libp2p.peerStore.update(peerInfo)`? `.put` would work fine for both.

vasco-santos: `update` would be better, but the `put` would call the `update`

src/registrar.js
vasco-santos: It makes sense yes. We iterate on this with the new versions of `connMgr` + `peer-store`

src/connection-manager/topology.js
vasco-santos: All good, I will make this change!

PR #472, opened 2019-11-03
docs: new api

vasco-santos:
New API documentation for the async refactor


API.md
```diff
@@ -0,0 +1,269 @@
...
+* [`stop`](#stop)
+* [`dial`](#dial)
+* [`dialProtocol`](#dialProtocol)
+* [`hangUp`](#hangUp)
+* [`handle`](#handle)
+* [`unhandle`](#unhandle)
+* [`peerRouting.findPeer`](#peerRouting.findPeer)
+* [`contentRouting.findProviders`](#contentRouting.findProviders)
+* [`contentRouting.provide`](#contentRouting.provide)
+* [`dht.put`](#dht.put)
```
daviddias: Suggestion: Avoid exposing the DHT methods as they are just methods from one impl of Peer Routing and Content Routing and instead stick with a clean Peer Routing & Content Routing interfaces.

API.md
```diff
@@ -0,0 +1,269 @@
...
+
+```js
+const Libp2p = require('libp2p')
+
+// specify options
+const options = {}
+
+// create libp2p
+const libp2p = new Libp2p (options)
+```
```
daviddias: js-ipfs had multiple rounds of iteration till it got its .create constructor https://github.com/ipfs/js-ipfs#ipfs-constructor, @alanshaw can shine some light here on why that was important. I would vote to have the same way of spawning a node just for consistency.

vasco-santos: I agree with the `create`. In `js-libp2p` we currently have `createLibp2p` and are moving into `create` now, same way as IPFS does.

I think this new API is consistent with `js-ipfs`. Same way, we also support using the `constructor`

API.md
vasco-santos: Shouldn't we make that clear instead?

As there are projects, including `IPFS` that use the `dht` API directly [interface-js-ipfs-core/SPEC/DHT.md](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/DHT.md), I think we should make your point clear in the docs, but also have the `dht` available

daviddias: A way to declare that the API is internal is by putting a `_` before the name (e.g. `_dht`), this way, folks are less tempted to use it.

API.md
```diff
@@ -0,0 +1,269 @@
...
+
+### `dial(peer, options)`
+
+#### Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| peer | [PeerInfo](https://github.com/libp2p/js-peer-info), [PeerId](https://github.com/libp2p/js-peer-id), [multiaddr](https://github.com/multiformats/js-multiaddr), `string` | peer to dial |
+| [options] | `Object` | dial options |
+| [options.abort] | `AbortSignal` | abort signal |
```
alanshaw: FYI these are named `options.signal` in a bunch of other places...including the `fetch` API. Personally I'd keep it consistent and familiar.

API.md
```diff
@@ -0,0 +1,269 @@
...
+|------|-------------|
+| `Promise` | Promise resolves when the node is fully stopped |
+
+#### Example
+
+```js
+const { create } = require('libp2p')
+
+// ...
+const libp2p = await create(options)
```
alanshaw: A matter of preference but IMHO calling create on a default export is clearer in examples (and in app code):

```suggestion
const libp2p = await Libp2p.create(options)
```

API.md
```diff
@@ -0,0 +1,269 @@
+# API
+
+* [`create`](#create)
```
alanshaw: This is a static function but the rest are instance methods however they appear the same here. Might be a good idea to differentiate :)

API.md
```diff
@@ -0,0 +1,552 @@
...
+  * [`dialProtocol`](#dialProtocol)
+  * [`hangUp`](#hangUp)
+  * [`handle`](#handle)
+  * [`unhandle`](#unhandle)
+  * [`peerRouting.findPeer`](#peerRouting.findPeer)
+  * [`contentRouting.findProviders`](#contentRouting.findProviders)
+  * [`contentRouting.provide`](#contentRouting.provide)
+  * [`_dht.put`](#_dht.put)
+  * [`_dht.get`](#_dht.get)
+  * [`_dht.getMany`](#_dht.getMany)
```
alanshaw: If these are not for public consumption they should not be included in the docs.

API.md
```diff
@@ -0,0 +1,552 @@
...
+const Libp2p = require('libp2p')
+
+// specify options
+const options = {}
+
+// create libp2p
+const libp2p = await Libp2p.create(options)
+```
+
+Note: It is important pointing out that with `create`, the `PeerInfo` option is not required and will be generated if it is not provided.
```
alanshaw:
```suggestion
Note: The `PeerInfo` option is not required and will be generated if it is not provided.
```

API.md
```diff
@@ -0,0 +1,552 @@
...
+
+Creates an instance of Libp2p.
+
+`create(options)`
+
+#### Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| options | `Object` | libp2p options |
```
alanshaw: What are the options?

API.md
```diff
@@ -0,0 +1,552 @@
...
+#### Example
+
+```js
+const Libp2p = require('libp2p')
+
+// specify options
+const options = {}
+
+// create libp2p
+const libp2p = new Libp2p (options)
```
alanshaw:
```suggestion
const libp2p = new Libp2p(options)
```

API.md
```diff
@@ -0,0 +1,552 @@
...
+```
+
+Required keys in the `options` object:
```
alanshaw: You just said above that `peerInfo` was not required.

API.md
```diff
@@ -0,0 +1,552 @@
...
+
+- `peerInfo`: instance of [PeerInfo][] that contains the [PeerId][], Keys and [multiaddrs][multiaddr] of the libp2p Node.
+- `modules.transport`: An array that must include at least 1 transport, such as `libp2p-tcp`.
```
alanshaw:
```suggestion
- `modules.transport`: An array that must include at least 1 transport, such as [`libp2p-tcp`](https://github.com/libp2p/js-libp2p-tcp).
```

API.md
```diff
@@ -0,0 +1,552 @@
...
+// specify options
+const options = {}
+
+// create libp2p
+const libp2p = await Libp2p.create(options)
+```
+
+Note: It is important pointing out that with `create`, the `PeerInfo` option is not required and will be generated if it is not provided.
+
+As an alternative, it is possible to create a Libp2p instance with the constructor:
```
alanshaw: In js-ipfs we did this:

Screenshot 2019-11-29 at 15 29 49

...then clicking to expand:

Screenshot 2019-11-29 at 15 30 10

This helps to reduce the amount of info the reader has to read and understand (whilst still keeping it available) for methods that are deprecated or not recommended.

API.md
alanshaw: There are many more options right? Are you planning on documenting them or at least linking to somewhere that does?

API.md
```diff
@@ -0,0 +1,552 @@
...
+
+`dial(peer, options)`
+
+#### Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| peer | [PeerInfo](https://github.com/libp2p/js-peer-info), [PeerId](https://github.com/libp2p/js-peer-id), [multiaddr](https://github.com/multiformats/js-multiaddr), `string` | peer to dial |
+| [options] | `Object` | dial options |
+| [options.signal] | `AbortSignal` | abort signal |
```
alanshaw:
```suggestion
| [options.signal] | [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) | An `AbortSignal` instance obtained from an [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) that can be used to abort the connection before it completes |
```

API.md
```diff
@@ -0,0 +1,552 @@
...
+`dialProtocol(peer, protocols, options)`
+
+#### Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| peer | [PeerInfo](https://github.com/libp2p/js-peer-info), [PeerId](https://github.com/libp2p/js-peer-id), [multiaddr](https://github.com/multiformats/js-multiaddr), `string` | peer to dial |
+| protocols | `String|Array<String>` | Strings that identifies the protocols (e.g '/ipfs/bitswap/1.1.0') |
+| [options] | `Object` | dial options |
+| [options.signal] | `AbortSignal` | abort signal |
```
alanshaw:
```suggestion
| [options.signal] | [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) | An `AbortSignal` instance obtained from an [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) that can be used to abort the connection before it completes |
```

API.md
```diff
@@ -0,0 +1,552 @@
...
+#### Example
+
+```js
+// ...
+const conn = await libp2p.dial(remotePeerInfo)
+```
+
+### dialProtocol
+
+Dials to another peer in the network and selects a protocol to communicate with that peer.
```
alanshaw: How does this work? Is it a stream that gets returned not a conneciton?

API.md
```diff
@@ -0,0 +1,552 @@
...
+
+`dialProtocol(peer, protocols, options)`
+
+#### Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| peer | [PeerInfo](https://github.com/libp2p/js-peer-info), [PeerId](https://github.com/libp2p/js-peer-id), [multiaddr](https://github.com/multiformats/js-multiaddr), `string` | peer to dial |
+| protocols | `String|Array<String>` | Strings that identifies the protocols (e.g '/ipfs/bitswap/1.1.0') |
```
alanshaw: From https://github.com/multiformats/js-multistream-select/#dialerselectprotocols:

```suggestion
| protocols | `String|Array<String>` |  A list of protocols (or single protocol) to negotiate with. Protocols are attempted in order until a match is made. (e.g '/ipfs/bitswap/1.1.0') |
```

API.md
```diff
@@ -0,0 +1,552 @@
...
+
+| Type | Description |
+|------|-------------|
+| `Promise<Connection>` | Promise resolves with the [Connection](https://github.com/libp2p/js-interfaces/tree/master/src/connection) instance |
+
+#### Example
+
+```js
+// ...
+const conn = await libp2p.dial(remotePeerInfo)
```
alanshaw: Maybe add some more to this example on how someone might use the connection.

API.md
```diff
@@ -0,0 +1,552 @@
...
+}
+
+libp2p.handle('/echo/1.0.0', handler)
+```
+
+### unhandle
+
+Unregisters all handlers with the given protocols
+
+`libp2p.unhandle(protocols)`
```
alanshaw: What if I only want to unregister a specific handle?

API.md
```diff
@@ -0,0 +1,552 @@
...
+// ...
+const peerInfo = await libp2p.peerRouting.findPeer(peerId, options)
+```
+
+### contentRouting.findProviders
+
+Iterates over all content routers in series to find providers of the given key.
+Once a content router succeeds, the iteration will stop.
+
+`libp2p.contentRouting.findProviders(cid, options)`
```
alanshaw: Please can this return an async iterable that yields `PeerInfo`.

API.md
```diff
@@ -0,0 +1,552 @@
...
+| `Promise<void>` | Promise resolves once notifications are sent |
+
+#### Example
+
+```js
+// ...
+await libp2p.contentRouting.provide(cid)
+```
+
+### _dht.put
```
alanshaw: Put each of these headings in backticks to avoid problems like this :)

API.md
```diff
@@ -0,0 +1,552 @@
...
+#### Example
+
+```js
+// ...
+
+const key = '/key'
+const { from, val } = await libp2p._dht.get(key)
+```
+
+### pubsub.getPeersSubscribed
```
alanshaw: Commonly being subscribed to something is referred to as being a subscriber. Suggestion: `pubsub.getSubscribers(topic)`?

API.md
```diff
@@ -0,0 +1,552 @@
...
+const data = Buffer.from('data')
+
+await libp2p.pubsub.publish(topic, data)
+```
+
+### pubsub.subscribe
+
+Subscribes the given handler to a pubsub topic.
+
+`libp2p.pubsub.subscribe(topic)`
```
alanshaw:
```suggestion
`libp2p.pubsub.subscribe(topic, handler)`
```

API.md
```diff
@@ -0,0 +1,552 @@
...
+
+#### Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| topic | `string` | topic to subscribe |
+| handler | `function(<Object>)` | handler for new data on topic |
```
alanshaw: Please describe the `Object`!

API.md
```diff
@@ -0,0 +1,552 @@
...
+
+#### Returns
+
+| Type | Description |
+|------|-------------|
+| `void` |  |
```
alanshaw: Just checking this is not an async operation?

API.md
```diff
@@ -0,0 +1,552 @@
...
+}
+
+libp2p.pubsub.subscribe(topic, handler)
+```
+
+### pubsub.unsubscribe
+
+Unsubscribes the given handler from a pubsub topic. If no handler is provided, all handlers for the topic are removed.
+
+`libp2p.pubsub.unsubscribe(topic)`
```
alanshaw:
```suggestion
`libp2p.pubsub.unsubscribe(topic, handler)`
```

API.md
vasco-santos: I understand that! 

However, taking into account that project such as `js-ipfs` use these API methods, I also don't like that we do not document something that will probably be widely used (also discussed previously on https://github.com/libp2p/js-libp2p/pull/472#discussion_r345011828)

API.md
vasco-santos: I would like to have a specific MD doc for the options/libp2p configuration, where this would point out. As there are so many options and instructions for configuration, I think it will be easier for people reading the docs to digest a lighter API, while being able to consult the options when needed. What do you think?

API.md
vasco-santos: cool, I will copy this approach! Thanks 🙏

API.md
vasco-santos: `peerInfo` is not required for the `create`. With the alternative approach,i.e. the `constructor` it is required. This logic is prior do this refactor, if you want to discuss this, maybe we can open an issue for it?

API.md
vasco-santos: answered on previous comment. Let me know your thoughts there

API.md
vasco-santos: we currently do not support it, but maybe you are right and we should support it. @jacobheun any opinions here for we not supporting it?

API.md
vasco-santos: Oh, I agree with you on this. We already have this on all the `pubsub` repos but I will see what I can do

API.md
alanshaw: Ok that makes sense.

API.md
```diff
@@ -0,0 +1,552 @@
...
+
+| Name | Type | Description |
+|------|------|-------------|
+| protocols | `Array<String>|String` | protocols to unregister |
+
+#### Example
+
+```js
+// ...
+libp2p.unhandle('/echo/1.0.0')
```
jacobheun: We'll convert to an array so I think we should be more specific in examples of our preference, even though we support just passing a string. I'd like to gradually reduce the footprint of accepted params as it increases the likelihood of bugs.
```suggestion
libp2p.unhandle(['/echo/1.0.0'])
```

API.md
```diff
@@ -0,0 +1,552 @@
...
+```
+
+### peerRouting.findPeer
+
+Iterates over all peer routers in series to find the given peer.
```
jacobheun:
```suggestion
Iterates over all peer routers in series to find the given peer. If the DHT is enabled, it will be tried first.
```

API.md
jacobheun: :+1: We haven't refactored this quite yet, should happen in the next day or two, and I intend to make all array based queries async iterable.

API.md
```diff
@@ -0,0 +1,552 @@
...
+
+```js
+// ...
+const peerInfo = await libp2p.peerRouting.findPeer(peerId, options)
+```
+
+### contentRouting.findProviders
+
+Iterates over all content routers in series to find providers of the given key.
+Once a content router succeeds, the iteration will stop.
```
jacobheun:
```suggestion
Once a content router succeeds, the iteration will stop. If the DHT is enabled, it will be queried first.
```

API.md
vasco-santos: I am updating to `contentRouting`

API.md
vasco-santos: I will create that in a new PR next week

API.md
vasco-santos: Yes, I forgot to update the docs here. We return `{ stream, protocol }`

API.md
vasco-santos: I will add a note on `handle`

API.md
vasco-santos: no, we do not need any async operation for any of the pubsub routers in this operation

API.md
vasco-santos: Will be changed in the next commit :)

README.md
```diff
@@ -209,321 +208,7 @@ class Node extends Libp2p {
...
-
-#### Enforcement
-
-Libp2p provides support for connection protection, such as for private networks. You can enforce network protection by setting the environment variable `LIBP2P_FORCE_PNET=1`. When this variable is on, if no protector is set via `options.connProtector`, Libp2p will throw an error upon creation.
-
-#### Protectors
-
-Some available network protectors:
-* [libp2p-pnet](https://github.com/libp2p/js-libp2p/tree/master/src/pnet)
+See [API.md](./doc/API.md).
```
vasco-santos: I will add in the section before the configuration doc, which will also be linked from the API when relevant

API.md
vasco-santos: So, after writting that in #495 , I think I will document the first level properties of the options object here and point to the other doc

doc/API.md
```diff
@@ -0,0 +1,611 @@
...
+
+const key = '/key'
+const { from, val } = await libp2p.contentRouting.get(key)
+```
+
+### pubsub.getPeersSubscribed
+
+Gets a list of the peer-ids that are subscribed to one topic.
+
+`libp2p.pubsub.getPeersSubscribed(topic)`
```
jacobheun:
```suggestion
`libp2p.pubsub.getSubscribers(topic)`
```

doc/API.md
```diff
@@ -0,0 +1,611 @@
...
+#### Example
+
+```js
+// ...
+
+const key = '/key'
+const { from, val } = await libp2p.contentRouting.get(key)
+```
+
+### pubsub.getPeersSubscribed
```
jacobheun:
```suggestion
### pubsub.getSubscribers
```

doc/API.md
```diff
@@ -0,0 +1,611 @@
...
+  * [`hangUp`](#hangUp)
+  * [`handle`](#handle)
+  * [`unhandle`](#unhandle)
+  * [`peerRouting.findPeer`](#peerRouting.findPeer)
+  * [`contentRouting.findProviders`](#contentRouting.findProviders)
+  * [`contentRouting.provide`](#contentRouting.provide)
+  * [`contentRouting.put`](#contentRouting.put)
+  * [`contentRouting.get`](#contentRouting.get)
+  * [`contentRouting.getMany`](#contentRouting.getMany)
+  * [`pubsub.getPeersSubscribed`](#pubsub.getPeersSubscribed)
```
jacobheun:
```suggestion
  * [`pubsub.getSubscribers`](#pubsub.getSubscribers)
```

PR #371, opened 2019-06-25
docs: update delegate example to include ipfs setup instructions

jacobheun:
This updates the delegate routing example to include instructions for setting up the go-ipfs node locally to support the needed API endpoints.

I also added an "Add Data" field to the UI and instructions to demonstrate content providing being delegable.

PR #475, opened 2019-11-10
Fix ipfs cat in delegated routing example 

phillmac:
`ipfs.files.cat` is incorrect. the correct function is `ipfs.cat`

### DIFF
--- a/examples/delegated-routing/src/App.js
+++ b/examples/delegated-routing/src/App.js
@@ -44,7 +44,7 @@ class App extends Component {
       isLoading: this.state.isLoading + 1
     })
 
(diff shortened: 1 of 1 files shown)


jacobheun:
The example is currently using js-ipfs ^0.34.x, did you verify the example runs with that? I don't recall when the API changed. 

There is an open PR, https://github.com/libp2p/js-libp2p/pull/371,  to update to a newer js-ipfs version, but the go-ipfs setup instructions need to get corrected. That PR uses the correct `cat` api.

phillmac:
Yup, I didn't modify the versions at all, using ~0.34.4, but I was getting `ipfs.files.cat` is not a function.
If the other pr is going to be merged then this one's redundant.

PR #477, opened 2019-11-14
refactor: circuit relay to async

jacobheun:
## Relay
- Refactors Circuit Relay to use async/await
- Makes Circuit Relay abide by the [Transport Interface](https://github.com/libp2p/js-interfaces/tree/master/src/transport)
  - *Note*: I left out the interface tests for the time being because they'll need to be hacked a bit to get them working because they only leverage a single transport/peer. Circuit needs 3 to work properly.
- All Circuit Relay connections are now created as [libp2p Connections](https://github.com/libp2p/js-interfaces/tree/master/src/connection). Their [MultiaddrConnection](https://github.com/libp2p/interface-transport#multiaddrconnection) is the circuit relay stream of the underlying connection they are using.
- **new** Adding a known relay address to your multiaddrs on startup (`/p2p-circuit/<relay multiaddr>/p2p/<relay peer id>` will cause the node to dial that relay.
- **new** If a relay dial fails and we were not already connected to the relay, we will now properly disconnect. This wasn't the case before.

**Note:** This does not include active relay dialing (Relays wont dial unconnected destination peers). However, this is a low priority. In most instances we shouldn't try dialing a peer we're not connected to, because this means the dialing peer should also be able to connect to them. The only instance this is not the case, is if the relay is in a private network with the destination node and is the only exposed node. Private networks should be explicitly connecting to their exposed relay in this instance.

## Other fixes
- The Registrar was not getting rid of its tracked connections when libp2p stopped. It now has a `close` method which is called on `libp2p.stop()`.
- Transport Manager was clearing it's listeners map on stop, which is created on libp2p instantiation, not startup which caused issues with restarts. This is fixed.
- If an internal dial error already had an error code it was being overridden with a generic dial error code. This makes debugging more difficult. Now the error will be rethrown if it already has a code


src/circuit/circuit/stream-handler.js
```diff
@@ -1,139 +1,88 @@
...
 
   /**
    * Get the raw Connection
    *
-   * @returns {null|Connection|*}
+   * @returns {*} A duplex iterable
    */
   getRawConn () {
-    return this.conn
+    return this.stream
```
vasco-santos: Can we leverage this to better name this function?

src/circuit/index.js
```diff
@@ -1,3 +1,187 @@
...
+const { relay: multicodec } = require('./multicodec')
+const createListener = require('./listener')
+const { handleCanHop, handleHop, hop } = require('./circuit/hop')
+const { handleStop } = require('./circuit/stop')
+const StreamHandler = require('./circuit/stream-handler')
+const toConnection = require('./stream-to-conn')
+
+class Circuit {
+  /**
+   * Creates an instance of Dialer.
```
vasco-santos:
```suggestion
   * Creates an instance of Circuit.
```

src/circuit/index.js
```diff
@@ -1,3 +1,187 @@
...
+          connection,
+          request,
+          streamHandler,
+          circuit
+        })
+        break
+      }
+      default: {
+        log('Request of type %s not supported', request.type)
+        virtualConnection = null
```
vasco-santos: do we need this line?

src/circuit/listener.js
```diff
@@ -1,94 +1,43 @@
...
+module.exports = (circuit) => {
+  const listener = new EventEmitter()
+  const listeningAddrs = new Map()
 
   /**
    * Add swarm handler and listen for incoming connections
    *
-   * @param {Multiaddr} ma
+   * @param {Multiaddr} addr
    * @param {Function} callback
```
vasco-santos:
```suggestion
```

src/circuit/listener.js
```diff
@@ -107,42 +56,12 @@ module.exports = (swarm, options, connHandler) => {
...
```
vasco-santos:
```suggestion
```

src/circuit/stream-to-conn.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
+
+const abortable = require('abortable-iterator')
+const log = require('debug')('libp2p:circuit:stream')
+
+// Convert a duplex iterable into a MultiaddrConnection
+// https://github.com/libp2p/interface-transport#multiaddrconnection
+module.exports = ({ stream, remoteAddr, localAddr }, options) => {
```
vasco-santos:
```suggestion
module.exports = ({ stream, remoteAddr, localAddr }, options = {}) => {
```

src/circuit/stream-to-conn.js
```diff
@@ -0,0 +1,51 @@
...
+  options = options || {}
```
vasco-santos:
```suggestion
```

src/circuit/circuit/stream-handler.js
jacobheun: It's not being used anywhere so I will just remove it.

jacobheun:
I switched tests over to use chai-as-promised to get better visibility into test failures, and then fixed an issue with needing to reset the multiaddrs of the test nodes for the relay suite due to the fix in #485

PR #478, opened 2019-11-19
refactor: core async

jacobheun:
This cleans up a few things in the main libp2p file for async.

* Removed the state machine as we really don't need it
* Removed promisify usage
* Ping is currently disabled, that will be refactored in a followup PR
* Autodial is now working with the async code
* `hangUp` now correctly interacts with the registrar to close connections
  * `hangUp` now only accepts a PeerId to simplify the api
* Fixed an issue with TransportManager for dial only nodes

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,21 +1,13 @@
 'use strict'
 
-const FSM = require('fsm-event')
 const { EventEmitter } = require('events')
 const debug = require('debug')
 const log = debug('libp2p')
 log.error = debug('libp2p:error')
-const errCode = require('err-code')
-const promisify = require('promisify-es6')
-
-const each = require('async/each')
 
 const PeerInfo = require('peer-info')
 const multiaddr = require('multiaddr')
-const Switch = require('./switch')
-const Ping = require('./ping')
 
-const { emitFirst } = require('./util')
 const peerRouting = require('./peer-routing')
 const contentRouting = require('./content-routing')
 const dht = require('./dht')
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -233,14 +157,23 @@ class Libp2p extends EventEmitter {
   }
 
   /**
-   * Starts the libp2p node and all sub services
+   * Starts the libp2p node and all subsystems
```
```suggestion
   * Starts the libp2p node and all its subsystems
```

test/peer-discovery/index.spec.js
```diff
@@ -0,0 +1,45 @@
...
+const baseOptions = require('../utils/base-options.browser')
+const { createPeerInfo } = require('../utils/creators/peer')
+
+describe('peer discovery', () => {
+  let peerInfo
+  let remotePeerInfo
+  let libp2p
+
+  before(async () => {
+    [peerInfo, remotePeerInfo] = await createPeerInfo(2)
```
vasco-santos:
```suggestion
    [peerInfo, remotePeerInfo] = await createPeerInfoFromFixture(2)
```

test/peer-discovery/index.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+const chai = require('chai')
+chai.use(require('dirty-chai'))
+const { expect } = chai
+const sinon = require('sinon')
+const defer = require('p-defer')
+
+const Libp2p = require('../../src')
+const baseOptions = require('../utils/base-options.browser')
+const { createPeerInfo } = require('../utils/creators/peer')
```
vasco-santos:
```suggestion
const { createPeerInfoFromFixture } = require('../utils/creators/peer')
```

PR #480, opened 2019-11-21
refactor: dht

vasco-santos:
Async refactor for the new async version of `libp2p-kad-dht`.

As discussed on #472 , the `dht` api is provided under `_dht` as it should not be directly used.

Needs:

- [x] https://github.com/libp2p/js-libp2p-kad-dht/pull/153


src/dht.js
```diff
@@ -1,43 +1,82 @@
...
-    put: promisify((key, value, callback) => {
-      if (!node._dht) {
-        return nextTick(callback, errCode(new Error(messages.DHT_DISABLED), codes.DHT_DISABLED))
-      }
+module.exports = (node, DHT, config) => {
+  const dht = new DHT({
+    dialer: {
+      dial: (peer, options) => node.dial(peer, options),
+      dialProtocol: (peer, protocols, options) => {
+        const recordedPeer = node.peerStore.get(peer.toB58String())
```
vasco-santos: When the dial is happening, if we try to `dial` using the `peer-id`, the `peer-info` stored gets replaced, removing the `multiaddrs` that previously existed:

https://github.com/libp2p/js-libp2p/blob/refactor/async-await/src/get-peer-info.js#L36-L41 

I think we can tackle this once we get rid of `peer-info`

src/peer-store/index.js
```diff
@@ -37,24 +37,28 @@ class PeerStore extends EventEmitter {
...
-      this.update(peerInfo)
+      peer = this.update(peerInfo)
     } else {
-      this.add(peerInfo)
+      peer = this.add(peerInfo)
 
       // Emit the new peer found
       this.emit('peer', peerInfo)
     }
+    return peer
```
vasco-santos: `getPeerInfo` expects that `peerStore.put()` returns the peer

src/dht.js
```diff
@@ -1,43 +1,82 @@
...
 
-module.exports = (node) => {
-  return {
-    put: promisify((key, value, callback) => {
-      if (!node._dht) {
-        return nextTick(callback, errCode(new Error(messages.DHT_DISABLED), codes.DHT_DISABLED))
-      }
+module.exports = (node, DHT, config) => {
+  const dht = new DHT({
+    dialer: {
```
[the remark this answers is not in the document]
vasco-santos: fixed in the dht pr

PR #483, opened 2019-11-25
async await version of examples/echo

dirkmc:
Note: depends on
- ~fixing [this line](https://github.com/libp2p/js-libp2p/blob/master/src/get-peer-info.js#L41) which is coming in https://github.com/libp2p/js-libp2p/pull/480~
- ~Latest release of [js-libp2p-secio](https://github.com/libp2p/js-libp2p-secio) being deployed to npm~

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ We've come a long way, but this project is still in Alpha, lots of development i
 
 ## Background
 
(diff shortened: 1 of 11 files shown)


jacobheun:
#480 has been merged!

dirkmc:
Agreed 👍 

I made that change

PR #485, opened 2019-11-26
Replace peerInfo addresses with listen addresses

dirkmc:


vasco-santos:
@dirkmc fyi I needed these changes in another PR, and as I think you are ooo today, I just fixed the linting 😛

PR #486, opened 2019-11-28
feat: discovery modules

vasco-santos:
Made necessary changes to integrate discovery modules, including the DHT random-walk.

Needs:

- [x] https://github.com/libp2p/js-libp2p/pull/485


src/index.js
```diff
@@ -160,6 +160,7 @@ class Libp2p extends EventEmitter {
     log('libp2p is starting')
     try {
       await this._onStarting()
+      this._isStarted = true
```
jacobheun: I think we should put this in `_onDidStart` as that keeps all post startup changes there.

src/index.js
```diff
@@ -319,6 +332,11 @@ class Libp2p extends EventEmitter {
    * @private
    */
   _onDidStart () {
+    // Peer discovery
+    if (this._modules.peerDiscovery) {
+      this._setupPeerDiscovery()
+    }
```
jacobheun: Is this if necessary? I think it would be good to have any necessary checks inside of `_setupPeerDiscovery`. I think it will be much easier for us to avoid bugs if we have the functions handle their checks.

PR #489, opened 2019-12-01
refactor: async routing

vasco-santos:
Added `content-routing` and `peer-routing` with the `delegated` modules with the new async versions.

I also refactored the test creators we have, as I am thinking on extracting them to a new module `js-libp2p-test-utils`, as they can be used across multiple repos


test/utils/creators/peer.js
```diff
@@ -1,32 +1,71 @@
...
-async function createPeerInfoFromFixture (length) {
-  const peers = await createPeerIdsFromFixture(length)
+/**
+ * Create Peer-ids.
+ * @param {Object} [properties]
+ * @param {number} [properties.number] number of peers (default: 1).
+ * @param {boolean} [properties.fixture] use fixture for peer-id generation (default: true)
+ * @return {Promise<Array<PeerInfo>>}
+ */
+async function createPeerInfo ({ number = 1, fixture } = {}) {
```
jacobheun: `createPeerId` is setting fixture to true, but it might be good to also set the default here since the docs say the default is true. This will prevent any changes to `createPeerId` from affecting this.
```suggestion
async function createPeerInfo ({ number = 1, fixture = true } = {}) {
```

vasco-santos:
Per https://github.com/libp2p/js-libp2p/pull/472/files#r352196509 I also switched `dht.put`, `dht.get` and `dht.getMany` under `contentRouting`

src/dht.js
```diff
@@ -15,54 +11,6 @@ module.exports = (node, DHT, config) => {
...
-      if (!node.isStarted() || !dht.isStarted) {
-        throw errCode(new Error(messages.NOT_STARTED_YET), codes.DHT_NOT_STARTED)
-      }
-
-      return dht.getMany(key, nVals, options)
-    },
-
     _dht: dht,
 
     start: () => dht.start(),
```
jacobheun: Is it worth having this file anymore? It's not doing much and is just adding complexity (libp2p._dht._dht instead of libp2p._dht). We could just have `libp2p._dht` be the actual DHT implementation.

test/content-routing/content-routing.node.js
```diff
@@ -0,0 +1,327 @@
...
+      const deferred = pDefer()
+
+      sinon.stub(nodes[0]._dht._dht, 'findProviders').callsFake(function * () {
+        deferred.resolve()
+        yield
+      })
+
+      for await (const _ of nodes[0].contentRouting.findProviders()) { // eslint-disable-line no-unused-vars
+        break
+      }
```
jacobheun: Is this for loop necessary? It waits for the first call and then exists the for loop, but we really only care about waiting for the deferred. Shouldn't you just be able to have `nodes[0].contentRouting.findProviders()` with no await?

test/content-routing/content-routing.node.js
```diff
@@ -0,0 +1,327 @@
...
+      const deferred = pDefer()
+
+      sinon.stub(delegate, 'findProviders').callsFake(function * () {
+        deferred.resolve()
+        yield
+      })
+
+      for await (const _ of node.contentRouting.findProviders()) { // eslint-disable-line no-unused-vars
+        break
+      }
```
jacobheun: Same question here as above.

test/content-routing/utils.js
```diff
@@ -0,0 +1,35 @@
...
+
+const mergeOptions = require('merge-options')
+
+const baseOptions = {
+  modules: {
+    transport: [Transport],
+    streamMuxer: [Muxer],
+    connEncryption: [Crypto]
+  }
+}
```
jacobheun: What about pulling in the baseOptions from the global test utils?

test/peer-routing/utils.js
```diff
@@ -0,0 +1,35 @@
...
+
+const mergeOptions = require('merge-options')
+
+const baseOptions = {
+  modules: {
+    transport: [Transport],
+    streamMuxer: [Muxer],
+    connEncryption: [Crypto]
+  }
+}
```
jacobheun: Leverage global test utils baseOptions?

src/dht.js
vasco-santos: you are right, I will remove it

test/content-routing/content-routing.node.js
vasco-santos: I can await for next. Just call the `findProviders` does not iterate

src/content-routing.js
```diff
@@ -1,16 +1,19 @@
...
-const promisify = require('promisify-es6')
+const { messages, codes } = require('./errors')
+
+const all = require('async-iterator-all')
+const get = require('dlv')
+const pAny = require('p-any')
 
 module.exports = (node) => {
   const routers = node._modules.contentRouting || []
+  const dht = get(node, '_dht')
```
jacobheun: `node` will always exist so we can drop `dlv` now right?

vasco-santos: yap, forgot this! Thanks for the reminder

PR #490, opened 2019-12-02
feat: token dialer

jacobheun:
This converts the Queue based dialer into a TokenDialer. This PR includes a dedicated readme for the dialer to improve the visibility on how the dialer works, and is intended to work (for any future work, like Multiaddr Confidence).

**Note**: The current dialer readme is a draft to show the general intent. This will be cleaned up and improved over the course of this PR.

## What is this solving?
There are two major problems we have right now with a Queue based dial. 
1. Queue congestion. It's currently possible for the dial queue to get flooded with requests. While dials should eventually run, the visibility of when that is executed for users is non existent.
2. Parallel dials per peer. Rate limiting dials for a Queue of Queues gets convoluted. Each item in the main queue is a dial request. That dial request will often have multiple addresses it needs to dial. In order to efficiently connect to a peer, we should try multiple addresses simultaneously to get the quickest connection. If we run multiple dials in parallel for a request, that number should be reflected in the main queue to avoid resource overallocation. 

A TokenDialer solves both of these problems.
1. Queue Congestion. The queue is gone. DialRequests will use tokens to perform 1 dial per token. If there are no tokens left, a DialRequest can simply error, informing the user of the high number of dials currently in progress. This enables users to act accordingly based on the needs of their application.
2. Parallel dials per peer. Parallel dials are easy to track because DialRequests must have a token for each request they make in parallel. Since the code is all internal to libp2p, this is a trustful system that depends on the DialRequest not abusing the token system. Once a DialRequest is done, or it no longer needs some of the tokens, they can be released individually or all at once.


RELEASE.md
```diff
@@ -50,8 +50,8 @@ Would you like to contribute to the libp2p project and don't know how? Well, the
 - Check the issues with the `help wanted` label in the [libp2p repo](https://github.com/libp2p/js-libp2p/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22)
 - Join an IPFS All Hands, introduce yourself and let us know where you would like to contribute - https://github.com/ipfs/team-mgmt#all-hands-call
 - Hack with IPFS and show us what you made! The All Hands call is also the perfect venue for demos, join in and show us what you built
-- Join the discussion at http://discuss.ipfs.io/ and help users finding their answers.
-- Join the [⚡️ⒿⓈ Core Dev Team Weekly Sync 🙌🏽 ](https://github.com/ipfs/team-mgmt/issues/650) and be part of the Sprint action!
+- Join the discussion at http://discuss.libp2p.io/ and help users finding their answers.
+- Join the [⚡️IPFS Core Implementations Weekly Sync 🙌🏽 ](https://github.com/ipfs/team-mgmt/issues/992) and be part of the Sprint action!
```
vasco-santos: should we change this for the libp2p sync?

doc/DIALER.md
```diff
@@ -0,0 +1,36 @@
...
+
+**Synopsis**
+* All Dial Requests in js-libp2p must request a token(s) from the Dialer.
+  * The number of tokens requested should be between 1 and the PER_PEER_LIMIT max set in the Dialer.
+  * If the number of available tokens is less than requested, the Dialer may return less than requested.
+* The number of tokens a DialRequest obtains reflects the maximum number of parallel Multiaddr Dials it can make.
+* If no tokens are available a DialRequest should immediately end and throw. This deviates from the existing queue system to avoid queue congestion and provide more visibility to users.
+* As tokens are limited, DialRequests should be given a prioritized list of Multiaddrs to minimize the potential request time.
+* Once a single Multiaddr Dial has succeeded, all pending dials in that Dial Request should be aborted. All tokens should be immediately released to the Dialer.
+* If all Multiaddr Dials fail, or the DIAL_TIMEOUT max is reached for the entire DialRequest, all in progress dials for that DialRequest should be aborted. All tokens should immediately be released to the Dialer.
```
vasco-santos: Should we try to release tokens as we don't need them?

Example: DialRequest gets 3 tokens and the peer has 3 multiaddr. As soon as the first dial attempt fails (from the 3 parallel requests), can we immediatly release that resource?

jacobheun: Yes we should definitely do this. I will update to reflect this.

RELEASE.md
jacobheun: Yeah, I think that's a better move 👍

jacobheun:
@vasco-santos @alanshaw this is the general approach I am looking at for the token dialing. I still need to add more tests and would like to do some cleanup, but I wanted to share this before I start that to get some feedback on the approach.

Just a note, I moved the identify call out of the dialer, as it is more of a triggered action when we get new connections, than a dialer role. This will also make sure inbound connections are identified.

src/constants.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   DIAL_TIMEOUT: 30e3, // How long in ms a dial attempt is allowed to take
   MAX_COLD_CALLS: 50, // How many dials w/o protocols that can be queued
   MAX_PARALLEL_DIALS: 100, // Maximum allowed concurrent dials
+  PER_PEER_LIMIT: 4, // Allowed parallel dials per DialRequest
```
vasco-santos: I think it would be more coherent with the remaining names

```suggestion
  MAX_PER_PEER_DIALS: 4, // Allowed parallel dials per DialRequest
```

src/dialer.js
```diff
@@ -29,105 +30,106 @@ class Dialer {
...
    * dialed, it will be used for encryption verification.
    *
-   * @async
    * @param {Multiaddr} addr The address to dial
    * @param {object} [options]
    * @param {AbortSignal} [options.signal] An AbortController signal
    * @returns {Promise<Connection>}
    */
-  async connectToMultiaddr (addr, options = {}) {
+  connectToMultiaddr (addr, options = {}) {
```
vasco-santos: should we keep the `async` keyword here?

src/dialer.js
```diff
@@ -29,105 +30,106 @@ class Dialer {
...
     } catch (err) {
-      if (err.name === 'TimeoutError') {
-        controller.abort()
-        err.code = codes.ERR_TIMEOUT
-      }
-      log.error('Error dialing address %s,', addr, err)
+      log.error(err)
+      timeoutController.abort()
       throw err
     }
```
vasco-santos: Should we log success dial?
Also, maybe do the return after the catch instead, so that it is easier to find what is returned

doc/DIALER.md
```diff
@@ -0,0 +1,37 @@
...
+* If a Multiaddr Dial fails and there are no more dials to use its token, that token should be immediately released to the Dialer.
+
+## Multiaddr Confidence
+
+**Not Yet Implemented**
+
+An effective dialing system should involve the inclusion of a Confidence system for Multiaddrs. This enables ranking of Multiaddrs so that a prioritized list can be passed to DialRequests to maximize usage of Dialer Tokens.
+
+* All Multiaddrs SHOULD start with a confidence of 1 (give it some trust by default)
```
vasco-santos: where are you planning to keep track of this? I know that we do not need to tackle it, but I would like to have this designed in a way that it will be easy to add this

src/dialer/dial-request.js
```diff
@@ -0,0 +1,197 @@
...
+  /**
+   * @async
+   * @param {object} options
+   * @param {AbortSignal} options.signal An AbortController signal
+   * @param {number} options.timeout The max dial time for each request
+   * @returns {Connection}
+   */
+  async run (options) {
+    // Determine how many tokens we need
+    const tokensWanted = Math.min(this.addrs.length, this.dialer.perPeerLimit)
```
vasco-santos: The `dialer.getTokens` already checks `perPeerLimit`. 

I think that only the dialer should take care of this validation, as in the `dialRequest` perspective, it wants the the maximum possible and we may end up iterating in the future to "runtime" decisions, instead of a fixed limit

src/dialer/dial-request.js
```diff
@@ -0,0 +1,197 @@
...
+    // Get the tokens
+    const tokens = this.dialer.getTokens(tokensWanted)
+    // If no tokens are available, throw
+    if (tokens.length < 1) {
+      throw Object.assign(new Error('No dial tokens available'), { code: 'ERR_NO_DIAL_TOKENS' })
```
vasco-santos: why no use `err-code`?

src/dialer/dial-request.js
```diff
@@ -0,0 +1,197 @@
...
+
+    const promise = this.dialAction(addr, { signal, timeout: options.timeout })
+    return {
+      abort: () => controller.abort(),
+      promise
+    }
+  }
+}
+
+class DialResolver {
```
vasco-santos: Please move this to a new file

vasco-santos: We can have both inside a `dial-request` folder I think

src/dialer/token-holder.js
```diff
@@ -0,0 +1,63 @@
...
+   */
+  constructor (tokens, release) {
+    this.originalTokens = tokens
+    this.tokens = [...tokens]
+    this._release = release
+  }
+
+  /**
+   * Resolves a token once once is available. Once the token is no
+   * longer needed it MUST be release with `releaseToken()`.
```
vasco-santos: Imagine this scenario: We ask for 3 tokens (and we have 5 multiaddr to try), but we only got 1 token.
Can we try to get another one during the dial process? We can make this improvement later on, but everytime we block in this function, we could try to ask for more first, in the event that we did not receive all of them in the first ask

doc/DIALER.md
jacobheun: I would expect this to exist in the PeerStore when we separate PeerInfo into its components. That's where we should be tracking metadata like this, imo.

src/dialer/token-holder.js
jacobheun: Yeah, I think there is potential here to clean a lot of the token handling up. So the dialer might give you a token generator, and it will eventually put the `n` tokens you asked for into the generator. The dial process can then pull from that generator and potentially push into a local generator. The local generator is actually a token cycler. So when a multiaddr dial is done but there are more to do, it pushes the token back into the cycler for other dials. Once there is no more dials to perform for a token, it's pushed into the dialer token pool. Psuedo code:

```js
pipe(
  dialer.getTokens(3),
  dialRequestTokenCycler(),  // Reuse tokens until they are not needed, then yield them
  dialer.consumeTokens()
)
```

doc/DIALER.md
jacobheun: I would expect this to exist in the PeerStore when we separate PeerInfo into its components. That's where we should be tracking metadata like this, imo.

src/dialer/token-holder.js
vasco-santos: this would be super cool

src/dialer/dial-request.js
```diff
@@ -0,0 +1,197 @@
...
+    // If there are tokens left, release them
+    // If there are multiaddrs left, wait for tokens to finish
+    const th = new TokenHolder(tokens, this.dialer.releaseToken)
+
+    // Create the dial functions
+    const dials = this.addrs.map(addr => {
+      return () => this._abortableDial(addr, options)
+    })
+
+    const dialResolver = new DialResolver()
```
alanshaw: This is quite a lot of complicated code, I wonder if we could essentially remove `TokenHolder` and `DialResolver` and simplify by using `p-fifo` (or something similar)? `p-fifo` is good because you can await on `shift` which resolves when a value gets pushed.

What about something like this?:

```js
const FIFO = require('p-fifo')

const th = new FIFO()
tokens.forEach(t => th.push(t))
const dialAbortControllers = this.addrs.map(() => new AbortController())

try {
  return await Promise.race(this.addrs.map(async (addr, i) => {
    const token = await th.shift() // get token
    let conn
    try {
      const signal = dialAbortControllers[i].signal
      if (signal.aborted) throw new Error('already aborted') // `dialAction` should do this check really...
      conn = await this.dialAction(addr, { ...options, signal: anySignal([signal, options.signal]) })
    } catch (err) {
      th.push(token) // return to token holder on error so another ma can be attempted
      throw err // idk maybe you want to collect these
    }
    return conn
  }))
} finally {
  dialAbortControllers.map(c => c.abort()) // success/failure happened, abort everything else
  tokens.forEach(t => this.dialer.releaseToken(t)) // release tokens back to the dialer
}
```

doc/DIALER.md
```diff
@@ -0,0 +1,37 @@
...
+* A failed dial SHOULD lower the confidence by 1, to a min of 0.
+* A successful dial SHOULD increase the confidence of a multiaddr by 1, to a max of 10.
+* Multiaddrs of confidence 0 MAY be pruned if higher confidence addresses exist. If this is done, it should be done an an actual failure. Aborted dials SHOULD NOT be considered failures, but timeouts SHOULD.
+
+## Notes
+
+* A DialRequest gets a set of tokens from the Dialer, up to the PER_PEER_LIMIT max.
+* A DialRequest SHOULD release its tokens after the DIAL_TIMEOUT has expired. This ensures that a peer with a large set of hanging addresses, does not block future dials.
```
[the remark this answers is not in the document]
jacobheun: I can work on clarifying this in the readme, I think it's getting convoluted in the various should statements. In reality it's just: `If a dial is completed and there are no other dials to do, release the token.`

doc/DIALER.md
```diff
@@ -0,0 +1,37 @@
...
+* Upon releasing the tokens, the Dialer may allocate them to another DialRequest.
+* A DialRequest SHOULD fail if no dial tokens are available. The DialRequest MAY proceed without tokens, but this should be reserved for High Priority actions and should be rare.
+* A DialRequest MUST NOT request more tokens than it has addresses to dial. Example: If the PER_PEER_LIMIT max is 4 and a DialRequest has 1 address, it should only request 1 token.
```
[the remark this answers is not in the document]
jacobheun: The difference is that in the token system you can't perform a dial if no tokens are available, it immediately fails. In the queue system, the dial will be added to the queue regardless of the queue saturation, so your dial may have to wait for `n` dials before its run.

doc/DIALER.md
```diff
@@ -0,0 +1,37 @@
...
+* A DialRequest MUST execute parallel dials for each of its addresses up the total number of tokens it has.
```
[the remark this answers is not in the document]
jacobheun: This should be a `should`. If a request takes more than 1 token, it shouldn't be dialing in serial.

src/dialer.js
```diff
@@ -29,105 +30,106 @@ class Dialer {
...
+      dialAction,
+      dialer: this
+    })
+
+    // Combine the timeout signal and options.signal, if provided
+    const timeoutController = new AbortController()
+    const signals = [timeoutController.signal]
+    options.signal && signals.push(options.signal)
+    const signal = anySignal(signals)
+    const timeoutPromise = delay.reject(this.timeout, {
```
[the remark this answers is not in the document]
jacobheun: I like this approach better, I'll refactor.

src/dialer/dial-request.js
jacobheun: I ended up leveraging this code which got rid of the TokenHolder and DialResolver, so it's now more succinct. I did make a few updates and changes:
1. `pAny` instead of `Promise.race` because we dont want to end after a single failure
2. Tokens are released early. If we have already run through all the addresses when a dial settles, we release it instead of pushing it back onto `p-fifo`. This ensures that if our last dial takes 5 seconds, we don't have to wait for the rest of the tokens to go back into the pool, we will just keep the 1 we're using.

doc/DIALER.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Multiaddr Confidence
+
+**Not Yet Implemented**
+
+An effective dialing system should involve the inclusion of a Confidence system for Multiaddrs. This enables ranking of Multiaddrs so that a prioritized list can be passed to DialRequests to maximize usage of Dialer Tokens.
+
+* All Multiaddrs SHOULD start with a confidence of 1 (give it some trust by default)
+* A failed dial SHOULD lower the confidence by 1, to a min of 0.
+* A successful dial SHOULD increase the confidence of a multiaddr by 1, to a max of 10.
```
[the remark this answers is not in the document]
vasco-santos: Maybe just leave your thoughts on an issue for now?

src/dialer.js
```diff
@@ -29,105 +29,88 @@ class Dialer {
...
+    const signals = [timeoutController.signal]
+    options.signal && signals.push(options.signal)
+    const signal = anySignal(signals)
+    const timeoutId = setTimeout(() => timeoutController.abort(), this.timeout)
 
     try {
-      conn = await this.queue.add(() => this.transportManager.dial(addr, options))
+      const dialResult = await dialRequest.run({ ...options, signal })
+      clearTimeout(timeoutId)
+      return dialResult
```
vasco-santos: Should we log success dial?
Also, maybe do the return after the catch instead, so that it is easier to find what is returned

src/dialer/dial-request.js
```diff
@@ -0,0 +1,77 @@
...
+   * @returns {Connection}
+   */
+  async run (options) {
+    const tokens = this.dialer.getTokens(this.addrs.length)
+    // If no tokens are available, throw
+    if (tokens.length < 1) {
+      throw errCode(new Error('No dial tokens available'), 'ERR_NO_DIAL_TOKENS')
+    }
+
+    const th = new FIFO()
```
vasco-santos: As we got rid of the `TokenHolder` class, maybe you can name this tokenHolder now? `th` is really abstract

src/dialer/dial-request.js
```diff
@@ -0,0 +1,197 @@
...
+   * @param {number} options.timeout The max dial time for each request
+   * @returns {Connection}
+   */
+  async run (options) {
+    // Determine how many tokens we need
+    const tokensWanted = Math.min(this.addrs.length, this.dialer.perPeerLimit)
+    // Get the tokens
+    const tokens = this.dialer.getTokens(tokensWanted)
+    // If no tokens are available, throw
+    if (tokens.length < 1) {
```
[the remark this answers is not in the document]
jacobheun: Right now we don't have a mechanism for this, other than observing `dialer.token.length`. I went with a throw approach when there are no tokens to simplify the dial logic. If a dial can happen right now, it will be excited, otherwise it just fails immediately. It does push retry/queue logic onto the user, if they wish to queue up dials. This does also have an impact on things like DHT queries. If we have enough queries to consume all available tokens, dial attempts for new queries will start to fail.

We could alleviate this by making `getTokens` a generator and allowing token dials to stack up until tokens are freed. This will create problems with queue saturation (which we already have in master). One thing I would like to eventually look into is adding Connection "prepurchasing" to Connection Management. This would enable subsystems to request a pool of connections prior to performing certain actions, such as a DHT query. Subsystems could then take a proactive approach to ratelimiting their actions.

The DHT and initial startup are the most common scenarios that would likely cause us to hit token limits within the system itself. Ideally, we should be avoiding scenarios where we're attempting to establish `MAX_PARALLEL_DIALS` all at once.

jacobheun:
Some of the tests were flakey due to slow shutdown times. I fixed this by adding global abort support for pending dials. The dialer will now track pending dials and when `libp2p.stop()` is called, libp2p will iterate over each pending dial and abort it, using the timeoutcontroller created for each dial request. This should help tremendously with issues we've had in the past where libp2p shutdown hangs.

src/dialer/dial-request.js
```diff
@@ -0,0 +1,76 @@
...
+const debug = require('debug')
+const errCode = require('err-code')
+const log = debug('libp2p:dialer:request')
+log.error = debug('libp2p:dialer:request:error')
+const FIFO = require('p-fifo')
+const pAny = require('p-any')
+
+class DialRequest {
+  /**
+   *
```
vasco-santos: Can you add a small description here?

package.json
```diff
@@ -42,6 +42,8 @@
   },
   "dependencies": {
     "abort-controller": "^3.0.0",
+    "aggregate-error": "^3.0.1",
+    "any-signal": "^1.0.0",
     "async": "^2.6.2",
     "async-iterator-all": "^1.0.0",
```
alanshaw: PSA, this module has changed name to `it-all`

package.json
```diff
@@ -42,6 +42,8 @@
   },
   "dependencies": {
     "abort-controller": "^3.0.0",
+    "aggregate-error": "^3.0.1",
+    "any-signal": "^1.0.0",
     "async": "^2.6.2",
```
alanshaw: 🙏 is it time to remove?

package.json
```diff
@@ -64,6 +66,7 @@
     "multistream-select": "^0.15.0",
     "once": "^1.4.0",
```
alanshaw: Can be removed?

package.json
```diff
@@ -64,6 +66,7 @@
...
     "p-any": "^2.1.0",
+    "p-fifo": "^1.0.0",
     "p-map": "^3.0.0",
     "p-queue": "^6.1.1",
     "p-settle": "^3.1.0",
```
alanshaw: Further down, there are pull stream modules and `promisify-es6` - can these be removed also?

src/dialer/dial-request.js
```diff
@@ -0,0 +1,80 @@
...
+    const dialAbortControllers = this.addrs.map(() => new AbortController())
+    let completedDials = 0
+
+    try {
+      return await pAny(this.addrs.map(async (addr, i) => {
+        const token = await tokenHolder.shift() // get token
+        let conn
+        try {
+          const signal = dialAbortControllers[i].signal
+          conn = await this.dialAction(addr, { ...options, signal: anySignal([signal, options.signal]) })
```
alanshaw: `options.signal` may be `undefined` - does `anySignal` account for this or do we need a guard?

src/dialer/dial-request.js
```diff
@@ -0,0 +1,80 @@
...
+          // Remove the successful AbortController so it is not aborted
+          dialAbortControllers.splice(i, 1)
+        } finally {
+          completedDials++
+          // If we have more dials to make, recycle the token, otherwise release it
+          if (completedDials < this.addrs.length) {
```
alanshaw: This doesn't necessarily mean that we're waiting for a token. We could have a token and be awaiting on `dialAction` and so will not have incremented `completedDials` yet. So we may give this token back to the token holder when actually it could be released.

Maybe a `startedDials` counter that gets incremented between `await tokenHolder.shift()` and `await this.dialAction(...)`?

src/dialer/index.js
```diff
@@ -0,0 +1,129 @@
...
+   * @param {Array<Multiaddr>|Multiaddr} addrs
+   * @param {object} [options]
+   * @param {AbortSignal} [options.signal] An AbortController signal
+   * @returns {Promise<Connection>}
+   */
+  async connectToMultiaddr (addrs, options = {}) {
+    if (!Array.isArray(addrs)) addrs = [multiaddr(addrs)]
+
+    const dialAction = (addr, options) => {
+      if (options.signal.aborted) throw errCode(new Error('already aborted'), 'ERR_ALREADY_ABORTED')
```
alanshaw:
```suggestion
      if (options && options.signal && options.signal.aborted) throw errCode(new Error('already aborted'), 'ERR_ALREADY_ABORTED')
```

src/dialer/index.js
```diff
@@ -0,0 +1,129 @@
...
+      dialAction,
+      dialer: this
+    })
+
+    // Combine the timeout signal and options.signal, if provided
+    const timeoutController = new AbortController()
+    const signals = [timeoutController.signal]
+    options.signal && signals.push(options.signal)
+    const signal = anySignal(signals)
+    const timeoutId = setTimeout(() => timeoutController.abort(), this.timeout)
```
alanshaw: I know this doesn't really save any LoC but it just makes it a bit more of a concrete thing that people might reach for instead of rolling their own timeouts. I'll need this in js-ipfs too:

```js
class TimeoutController extends AbortController {
  constructor (ms) {
    super()
    this._timeoutId = setTimeout(() => this.abort(), ms)
  }
  abort () {
    clearTimeout(this._timeoutId)
    return super.abort()
  }
  clear () {
    clearTimeout(this._timeoutId)
  }
}
```

...also, since this is the controller being used to abort dials when libp2p stops, this class actually clears the timeout if `abort` is called before it expires - currently these timeouts remain running and may cause `abort` to be called multiple times. I think that's not actually a problem but having timers still running might keep the process running (although you might not seeing this right now because `pendingDials` are deleted and all of the timers are probably all garbage collected).

src/index.js
```diff
@@ -188,8 +194,15 @@ class Libp2p extends EventEmitter {
...
     try {
-      this.pubsub && await this.pubsub.stop()
-      this._dht && await this._dht.stop()
+      await Promise.all([
+        this.pubsub && this.pubsub.stop(),
+        this._dht && this._dht.stop()
+      ])
+
+      for (const dial of this.dialer.pendingDials.values()) {
+        dial.abort()
```
alanshaw: `dial.controller.abort()`

src/index.js
```diff
@@ -188,8 +194,15 @@ class Libp2p extends EventEmitter {
...
 
     try {
-      this.pubsub && await this.pubsub.stop()
-      this._dht && await this._dht.stop()
+      await Promise.all([
+        this.pubsub && this.pubsub.stop(),
+        this._dht && this._dht.stop()
+      ])
+
+      for (const dial of this.dialer.pendingDials.values()) {
```
alanshaw: I would add a `destroy` function to `this.dialer` that does this and rename to `this.dialer._pendingDials`. This is state that does not need to be public or shared - right?

src/peer-store/index.js
```diff
@@ -217,6 +217,16 @@ class PeerStore extends EventEmitter {
...
     })
   }
+
+  /**
+   * Returns the known multiaddrs for a given `PeerId`
+   * @param {PeerId} peerId
+   * @returns {Array<Multiaddr>}
+   */
+  multiaddrsForPeer (peerId) {
+    const peerInfo = this.get(peerId.toB58String())
```
alanshaw: We really need to move away from `toB58String`!

package.json
jacobheun: I've got a cleanup branch that gets rid of all of these old modules, code and turns bundlesize back on: https://github.com/libp2p/js-libp2p/pull/497. I'll add the `it-all` switch to that.

src/dialer/dial-request.js
jacobheun: DialRequest requires signal, and the dialer is currently protecting against it so this isn't currently an issue. I will also add a patch to anySignal to help protect against non signals for.

src/dialer/index.js
jacobheun: This isn't necessary. `DialRequest` will **always** call `dialAction` with a signal.

src/peer-store/index.js
jacobheun: We can do a followup PR to replace `toB58String` with `toString` everywhere. One of the things we started moving towards is just taking `PeerId` in the APIs to avoid the discrepancy. Then internally any PeerId string can be used, as long as it's consistent in that system.

src/dialer/index.js
alanshaw: ...then it is not an option 😜

jacobheun: Well, technically an option isn't necessarily optional, and we are internally requiring the optional values of `Transport.dial()`. English is fun.

src/dialer/dial-request.js
jacobheun: Good catch, I fixed this and added a test.

alanshaw:
>     * I created `timeout-abort-controller` (timeout-controller was too similar to an existing, different module) and we are using that here now. It uses `retimer` internally to avoid spinning up numerous timers via `setTimeout`

LOVE IT

> 
>     * All the legacy modules are being removed in the cleanup PR #497

ACK

> The only outstanding item should be the dialAction `options` naming. I am happy to take suggestions on what that should be changed to if it's confusing, I was mainly attempting to keep that consistent with what ultimately gets passed to the `transport.dial` `options`.

🤷‍♂️ meh, it minor (and possibly my own preference) and I don't care enough to argue. Let merge this puppy.

PR #494, opened 2019-12-03
fix: clean up peer discovery flow

jacobheun:
* `peerStore.update` no longer replaces existing protocols and multiaddrs, this is inline with PeerBook.
* Identify will now call `peerStore.replace()` instead of `peerStore.update()`. As we are talking directly with the target peer, we should consider them the source of truth. With Multiaddr confidence in the future and PeerStore v2, this may may change.
* All PeerDiscovery `peer` events will now be directed to the peerStore directly.
* `peer:discovery` events are now being Proxied from the PeerStore. This also ensures we won't discover a peer twice.
* `PeerStore.replace()` will always trigger `change:multiaddrs` and `change:protocols` events. This is an oversimplification and may not indicate an actual change, but this is probably better reserved for PeerStore v2 work.

### DIFF
--- a/src/peer-store/index.js
+++ b/src/peer-store/index.js
@@ -45,7 +45,7 @@ class PeerStore extends EventEmitter {
 
     let peer
     // Already know the peer?
-    if (this.peers.has(peerInfo.id.toB58String())) {
+    if (this.has(peerInfo.id)) {
       peer = this.update(peerInfo)
     } else {
       peer = this.add(peerInfo)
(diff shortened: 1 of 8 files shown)


src/peer-store/index.js
```diff
@@ -217,6 +206,16 @@ class PeerStore extends EventEmitter {
 
     this.remove(peerInfo.id.toB58String())
     this.add(peerInfo)
+
+    // This should be cleaned up in PeerStore v2
+    this.emit('change:multiaddrs', {
```
vasco-santos: the `multiaddrs` may not change, right?

Issue #488, opened 2019-12-01
[Documentation] "Install" section in README does not make sense


mcclure:
Say I am new to libp2p and just want to install whatever npm package I need to get started so I can run example code in tutorials. I come to the js-libp2p page on github. I look under "Install". It gives the npm line to run to install, then says,

> this module is only a skeleton and should not be used directly other than libp2p bundle implementors that want to extend its code

So… wait. I *shouldn't* install it? The line is unclear; it says "don't use", not "don't install".

It also says "as noted above" so I look above. Above it nicely explains that I should use "libp2p-ipfs-nodejs" or "libp2p-ipfs-browser". Great! I'll install one of those. Except those links are links to… javascript files? I search the npm registry. There is no "libp2p-ipfs-browser" there. There is a "libp2p-ipfs-nodejs" but it is marked "deprecated" with a big banner that says to go to… [github.com/libp2p/js-libp2p](https://github.com/libp2p/js-libp2p), in other words this page. Later the README suggests I "create my own" libp2p "bundle" but it doesn't explain what thing or collection of things I need to install to create one of those. I don't know exactly what a "bundle" is at this point.

It is clear I am probably looking at the wrong page. Unfortunately you are the top hit on Google for "javascript libp2p". So you have a page which is clearly not the first thing a new user should look at, but because of Google you **are** the first thing a new user is likely to look at, and you don't link which page within the sprawling libp2p web presence **is** the first thing a new user should look at.

My "**expected behavior**"/suggestion is that you change "this module… should not be used" to "this module… should not be installed" (if that is what you mean), and that you provide links in the install section to one of:

* What to install instead
* What document I should read to understand that "what should I install instead?" is a nonsensical question
* If you're going to suggest using the "libp2p-ipfs-nodejs" or "libp2p-ipfs-browser" "bundles", you should say what package(s) to install to use those two "bundles".

"Type: Very Low - Translation or documentation mistake. Something that won't give anyone a bad day." My day is ok so far

(Maybe this is redundant with #231. But that issue is over a year old so maybe resolving #231 is hard and it's worth addressing the smaller problem that you have a "install" section that is not actionable.)

jacobheun:
@mcclure thank you for the write up, it's very helpful hearing your experience attempting to get up and running. We are currently finishing up work on the async/await refactor and part of this is rewriting the readme. I would like to include a **Getting Started** section / standalone readme that covers what modules to install and how to use them for both (Node.js and the Browser). 

[...]

Issue #350, opened 2019-04-08
Discuss: "bundling" terminology is confusing


raulk:
The term "bundle" in modern Javascript land often implies things like packaging, minification, tree shaking, etc. all of which don't apply in our usage of the term. It evokes Webpack, Parcel, etc. I've received feedback from developers that this is confusing.

Really, what we are trying to recommend is putting libp2p imports and configuration in one file, and exporting it from there. Calling that a "bundle" is unnecessarily complex IMO.

Can we simplify this?

raulk:
cc @mikerah

raulk:
I actually think it's great to point to real examples of libp2p node configurations like the ones in IPFS. Can we point to other projects too?

raulk:
Alternatively we may want to bring in those examples into libp2p itself as "blueprints" or something.

jacobheun:
This is a duplicate of https://github.com/libp2p/js-libp2p/issues/231.

It would be ideal to do a full audit of the readme and language used, including bundling. There is an open issue at https://github.com/libp2p/js-libp2p/issues/237.

raulk:
@jacobheun gotcha, let's dislodge all this stuff! Should we create a `dx` label and group all these tasks and discussions so we can address them ASAP?

jacobheun:
@raulk yes, let's group them at https://github.com/libp2p/js-libp2p/issues/308 which is to track the OKR around improved docs and examples. I'll comment there with links to each of the issues for `js-libp2p` itself and tag each of those with the new label.

yusefnapora:
I agree that this is a confusing bit of terminology -  we should either publish common "bundles" as npm packages, or just get rid of the term entirely and refer to "configuring libp2p" instead as suggested [here](https://github.com/libp2p/js-libp2p/issues/231#issuecomment-410865609).

Mikerah:
I actually just got into this very issue a few days ago. I used the word __bundle__ in the same paragraph both in the JS sense and libp2p sense. A suggestion I have would just to refer to them as libp2p modules or libp2p configurations.

vasco-santos:
We got away of the `bundle` terminology on the latest release, as a consequence of it has been reported as confusing by several issues. Closing this as it is not relevant anymore

Issue #231, opened 2018-08-06
Clarify readme "bundles" section


lgierth:
We just had someone in IRC confused about the relationship between the bundles and js-libp2p. I feel like the section should avoid the word skeleton, which sounds like death and lack of substance, and feels more like a quantification of project status vs. a qualification of the js-libp2p repo's function within the js-libp2p project.

Instead it should probably say something like:

- This is an umbrella repo and the actual functionality is in separate packages, which you pull in and wire up based on your needs
- For most purposes, use the wiring from js-ipfs (ipfs-browser and ipfs-node bundles)
- If that doesn't fit you, modify that wiring code as you please

(I think the concept of bundles generally needs clarification and rethinking. Maybe "snippets" get closer to what we're trying to achieve there.)

cc @Mr0grog

Isan-Rivkin:
It's me on the IRC and agreed :D

Mr0grog:
Hmmm, what’s the story with the `libp2p-ipfs-nodejs` and `libp2p-ipfs-browser` links? They read like they should link to actual packages (which seemed surprising, since I know how js-ipfs uses libp2p), but then I clicked them and realized they just linked to js-ipfs’s internal configuration. Were these once intended to become wrapper packages of some sort? I definitely agree it’s confusing.

Are there any published “bundles” that wrap up groups of libp2p modules so you can just create an instance [of the bundle] instead of manually composing a bunch of libp2p parts? If not, I don’t think we should even call this section “bundles.” Maybe it could be “composing functionality” or something similar.

My take:

> ## Composing Functionality
> 
> `js-libp2p` itself is just a tool for plugging together different pieces of a peer-to-peer networking stack. To create a working `libp2p` instance, you’ll need to select a set of transports (like TCP, WebSockets, etc.), multiplexers, encryption mechanisms, peer discovery mechanisms, and DHT implementations to use. The `libp2p` project contains packages that implement all of these, but it’s up to you to import those packages and send them to the `js-libp2p` constructor.
>
> For example, you might use the TCP and WebSocket transports, MPLEX multiplexing, SecIO encryption, Multicast discovery, and Kademlia DHT:
>
> ```js
> const libp2p = require('libp2p')
> const TCP = require('libp2p-tcp')
> const WS = require('libp2p-websockets')
> const MPLEX = require('libp2p-mplex')
> const SECIO = require('libp2p-secio')
> const MulticastDNS = require('libp2p-mdns')
[...]

PR #495, opened 2019-12-04
docs: config

vasco-santos:
One of the biggest issues that `js-libp2p` users usually have relies on being complicated to understand:

- what is `js-libp2p` in the middle of so many repos?
- what is a libp2p bundle?
- how to configure `js-libp2p`?
- how to configure X in `js-libp2p`?

Some decisions:

- Get rid of the `bundle` naming

Some topics to discuss in the future:

- Separate modules configuration from libp2p core configuration (will open an issue for this later)

Relevant issues:

- https://github.com/libp2p/js-libp2p/issues/231
- https://github.com/libp2p/js-libp2p/issues/350
- https://github.com/libp2p/js-libp2p/issues/488


doc/CONFIGURATION.md
```diff
@@ -0,0 +1,289 @@
...
+  * [Content Routing](#content-routing)
+  * [Peer Routing](#peer-routing)
+  * [DHT](#dht)
+  * [Pubsub](#pubsub)
+* [Libp2p node](#libp2p-node)
+* [Configuration examples](#configuration-examples)
+
+## Overview
+
+Considering libp2p's modular nature, it can be composed by a set of building blocks that enable users to create any type of p2p systems without needing to implement its network layer. Taking into account that each p2p system has its own needs and requirements, libp2p can be configured accordingly.
```
jacobheun:
```suggestion
libp2p is a modular networking stack. It's designed to be able to suit a variety of project needs. The configuration of libp2p is a key part of its structure. It enables you to bring exactly what you need, and only what you need. This document is a guide on how to configure libp2p for your particular project. Check out the [Configuration examples](#configuration-examples) section if you're just looking to leverage an existing configuration.
```

doc/CONFIGURATION.md
```diff
@@ -0,0 +1,289 @@
...
+
+`js-libp2p` acts as the composer for this modular p2p networking stack using libp2p compatible modules as its building blocks.
```
jacobheun: I think this is superfluous. 
```suggestion
```

doc/CONFIGURATION.md
```diff
@@ -0,0 +1,289 @@
...
+While libp2p building blocks may vary, `js-libp2p` top level API will be consistent.
```
jacobheun:
```suggestion
Regardless of how you configure libp2p, the top level [API](./API.md) will always remain the same. **Note**: if some modules are not configured, like Content Routing, using those methods will throw errors.
```

doc/CONFIGURATION.md
```diff
@@ -0,0 +1,289 @@
...
+
+## Modules
+
+For getting an instance of `js-libp2p` compliant with all types of networking requirements, it is possible to specify the following building blocks:
```
jacobheun: I think using `subsystems` instead of `building blocks` is a bit more descriptive.

doc/CONFIGURATION.md
vasco-santos: Maybe I should move this under the next section? Its main intend is to introduce the `modules`

doc/CONFIGURATION.md
```diff
@@ -0,0 +1,289 @@
...
+
+If none of the available content routing protocols fulfills your needs, you can create a libp2p compatible one. A libp2p content routing protocol just needs to be compliant with the [interface-content-routing](https://github.com/libp2p/js-interfaces/tree/master/src/content-routing). **(WIP: This module is not yet implemented)**
+
+If you want to get to know more about libp2p content routing, you should read the following content:
+
+- https://docs.libp2p.io/concepts/content-routing
+
+### Peer Routing
+
+> Peer Routing offers a way to find other peers in the network by intentionally issuing queries, iterative or recursive, until a Peer is found or the closest Peers, given the Peer Routing algorithm strategy, are found.
```
yusefnapora:
```suggestion
> Peer Routing offers a way to find other peers in the network by issuing queries using a Peer Routing algorithm, which may be iterative or recursive. If the algorithm is unable to find the target peer, it will return the peers that are "closest" to the target peer, using a distance metric defined by the algorithm.
```

doc/CONFIGURATION.md
```diff
@@ -0,0 +1,289 @@
...
+  connEncryption: [],
+  contentRouting: [],
+  peerRouting: [],
+  peerDiscovery: [],
+  dht: dhtImplementation,
+  pubsub: pubsubImplementation
+}
+```
+
+Moreover, the majority of the modules can be customized via option parameters. This way, it is also possible to provide this options through a config object. This config object should have the property name of each building block to configure, the same way as the modules specification.
```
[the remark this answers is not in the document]
vasco-santos: Yeah, I agree with you! I will break this in smaller pieces

mcclure:
My thoughts as a libp2p newbie:

This document is very clear! This would have helped me a lot if it had been linked when I first found js-libp2p, and honestly it helps me a lot even now. I have some comments:

- I think it is okay to use the "bundle" terminology as long as you clearly explain "In the libp2p community, a 'bundle' is…" (…"a list of modules selected for a particular project?")

- I would recommend listing, for each of your "transports", which platforms they work on, eg (desktop) or (browser) or (desktop/browser).

- Are you a native German speaker? There were a small number of sentences that did not read to me like natural English. I recommend these fixes:

    "This way, the user should install and import the modules that are relevant for their requirements" In this sentence I would remove "This way,", the sentence is entirely clear without it.

    "Bear in mind that only a \*\*transport\*\* is required, being all the other subsystems optional." I would replace the word "being" with "with"

    I noticed a spelling error "multiplxers" for "multiplexers"

    Otherwise it was good!

vasco-santos:
Thanks for your feedback @mcclure 

> My thoughts as a libp2p newbie:
> 
> This document is very clear! This would have helped me a lot if it had been linked when I first found js-libp2p, and honestly it helps me a lot even now. I have some comments:
> 
> * I think it is okay to use the "bundle" terminology as long as you clearly explain "In the libp2p community, a 'bundle' is…" (…"a list of modules selected for a particular project?")

The bundle naming caused confusion to several people and in the web space bundle is also used in other context. As a consequence, we thought it is a good timing to get rid of the bundle terminology.

> * I would recommend listing, for each of your "transports", which platforms they work on, eg (desktop) or (browser) or (desktop/browser).

I think that we can add that. I was thinking on writing that in the `getting started` document, but since it does not add too much noise here, I will add it

> * Are you a native German speaker? There were a small number of sentences that did not read to me like natural English. I recommend these fixes:

Not really, native portuguese speaker, but thanks for the recommendations. I will work on them

>   "This way, the user should install and import the modules that are relevant for their requirements" In this sentence I would remove "This way,", the sentence is entirely clear without it.
>   "Bear in mind that only a **transport** is required, being all the other subsystems optional." I would replace the word "being" with "with"
>   I noticed a spelling error "multiplxers" for "multiplexers"
>   Otherwise it was good!

Thanks for having a look on this and hopefully it will help you and other folks interested in getting started with libp2p

doc/CONFIGURATION.md
```diff
@@ -0,0 +1,384 @@
...
+- Content routing protocols
+- Peer routing protocols
+- DHT implementation
+- Pubsub router
+
+The libp2p ecosystem contains at least one module for each of these subsystems. The user should install and import the modules that are relevant for their requirements. Moreover, thanks to the existing interfaces it is easy to create a libp2p compatible module and use it.
+
+After selecting the modules to use, it is also possible to configure each one according to your needs.
+
+Bear in mind that only a **transport** and a **connection encryption** are required, while all the other subsystems are optional.
```
jacobheun:
```suggestion
Bear in mind that only a **transport** and **connection encryption** are required, while all the other subsystems are optional.
```

PR #496, opened 2019-12-07
refactor: examples

vasco-santos:


vasco-santos:
Closing this in favor of a PR per example, as a result of hitting some issues in some of the examples

PR #497, opened 2019-12-08
refactor: cleanup

jacobheun:
This uses the token dialer branch as a base.

- Removes the old switch code as it is no longer needed
- The Stats code has been moved to a root directory, as it will be reincorporated soon.
- All unused packages have been removed from package.json
- Adds the bundlesize check back to ci

**Note**: The deps check in ci is excluding pull-stream and pull-handshake. Ping still needs to be refactored and as such is requiring these modules. Once that refactor is done, we can remove the exclusion from the ci deps check.


alanshaw:
26E3A1C9-593C-4F22-90FC-F054EC48AFAE 🙌

PR #499, opened 2019-12-09
refactor: examples/encrypted-communications

vasco-santos:
This PR refactors the `encrypted-communications` example to the refactored async `js-libp2p`.

Needs:


examples/encrypted-communications/1.js
```diff
@@ -1,63 +1,56 @@
...
+    pipe(
+      stream,
+      toBuffer,
+      map(String),
+      source => (async function () {
+        for await (const msg of source) {
+          console.log(msg)
+        }
+      })(),
     )
```
jacobheun: You can `toString` a BufferList, so this could be simplified. Also, the sink function is more complex than it needs to be:
```js
pipe(
  stream,
  async function (source) {
      console.log(msg.toString())
    }
  }
)
```

vasco-santos: I tried to mimic what was before, but considering that decision again, I think we should simplify :)

vasco-santos: changed

PR #501, opened 2019-12-09
refactor: stats

jacobheun:
Refactors bandwidth stats for the async / await endeavor. 

* Metrics are now optional and will be disabled by default! If your application needs metrics, you can still enable them with the config. If metrics are disabled, the Connection Manager will only use latency monitoring and connected peer numbers to manage connections. An upcoming PR for connection manager will explain the changes in more depth.
* Metrics have moved from `libp2p.stats` to `libp2p.metrics`
* Metrics no longer emit events and are now request based. This greatly improves performance as multiple events would be emitted for every chunk that went through libp2p.
* Individual `Stats` objects now have a `toJSON` method. This makes it easy to serialize and view a snapshot of a particular stat.
* Per transport `Stats` are no longer available. This data can be aggregated via monitoring connection addresses and stats per peer.


doc/METRICS.md
```diff
@@ -0,0 +1,36 @@
...
+* Metrics gathering should be optional, as there is a performance hit to using it.
+* Metrics do NOT currently contain OS level stats, only libp2p application level metrics. For example, TCP messages (ACK, FIN, etc) are not accounted for.
+
+## Tracking
+* When a transport hands off a connection for upgrading, metrics are hooked up if turned on.
+* When a stream is created, metrics will be tracked on that stream and associated to that streams protocol.
+
+### Questions?
+* Should we track crypto protocol bandwidth for crypto handshaking? We won't know the remote peer id for sure until crypto is done, so we will need some way to lazily track by id.
+* Should we track muxer protocol bandwidth for muxer selection? This is likely not worth it.
```
vasco-santos: I think it would be nice to track the crypto negotiation. We can postpone a decision for this for when we iterate on `connMgr`.

Same for `muxer`, I don't think it is relevant either, but we can revisit

doc/METRICS.md
```diff
@@ -0,0 +1,36 @@
...
+* If metrics is disabled, how will Connection Manager cull connections?
+  * Topology peers first? What happens when a Topology has too many peers, which ones are pruned first?
+
+## Notes
+* Tracking stats on a connection would result in stat loss when the connection is destroyed.
+* Global stats tracking doesn't account for any per peer cleanup. (If we haven't talked with a peer in days or weeks, should we keep that info? There are memory implications in doing so.)
```
vasco-santos: We should definitely have a pruning mechanism here. It should be configurable though

jacobheun: The old stats are actually using an LRU cache with a default of 50 peers. I will leverage this as I don't see any reason to change it at this point.

doc/METRICS.md
jacobheun: I think muxer selection won't end up being needed. We should be able to extrapolate this information by getting the stats for a peer and then checking the muxer associated with their connection(s).

I'll start without crypto handshake tracking for now, and then add that in once the base is done.

doc/METRICS.md
vasco-santos: Agree that it can work for now

jacobheun:
@vasco-santos this should be ready for review. I am working on updating the API and Config docs for metrics, but wanted to get a review earlier in case we need any changes to the API.

src/metrics/index.js
```diff
@@ -0,0 +1,244 @@
...
+    return this._protocolStats.get(protocol)
+  }
+
+  /**
+   * Should be called when all connections to a given peer
+   * have closed. The `Stats` collection for the peer will
+   * be stopped and moved to an LRU for temporary retention.
+   * @param {PeerId} peerId
+   */
+  onPeerDisconnected (peerId) {
```
vasco-santos: This should be called

src/metrics/index.js
```diff
@@ -0,0 +1,244 @@
...
+    const idString = peerId.toString()
+    const peerStats = this._peerStats.get(idString)
+    if (peerStats) {
+      peerStats.stop()
+      this._peerStats.delete(idString)
+      this._oldPeers.set(idString, peerStats)
+    }
+  }
+
+  onMessage ({ remotePeer, protocol, direction, dataLength }) {
```
vasco-santos: Can you also add docs here? It is the only function missing

src/metrics/index.js
```diff
@@ -0,0 +1,244 @@
...
+      this._globalStats.push(key, dataLength)
+    }
+  }
+
+  /**
+   * Replaces the `PeerId` string with the given `peerId`.
+   * If stats are already being tracked for the given `peerId`, the
+   * placeholder stats will be merged with the existing stats.
+   *
+   *
```
vasco-santos:
```suggestion
```

src/metrics/index.js
```diff
@@ -0,0 +1,244 @@
...
+      peerStats.push(key, dataLength)
+      this._globalStats.push(key, dataLength)
+    }
+  }
+
+  /**
+   * Replaces the `PeerId` string with the given `peerId`.
+   * If stats are already being tracked for the given `peerId`, the
+   * placeholder stats will be merged with the existing stats.
+   *
```
vasco-santos:
```suggestion
```

src/metrics/index.js
```diff
@@ -0,0 +1,244 @@
...
+
+    this._peerStats.delete(placeholder.toString())
+    this._peerStats.set(peerIdString, mergedStats)
+    mergedStats.start()
+  }
+
+  /**
+   * Tracks data running through a given Duplex Iterable `stream`. If
+   * the `peerId` is not provided, a placeholder string will be created and
+   * returned. This allow lazy tracking of a peer when the peer is not yet known.
```
vasco-santos:
```suggestion
   * returned. This allows lazy tracking of a peer when the peer is not yet known.
```

src/metrics/index.js
```diff
@@ -0,0 +1,244 @@
...
+   * @param {PeerId} [options.peerId] The id of the remote peer that's connected
+   * @param {string} [options.protocol] The protocol the stream is running
+   * @returns {string} The peerId string or placeholder string
+   */
+  trackStream ({ stream, remotePeer, protocol }) {
+    const metrics = this
+    const _source = stream.source
+    stream.source = (async function * () {
+      for await (const chunk of _source) {
+        metrics.onMessage({ remotePeer, protocol, direction: 'in', dataLength: chunk.length })
```
vasco-santos: it would be nice to use [streaming-iterables#tap](https://github.com/bustle/streaming-iterables#tap) with pipe here

src/metrics/index.js
```diff
@@ -0,0 +1,244 @@
...
+        yield chunk
+      }
+    })()
+
+    const _sink = stream.sink
+    stream.sink = source => {
+      pipe(
+        source,
+        source => (async function * () {
```
vasco-santos: use [streaming-iterables#tap](https://github.com/bustle/streaming-iterables#tap)

src/upgrader.js
```diff
@@ -63,6 +66,15 @@ class Upgrader {
     let muxedConnection
     let Muxer
     let cryptoProtocol
+    let placeholderPeer
+
+    if (this.metrics) {
+      const idString = (parseInt(Math.random() * 1e9)).toString(36) + Date.now()
+      placeholderPeer = {
```
vasco-santos: let's create a `getPlaceholderPeer` function instead and call it everywhere

test/metrics/index.node.js
```diff
@@ -0,0 +1,90 @@
...
+
+    expect(libp2p.metrics).to.not.exist()
+  })
+
+  it('should start/stop metrics on startup/shutdown when enabled', async () => {
+    const config = { ...baseOptions }
+    config.metrics = {
+      enabled: true
+    }
+    libp2p = await Libp2p.create(config)
```
vasco-santos: use `createPeer` utility to avoid creating a `PeerId`

test/metrics/index.spec.js
```diff
@@ -0,0 +1,206 @@
...
+
+const Metrics = require('../../src/metrics')
+const { createPeerId } = require('../utils/creators/peer')
+
+describe('Metrics', () => {
+  let peerId
+  let peerId2
+
+  before(async () => {
+    [peerId, peerId2] = await createPeerId({ number: 2, fixture: true })
```
vasco-santos:
```suggestion
    [peerId, peerId2] = await createPeerId({ number: 2 })
```

`fixture: true` is the default

doc/METRICS.md
```diff
@@ -0,0 +1,71 @@
...
+  * The current implementation uses an LRU cache of 50 old peers. So we will only retain the 50 most recently disconnected peers.
+* Total bandwidth - cumulative protocol bandwidth = crypto negotiation + cumulative multiselect negotiation
+* Stats to this point has been a push based system, it should be moved to poll. This would help avoid unnecessary events and allow users to better customize their metrics collection.
+* RemotePeer can't be reliably determined until crypto has been completed.
+* Metrics tracking per connection would provide better visibility into multiple connections to a single peer. Is this needed? Ideally we should be converging to a single connection. Converge on lowest hash?
```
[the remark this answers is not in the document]
jacobheun: I'll go through and this give doc another pass now that things are mostly settled. The metrics don't currently care about individual connections. Everything is equated to a specific peer, or is global. Stats are merged once we have verified the remote peer of the connection. Moving Averages could be affected because we don't have a clean way of merging them right now, but in theory this should rarely be needed as Stats for a new connection shouldn't get computed until after the merge has happened.

src/metrics/index.js
jacobheun: Yes, good catch. I will think through some more test cases to catch these things. I'm not doing any "old peer" verification yet.

test/metrics/index.node.js
jacobheun: 👍 I just noticed the `started` option, I will switch back to using that. I didn't see that option when I was using it before and needed the nodes returned stopped.

PR #502, opened 2019-12-09
refactor: examples/protocol-and-stream-muxing

vasco-santos:
This PR refactors the `protocol-and-stream-muxing` example to the refactored async `js-libp2p`.

Needs:

### DIFF
--- a/examples/protocol-and-stream-muxing/1.js
+++ b/examples/protocol-and-stream-muxing/1.js
@@ -44,9 +44,13 @@ const createNode = async () => {
     )
   })
 
-  // semver matching
+  // multiple protocols
   /*
-  node2.handle('/another-protocol/1.0.1', ({ stream }) => {
+  node2.handle(['/another-protocol/1.0.0', '/another-protocol/2.0.0'], ({ protocol, stream }) => {
+    if (protocol === '/another-protocol/2.0.0') {
+      // handle backwards compatibility
+    }
+
     pipe(
       stream,
       async function (source) {
--- a/examples/protocol-and-stream-muxing/1.js
+++ b/examples/protocol-and-stream-muxing/1.js
@@ -1,102 +1,81 @@
 'use strict'
 
-const libp2p = require('../../')
+const Libp2p = require('../../')
 const TCP = require('libp2p-tcp')
+const MPLEX = require('libp2p-mplex')
+const SECIO = require('libp2p-secio')
 const PeerInfo = require('peer-info')
-const waterfall = require('async/waterfall')
-const parallel = require('async/parallel')
-const pull = require('pull-stream')
-const defaultsDeep = require('@nodeutils/defaults-deep')
 
-class MyBundle extends libp2p {
-  constructor (_options) {
-    const defaults = {
-      modules: {
-        transport: [ TCP ]
-      }
+const pipe = require('it-pipe')
+const { map } = require('streaming-iterables')
+const { toBuffer } = require('it-buffer')
+
+const createNode = async () => {
+  const peerInfo = await PeerInfo.create()
+  peerInfo.multiaddrs.add('/ip4/0.0.0.0/tcp/0')
+
+  const node = await Libp2p.create({
+    peerInfo,
+    modules: {
+      transport: [TCP],
+      streamMuxer: [MPLEX], // TODO: should not need this
+      connEncryption: [SECIO] // TODO: should not need this
     }
+  })
 
-    super(defaultsDeep(_options, defaults))
-  }
+  await node.start()
+
+  return node
 }
 
-function createNode (callback) {
-  let node
-
-  waterfall([
-    (cb) => PeerInfo.create(cb),
-    (peerInfo, cb) => {
-      peerInfo.multiaddrs.add('/ip4/0.0.0.0/tcp/0')
-      node = new MyBundle({
-        peerInfo
-      })
-      node.start(cb)
-    }
-  ], (err) => callback(err, node))
-}
-
-parallel([
-  (cb) => createNode(cb),
-  (cb) => createNode(cb)
-], (err, nodes) => {
-  if (err) { throw err }
-
-  const node1 = nodes[0]
-  const node2 = nodes[1]
+;(async () => {
+  const [node1, node2] = await Promise.all([
+    createNode(),
+    createNode()
+  ])
 
   // exact matching
-  node2.handle('/your-protocol', (protocol, conn) => {
-    pull(
-      conn,
-      pull.map((v) => v.toString()),
-      pull.log()
+  node2.handle('/your-protocol', ({ stream }) => {
+    pipe(
+      stream,
+      toBuffer,
+      map(String),
+      source => (async function () {
+        for await (const msg of source) {
+          console.log(msg)
+        }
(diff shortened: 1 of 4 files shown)


examples/protocol-and-stream-muxing/1.js
```diff
@@ -1,102 +1,81 @@
...
-      node = new MyBundle({
-        peerInfo
-      })
-      node.start(cb)
+  const node = await Libp2p.create({
+    peerInfo,
+    modules: {
+      transport: [TCP],
+      streamMuxer: [MPLEX], // TODO: should not need this
+      connEncryption: [SECIO] // TODO: should not need this
```
vasco-santos: The handle throws if we do not include there

examples/protocol-and-stream-muxing/README.md
```diff
@@ -6,57 +6,76 @@ The feature of agreeing on a protocol over an established connection is what we
...
+  ['my own protocol, wow!'],
+  stream
+)
 ```
 
 This feature is super power for network protocols. It works in the same way as versioning your RPC/REST API, but for anything that goes in the wire. We had to use this feature to upgrade protocols within the IPFS Stack (i.e Bitswap) and we successfully managed to do so without any network splits.
 
 There is still one last feature, you can create your custom match functions.
 
+TODO: should we remove this?
```
vasco-santos: ideas ?

jacobheun: Yes, remove it, we don't support custom match the same way anymore. The new method is to provide each of the protocols you want to handle for a specific handler. It may be valuable to replace it with an example of this. If a user has a backwards incompatible change, but it only requires a few changes to their code, they may wish to do protocol checking instead of having multiple handlers. An example might be a protobuf (or other) data structure having a breaking change. It might be easy to change the serializer being used and keep the remaining code.
```js
libp2p.handle(['/my-protocol/1.0.0', '/my-protocol/2.0.0'], myProtocolHandler)

function myProtocolHandler ({ connection, stream, protocol }) {
  if (protocol === '/my-protocol/2.0.0') {
    // handle backwards compatibility
  }
  ...
}
```

vasco-santos: I have an example with an array below, but I think it makes sense to mention that it can be used for `backwards incompatible change`

examples/protocol-and-stream-muxing/README.md
```diff
@@ -6,155 +6,161 @@ The feature of agreeing on a protocol over an established connection is what we
...
+    stream,
+    toBuffer,
+    map(String),
+    source => (async function () {
+      for await (const msg of source) {
+        console.log(msg)
+      }
+    })()
   )
 })
```
jacobheun: This can be:
```js
node2.handle('/your-protocol', ({ stream }) => {
  pipe(
    stream,
    async function (source) {
        console.log(msg.toString())
      }
    }
  )
})
```

examples/protocol-and-stream-muxing/README.md
```diff
@@ -6,155 +6,161 @@ The feature of agreeing on a protocol over an established connection is what we
...
+  pipe(
+    stream,
+    toBuffer,
+    map(String),
+    source => (async function () {
+      for await (const msg of source) {
+        console.log(msg)
+      }
+    })()
   )
```
jacobheun: Same here, the pipe can be simplified as above

examples/protocol-and-stream-muxing/README.md
```diff
@@ -6,155 +6,161 @@ The feature of agreeing on a protocol over an established connection is what we
...
+  pipe(
+    stream,
+    toBuffer,
+    map(String),
+    source => (async function () {
+      for await (const msg of source) {
+        console.log(msg)
+      }
+    })()
+  )
```
jacobheun: Same here.

examples/protocol-and-stream-muxing/README.md
```diff
@@ -6,155 +6,161 @@ The feature of agreeing on a protocol over an established connection is what we
...
+The examples above would require a node to create a whole new connection for every time it dials in one of the protocols, this is a waste of resources and also it might be simply not possible (e.g lack of file descriptors, not enough ports being open, etc). What we really want is to dial a connection once and then multiplex several virtual connections (stream) over a single connection, this is where _stream multiplexing_ comes into play.
 
-Stream multiplexing is a old concept, in fact it happens in many of the layers of the [OSI System](https://en.wikipedia.org/wiki/OSI_model), in libp2p we make this feature to our avail by letting the user pick which module for stream multiplexing to use.
+Stream multiplexing is a old concept, in fact it happens in many of the layers of the [OSI System](https://en.wikipedia.org/wiki/OSI_model). In libp2p, we make this feature to our avail by letting the user pick which module for stream multiplexing to use.
```
jacobheun:
```suggestion
Stream multiplexing is an old concept, in fact it happens in many of the layers of the [OSI System](https://en.wikipedia.org/wiki/OSI_model). In libp2p, we make this feature to our avail by letting the user pick which module for stream multiplexing to use.
```

examples/protocol-and-stream-muxing/README.md
```diff
@@ -6,155 +6,161 @@ The feature of agreeing on a protocol over an established connection is what we
...
 
-Currently, we have two available [libp2p-spdy](https://github.com/libp2p/js-libp2p-spdy) and [libp2p-mplex](https://github.com/libp2p/js-libp2p-mplex) and pluging them in is as easy as adding another transport. Let's revisit our libp2p bundle.
+Currently, we have [libp2p-mplex](https://github.com/libp2p/js-libp2p-mplex) and pluging it in is as easy as adding another transport. Let's revisit our libp2p configuration.
```
jacobheun: Similar to adding a transport, not like adding *another* transport.
```suggestion
Currently, we have [libp2p-mplex](https://github.com/libp2p/js-libp2p-mplex) and pluging it in is as easy as adding a transport. Let's revisit our libp2p configuration.
```

examples/protocol-and-stream-muxing/README.md
```diff
@@ -6,155 +6,161 @@ The feature of agreeing on a protocol over an established connection is what we
...
-    pull.log()
+node2.handle(['/a', '/b'], ({ protocol, stream }) => {
+  pipe(
+    stream,
+    toBuffer,
+    source => (async function () {
+      for await (const msg of source) {
+        console.log(`from: ${protocol}, msg: ${msg.toString()}`)
+      }
+    })()
```
jacobheun: Same here

examples/protocol-and-stream-muxing/3.js
```diff
@@ -1,88 +1,72 @@
...
-      pull.log()
+  node2.handle('/node-2', ({ stream }) => {
+    pipe(
+      stream,
+      toBuffer,
+      source => (async function () {
+        for await (const msg of source) {
+          console.log(msg.toString())
+        }
+      })()
```
jacobheun: We can simplify the pipes in each of the example files as mentioned in the readme below.

vasco-santos:
Thanks for pushing it @jacobheun ! Looks good

PR #503, opened 2019-12-10
refactor: examples transports

vasco-santos:
This PR refactors the `transports` example to the refactored async `js-libp2p`.

Needs:

### DIFF
--- a/examples/transports/2.js
+++ b/examples/transports/2.js
@@ -8,7 +8,7 @@ const MPLEX = require('libp2p-mplex')
 const PeerInfo = require('peer-info')
 
 const pipe = require('it-pipe')
-const { toBuffer } = require('it-buffer')
+const concat = require('it-concat')
 
 const createNode = async (peerInfo) => {
   // To signall the addresses we want to be available, we use
@@ -46,16 +46,12 @@ function printAddrs (node, number) {
   printAddrs(node1, '1')
   printAddrs(node2, '2')
 
-  node2.handle('/print', ({ stream }) => {
-    pipe(
+  node2.handle('/print', async ({ stream }) => {
+    const result = await pipe(
       stream,
-      toBuffer,
-      source => (async function () {
-        for await (const msg of source) {
-          console.log(msg.toString())
-        }
-      })(),
+      concat
     )
+    console.log(result.toString())
   })
 
   const { stream } = await node1.dialProtocol(node2.peerInfo, '/print')
--- a/examples/transports/2.js
+++ b/examples/transports/2.js
@@ -1,39 +1,31 @@
 /* eslint-disable no-console */
 'use strict'
 
-const libp2p = require('../../')
+const Libp2p = require('../..')
 const TCP = require('libp2p-tcp')
+const SECIO = require('libp2p-secio')
+const MPLEX = require('libp2p-mplex')
 const PeerInfo = require('peer-info')
-const waterfall = require('async/waterfall')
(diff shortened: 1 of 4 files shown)


examples/transports/3.js
```diff
@@ -1,70 +1,67 @@
...
 
-function print (protocol, conn) {
-  pull(
-    conn,
-    pull.map((v) => v.toString()),
-    pull.log()
+function print ({ stream }) {
+  pipe(
+    stream,
+    toBuffer,
```
jacobheun: No need to `toBuffer` if you're just doing `toString` on `msg`
```suggestion
```

examples/transports/3.js
```diff
@@ -1,70 +1,67 @@
...
+    source => (async function () {
+      for await (const msg of source) {
+        console.log(msg.toString())
+      }
+    })(),
```
jacobheun: Can be:
```js
async function (source) {
    console.log(msg.toString())
  }
}
```

examples/transports/README.md
```diff
@@ -96,78 +76,65 @@ That `QmW2cKTakTYqbQkUzBTEGXgWYFj1YEPeUndE1YWs6CBzDQ` is the PeerId that was cre
...
-      conn,
-      pull.map((v) => v.toString()),
-      pull.log()
+  node2.handle('/print', ({ stream }) => {
+    pipe(
+      stream,
+      toBuffer,
+      source => (async function () {
+        for await (const msg of source) {
+          console.log(msg.toString())
```
jacobheun: This will print the messages all on new lines so it looks like:
```sh
Hello
 
p2p
 
world
!
```

It might be better to collect everything and then do the console log. You can either use `toBuffer` and `concat` from streaming-iterables, or use [it-concat](https://github.com/alanshaw/it-concat) which will allow you to collect the BufferList and then just toString it.

```js
const concat = require('it-concat')
const result = await pipe(
  stream,
  concat
)
console.log(result.toString())
```

examples/transports/2.js
```diff
@@ -1,69 +1,67 @@
...
+  node2.handle('/print', ({ stream }) => {
+    pipe(
+      stream,
+      toBuffer,
+      source => (async function () {
+        for await (const msg of source) {
+          console.log(msg.toString())
+        }
+      })(),
     )
```
jacobheun: Improvement here mentioned in the readme comment below

examples/transports/README.md
```diff
@@ -243,58 +195,58 @@ As a rule, a libp2p node will only be capable of using a transport if: a) it has
...
+  ['node 2 dialed to node 3 successfully'],
+  stream2
+)
+
+// node 3 (WebSockets) attempts to dial to node 1 (TCP)
+try {
+  await node3.dialProtocol(node1.peerInfo, '/print')
+} catch (err) {
+  console.log('node 3 failed to dial to node 1 with:', err.message)
+}
```
jacobheun: The README no longer matches up with the file in 3. The readme example attempts to show a failure dialing from node 3 to node 1. This previously failed because we would remove transports we didn't have addresses for, which is unnecessary. Since we didn't have a TCP transport to dial to node 1 from 3, the dial would fail. Now that node3 has the transport it passes now. To properly demonstrate the failure we should create node 3 without a TCP transport, and node 1 without a websocket transport.

vasco-santos: changed it back to fail

PR #505, opened 2019-12-10
refactor: ping

vasco-santos:
This PR refactors ping to use `async-it` instead of `pull-streams`

Needs:

- [x] #490


src/ping/handler.js
```diff
@@ -1,48 +1,46 @@
...
 
-function mount (swarm) {
-  swarm.handle(PROTOCOL, (protocol, conn) => {
-    const stream = handshake({ timeout: 0 })
-    const shake = stream.handshake
+/**
+ * Subscribe ping protocol handler.
+ * @param {Libp2p} node
+ */
+function mount (node) {
```
vasco-santos: @jacobheun what do you think about these namings?

I thought that we could make this more consistent with the transports and have a `createListener` function or similar. But, I decided to gather your feedback first.

jacobheun: I think mount/unmount is okay. We're not so much creating listeners as we are attaching an application protocol to libp2p. Different services/protocols may have different needs so keeping it generic might be good. If we ever create an interface for this type of service we could reevaluate this then.

vasco-santos: agreed

src/index.js
```diff
@@ -280,17 +286,16 @@ class Libp2p extends EventEmitter {
...
-  //   const peerInfo = await getPeerInfoRemote(peer, this)
-  //   return new Ping(this._switch, peerInfo)
-  // }
+  /**
+   * Pings the given peer
+   * @param {PeerInfo|PeerId|Multiaddr|string} peer The peer to ping
+   * @returns {Promise<number>}
+   */
+  async ping (peer) {
+    const peerInfo = await getPeerInfoRemote(peer, this)
```
jacobheun: I think we should just use `getPeerInfo` for this, as ping really shouldn't support peers we don't know about locally. `getPeerInfoRemote` will search the network if it can't find it locally.

PR #507, opened 2019-12-11
refactor: examples/delegated-routing

vasco-santos:
This PR refactors the `delegated-routing` example to the refactored async `js-libp2p`.

### DIFF
--- a/examples/delegated-routing/.env
+++ b/examples/delegated-routing/.env
@@ -1,3 +1,3 @@
 # required because react-scripts scans *up* the tree from this project and finds
-# a conflicting version of eslint in the node_modules dir for js-ipfs.
+# a conflicting version of webpack in the node_modules dir for js-libp2p (aegir).
 SKIP_PREFLIGHT_CHECK=true
--- a/examples/delegated-routing/.env
+++ b/examples/delegated-routing/.env
@@ -1,0 +1,3 @@
+# required because react-scripts scans *up* the tree from this project and finds
+# a conflicting version of eslint in the node_modules dir for js-ipfs.
+SKIP_PREFLIGHT_CHECK=true
--- a/examples/delegated-routing/package.json
+++ b/examples/delegated-routing/package.json
@@ -3,19 +3,18 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "ipfs": "~0.34.4",
-    "libp2p": "github:libp2p/js-libp2p#master",
-    "libp2p-delegated-content-routing": "~0.2.2",
-    "libp2p-delegated-peer-routing": "~0.2.2",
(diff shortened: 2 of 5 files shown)


examples/delegated-routing/package.json
```diff
@@ -3,19 +3,18 @@
...
-    "libp2p-kad-dht": "~0.14.12",
-    "libp2p-mplex": "~0.8.5",
-    "libp2p-secio": "~0.11.1",
-    "libp2p-webrtc-star": "~0.15.8",
-    "libp2p-websocket-star": "~0.10.2",
-    "libp2p-websockets": "~0.12.2",
-    "react": "^16.8.6",
-    "react-dom": "^16.8.6",
-    "react-scripts": "2.1.8"
+    "ipfs": "~0.40.0",
```
vasco-santos: We need to get `libp2p` to `js-ipfs` before getting this to work. 

However, I could also not test this with #master . Do we need to run `go-ipfs` in the specific branch mentioned in the README?

jacobheun: There are some configuration issues that need to be resolved, see https://github.com/libp2p/js-libp2p/pull/371#pullrequestreview-255991314. 

In regards to requiring `ipfs`, we may want to revisit this example and only have libp2p in the client. The example cats a file right now, but in reality we just need to find things on the network via the delegate node.

vasco-santos: For this example, should we have a `js-libp2p` node running and providing a few cid that we can then find with the app example?

examples/delegated-routing/.env
```diff
@@ -0,0 +1,3 @@
+# required because react-scripts scans *up* the tree from this project and finds
+# a conflicting version of eslint in the node_modules dir for js-ipfs.
```
achingbrain, line 2: Is this necessary?  js-ipfs doesn't have a runtime dependency on eslint?  Maybe if this example was in the js-ipfs repo?

vasco-santos: This is needed. I copied this from the react example of `js-ipfs`, but did not read the comment: https://github.com/ipfs/js-ipfs/blob/master/examples/browser-create-react-app/.env
It is aegir that installs a version of `webpack` that is not compatible with `react-scripts`. I am going to change that comment

vasco-santos:
After syncing with @jacobheun , we will start by having a public delegated router configured.

We were having some issues finding peer via the delegate because of the lack of support for HTTPS in the `go-ipfs` API. This needs to be properly configured using a reverse proxy and generated SSL certificates. I will create an issue to track adding documentation for infrastructure for going to production to unblock this PR for now.

achingbrain:
Superseded by #1390

PR #510, opened 2019-12-11
feat: discovery modules from transports should be added

vasco-santos:
Discovery modules within Transports automatically added to discovery if enabled

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -110,7 +110,7 @@ Some available peer discovery modules are:
 - [js-libp2p-kad-dht](https://github.com/libp2p/js-libp2p-kad-dht)
 - [js-libp2p-webrtc-star](https://github.com/libp2p/js-libp2p-webrtc-star)
 
-Take into consideration that `peer-discovery` services within transports (such as `js-libp2p-webrtc-star`) are automatically gathered from the `transport` and do not need to be added in the discovery modules. However, these transports can also be configured and disabled as the other ones.
+**Note**: `peer-discovery` services within transports (such as `js-libp2p-webrtc-star`) are automatically gathered from the `transport`, via it's `discovery` property. As such, they do not need to be added in the discovery modules. However, these transports can also be configured and disabled as the other ones.
 
 If none of the available peer discovery protocols fulfills your needs, you can create a libp2p compatible one. A libp2p peer discovery protocol just needs to be compliant with the [Peer Discovery Interface](https://github.com/libp2p/js-interfaces/tree/master/src/peer-discovery).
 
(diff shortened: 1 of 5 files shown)


doc/CONFIGURATION.md
```diff
@@ -110,6 +110,8 @@ Some available peer discovery modules are:
 - [js-libp2p-kad-dht](https://github.com/libp2p/js-libp2p-kad-dht)
 - [js-libp2p-webrtc-star](https://github.com/libp2p/js-libp2p-webrtc-star)
 
+Take into consideration that `peer-discovery` services within transports (such as `js-libp2p-webrtc-star`) are automatically gathered from the `transport` and do not need to be added in the discovery modules. However, these transports can also be configured and disabled as the other ones.
```
jacobheun:
```suggestion
**Note**: `peer-discovery` services within transports (such as `js-libp2p-webrtc-star`) are automatically gathered from the `transport`, via it's `discovery` property. As such, they do not need to be added in the discovery modules. However, these transports can also be configured and disabled as the other ones.
```

src/index.js
```diff
@@ -448,7 +448,8 @@ class Libp2p extends EventEmitter {
         config = { ...config, ...this._config.peerDiscovery[DiscoveryService.tag] }
       }
 
-      if (config.enabled) {
+      if (config.enabled &&
+        !this._discovery.filter((service) => service.tag === DiscoveryService.tag).length) { // not already added
```
jacobheun: Should we just make `this._discovery` a Map? They require a tag, it might be easier for checking existence and getting specific ones should we need to do that.

test/peer-discovery/index.spec.js
```diff
@@ -10,57 +10,108 @@ const defer = require('p-defer')
...
+            webRTCStar: {
+              enabled: true
+            }
+          }
+        }
+      }))
+
+      await libp2p.start()
+
+      expect(libp2p._discovery).to.have.lengthOf(1)
```
jacobheun: We should verify this is the discovery service we expect. Making `libp2p. _discovery` a map would make this easy to do.

src/index.js
```diff
@@ -458,11 +459,23 @@ class Libp2p extends EventEmitter {
...
+
+    // Transport modules with discovery
+    for (const Transport of this.transportManager.getTransports()) {
+      if (Transport.discovery) {
+        setupService(Transport.discovery)
       }
     }
 
-    return this._discovery.map(d => d.start())
+    return Array.from(this._discovery.values()).map(d => d.start())
```
jacobheun: We should promisify these calls since `start` is supposed to be async, right now it's just an array of promises. The jsdocs already mention yielding a promise, so this will actually do that. Additionally, you can just pass the map function to `Array.from`, which should avoid double iteration.
```suggestion
    return Promise.all(Array.from(this._discovery.values(), d => d.start()))
```

PR #514, opened 2019-12-12
docs: getting started

vasco-santos:
This getting started doc aims to be the entry point for new users of `js-libp2p` since it goes deeper in detail within each concept.

For more advanced users, we expect that they consult the API and CONFIG docs mostly, as they are more straightforward documents

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -28,6 +28,7 @@
   * [`metrics.protocols`](#metricsprotocols)
   * [`metrics.forPeer`](#metricsforpeer)
   * [`metrics.forProtocol`](#metricsforprotocol)
+* [Events](#events)
 * [Types](#types)
   * [`Stats`](#stats)
 
@@ -93,39 +94,6 @@ Required keys in the `options` object:
 
 - `peerInfo`: instance of [PeerInfo][] that contains the [PeerId][], Keys and [multiaddrs][multiaddr] of the libp2p Node (optional when using `.create`).
 - `modules.transport`: An array that must include at least 1 compliant transport. See [modules that implement the transport interface](https://github.com/libp2p/js-interfaces/tree/master/src/transport#modules-that-implement-the-interface).
-
-</details>
-
-Once you have a libp2p instance, you are able to listen to several events it emits, so that you can be noticed of relevant network events.
-
-<details><summary>Events</summary>
-
-#### An error has occurred
-
-`libp2p.on('error', (err) => {})`
-
-- `err`: instance of `Error`
-
-#### A peer has been discovered
-
-`libp2p.on('peer:discovery', (peer) => {})`
-
-If `autoDial` option is `true`, applications should **not** attempt to connect to the peer
-unless they are performing a specific action. See [peer discovery and auto dial](./PEER_DISCOVERY.md) for more information.
-
-- `peer`: instance of [PeerInfo][https://github.com/libp2p/js-peer-info]
-
-#### We have a new connection to a peer
-
-`libp2p.on('peer:connect', (peer) => {})`
-
-- `peer`: instance of [PeerInfo][https://github.com/libp2p/js-peer-info]
-
-#### We have closed a connection to a peer
-
-`libp2p.on('peer:disconnect', (peer) => {})`
-
-- `peer`: instance of [PeerInfo][https://github.com/libp2p/js-peer-info]
 
 </details>
 
@@ -776,6 +744,37 @@ Returns the [`Stats`](#stats) object for a given protocol if it is being tracked
 const peerStats = libp2p.metrics.forProtocol('/meshsub/1.0.0')
 console.log(peerStats.toJSON())
 ```
+
+## Events
+
+Once you have a libp2p instance, you are able to listen to several events it emits, so that you can be noticed of relevant network events.
+
+#### An error has occurred
+
+`libp2p.on('error', (err) => {})`
+
+- `err`: instance of `Error`
+
+#### A peer has been discovered
+
+`libp2p.on('peer:discovery', (peer) => {})`
+
+If `autoDial` option is `true`, applications should **not** attempt to connect to the peer
+unless they are performing a specific action. See [peer discovery and auto dial](./PEER_DISCOVERY.md) for more information.
+
+- `peer`: instance of [PeerInfo][https://github.com/libp2p/js-peer-info]
+
+#### We have a new connection to a peer
+
+`libp2p.on('peer:connect', (peer) => {})`
+
+- `peer`: instance of [PeerInfo][https://github.com/libp2p/js-peer-info]
+
+#### We have closed a connection to a peer
+
+`libp2p.on('peer:disconnect', (peer) => {})`
+
+- `peer`: instance of [PeerInfo][https://github.com/libp2p/js-peer-info]
 
 ## Types
 
--- a/doc/API.md
+++ b/doc/API.md
@@ -747,7 +747,7 @@ console.log(peerStats.toJSON())
 
 ## Events
 
-Once you have a libp2p instance, you are able to listen to several events it emits, so that you can be noticed of relevant network events.
+Once you have a libp2p instance, you can listen to several events it emits, so that you can be notified of relevant network events.
 
 #### An error has occurred
 
@@ -770,7 +770,7 @@ unless they are performing a specific action. See [peer discovery and auto dial]
 
 - `peer`: instance of [PeerInfo][https://github.com/libp2p/js-peer-info]
 
-#### We have closed a connection to a peer
+#### An existing connection to a peer has been closed
 
 `libp2p.on('peer:disconnect', (peer) => {})`
 
--- a/doc/API.md
+++ b/doc/API.md
@@ -764,13 +764,17 @@ unless they are performing a specific action. See [peer discovery and auto dial]
 
 - `peer`: instance of [PeerInfo][https://github.com/libp2p/js-peer-info]
 
-#### We have a new connection to a peer
+#### A new connection to a peer has been opened
+
+This event will be triggered anytime a new Connection is established to another peer.
 
 `libp2p.on('peer:connect', (peer) => {})`
 
 - `peer`: instance of [PeerInfo][https://github.com/libp2p/js-peer-info]
 
 #### An existing connection to a peer has been closed
+
+This event will be triggered anytime we are disconnected from another peer, regardless of the circumstances of that disconnection. If we happen to have multiple connections to a peer, this event will **only** be triggered when the last connection is closed.
 
 `libp2p.on('peer:disconnect', (peer) => {})`
 
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -129,7 +129,7 @@ If you want to know more about libp2p peer discovery, you should read the follow
 Some available content routing modules are:
 
 - [js-libp2p-kad-dht](https://github.com/libp2p/js-libp2p-kad-dht)
-- [js-libp2p-delegated-peer-routing](https://github.com/libp2p/js-libp2p-delegated-peer-routing)
+- [js-libp2p-delegated-content-routing](https://github.com/libp2p/js-libp2p-delegated-content-routing)
 
 If none of the available content routing protocols fulfills your needs, you can create a libp2p compatible one. A libp2p content routing protocol just needs to be compliant with the [Content Routing Interface](https://github.com/libp2p/js-interfaces/tree/master/src/content-routing). **(WIP: This module is not yet implemented)**
 
--- a/doc/GETTING_STARTED.md
+++ b/doc/GETTING_STARTED.md
@@ -1,25 +1,21 @@
-# GETTING STARTED
+# Getting Started
 
-Welcome to libp2p, let's get you setup your first and fully functional libp2p node 🚀
+Welcome to libp2p, this guide will walk you through setting up a fully functional libp2p node 🚀
 
 - [Getting Started](#getting-started)
   - [Install](#install)
-  - [Usage](#usage)
+  - [Configuring libp2p](#configuring-libp2p)
     - [Basic setup](#basic-setup)
-      - [Transport](#transport)
+      - [Transports](#transports)
       - [Connection Encryption](#connection-encryption)
-      - [Overview](#overview)
+      - [Running Libp2p](#running-libp2p)
     - [Custom setup](#custom-setup)
       - [Multiplexing](#multiplexing)
       - [Peer Discovery](#peer-discovery)
       - [Peer Routing](#peer-routing)
       - [Content Routing](#content-routing)
-      - [Pubsub](#pubsub)
+    - [Pubsub](#pubsub)
   - [What is next](#what-is-next)
-
-libp2p is a composable and modular networking stack, which empowers its users to just take the modules they need. All the flexibility it provides, in order to fulfil every network requirement comes with a cost. For getting a libp2p node running, users are faced with a large number of possible configuration setups, which involves a considerable number of concepts that they might not be familiar with.
-
-This document intends to help in the first contact with `js-libp2p`.
 
 ## Install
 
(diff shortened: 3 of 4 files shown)


vasco-santos:
cc @terichadbourne

vasco-santos:
> I think it would be good to have the final code of this in ./examples. If anyone gets stuck, they can go reference the example code and we can more easily ensure this is running properly in future releases, as we test all examples.

👍

vasco-santos:
I think I addressed all the feedback except for:

> I think it would be good to have the final code of this in ./examples. If anyone gets stuck, they can go reference the example code and we can more easily ensure this is running properly in future releases, as we test all examples.

For that one in specific, I will create an issue to track all the remaining endeavours. I want to get all the examples being tested in CI, in order to ease the release process

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,482 @@
...
+
+### Basic setup
+
+Now that we have libp2p installed, let's configure the absolute minimum needed to get your node running. The only modules libp2p requires are a [**Transport**][transport] and [**Crypto**][crypto] module. Moreover, we recommend that a basic setup should also have a [**Stream Multiplexer**](streamMuxer) configured. Let's start by setting up a Transport.
+
+#### Transports
+
+Libp2p uses Transports to establish connections between peers over the network. You can configure 1 Transport, or as many as you like. Supporting more Transports will improve the ability for other nodes on the network to communicate with you.
+
+You should select Transports according to the runtime where your application will run. You can see a list of some of the available Transports in the [configuration readme](./CONFIGURATION.md#transport). We are going to install `libp2p-websockets`, as it can be used in both Node.js and the browser.
```
jacobheun:
```suggestion
You should select Transports according to the runtime of your application (Node.js / browser). You can see a list of some of the available Transports in the [configuration readme](./CONFIGURATION.md#transport). We are going to install `libp2p-websockets`, as it can be used in both Node.js and the browser.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,482 @@
...
+```
+
+<details><summary>Read More</summary>
+If you want to know more about libp2p connection encryption, you should read the following content:
+
+- https://docs.libp2p.io/concepts/secure-comms
+- https://github.com/libp2p/specs/tree/master/connections
+</details>
+
+#### Multiplexing
```
jacobheun: I think we need to split multiplexing configuration and protocol creation into separate sections.

This Multiplexing section also feels like it's diving a little too deep into what multiplexing is. I think we should reduce the up front language and the leverage the `Read More` section to  send folks to more detailed information.

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,482 @@
...
+
+In this context, your node will not be efficient if you open multiple connections with a single peer to use different protocols. Therefore, each protocol works as an independent logical stream and you can use a stream multiplexer, in order to share a common underlying transport medium between streams (single connection).
+
+Looking at the [available stream multiplexing](./CONFIGURATION.md#stream-multiplexing) protocols, we are only able to use `libp2p-mplex` for the time being. Bear in mind that future libp2p transports might have `multiplexing` capabilities already built-in (such as `QUIC`).
+
+You can install `libp2p-mplex` and add it to your libp2p node as follows in the next example. You should revisit the [API](./API.md) document, more specifically [API#dial](./API.md#dial) and [API#handle](./API.md#handle).
+
+We will also install the next dependencies for easily work with [async iterable duplex streams](https://gist.github.com/alanshaw/591dc7dd54e4f99338a347ef568d6ee9).
```
jacobheun: We should point to the local README we have for this, as we'll also be including more examples of using iterable streams in libp2p https://github.com/libp2p/js-libp2p/blob/refactor/async-await/doc/STREAMING_ITERABLES.md

doc/GETTING_STARTED.md
jacobheun: I think the protocol creation section should also come after starting and stopping your node. That way if people have issues with their early configuration, it can be caught prior to building out a custom protocol.

Do you think it would be more helpful to point people to the protocol example, instead of building that out here? I think `Getting Started` can focus on creating a first node, and connecting it to the public network with basic peer discovery and new connection logging. Then we could point to the examples folder. This would also put us in a good spot to revamp those in the future, where all examples are based off of the configuration from the getting started guide.

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+- https://docs.libp2p.io/concepts/secure-comms
+- https://github.com/libp2p/specs/tree/master/connections
+</details>
+
+#### Multiplexing
+
+With libp2p you are able to create your own protocols so that peers can communicate in a meaningful way to your application. Each protocol works as an independent logical stream and you can use a stream multiplexer, in order to share a common underlying transport medium between streams (single connection). As an example, you can create a chat protocol on top of libp2p and use it to exchange chat messages.
+
+Looking at the [available stream multiplexing](./CONFIGURATION.md#stream-multiplexing) protocols, we are only able to use `libp2p-mplex` for the time being. Bear in mind that future libp2p transports might have `multiplexing` capabilities already built-in (such as `QUIC`).
```
jacobheun:
```suggestion
Looking at the [available stream multiplexing](./CONFIGURATION.md#stream-multiplexing) modules, js-libp2p currently only supports `libp2p-mplex`, so we will use that here. Bear in mind that future libp2p transports might have `multiplexing` capabilities already built-in (such as `QUIC`).
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+<details><summary>Read More</summary>
+If you want to know more about libp2p connection encryption, you should read the following content:
+
+- https://docs.libp2p.io/concepts/secure-comms
+- https://github.com/libp2p/specs/tree/master/connections
+</details>
+
+#### Multiplexing
+
+With libp2p you are able to create your own protocols so that peers can communicate in a meaningful way to your application. Each protocol works as an independent logical stream and you can use a stream multiplexer, in order to share a common underlying transport medium between streams (single connection). As an example, you can create a chat protocol on top of libp2p and use it to exchange chat messages.
```
jacobheun:
```suggestion
While multiplexers are not strictly required, they are highly recommended as they improve the effectiveness and efficiency of connections for the various protocols libp2p runs. Including a multiplexer will allow libp2p to run several of its internal protocols, like Identify, as well as allowing your application to easily run any number of protocols over a single connection.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+const node = await Libp2p.create({
+  modules: {
+    transport: [WEBSOCKETS],
+    connEncryption: [SECIO],
+    streamMuxer: [MPLEX]
+  }
+})
+```
+
+Protocols design is similar to an http api, where each protocol has an handler, receives a message and returns a response. If you want to create your own protocol, start by having a look at the [Protocol and Stream Muxing Example](../examples/protocol-and-stream-muxing).
```
jacobheun:
```suggestion
Protocol design is similar to an http api, where each protocol has a registered handler that receives and responds to requests. If you want to create your own protocol, start by having a look at the [Protocol and Stream Muxing Example](../examples/protocol-and-stream-muxing).
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+})
+
+// start libp2p
+await node.start()
+
+// stop libp2p
+await node.stop()
+```
+
+If you are not familiar with multiaddrs, you should read the [multiformats/js-multiaddr](https://github.com/multiformats/js-multiaddr) README.
```
jacobheun: This seems out of place, not sure where this should go.

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+### Custom setup
+
+After having your libp2p node running, it is time to configure all its features according to your project needs.
```
jacobheun:
```suggestion
Once your libp2p node is running, it is time to get it connected to the public network. We can do this via peer discovery.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+#### Peer Discovery
+
+In a P2P context, peer discovery is critical to a functioning system. It is not pratical to connect to all the peers we know manually, nor to get to know most of the peers' addresses in advance.
```
jacobheun:
```suggestion
Peer discovery is an important part of creating a well connected libp2p node. A static list of peers will often be used to join the network, but it's useful to couple other discovery mechanisms to ensure you're able to discover other peers that are important to your application.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+Looking at the [available peer discovery](./CONFIGURATION.md#peer-discovery) protocols, there are several options to be considered:
+- If you already know the addresses of some other network peers, you should consider using `libp2p-bootstrap` as this is the easiest way of getting your peer into the network.
+- If it is likely that you will have other peers on your network, `libp2p-mdns` is a must. It allows peers to discover each other when on the same local network with zero configuration. mDNS uses a multicast system of DNS records.
```
jacobheun:
```suggestion
- If it is likely that you will have other peers on your local network, `libp2p-mdns` is a must if you're not using the browser. It allows peers to discover each other when on the same local network with zero configuration. mDNS uses a multicast system of DNS records.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+- In the event of working on a browser runtime, you must go with `js-libp2p-webrtc-star`.
```
jacobheun:
```suggestion
- If your application is browser based you can use the transport `js-libp2p-webrtc-star`, which includes a rendezvous based peer sharing mechanism.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+- if you need to crawl the network and discover a large number of peers, `js-libp2p-kad-dht`.
```
jacobheun:
```suggestion
- A random walk approach can be used via `js-libp2p-kad-dht`, to crawl the network and find new peers along the way.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+Reviewing the events specification on [Libp2p.create](./API.md#create), each time a peer is discovered a `peer:discovery` event will be emitted by the node.
```
jacobheun: Should we update the API doc so that we can deep link to the Events section? I think it convolutes this section and would make it easier of we could just link directly to that events declaration there.

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+```sh
+npm install libp2p-bootstrap libp2p-mdns
+```
+
+We can provide specific configurations for each protocol within a `config.peerDiscovery` property in the options.
+
+```js
+const Libp2p = require('libp2p')
+const TCP = require('libp2p-tcp')
```
jacobheun: This switches over to TCP, any reason for that? We should keep it consistent with the rest of the guide.

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+If you want to know more about libp2p peer discovery, you should read the following content:
+
+- https://github.com/libp2p/specs/blob/master/discovery/mdns.md
+</details>
+
+#### Peer Routing
+
+Peer Routing offers a way to find other peers in the network by issuing queries using a Peer Routing algorithm, through their peer ids.
+
+Looking at the [available peer routing](./CONFIGURATION.md#peer-routing) protocols, we might choose `libp2p-kad-dht` or `libp2p-delegated-peer-routing`. If your runtime is a browser or a low power device, you might prefer to use `libp2p-delegated-peer-routing` as you are able to delegate other peer to issue the queries for you. However, you need to know a peer able to do that. Otherwise, you should use `libp2p-kad-dht`.
```
jacobheun:
```suggestion
Looking at the [available peer routing](./CONFIGURATION.md#peer-routing) modules, we might choose `libp2p-kad-dht` or `libp2p-delegated-peer-routing`. If your runtime is a browser or a low power device, you might prefer to use `libp2p-delegated-peer-routing` as you are able to delegate to another peer to issue the queries for you. However, you need to know a peer able to do that. Otherwise, you should use `libp2p-kad-dht`.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+In this context, we will install the dht. You should revisit the [API](./API.md) document, more specifically [API#peerRouting.findPeer](./API.md#peerRoutingfindPeer).
```
jacobheun:
```suggestion
In this guide, we will install `libp2p-kad-dht`. You should revisit the [API](./API.md) document, more specifically [API#peerRouting.findPeer](./API.md#peerRoutingfindPeer).
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+```sh
+npm install libp2p-kad-dht
+```
+
+`js-libp2p` configuration is not well defined yet regarding peer routing. Therefore, the `dht` should be plugged through a `modules.dht` property, while the `libp2p-delegated-peer-routing` should be plugged into `modules.peerRouting` array.
```
jacobheun: It is defined, but the DHT isn't currently separable into those components so it has it's own configuration parameter.

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+<details><summary>Read More</summary>
+If you want to know more about libp2p peer routing, you should read the following content:
+
+- https://docs.libp2p.io/concepts/peer-routing
+</details>
+
+#### Content Routing
+
+Content routing provides a way to find where content lives in the network. It works in two steps: 1) Peers provide (announce) to the network that they are holders of specific content and 2) Peers issue queries to find where that content lives. A Content Routing mechanism could be as complex as a DHT or as simple as a registry somewhere in the network.
```
jacobheun:
```suggestion
Content routing provides a way to find and announce where content lives in the network. It works in two steps: 1) Peers announce to the network that they are a provider of a certain piece of content and 2) Peers issue queries to find where that content lives. A Content Routing mechanism could be as complex as a DHT or as simple as a registry somewhere in the network.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+Looking at the [available content routing](./CONFIGURATION.md#content-routing) protocols, we might choose `libp2p-kad-dht` or `libp2p-delegated-content-routing`. If of your runtime is a browser or a low power device, you might prefer to use `libp2p-delegated-peer-routing` as you are able to delegate other peer to issue the queries for you. However, you need to know a peer able to do that. Otherwise, you should use `libp2p-kad-dht`.
```
jacobheun:
```suggestion
Looking at the [available content routing](./CONFIGURATION.md#content-routing) modules, we might choose `libp2p-kad-dht` or `libp2p-delegated-content-routing`. If your runtime is a browser or a low power device, you might prefer to use `libp2p-delegated-content-routing` as you are able to delegate to another peer to issue the queries for you. However, you need to know a peer able to do that. Otherwise, you should use `libp2p-kad-dht`.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+In this context, we will install the dht. You should revisit the [API](./API.md) document, more specifically [API#contentrouting.findProviders](./API.md#contentroutingfindproviders), [API#contentrouting.provide](./API.md#contentroutingprovide), [API#contentrouting.put](./API.md#contentroutingput), [API#contentrouting.get](./API.md#contentroutingget) and [API#contentrouting.contentrouting.getMany](./API.md#contentroutingcontentroutinggetmany).
```
jacobheun:
```suggestion
In this guide, we will install `libp2p-kad-dht`. If you'd like to review the available content routing methods, you should revisit the [API](./API.md) document, more specifically [API#contentrouting.findProviders](./API.md#contentroutingfindproviders), [API#contentrouting.provide](./API.md#contentroutingprovide), [API#contentrouting.put](./API.md#contentroutingput), [API#contentrouting.get](./API.md#contentroutingget) and [API#contentrouting.contentrouting.getMany](./API.md#contentroutingcontentroutinggetmany).
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+```sh
+npm install libp2p-kad-dht
+```
+
+`js-libp2p` configuration is not well defined yet regarding content routing routing. Therefore, the `dht` should be plugged through a `modules.dht` property, while the `libp2p-delegated-content-routing` should be plugged into `modules.contentRouting` array.
```
jacobheun: Same here, it is defined, but the DHT does not yet conform to the separate configuration.

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+      }
+    }
+  }
+})
+
+// start libp2p
+await node.start()
+
+// provide cid
+await libp2p.contentRouting.provide(cid)
```
jacobheun: Since we're aiming to have people join the live network we should probably not have them provide content, as their node is going to go offline pretty quickly. What do you think?

For findProviders we should just use a well known cid, and we should include the cid module here. A website or the ipfs readme might be good options.

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+<details><summary>Read More</summary>
+If you want to know more about libp2p content routing, you should read the following content:
+
+- https://docs.libp2p.io/concepts/content-routing
+</details>
+
+### Pubsub
+
+If you are looking for real time message exchange between peers, pubsub may be what you are looking for. Publish/Subscribe is a system where peers congregate around topics they are interested in. Peers interested in a topic are said to be subscribed to that topic and should receive the data published on it from other peers.
```
jacobheun:
```suggestion
If you are looking for real time message exchange between multiple peers, pubsub may be what you are looking for. Publish/Subscribe is a system where peers congregate around topics they are interested in. Peers interested in a topic will subscribe to that topic and receive the data published on it from other peers.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+
+Looking at the [available pubsub routers](./CONFIGURATION.md#pubsub), you might choose `libp2p-gossipsub` first, and consider other options later on. You should revisit the [API](./API.md) document, more specifically [API#pubsub.getSubscribers](./API.md#pubsubgetsubscribers), [API#pubsub.getTopics](./API.md#pubsubgettopics), [API#pubsub.publish](./API.md#pubsubpublish) and [API#pubsub.subscribe](./API.md#pubsubsubscribe).
```
jacobheun:
```suggestion
You can see the [available pubsub routers](./CONFIGURATION.md#pubsub) in the configuration readme. In this guide we will use `libp2p-gossipsub` as it is a more efficient alternative to `libp2p-floodsub`. You can revisit the [API](./API.md) document to familiarize yourself with the available methods, more specifically [API#pubsub.getSubscribers](./API.md#pubsubgetsubscribers), [API#pubsub.getTopics](./API.md#pubsubgettopics), [API#pubsub.publish](./API.md#pubsubpublish) and [API#pubsub.subscribe](./API.md#pubsubsubscribe).
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,428 @@
...
+```js
+const Libp2p = require('libp2p')
+const TCP = require('libp2p-tcp')
+const MPLEX = require('libp2p-mplex')
+const SECIO = require('libp2p-secio')
+const GossipSub = require('libp2p-gossipsub')
+
+const node = await Libp2p.create({
+  modules: {
+    transport: [TCP],
```
jacobheun: Same thing here, why the switch to TCP?

doc/GETTING_STARTED.md
vasco-santos: I think we can just remove it

doc/GETTING_STARTED.md
vasco-santos: Good call!

doc/GETTING_STARTED.md
vasco-santos: Should we add it in the same level as the Instance Methods and Types?

jacobheun: Yeah, I think so

jacobheun:
I went ahead and cleaned up some of the language. I also removed mdns, DHT and Pubsub setup from the guide for now. The problem with having these included is that the setup as it was, wasn't sufficient for these things to work properly. For example, you could start and run pubsub, but you wouldn't be publishing to anything. I think it's more valuable to get people running a node and connecting it to the public network, and then referencing the Configuration readme and examples folder for additional sources of configuration. The examples do a better job of ensuring each piece will yield results by doing things like running multiple nodes when necessary.

I think we can add additional components to this in the future, but they should work without needing any additional configuration/resources.

doc/API.md
```diff
@@ -777,6 +745,37 @@ const peerStats = libp2p.metrics.forProtocol('/meshsub/1.0.0')
...
+#### A peer has been discovered
+
+`libp2p.on('peer:discovery', (peer) => {})`
+
+If `autoDial` option is `true`, applications should **not** attempt to connect to the peer
+unless they are performing a specific action. See [peer discovery and auto dial](./PEER_DISCOVERY.md) for more information.
+
+- `peer`: instance of [PeerInfo][https://github.com/libp2p/js-peer-info]
+
+#### We have a new connection to a peer
```
alanshaw:
```suggestion
#### An incoming connection from a peer has been established
```

Is it only incoming or outgoing also?

doc/API.md
```diff
@@ -777,6 +745,37 @@ const peerStats = libp2p.metrics.forProtocol('/meshsub/1.0.0')
...
+
+`libp2p.on('peer:connect', (peer) => {})`
+
+- `peer`: instance of [PeerInfo][https://github.com/libp2p/js-peer-info]
+
+#### We have closed a connection to a peer
```
alanshaw:
```suggestion
#### An existing connection to a peer has been closed
```

Can we explain why this might happen? Is it only if the remote disconnects or do we also get this event when we disconnect?

doc/CONFIGURATION.md
```diff
@@ -129,7 +129,7 @@ If you want to know more about libp2p peer discovery, you should read the follow
 Some available content routing modules are:
 
 - [js-libp2p-kad-dht](https://github.com/libp2p/js-libp2p-kad-dht)
-- [js-libp2p-delegated-peer-routing](https://github.com/libp2p/js-libp2p-delegated-peer-routing)
+- [js-libp2p-delegated-content-routing](https://github.com/libp2p/js-libp2p-delegated-content-routing)
```
alanshaw: lol

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,247 @@
...
+npm install libp2p
+```
+
+## Configuring libp2p
+
+If you're new to libp2p, we recommend configuring your node in stages, as this can make troubleshooting configuration issues much easier. In this guide, we'll do just that. If you're more experienced with libp2p, you may wish to jump to the [Configuration readme](./CONFIGURATION.md).
+
+### Basic setup
+
+Now that we have libp2p installed, let's configure the minimum needed to get your node running. The only modules libp2p requires are a [**Transport**][transport] and [**Crypto**][crypto] module. However, we recommend that a basic setup should also have a [**Stream Multiplexer**](streamMuxer) configured, which we will explain shortly. Let's start by setting up a Transport.
```
alanshaw: Is the muxer really optional?

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,247 @@
...
+const WebSockets = require('libp2p-websockets')
+
+const node = await Libp2p.create({
+  modules: {
+    transport: [WebSockets]
+  }
+})
+```
+
+As new Transports are created, you can reevaluate the needs of your appliction and select the Transports that best suit your requirements. You may wish to remove the Transports you had before, or simply append the new Transports to `modules.transport` in order to establish connections with peers that support either.
```
alanshaw:
```suggestion
There are multiple libp2p transports available, you should evaluate the needs of your application and select the Transport(s) that best suit your requirements. You can add as many transports as you like to `modules.transport` in order to establish connections with as many peers as possible.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,247 @@
...
+
+If you're new to libp2p, we recommend configuring your node in stages, as this can make troubleshooting configuration issues much easier. In this guide, we'll do just that. If you're more experienced with libp2p, you may wish to jump to the [Configuration readme](./CONFIGURATION.md).
+
+### Basic setup
+
+Now that we have libp2p installed, let's configure the minimum needed to get your node running. The only modules libp2p requires are a [**Transport**][transport] and [**Crypto**][crypto] module. However, we recommend that a basic setup should also have a [**Stream Multiplexer**](streamMuxer) configured, which we will explain shortly. Let's start by setting up a Transport.
+
+#### Transports
+
+Libp2p uses Transports to establish connections between peers over the network. You can configure any number of Transports, but you only need 1 to start with. Supporting more Transports will improve the ability of your node to speak to a larger number of nodes on the network.
```
alanshaw: Could we explain a little what a transport is?

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,247 @@
...
+
+- https://docs.libp2p.io/concepts/transport
+- https://github.com/libp2p/specs/tree/master/connections
+</details>
+
+#### Connection Encryption
+
+Encryption is an important part of communicating on the libp2p network. Every connection must be encrypted to help ensure security for everyone. As such, Connection Encryption (Crypto) is a required component of libp2p.
+
+There are a growing number of Crypto modules being developed for libp2p. As those are released they will be tracked in the [available Connection Encryption](./CONFIGURATION.md#connection-encryption) section of the configuration readme. For now, we are going to configure our node to use the `libp2p-secio` module, which is widely supported across the various libp2p implementations.
```
alanshaw:
```suggestion
There are a growing number of Crypto modules being developed for libp2p. As those are released they will be tracked in the [Connection Encryption section of the configuration readme](./CONFIGURATION.md#connection-encryption). For now, we are going to configure our node to use the `libp2p-secio` module, which is widely supported across the various libp2p implementations.
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,247 @@
...
+const Libp2p = require('libp2p')
+const WebSockets = require('libp2p-websockets')
+const SECIO = require('libp2p-secio')
+const MPLEX = require('libp2p-mplex')
+
+const Bootstrap = require('libp2p-bootstrap')
+
+// Known peers addresses
+const bootstrapMultiaddrs = [
+  '/dns4/ams-1.bootstrap.libp2p.io/tcp/443/wss/ipfs/QmSoLer265NRgSp2LA3dPaeykiS1J6DifTC88f5uVQKNAd',
```
alanshaw:
```suggestion
  '/dns4/ams-1.bootstrap.libp2p.io/tcp/443/wss/p2p/QmSoLer265NRgSp2LA3dPaeykiS1J6DifTC88f5uVQKNAd',
```

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,247 @@
...
+  '/dns4/lon-1.bootstrap.libp2p.io/tcp/443/wss/ipfs/QmSoLMeWqB7YGVLJN3pNLQpmmEk35v6wYtsMGLzSr5QBU3'
```
alanshaw:
```suggestion
  '/dns4/lon-1.bootstrap.libp2p.io/tcp/443/wss/p2p/QmSoLMeWqB7YGVLJN3pNLQpmmEk35v6wYtsMGLzSr5QBU3'
```

doc/API.md
jacobheun: It is both.

doc/API.md
jacobheun: It occurs anytime we are disconnected from a peer, regardless of what caused it.

doc/GETTING_STARTED.md
jacobheun: Yes, you should be able to run an unmuxed connection. The primary reason would be to support applications that don't need/want it, such as lower energy/power devices, although this may not be a valid use case for js-libp2p.

doc/API.md
alanshaw: Gotcha, might be an idea just to be explicit that it is!

PR #517, opened 2019-12-13
fix: upgrader should not need muxers

vasco-santos:
With the current codebase, having a `streamMuxer` is mandatory, which was not the case before. This PR allows a connection to be upgraded just with crypto.

Note: the `connection` properties regarding multiplexers are throwing an error now.

### DIFF
--- a/src/upgrader.js
+++ b/src/upgrader.js
@@ -64,7 +64,7 @@ class Upgrader {
   async upgradeInbound (maConn) {
     let encryptedConn
     let remotePeer
-    let muxedConnection
+    let upgradedConn
     let Muxer
     let cryptoProtocol
     let setPeer
@@ -94,7 +94,11 @@ class Upgrader {
       } = await this._encryptInbound(this.localPeer, protectedConn, this.cryptos))
 
       // Multiplex the connection
-      ;({ stream: muxedConnection, Muxer } = await this._multiplexInbound(encryptedConn, this.muxers))
+      if (this.muxers.size) {
+        ({ stream: upgradedConn, Muxer } = await this._multiplexInbound(encryptedConn, this.muxers))
+      } else {
+        upgradedConn = encryptedConn
+      }
     } catch (err) {
       log.error('Failed to upgrade inbound connection', err)
       await maConn.close(err)
@@ -112,7 +116,7 @@ class Upgrader {
       cryptoProtocol,
       direction: 'inbound',
       maConn,
-      muxedConnection,
+      upgradedConn,
       Muxer,
       remotePeer
     })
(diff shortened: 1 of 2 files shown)


test/upgrading/upgrader.spec.js
```diff
@@ -116,6 +116,26 @@ describe('Upgrader', () => {
...
+    sinon.stub(localUpgrader, 'cryptos').value(cryptos)
+    sinon.stub(remoteUpgrader, 'cryptos').value(cryptos)
+
+    const connections = await Promise.all([
+      localUpgrader.upgradeOutbound(outbound),
+      remoteUpgrader.upgradeInbound(inbound)
+    ])
+
+    expect(connections).to.have.length(2)
+  })
```
vasco-santos: Should I test the errors with multiplexer operations in the `connection`?

src/upgrader.js
```diff
@@ -205,10 +213,34 @@ class Upgrader {
...
     direction,
     maConn,
-    muxedConnection,
+    upgradedConn,
     Muxer,
     remotePeer
   }) {
+    if (!Muxer) {
+      // Create the connection
+      maConn.timeline.upgraded = Date.now()
```
jacobheun: This will miss the timeline proxy below. That should probably get moved up to the top of the function

test/upgrading/upgrader.spec.js
jacobheun: It might be good to add a quick test. They should also throw errors with codes.

src/upgrader.js
```diff
@@ -244,20 +292,9 @@ class Upgrader {
     }
 
     // Pipe all data through the muxer
-    pipe(muxedConnection, muxer, muxedConnection)
+    pipe(upgradedConn, muxer, upgradedConn)
 
     maConn.timeline.upgraded = Date.now()
```
jacobheun: We should just set this right after we created the timeline Proxy, that way we don't need to set it here and in the `!Muxer` if statement above.

src/upgrader.js
```diff
@@ -195,7 +203,7 @@ class Upgrader {
    * @param {string} cryptoProtocol The crypto protocol that was negotiated
    * @param {string} direction One of ['inbound', 'outbound']
    * @param {MultiaddrConnection} maConn The transport layer connection
-   * @param {*} muxedConnection A duplex connection returned from multiplexer selection
+   * @param {*} upgradedConn A duplex connection returned from multiplexer selection
```
jacobheun:
```suggestion
   * @param {*} upgradedConn A duplex connection returned from multiplexer and/or crypto selection
```

test/upgrading/upgrader.spec.js
jacobheun: I think it would be good to also add a test for closing. It's going to the MultiaddrConn so it should be fine, but I think having it for avoid regressions in the future would be good.

vasco-santos: I added to this test. I think it is better than create a new test

PR #518, opened 2019-12-13
feat: coalescing dial support

jacobheun:
* Parallel dials to the same peer / multiaddr will now settle at the same time with the same result
* Removed `connectToMultiaddr`, all dials now go directly through `connectToPeer`
* Fixed an issue with closed inbound connections not getting removed from tracking
* `PeerStore.put` now allows you to update a peer without firing the `peer` event. This helps prevent duplicate dialing when dialing a new multiaddr directly. (For example: `ipfs swarm connect [addr]`)

### DIFF
--- a/src/dialer/index.js
+++ b/src/dialer/index.js
@@ -4,6 +4,8 @@ const multiaddr = require('multiaddr')
 const errCode = require('err-code')
 const TimeoutController = require('timeout-abort-controller')
 const anySignal = require('any-signal')
+const PeerId = require('peer-id')
+const PeerInfo = require('peer-info')
 const debug = require('debug')
 const log = debug('libp2p:dialer')
 log.error = debug('libp2p:dialer:error')
@@ -38,7 +40,7 @@ class Dialer {
     this.timeout = timeout
     this.perPeerLimit = perPeerLimit
     this.tokens = [...new Array(concurrency)].map((_, index) => index)
-    this._pendingDials = new Set()
+    this._pendingDials = new Map()
   }
 
   /**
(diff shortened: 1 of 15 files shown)


src/dialer/index.js
```diff
@@ -56,72 +58,112 @@ class Dialer {
...
+   */
+  _createDialTarget (peer) {
+    const dialable = Dialer.getDialable(peer)
+    if (multiaddr.isMultiaddr(dialable)) {
+      return {
+        id: dialable.toString(),
+        addrs: [dialable]
+      }
+    }
+    // Check the peerstore first, and then fallback to the instance
```
vasco-santos: What do you mean by fallback here?

jacobheun: I can remove that I think it's just residual from when I was moving stuff around and trying some things out. It will now just do a PeerStore update and get the full list of existing and new addrs.

src/peer-store/index.js
```diff
@@ -36,11 +36,15 @@ class PeerStore extends EventEmitter {
...
+   * If already exist, its info is updated. If `silent` is set to
+   * true, no 'peer' event will be emitted. This can be useful if you
+   * are already in the process of dialing the peer. The peer is technically
+   * known, but may not have been added to the PeerStore yet.
    * @param {PeerInfo} peerInfo
+   * @param {boolean} [silent] (Default=false)
    * @return {PeerInfo}
    */
-  put (peerInfo) {
+  put (peerInfo, silent = false) {
```
vasco-santos: what do you think on transforming silent in an object? I think it helps to understand in the `peerStore.put(peerInfo, true)` what the `true` is

jacobheun: Sounds good, I refactored and pushed a change to make it take an options object.

PR #519, opened 2019-12-16
Adding HOP_RELAY_REFUSED code.

Jorropo:
That doesn't add circuit filter but at least this follow libp2p/go-libp2p-circuit#86 code.
Its not critical to merge this because if you got this code the relay attempt is over anyway.


jacobheun:
Closing this as its not really necessary, and not part of the spec.

Issue #512, opened 2019-12-11
[Documentation] Community sample code list?


mcclure:
I ported the "libp2p-in-the-browser" example from the js-libp2p repo to use Preact. I intend to release this Preact version as sample code for others.

I asked if in your IRC channel if it would be helpful for me to submit this modified sample code "upstream" anywhere, and @vasco-santos said:

> I think it would be nice if we have a list of examples/projects created by the community, which would reference those projects so that other people could collaborate or learn from them

I think having a list of community-supported js-libp2p examples, or a libp2p/js-libp2p-examples repo people could submit PRs to, would be neat.

jacobheun:
> libp2p/js-libp2p-examples repo people could submit PRs to, would be neat

I really like this idea. We do have [js-libp2p-examples](https://github.com/libp2p/js-libp2p-examples) that currently just has an in depth chat example. I think transitioning that into a community driven examples repo is a great idea.

vasco-santos:
I slept on this and I drafted a proposal for this. The goal is to create a `js-libp2p-community` repo with a similar concept to the `*-awesome` repos.

## js-libp2p-community

A catalog of libp2p awesomeness from the community to the community

## Structure

This catalog is divided into 4 main categories: examples, protocols, presentations and articles

### Examples

Examples section should contain code examples using `js-libp2p` for all types of purposes. This list can include references to other github repositories and codesandbox projects. If a github repository is referenced for an "in browser" example, it is suggested that the example also link to a codesandbox project.


### Protocols

List of protocols implemented on top of libp2p.

- `js-ipfs-bitswap`

### Presentations

List of relevant presentations

### Articles

List of relevant articles

---

If you feel this is a good idea, I can create the repo and move this proposal to a `README.md` so that we can iterate it. 

With this, we should rename `js-libp2p-examples` into `js-libp2p-workshop-ipfs-camp`, or similar and add to this.

daviddias:
I like the general way this is headed, a few thoughts:

- What about reusing https://github.com/ipfs/awesome-ipfs as awesome-libp2p?
- "we should rename js-libp2p-examples into js-libp2p-workshop-ipfs-camp" -> 👍🏽
- Caveat: You will destroy the onboarding experience if you throw new users into an ocean of examples. Keep a few (3~5) powerful tutorials of things built with libp2p close to the implementation, those will expose new users to the ways of doing things with libp2p and encourage them to experiment.

vasco-santos:
[...]

PR #520, opened 2019-12-16
docs: review getting started

jacobheun:
This PR is to supplement review of #514

### DIFF
--- a/doc/GETTING_STARTED.md
+++ b/doc/GETTING_STARTED.md
@@ -1,8 +1,8 @@
-# GETTING STARTED
+# Getting Started
 
 Welcome to libp2p, this guide will walk you through setting up a fully functional libp2p node 🚀
 
-- [GETTING STARTED](#getting-started)
+- [Getting Started](#getting-started)
   - [Install](#install)
   - [Configuring libp2p](#configuring-libp2p)
     - [Basic setup](#basic-setup)
--- a/doc/GETTING_STARTED.md
+++ b/doc/GETTING_STARTED.md
@@ -27,6 +27,8 @@ npm install libp2p
 
 ## Configuring libp2p
 
+If you're new to libp2p, we recommend configuring your node in stages, as this can make troubleshooting configuration issues much easier. In this guide, we'll do just that. If you're more experienced with libp2p, you may wish to jump to the [Configuration readme](./CONFIGURATION.md).
+
 ### Basic setup
 
 Now that we have libp2p installed, let's configure the absolute minimum needed to get your node running. The only modules libp2p requires are a [**Transport**][transport] and [**Crypto**][crypto] module. Let's start by setting up a Transport.
@@ -43,7 +45,7 @@ Start by installing `libp2p-websockets`:
 npm install libp2p-websockets
 ```
 
-Now that we have the module installed, let's configure libp2p to use the Transport. `Libp2p.create` takes a single configuration object as it's only parameter. We can add the Transport by passing it into the `modules.transport` array:
+Now that we have the module installed, let's configure libp2p to use the Transport. We'll use the [`Libp2p.create`](./API.md#create) method, which takes a single configuration object as it's only parameter. We can add the Transport by passing it into the `modules.transport` array:
 
 ```js
 const Libp2p = require('libp2p')
--- a/doc/GETTING_STARTED.md
+++ b/doc/GETTING_STARTED.md
@@ -45,7 +45,7 @@ Start by installing `libp2p-websockets`:
 npm install libp2p-websockets
 ```
 
-Now that we have the module installed, let's configure libp2p to use the Transport. We'll use the [`Libp2p.create`](./API.md#create) method, which takes a single configuration object as it's only parameter. We can add the Transport by passing it into the `modules.transport` array:
+Now that we have the module installed, let's configure libp2p to use the Transport. We'll use the [`Libp2p.create`](./API.md#create) method, which takes a single configuration object as its only parameter. We can add the Transport by passing it into the `modules.transport` array:
 
 ```js
 const Libp2p = require('libp2p')
@@ -70,7 +70,7 @@ There are a growing number of Crypto modules being developed for libp2p. As thos
 npm install libp2p-secio
 ```
 
-With `libp2p-secio` installed, we can add it to our existing configuration by important it and adding it to the `modules.connEncryption` array:
+With `libp2p-secio` installed, we can add it to our existing configuration by importing it and adding it to the `modules.connEncryption` array:
 
 ```js
 const Libp2p = require('libp2p')
--- a/doc/GETTING_STARTED.md
+++ b/doc/GETTING_STARTED.md
@@ -1,25 +1,21 @@
 # GETTING STARTED
 
-Welcome to libp2p, let's get you setup your first and fully functional libp2p node 🚀
+Welcome to libp2p, this guide will walk you through setting up a fully functional libp2p node 🚀
 
-- [Getting Started](#getting-started)
+- [GETTING STARTED](#getting-started)
   - [Install](#install)
-  - [Usage](#usage)
+  - [Configuring libp2p](#configuring-libp2p)
     - [Basic setup](#basic-setup)
-      - [Transport](#transport)
+      - [Transports](#transports)
       - [Connection Encryption](#connection-encryption)
-      - [Overview](#overview)
+      - [Running Libp2p](#running-libp2p)
     - [Custom setup](#custom-setup)
       - [Multiplexing](#multiplexing)
       - [Peer Discovery](#peer-discovery)
       - [Peer Routing](#peer-routing)
       - [Content Routing](#content-routing)
-      - [Pubsub](#pubsub)
+    - [Pubsub](#pubsub)
   - [What is next](#what-is-next)
-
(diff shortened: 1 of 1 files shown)


doc/GETTING_STARTED.md
```diff
@@ -29,87 +25,83 @@ The first step is to install libp2p in your project:
 npm install libp2p
 ```
 
-## Usage
-
-Configuring libp2p should not be a single step, but a continuous task. This includes configuring libp2p to achieve all the project requirements, as well as possible optimizations. Regardless of how you configure libp2p, the top level API will always remain the same.
```
vasco-santos: I think that we should have at least part of this. Explicitly say that configuring `js-libp2p` should be an iterative process rather than a let's configure everything before running

doc/GETTING_STARTED.md
```diff
@@ -29,87 +25,83 @@ The first step is to install libp2p in your project:
...
-
-For creating a `js-libp2p` node you should use [Libp2p.create](./API.md#create). As stated in the API docs, we need to provide the [libp2p configuration](./CONFIGURATION.md) as a parameter. If you did not read the configuration document yet, it is worth you read it before continuing.
```
vasco-santos: I added this because I think we should guide people to use the API and CONFIGURATION docs, because as soon as they are experienced with `js-libp2p`, they should go through those docs instead

doc/GETTING_STARTED.md
jacobheun: To me this read like, "The libp2p setup is more than a one time thing, you're never done configuring libp2p". It sounds daunting. 

It sounds like you meant something more like:

```
If you're new to libp2p, we recommend configuring your node in stages, as this can make
troubleshooting configuration issues much easier. In this guide, we'll do just that. 
If you're more experienced with libp2p, you may wish to jump to the [Configuration readme](#).
```

doc/GETTING_STARTED.md
jacobheun: Since the main readme is the landing page for most users I think it makes sense to have that distinction there. My concern with recommending people go read configuration first is that can be an overwhelming amount of information. We should help users get started quickly, and then send them to docs that will build on what they've learned here, and help them customize their nodes.

doc/GETTING_STARTED.md
vasco-santos: Yes, I think that is a better way of putting this! Can you change to that in this PR?

doc/GETTING_STARTED.md
vasco-santos: I agree that it can be overwhelming to recommend people to read configuration first.
My idea here is to point them to subsections of the other documents, instead, of all document. So I agree with you on the configuration part, but I think we should keep the pointer do `./API.md#create`

doc/GETTING_STARTED.md
jacobheun: Added that in.

doc/GETTING_STARTED.md
jacobheun: I added the link back to the first mention of `create`.

terichadbourne:
Just some quick feedback after skimming through  #514 at @vasco-santos's request. My main questions is where you expect people to be coming from when they get here and what they need to have already read in order for this Getting Started guide to make sense. 

Personally, having very little libp2p knowledge, I appreciated the little paragraphs at the start of each section but wished there were more of that context. I'm sure there are others who'd want less of it and more direct jumps into the lines of code needed to make things happen. 

I went to the main readme of this repo to see if it had the context I was looking for and it didn't. It did point to a ton of different websites for reference, with this disclaimer: 
`We are in the process of writing better documentation, blog posts, tutorials and a formal specification. Today you can find:`
which felt like "here's a list of our subpar docs; good luck!" (totally underselling what you do have available). I would have preferred to see that list prefaced with a "Here are our docs" and a note next to each one about what type of info I could info at that site, followed up with a "We're always improving our docs and we'd love your help" call to action. 

https://docs.libp2p.io/ turned out to have just the kind of explanations I was looking for, and looks very professional to boot, so it might make sense to link from certain sections here in the js-libp2p getting started guide to the concept explainers over there for those who'd like more background.

vasco-santos:
Thanks for your time @terichadbourne 

> Just some quick feedback after skimming through #514 at @vasco-santos's request. My main question is where you expect people to be coming from when they get here and what they need to have already read in order for this Getting Started guide to make sense. (I recommend adding links to prereqs if they exist.)

I expect them to come from: [libp2p/js-libp2p/refactor/async-await#usage](https://github.com/libp2p/js-libp2p/tree/refactor/async-await#usage).  Maybe before tutorials and examples. I think that we would not have prereqs here, people would have to know JS, as well as some concepts of networking, but when they come to a lib like this, I think that it is already implicit.

> Personally, having very little libp2p knowledge, I appreciated the little paragraphs at the start of each section but wished there were more of that context. I'm sure there are others who'd want less of it and more direct jumps into the lines of code needed to make things happen.

@jacobheun what do you think on adding toggles with "more details" for people to click and the section introductions would be available? I agree with teri that this introduction might be important to some people, but I also think if might be noisy for people that prefer to directly look into code.

> I went to the main readme of this repo to see if it had the context I was looking for and it didn't. It did point to a ton of different websites for reference, with this disclaimer:
We are in the process of writing better documentation, blog posts, tutorials and a formal specification. Today you can find:
which felt like "here's a list of our subpar docs; good luck!" (totally underselling what you do have available). I would have preferred to see that list prefaced with a "Here are our docs" and a note next to each one about what type of info I could info at that site, followed up with a "We're always improving our docs and we'd love your help" call to action.

My idea is to simply remove that and reference my proposal on [libp2p/js-libp2p#512#issuecomment-564929701](https://github.com/libp2p/js-libp2p/issues/512#issuecomment-564929701)

> https://docs.libp2p.io/ turned out to have just the kind of explanations I was looking for, and looks very professional to boot, so it might make sense to link from certain sections here in the js-libp2p getting started guide to the concept explainers over there for those who'd like more background.

In this document, we reference parts of [CONFIGURATION.md](https://github.com/libp2p/js-libp2p/blob/refactor/async-await/doc/CONFIGURATION.md) which have references to the docs as you suggested. I know that people need to navigate from Getting started to Configuration to docs website, but I prefer that approach rather than having those references to the docs website in both, as it would introduce a lot of noise in the documentation. Other option is to move those reference links to the Getting started, but I think they are more useful next to the Configuration part of things. Let me know your thoughts

jacobheun:
> @jacobheun what do you think on adding toggles with "more details" for people to click and the section introductions would be available? I agree with teri that this introduction might be important to some people, but I also think if might be noisy for people that prefer to directly look into code.

I think we can do lead ins to each section with a brief introduction to what we're building and why we needed. I don't think this needs to be more than a few sentences. I think we should follow those with "read more" links to docs.libp2p.io. That would enable us to focus on keeping those docs up to date, and avoid duplicating that information here if we don't need to.

I think the goal should be for users have just enough information to get them started and 
 for them to understand what they're doing, with clear calls to more information for people who want to learn more about that topic, or specific js usage.

For example, when I start adding the code for Transports, I should understand that they are the components of libp2p that do they actual communicating between nodes, and if I was so inclined, should have easily been able to read up more on transports at https://docs.libp2p.io/concepts/transport/. I think having each section end with a "call to action" for users to see more about configuring that component in the Configuration readme, and the API where it makes sense, will go a long way in creating a nice starting point for new users, and existing users.

doc/GETTING_STARTED.md
```diff
@@ -29,87 +25,85 @@ The first step is to install libp2p in your project:
...
 
-Taking into account a context where we are in a node.js environment, but have a requirement to just use `websockets`, we should go with `libp2p-websockets`. Start by installing that libp2p module:
+Start by installing `libp2p-websockets`:
 
 ```sh
 npm install libp2p-websockets
 ```
 
-and finally add it into the `modules.transport`, as an array:
+Now that we have the module installed, let's configure libp2p to use the Transport. We'll use the [`Libp2p.create`](./API.md#create) method, which takes a single configuration object as it's only parameter. We can add the Transport by passing it into the `modules.transport` array:
```
vasco-santos:
```suggestion
Now that we have the module installed, let's configure libp2p to use the Transport. We'll use the [`Libp2p.create`](./API.md#create) method, which takes a single configuration object as its only parameter. We can add the Transport by passing it into the `modules.transport` array:
```

doc/GETTING_STARTED.md
```diff
@@ -29,87 +25,85 @@ The first step is to install libp2p in your project:
...
+Encryption is an important part of communicating on the libp2p network. Every connection must be encrypted to help ensure security for everyone. As such, Connection Encryption (Crypto) is a required component of libp2p.
 
-Looking at the [available connection encryption](./CONFIGURATION.md#connection-encryption) protocols, you can start by using `libp2p-secio` and revisit this choice later on, as you get to know more about libp2p, as well as new other protocols for securing connections appear. You can install `libp2p-secio` and add it to your libp2p node as follows:
+There are a growing number of Crypto modules being developed for libp2p. As those are released they will be tracked in the [available Connection Encryption](./CONFIGURATION.md#connection-encryption) section of the configuration readme. For now, we are going to configure our node to use the `libp2p-secio` module.
 
 ```sh
 npm install libp2p-secio
 ```
 
+With `libp2p-secio` installed, we can add it to our existing configuration by important it and adding it to the `modules.connEncryption` array:
```
vasco-santos:
```suggestion
With `libp2p-secio` installed, we can add it to our existing configuration by importing it and adding it to the `modules.connEncryption` array:
```

jacobheun:
@vasco-santos sounds good, added your changes, fill free to merge if it looks good.

PR #522, opened 2019-12-16
feat: add libp2p.connections getter

jacobheun:
* Adds `libp2p.connections` getter to allow users to more easily get existing connections.
* Fixes `libp2p.hangUp(peer)` to take arguments that the API doc specifies.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -297,7 +297,7 @@ class Libp2p extends EventEmitter {
   /**
    * Disconnects all connections to the given `peer`
    *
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -288,12 +297,13 @@ class Libp2p extends EventEmitter {
   /**
    * Disconnects all connections to the given `peer`
    *
-   * @param {PeerId} peer The PeerId to close connections to
+   * @param {PeerInfo|PeerId|multiaddr|string} peer The the to close connections to
```
```suggestion
   * @param {PeerInfo|PeerId|multiaddr|string} peer the to close connections to
```

PR #523, opened 2019-12-19
refactor: examples/pnet

vasco-santos:
This PR refactors the `pnet` example to the refactored async `js-libp2p`. Moreover, it now does not use ipfs and relies only on `libp2p`


examples/pnet/index.js
```diff
@@ -0,0 +1,48 @@
...
+generate(swarmKey)
+
+// This key is for testing a different key not working
+const otherSwarmKey = Buffer.alloc(95)
+generate(otherSwarmKey)
+
+;(async () => {
+  const node1 = await privateLibp2pNode(swarmKey)
+  const node2 = await privateLibp2pNode(swarmKey)
+  // const node2 = await privateLibp2pNode(otherSwarmKey)
```
vasco-santos: @jacobheun I am having an issue using a invalid key. 

On the transport listener, for example `libp2p-tcp` https://github.com/libp2p/js-libp2p-tcp/blob/master/src/listener.js#L23 the `upgradeInbound` throws and the promise rejection is not handled. What do you recommend to do here? A `try...catch` in the encrypt and just log the error?

vasco-santos: btw, the error comes from:

https://github.com/libp2p/js-libp2p/blob/refactor/async-await/src/upgrader.js#L328
https://github.com/multiformats/js-multistream-select/blob/master/src/handle.js#L14
https://github.com/multiformats/js-multistream-select/blob/master/src/multistream.js#L37

jacobheun: @vasco-santos the upgradeInbound should probably still throw. The Transport needs to know that there has been a failure so that it doesn't do any emitting. I think the best thing to do is to make sure that the transport listeners are catching that and logging errors. We should add a test for this to the transport interface and roll out patches to each of the transports.

vasco-santos: yes, I agree that should be the transports to handle this. Will PR everything

examples/pnet-ipfs/index.js
```diff
@@ -1,145 +0,0 @@
...
-
-// This key is for the `TASK` mentioned in the writeFileSync calls below
-const otherSwarmKey = Buffer.alloc(95)
-writeKey(otherSwarmKey)
-
-// Add the swarm key to both repos
-const swarmKey1Path = path.resolve(repo1, 'swarm.key')
-const swarmKey2Path = path.resolve(repo2, 'swarm.key')
-fs.writeFileSync(swarmKey1Path, swarmKey)
-// TASK: switch the commented out line below so we're using a different key, to see the nodes fail to connect
```
jacobheun: The README mentions the `TASK` comments in the code for trying out the different swarm keys. How this is being done has changed, can we add a `TASK` comment for replacing `const node2 = await privateLibp2pNode(swarmKey)` with `const node2 = await privateLibp2pNode(otherSwarmKey)`. The code is there, there's just no comment saying to do this.

vasco-santos: good catch

Issue #509, opened 2019-12-11
Shrink the spread of modules


daviddias:
This is a continuation of the work initiated by https://github.com/ipfs/js-ipfs/issues/2222

Outstanding proposals
- [x] Merge js-libp2p-peer-book into js-libp2p (done on a branch) https://github.com/libp2p/js-libp2p/issues/453
- [ ] Merge js-libp2p-peer-id and js-libp2p-peer-info into a new module - js-libp2p-peer
- [x] Merge https://github.com/libp2p/js-libp2p-keychain into js-libp2p
- [ ] Deprecate https://github.com/libp2p/js-libp2p-nat-mngr and/or merge into js-libp2p
- [x] ~~Deprecate https://github.com/libp2p/js-libp2p-utils (not used anywhere)~~ seems that it is actually used in unreleased branches

vasco-santos:
👍 

Just want to point out that [libp2p/js-libp2p-utils](https://github.com/libp2p/js-libp2p-utils) is being used in some transports and the idea would be to continue collecting specific utils (unlikely to change) common to several repos and move them there. As a result, I think we should continue with this module

Regarding the other ones, I can take ownership of getting them done over the next weeks

daviddias:
Thanks for the clarification, @vasco-santos. I scratched it off the list

daviddias:
Hi folks! Thought of passing by and doing a check in, here are a few questions: 

<img width="942" alt="image" src="https://user-images.githubusercontent.com/1211152/73827335-ee7d2e00-4831-11ea-8b2e-52b07869a1ac.png">

daviddias:
Ah, there is one more. What about https://github.com/libp2p/js-libp2p-keychain ? Will it be merged to js-libp2p as well?

vasco-santos:
> `libp2p-peer`

We still not decided the exact way we are going to do this, but we will get rid of `peer-info` and keep `peer-id` as it is, or rename it if appropriate. If everything goes as planned, I should start working on `peerstore v2` during this quarter and one of the things we intend to do during that initiative is this.

> `libp2p-pubsub`

[...]

PR #388, opened 2019-07-29
refactor: add js-libp2p-switch to the libp2p codebase

jacobheun:
This PR adds js-libp2p-switch to js-libp2p as a part of https://github.com/libp2p/js-libp2p/issues/384.

Things to note:
- The entire libp2p-switch history has been added
- Additional commits were made to move switch under `./src/switch` and its tests under `./test/switch`.
- Some of the unneeded files were removed (like LICENSE) since libp2p already has a version of that file
- The readme and package.json from switch were kept, to simplify the transition. Ideally this should all get cleaned up and duplicate tests should be removed in a future PR

## **Note on Merging**: Follow steps mentioned at https://github.com/libp2p/js-libp2p/pull/388#issuecomment-516821568.

Issue #384, opened 2019-07-20
Merge js-libp2p-{core pieces} into js-libp2p repo


daviddias:
Following https://github.com/ipfs/js-ipfs/issues/2222, the proposal is to:

Merge into js-libp2p, the following modules:
- [x] [js-libp2p-connection-manager](https://github.com/libp2p/js-libp2p-connection-manager)
- [x] [js-libp2p-identify](https://github.com/libp2p/js-libp2p-identify)
~~- [ ] [js-libp2p-keychain](https://github.com/libp2p/js-libp2p-keychain)~~
~~- [ ] [js-libp2p-nat-mngr](https://github.com/libp2p/js-libp2p-nat-mngr)~~
- [x] [js-libp2p-pnet](https://github.com/libp2p/js-libp2p-pnet)
- [x] [js-libp2p-ping](https://github.com/libp2p/js-libp2p-ping)
- [x] [js-libp2p-switch](https://github.com/libp2p/js-libp2p-switch) https://github.com/libp2p/js-libp2p/pull/388
- [x] [js-libp2p-circuit](https://github.com/libp2p/js-libp2p-circuit)

### NPM Deprecation Notice
- [x] https://www.npmjs.com/package/libp2p-circuit
- [x] https://www.npmjs.com/package/libp2p-connection-manager
- [x] https://www.npmjs.com/package/libp2p-identify
- [x] https://www.npmjs.com/package/libp2p-pnet
- [x] https://www.npmjs.com/package/libp2p-ping
- [x] https://www.npmjs.com/package/libp2p-switch

### Readme Deprecation Notice
- [x] [js-libp2p-circuit](https://github.com/libp2p/js-libp2p-circuit)
- [x] [js-libp2p-connection-manager](https://github.com/libp2p/js-libp2p-connection-manager)
- [x] [js-libp2p-identify](https://github.com/libp2p/js-libp2p-identify)
- [x] [js-libp2p-pnet](https://github.com/libp2p/js-libp2p-pnet)
- [x] [js-libp2p-ping](https://github.com/libp2p/js-libp2p-ping)
- [x] [js-libp2p-switch](https://github.com/libp2p/js-libp2p-switch)

### Github Archival
Description replacement: `[DEPRECATED]: now part of the https://github.com/libp2p/js-libp2p repo`
- [x] [js-libp2p-circuit](https://github.com/libp2p/js-libp2p-circuit)
  - [x] Update description
[...]

PR #525, opened 2019-12-23
refactor: add js-libp2p-keychain to the libp2p codebase #388

vasco-santos:
This PR adds js-libp2p-switch to js-libp2p as a part of #384 and #509 .

Things to note:
- The entire libp2p-keychain history has been added
- Additional commits were made to move switch under `./src/keychain` and its tests under `./test/keychain`.
- Some of the unneeded files were removed (like LICENSE) since libp2p already has a version of that file
- The readme and package.json from switch were kept, to simplify the transition. Ideally this should all get cleaned up and duplicate tests should be removed in a future PR

## **Note on Merging**: 

Right now the plan (obtained from `libp2p-keychain` PR) is to:

- squash the merge to minimize the commit pollution
- update the readme to point to the original repo
- append the Co-Author list to the squashed commit to maintain author credit
  - This co author list was generated from the libp2p-keychain package.json which is auto generated from the commit history there.

```
Co-authored-by: Alan Shaw <alan.shaw@protocol.ai>
Co-authored-by: Alberto Elias <hi@albertoelias.me>
Co-authored-by: Alex Potsides <alex@achingbrain.net>
Co-authored-by: David Dias <daviddias.p@gmail.com>
Co-authored-by: Hugo Dias <mail@hugodias.me>
Co-authored-by: Jacob Heun <jacobheun@gmail.com>
Co-authored-by: Maciej Krüger <mkg20001@gmail.com>
Co-authored-by: Masahiro Saito <camelmasa@gmail.com>
Co-authored-by: Richard Schneider <makaretu@gmail.com>
Co-authored-by: Vasco Santos <vasco.santos@moxy.studio>
Co-authored-by: Vasco Santos <vasco.santos@ua.pt>
Co-authored-by: Victor Bjelkholm <victorbjelkholm@gmail.com>
```


vasco-santos:
@jacobheun `js-libp2p` users, such as `js-ipfs` rely on `js-libp2p-keychain` directly. I wonder if we should continue that path, and just modify the require, or if we should make this only usable through `js-libp2p` API. Let me know your thoughts

vasco-santos:
@jacobheun just rebased this. We should document how to use `libp2p-keychain`. What do you think to be the best way?

jacobheun:
> js-libp2p users, such as js-ipfs rely on js-libp2p-keychain directly. I wonder if we should continue that path, and just modify the require, or if we should make this only usable through js-libp2p API. Let me know your thoughts.

We should integrate this into the libp2p api, this would likely look like `libp2p.keychain`, and expose the necessary methods from there. Users shouldn't need to worry about importing this or using it directly it just complicates things. 

I would expect users to be able to provide a datastore to libp2p and it would just handle key storage. Users should probably also be able to specify a separate datastore for libp2p-keychain. (We should likely always want these to be "separate" datastores so we're not putting dht/peer/etc data in the same place as our keys).

> just rebased this. We should document how to use libp2p-keychain. What do you think to be the best way?

At the bare minimum it needs to be documented in the API and Configuration readme's, and we really need more tests around using this directly with libp2p.

* It would be good to test the persistence. (If `Libp2p.create()` is given a keystore, it should retain its previous identity instead of creating a new peer id)

PR #526, opened 2020-01-03
chore: interop tests job in ci

vasco-santos:
This PR adds a CI job to run the interop tests 🎉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,5 +39,12 @@ jobs:
       script:
         - npx aegir test -t browser -t webworker -- --browsers FirefoxHeadless
 
+    - stage: test
+      name: interop
+      script:
+        - cd node_modules/interop-libp2p
+        - npm install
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -40,5 +40,12 @@ jobs:
         - npx aegir test -t browser -- --browsers FirefoxHeadless
         - npx aegir test -t webworker -- --browsers FirefoxHeadless
 
+    - stage: test
+      name: interop
+      script:
+        - cd node_modules/interop-libp2p
+        - npm install
+        - LIBP2P_JS=./../../../../../src/index.js npx aegir test -t node --bail
```
jacobheun: I think we should be able to use the travis build directory to simplify this.
```suggestion
        - LIBP2P_JS=${TRAVIS_BUILD_DIR}/src/index.js npx aegir test -t node --bail
```

PR #543, opened 2020-01-29
docs: add migration guide from 0.26 to 0.27

jacobheun:
This adds a basic migration guide from 0.26.x to 0.27.

### DIFF
--- a/doc/migrations/v0.26-v0.27.md
+++ b/doc/migrations/v0.26-v0.27.md
@@ -66,9 +66,6 @@ libp2p.handle(['/echo/1.0.0'], ({ protocol, stream }) => pipe(stream, stream))
 const pull = require('pull-stream')
 libp2p.dialProtocol(peerInfo, '/echo/1.0.0', (err, conn) => {
   if (err) { throw err }
-
-  console.log('nodeA dialed to nodeB on protocol: /echo/1.0.0')
-
   pull(
     pull.values(['hey']),
     conn,
--- a/doc/migrations/v0.26-v0.27.md
+++ b/doc/migrations/v0.26-v0.27.md
@@ -1,0 +1,182 @@
+# Migrating to the new API
+
+A migration guide for refactoring your application code from libp2p v0.26.x to v0.27.0.
+
+## Table of Contents
+
+- [Migrating to the new API](#migrating-to-the-new-api)
+  - [Table of Contents](#table-of-contents)
+  - [Migrating from callbacks](#migrating-from-callbacks)
+  - [Pull Streams to Streaming Iterables](#pull-streams-to-streaming-iterables)
+  - [Sample API Migrations](#sample-api-migrations)
+    - [Registering Protocol Handlers](#registering-protocol-handlers)
+    - [Dialing and Sending Data](#dialing-and-sending-data)
+    - [Checking if a peer is connected](#checking-if-a-peer-is-connected)
+    - [Pinging another peer](#pinging-another-peer)
+    - [Pubsub](#pubsub)
+      - [Getting subscribers](#getting-subscribers)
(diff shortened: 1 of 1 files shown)


doc/migrations/v0.26-v0.27.md
```diff
@@ -0,0 +1,182 @@
+# Migrating to the new API
+
+A migration guide for refactoring your application code from libp2p v0.26.x to v0.27.0.
+
+## Table of Contents
+
+- [Migrating to the new API](#migrating-to-the-new-api)
+  - [Table of Contents](#table-of-contents)
```
alanshaw: I like your style, but do these two _really_ need to be included here?

doc/migrations/v0.26-v0.27.md
```diff
@@ -0,0 +1,182 @@
...
+const pull = require('pull-stream')
+libp2p.dialProtocol(peerInfo, '/echo/1.0.0', (err, conn) => {
+  if (err) { throw err }
+
+  console.log('nodeA dialed to nodeB on protocol: /echo/1.0.0')
+
+  pull(
+    pull.values(['hey']),
+    conn,
+    pull.collect((err, data) => {
```
alanshaw: `.collect` is not quite what's happening in the example below - do you want `.onEnd(op, done)` instead?

doc/migrations/v0.26-v0.27.md
```diff
@@ -0,0 +1,182 @@
...
+    for await (const data of source) {
+      console.log('received echo:', data.toString())
+    }
+  }
+)
+```
+
+### Checking if a peer is connected
+
+`peerInfo.isConnected` is unreliable. It requires an address be set via `PeerInfo.connect(addr)`, which requires the PeerInfo to be already tracking that address. This is problematic for incoming connections. The remoteAddr of an incoming connection is not reliable, and shouldn't be stored on the PeerInfo. Instead, we should simply check for an actual open connection. Since libp2p has better tracking of connections, this is trivial to do.
```
alanshaw: I would maybe rephrase this around `peerInfo.isConnected` being deprecated and that consumers should not use it because libp2p is now tracking this information centrally and it will not be correct.

doc/migrations/v0.26-v0.27.md
jacobheun: Not sure I follow, do you mean to run `.onEnd` preceded by `.map` or similar? Replacing `collect` with `onEnd` won't return the results of the echo. The issue with collect is that `data` is an array, so using a through stream for logging the response and a sink such as `.onEnd` to initiate the pull and catch the error would be more analogous.

doc/migrations/v0.26-v0.27.md
jacobheun: I'll remove, the TOC plugin I'm using auto adds them.

doc/migrations/v0.26-v0.27.md
jacobheun: nvm, I was looking at something else, you're right `onEnd` is the appropriate change.

doc/migrations/v0.26-v0.27.md
```diff
@@ -0,0 +1,181 @@
...
+
+`dialProtocol` no longer takes a callback, and will now return a [Streaming Iterable][streaming_iterable] and the protocol that was successfully negotiated. The new stream can be used with async iterator modules, see [it-awesome][it_awesome], instead of pull streams.
+
+**Before**
+```js
+const pull = require('pull-stream')
+libp2p.dialProtocol(peerInfo, '/echo/1.0.0', (err, conn) => {
+  if (err) { throw err }
+
+  console.log('nodeA dialed to nodeB on protocol: /echo/1.0.0')
```
alanshaw:
```suggestion
```

PR #544, opened 2020-01-29
chore: README build badges fixed

eordano:


jacobheun:
The table is created via the package-list.json file using the `package-table` module. We'll need to fix the problem there to avoid accidentally overwriting the table in the future. I created a PR there to correct the badges in the table https://github.com/ipfs-shipyard/package-table/pull/11.

jacobheun:
resolved by https://github.com/libp2p/js-libp2p/pull/565

PR #551, opened 2020-02-05
fix: ensure identify streams are closed

jacobheun:
- Identify streams were not being closed properly. They were either being only read from or written to, which would lave them half open. This adds tests to check for open streams after identify runs and fixes it for both identify and identify push.
- `connection.addStream()` wasn't being called properly in the Upgrader. This resulted in protocols being tracked as `undefined`. It was also tracking the multistream selected stream, instead of the MuxedStream. The call now matches the connection interface https://github.com/libp2p/js-interfaces/tree/master/src/connection#add-stream-metadata.

### DIFF
--- a/src/identify/index.js
+++ b/src/identify/index.js
@@ -4,7 +4,7 @@ const debug = require('debug')
 const pb = require('it-protocol-buffers')
 const lp = require('it-length-prefixed')
 const pipe = require('it-pipe')
-const { collect, take } = require('streaming-iterables')
+const { collect, take, consume } = require('streaming-iterables')
 
 const PeerInfo = require('peer-info')
 const PeerId = require('peer-id')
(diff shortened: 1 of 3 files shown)


src/identify/index.js
```diff
@@ -159,6 +160,8 @@ class IdentifyService {
       toBuffer,
       collect
     )
+    // close the stream, no need to wait
+    stream.sink([])
```
alanshaw: Pass `[]` to the pipe above like:

```js
    const [data] = await pipe(
      [],
      stream,
      lp.decode(),
      take(1),
      toBuffer,
      collect
    )
```

src/identify/index.js
```diff
@@ -261,6 +265,8 @@ class IdentifyService {
       toBuffer,
       collect
     )
+    // close the stream, but no need to wait
+    stream.sink([])
```
alanshaw: ...and here

PR #552, opened 2020-02-06
fix: protocols change event

alanshaw:
I was spelunking last night and found that pubsub was constantly having it's `onConnect` handler fired from the multicodec topology.

It was because bitswap was calling `dialProtocol` to connect to a peer. `dialProtocol` makes a `PeerInfo` out of the multiaddr it's given but it has no `protocols` i.e. `[]`. This is passed to `peerStore.update()` where we compare `[]` with an array of populated protocols. Obviously there is no intersection here so `change:protocols` was being emitted, even though no protocols were added or removed.

This logic needs to be improved and tested properly but I just wanted to send a PR to document my findings.

### DIFF
--- a/src/peer-store/index.js
+++ b/src/peer-store/index.js
@@ -132,19 +132,16 @@ class PeerStore extends EventEmitter {
         multiaddrs: recorded.multiaddrs.toArray()
       })
     }
-
-    // Update protocols
-    // TODO: better track added and removed protocols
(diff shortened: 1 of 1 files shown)


vasco-santos:
Thanks @alanshaw for the visibility ! With the latest updates for `0.28` with the AddressBook, this should not be a problem anymore, as we keep track of everything in isolation :)

I am closing this now

PR #565, opened 2020-02-17
chore: remove websocket star from package table

vasco-santos:
With the modules refactor, we do not support `websocket-star` anymore and it should be removed from the table


jacobheun:
related https://github.com/libp2p/js-libp2p/pull/544

@vasco-santos the package-list file needs to be updated as well. #544 includes an update to a lot of the tables. Can we make sure the table generation is working properly and update the json file?

vasco-santos:
Yeah, my bad! I always forget about the `package-table` 🤦‍♂ 

I fixed it now, but it created one issue: the script is gathering the `leadMaintainer` for `libp2p-utp` and `libp2p-rendezvous`, which supposedly are my responsibility. The thing is that they needed to be released so that the package.json published is updated. However, they are still missing implementation. Should we keep them for now as `N/A` or should we do that manually?

Regarding the build, that needs to be a PR to the `package-table` module. I can get a PR for it to fix the build links

jacobheun:
> However, they are still missing implementation. Should we keep them for now as N/A or should we do that manually?

We should probably just remove them for the time being, until they are active modules.

> Regarding the build, that needs to be a PR to the package-table module. I can get a PR for it to fix the build links

I already submitted a PR there and it should be in the latest package-table version.

just FYI, I am separately looking into the flakey CI test failures.

PR #585, opened 2020-03-10
chore: use libp2p-noise

### LINKED ISSUE #584: Add libp2p-noise to the connection encryption modules
- **Version**: N/A
- **Platform**: N/A
- **Subsystem**: N/A

#### Description:

Once `libp2p-noise` is fully operational, it needs to be properly documented in `js-libp2p`.

- [x] add it to `js-libp2p` packages table
- [x] add it to the `CONFIGURATION.MD` document
- [x] recommend its usage on the `GETTING_STARTED.MD` document
- [x] replace `libp2p-secio` on the tests
- [x] replace `libp2p-secio` on the `examples/encrypted-communications`, and other relevant examples

Needs:

- [x] final `libp2p-noise` release

vasco-santos:
This PR adds doc references for `libp2p-noise` and uses it in the tests/examples instead of `libp2p-secio`.

Needs:

- [ ] update examples
- [x] get CI green
- [x] final noise release

Closes #584


vasco-santos:
@jacobheun @mpetrunic FYI I went ahead and created this PR to also get `js-libp2p` tests running with `libp2p-noise` instead of `libp2p-secio` to speed up noise integration. I found out a few issues with this.

This [test](https://github.com/libp2p/js-libp2p/blob/master/test/upgrading/upgrader.spec.js#L61) is failing in `node`, as the `secureOutbound` returns undefined. Moreover, there are a few browser tests failing with an `handshake` error from `libp2p-noise`. We should guarantee that this is working before the final release.

vasco-santos:
Thanks for having a look at the issues @mpetrunic ❤️

vasco-santos:
FYI This was rebased with the `0.28.x` branch and the target branch of this PR updated for it.

In the context of updating the example, we also have #625 that fixes the examples. So, it will probably be better to do a follow up changing secio to noise in all the examples

PR #587, opened 2020-03-16
fix: remove node global

hugomrdias:
- adds buffer require
- adds globalThis where needed
- streaming-iterables was remove and new utils created, this will be consolidated in `ipfs-utils` and backported here to normalize all these iterator helper functions
- latency-monitor was copied inside the repo
- iso-random-stream is now used instead of node crypto
- the test `should ignore self on discovery` was moved to node only 

Size: 172.97KB
47.03KB below the 220KB limit.

`aegir build --node false` and `aegir test -t browser --node false` now work 🎉

related to this https://github.com/ipfs/js-ipfs/issues/2924


src/connection-manager/latency-monitor.js
```diff
@@ -0,0 +1,242 @@
+'use strict'
```
jacobheun, line 1: The latency monitor was pulled from the npm package right? Can we get attribution added, I'm not seeing it anywhere.

src/utils.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+  return Array.from(it)
+}
+
+module.exports = {
+  collect,
+  consume,
+  tap,
+  take
+}
```
jacobheun, line 78: I think this is a mistake. These functions are used in other downstream codebases. While we're getting a bundle reduction here, there's a high risk we're going to end up with duplicate code elsewhere as downstream packages pull in these or other sources causing those bundles to bloat unnecessarily.

src/pnet/key-generator.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const crypto = require('crypto')
+const randomBytes = require('iso-random-stream/src/random')
```
jacobheun, line 3: Why not just use `libp2p-crypto` random bytes? It uses the same thing under the hood and avoids us needing to worry about another dependency since we have to use `libp2p-crypto` anyway.

src/connection-manager/latency-monitor.js
hugomrdias: looks better now ?

jacobheun:
Released in v0.27.7

Issue #582, opened 2020-03-09
PeerStore improvements


vasco-santos:
# PeerStore improvements

Goal: Road to PeerStore v2 and improve browser connectivity

## Milestones

| Milestone | Issue | PR | State |
|----------|:-------------:|------:|------:|
| `0) Gather relevant metrics -- testground` |  TODO | TODO | TODO |
| `1) AddressBook and ProtoBook` |  N/A | [libp2p/js-libp2p#590](https://github.com/libp2p/js-libp2p/pull/590) | DONE |
| `2) Deprecate peerInfo` |  [libp2p/js-libp2p#589](https://github.com/libp2p/js-libp2p/issues/589) | In the tracking issue | DONE |
| `3) Backed Datastore` |  [libp2p/js-libp2p#591](https://github.com/libp2p/js-libp2p/issues/591) |[libp2p/js-libp2p#619](https://github.com/libp2p/js-libp2p/pull/619) | DONE |
| `4) keybook and libp2p-keychain` | [libp2p/js-libp2p#592](https://github.com/libp2p/js-libp2p/issues/592) | [libp2p/js-libp2p#634](https://github.com/libp2p/js-libp2p/pull/634) | DONE |
| `5) MetadataBook` | NA | [libp2p/js-libp2p#638](https://github.com/libp2p/js-libp2p/pull/638) | DONE |
| `6) Multiaddr confidence` |  TODO | TODO | TODO |
| `7) Peers Priority` |  TODO | TODO | TODO |

## Open minor issues

- Uncomment testing public key in keyBook after connection upgraded

## Problem

In `js-libp2p`, we are currently relying on `peer-info` to track the information of a peer, like its addresses and running protocols. We are moving different `PeerInfo` objects around, which in some cases are also not properly updated when the `identify-push`/ `identify-delta` are triggered, instead of having a single source of truth to handle this metadata, i.e. the PeerStore. After the release of `js-libp2p@0.27` (async refactor) we got some improvements on this, but it was just an initial step.

PeerStore provides us several opportunities to improve peers' connectivity and connection management. From peer scoring, to multiaddr confidence, there are several improvements that we can get from a store to manage peers metadata.

### Persistence

Centralizing all the information a peer has about its environment enables us to easily persist this data. This will have a big impact on browser nodes, which will be able to connect to the previous peers they were connected to.

With persistence, we will also be able to have Less load on the bootstrap nodes as we can immeditaly connect to other peers without them.

[...]

PR #590, opened 2020-03-24
feat: address and proto books

vasco-santos:
# AddressBook and ProtoBook

This is the initial milestone regarding `js-libp2p` PeerStore improvements described on [libp2p/js-libp2p#582](https://github.com/libp2p/js-libp2p/issues/582)

This PR restructures the initial version of `peerStore`, where we simply had a record of PeerInfos into data structures with different concerns and own logic.

We will start by introducing the `AddressBook` and `ProtoBook` since this was the content being used from the PeerInfos previously stored. In next PRs we will add other "books", namely the keybook and metadata book.

- [x] AddressBook implementation (initial version implementated)
- [x] ProtoBook implementation (initial version implementated)
- [x] PeerStore core
- [x] Change PeerStore usage to be compliant with the new implementation
- [x] Document PeerStore (initial version written)
- [x] Update docs/API.md
- [x] Remove `peer-info` from Ping (initial version implemented)

#### Change Peers data

- identifyService
  - pushToPeerStore - identify push (on libp2p.handle, libp2p.unhandle)
    - iterate peerStore peers, can iterate on ProtoBook instead?
  - After identify, **replace** peer's data
  - After handle push message, **replace** peer's data

- Upgrader: onConnection handler
  - Silent put to PeerStore
    - **add** to AddressBook
- Discovery
  - **add** Peer with multiaddr to peerStore
- DHT
  - Get/has operations
  - Put operations when receiving peers from other peers
    - DHT Message contains addresses to peers

#### Get Peers data

- Registrar
  - topology needs to listen for `protocol:change` event and to iterate on all the known peers to check if they run the protocol
  - topology iterates on all the peers. if they support the protocol, they add it to the topology

- dialProtocol
  - if received PeerInfo to dial (deprecate!)
- Dialer
  - Uses `peerStore.multiaddrsForPeer` to get the multiaddrs for a PeerInfo
  - PeerInfo widely used internally, need to remove it and use peerId + addressBook content
- getPeerInfo
  - Change with PeerInfo refactor?
- ping
  - receives peerInfo and libp2p, but can simply receive peerId
  
## Datastructure

The books datastructures where designed with consistency in mind among them and with their underlying data structure (Map). All should contain the main map methods `set`, `get`, `has` and `delete`, which should work in the same way as the map methods by default, but with some extra optional parameters to add additional super powers.

They were separated on their own instead of living on the top level `peerStore` for some reasons:
- separation of concerns and logic
- in the future some will need runtime pruning, such as multiaddr cleanning on ttls, scorings, ...

#### set(peerId, data, [options = { replace = true }]) : `Array<data>`

Set known data `{Array<multiaddr>, Array<string>, ...}` of a provided peer. By default, if previous data exists, it will replace it. If the `replace` options is explicitly defined to `false`, the data will be merged.

When the recorded data is inserted a new event should be triggered for the interested parties. If a record for the provided peer already exists and the new data is not relevant (i.e the final content will be the same) the operation just returns the final content, but no event is triggered for efficiency.

#### get(peerId) : `Array<data>`

Get known data of a provided peer. If the peer does not exist, it returns `undefined`.

#### has(peerId) : `boolean`

Do we have data for this peer?

#### delete(peerId, [data]) : `boolean`

Deletes the provided peer from the book. If data is provided, just remove the data 

Returns true if the peer's data was fond and the data was deleted (or the partial data did not already exist). If data is modified or removed, the proper event change is triggered to the interest parties.

**Concern**: what if the array of data gets empty? should we remove the peer? I don't see any good reason to keep it!

## General open questions

- Who handles the emit `peer`?
  - When we known an address, so it should be when added to the AddressBook!

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -17,16 +17,17 @@
   * [`contentRouting.put`](#contentroutingput)
   * [`contentRouting.get`](#contentroutingget)
   * [`contentRouting.getMany`](#contentroutinggetmany)
+  * [`peerStore.addressBook.add`](#peerstoreaddressbookadd)
   * [`peerStore.addressBook.delete`](#peerstoreaddressbookdelete)
   * [`peerStore.addressBook.get`](#peerstoreaddressbookget)
   * [`peerStore.addressBook.getMultiaddrsForPeer`](#peerstoreaddressbookgetmultiaddrsforpeer)
   * [`peerStore.addressBook.set`](#peerstoreaddressbookset)
+  * [`peerStore.protoBook.add`](#peerstoreprotobookadd)
   * [`peerStore.protoBook.delete`](#peerstoreprotobookdelete)
   * [`peerStore.protoBook.get`](#peerstoreprotobookget)
   * [`peerStore.protoBook.set`](#peerstoreprotobookset)
-  * [`peerStore.protoBook.supports`](#peerstoreprotobooksupports)
   * [`peerStore.delete`](#peerstoredelete)
-  * [`peerStore.find`](#peerstorefind)
+  * [`peerStore.get`](#peerstoreget)
   * [`peerStore.peers`](#peerstorepeers)
   * [`pubsub.getSubscribers`](#pubsubgetsubscribers)
   * [`pubsub.getTopics`](#pubsubgettopics)
@@ -55,13 +56,13 @@ Creates an instance of Libp2p.
 
 | Name | Type | Description |
 |------|------|-------------|
-| options | `Object` | libp2p options |
-| options.modules | `Array<Object>` | libp2p modules to use |
-| [options.config] | `Object` | libp2p modules configuration and core configuration |
-| [options.connectionManager] | `Object` | libp2p Connection Manager configuration |
-| [options.datastore] | `Object` | must implement [ipfs/interface-datastore](https://github.com/ipfs/interface-datastore) (in memory datastore will be used if not provided) |
-| [options.dialer] | `Object` | libp2p Dialer configuration
-| [options.metrics] | `Object` | libp2p Metrics configuration
+| options | `object` | libp2p options |
+| options.modules | `Array<object>` | libp2p modules to use |
+| [options.config] | `object` | libp2p modules configuration and core configuration |
+| [options.connectionManager] | `object` | libp2p Connection Manager configuration |
+| [options.datastore] | `object` | must implement [ipfs/interface-datastore](https://github.com/ipfs/interface-datastore) (in memory datastore will be used if not provided) |
+| [options.dialer] | `object` | libp2p Dialer configuration
+| [options.metrics] | `object` | libp2p Metrics configuration
 | [options.peerInfo] | [`PeerInfo`][peer-info] | peerInfo instance (it will be created if not provided) |
 
 For Libp2p configurations and modules details read the [Configuration Document](./CONFIGURATION.md).
@@ -192,7 +193,7 @@ for (const [peerId, connections] of libp2p.connections) {
 | Name | Type | Description |
 |------|------|-------------|
 | peer | [`PeerInfo`][peer-info]\|[`PeerId`][peer-id]\|[`Multiaddr`][multiaddr]\|`string` | The peer to dial. If a [`Multiaddr`][multiaddr] or its string is provided, it **must** include the peer id |
-| [options] | `Object` | dial options |
+| [options] | `object` | dial options |
 | [options.signal] | [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) | An `AbortSignal` instance obtained from an [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) that can be used to abort the connection before it completes |
 
 #### Returns
@@ -227,8 +228,8 @@ Dials to another peer in the network and selects a protocol to communicate with 
 | Name | Type | Description |
 |------|------|-------------|
 | peer | [`PeerInfo`][peer-info]\|[`PeerId`][peer-id]\|[`Multiaddr`][multiaddr]\|`string` | The peer to dial. If a [`Multiaddr`][multiaddr] or its string is provided, it **must** include the peer id |
-| protocols | `String|Array<String>` |  A list of protocols (or single protocol) to negotiate with. Protocols are attempted in order until a match is made. (e.g '/ipfs/bitswap/1.1.0') |
-| [options] | `Object` | dial options |
+| protocols | `string|Array<string>` |  A list of protocols (or single protocol) to negotiate with. Protocols are attempted in order until a match is made. (e.g '/ipfs/bitswap/1.1.0') |
+| [options] | `object` | dial options |
 | [options.signal] | [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) | An `AbortSignal` instance obtained from an [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) that can be used to abort the connection before it completes |
 
 #### Returns
@@ -286,7 +287,7 @@ In the event of a new handler for the same protocol being added, the first one i
 
 | Name | Type | Description |
 |------|------|-------------|
-| protocols | `Array<String>|String` | protocols to register |
+| protocols | `Array<string>|string` | protocols to register |
 | handler | `function({ connection:*, stream:*, protocol:string })` | handler to call |
 
 
@@ -311,7 +312,7 @@ Unregisters all handlers with the given protocols
 
 | Name | Type | Description |
 |------|------|-------------|
-| protocols | `Array<String>|String` | protocols to unregister |
+| protocols | `Array<string>|string` | protocols to unregister |
 
 #### Example
 
@@ -356,7 +357,7 @@ Iterates over all peer routers in series to find the given peer. If the DHT is e
 | Name | Type | Description |
 |------|------|-------------|
 | peerId | [`PeerId`][peer-id] | ID of the peer to find |
-| options | `Object` | operation options |
+| options | `object` | operation options |
 | options.timeout | `number` | maximum time the query should run |
 
 #### Returns
@@ -384,7 +385,7 @@ Once a content router succeeds, the iteration will stop. If the DHT is enabled, 
 | Name | Type | Description |
 |------|------|-------------|
 | cid | [`CID`][cid] | cid to find |
-| options | `Object` | operation options |
+| options | `object` | operation options |
 | options.timeout | `number` | maximum time the query should run |
 | options.maxNumProviders | `number` | maximum number of providers to find |
 
@@ -438,9 +439,9 @@ Writes a value to a key in the DHT.
 
 | Name | Type | Description |
 |------|------|-------------|
-| key | `String` | key to add to the dht |
+| key | `string` | key to add to the dht |
 | value | `Buffer` | value to add to the dht |
-| [options] | `Object` | put options |
+| [options] | `object` | put options |
 | [options.minPeers] | `number` | minimum number of peers required to successfully put (default: closestPeers.length) |
 
 #### Returns
@@ -469,8 +470,8 @@ Queries the DHT for a value stored for a given key.
 
 | Name | Type | Description |
 |------|------|-------------|
-| key | `String` | key to get from the dht |
-| [options] | `Object` | get options |
+| key | `string` | key to get from the dht |
+| [options] | `object` | get options |
 | [options.timeout] | `number` | maximum time the query should run |
 
 #### Returns
@@ -498,9 +499,9 @@ Queries the DHT for the n values stored for the given key (without sorting).
 
 | Name | Type | Description |
 |------|------|-------------|
-| key | `String` | key to get from the dht |
+| key | `string` | key to get from the dht |
 | nvals | `number` | number of values aimed |
-| [options] | `Object` | get options |
+| [options] | `object` | get options |
 | [options.timeout] | `number` | maximum time the query should run |
 
 #### Returns
@@ -518,6 +519,31 @@ const key = '/key'
 const { from, val } = await libp2p.contentRouting.get(key)
 ```
 
+### peerStore.addressBook.Add
+
+Adds known `multiaddrs` of a given peer. If the peer is not known, it will be set with the provided multiaddrs.
+
+`peerStore.addressBook.add(peerId, multiaddrs)`
+
+#### Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| peerId | [`PeerId`][peer-id] | peerId to set |
+| multiaddrs | |`Array<Multiaddr>` | [`Multiaddrs`][multiaddr] to add |
+
+#### Returns
+
+| Type | Description |
+|------|-------------|
+| `Map<string, MultiaddrInfo>` | Map of known peers' string identifier with their relevant information [`MultiaddrInfo`](multiaddr-info) |
+
+#### Example
+
+```js
+peerStore.addressBook.add(peerId, multiaddr)
+```
+
 ### peerStore.addressBook.delete
 
 Delete the provided peer from the book.
@@ -528,7 +554,7 @@ Delete the provided peer from the book.
 
 | Name | Type | Description |
 |------|------|-------------|
-| peerId | `peerid` | peerId to remove |
+| peerId | [`PeerId`][peer-id] | peerId to remove |
 
 #### Returns
 
@@ -541,14 +567,14 @@ Delete the provided peer from the book.
 ```js
 peerStore.addressBook.delete(peerId)
 // false
-peerStore.addressBook.set(peerId, discoveredMultiaddr)
+peerStore.addressBook.set(peerId, multiaddr)
 peerStore.addressBook.delete(peerId)
 // true
 ```
 
 ### peerStore.addressBook.get
 
-Get the known `multiaddrInfos` of a provided peer.
+Get the known [`MultiaddrInfos`](multiaddr-info) of a provided peer.
 
 `peerStore.addressBook.get(peerId)`
 
@@ -556,20 +582,20 @@ Get the known `multiaddrInfos` of a provided peer.
 
 | Name | Type | Description |
 |------|------|-------------|
-| peerId | `peerid` | peerId to get |
+| peerId | [`PeerId`][peer-id] | peerId to get |
 
 #### Returns
 
 | Type | Description |
 |------|-------------|
-| `Array<multiaddrInfo>` | Array of peer's multiaddr with their relevant information |
+| `Array<MultiaddrInfo>` | Array of peer's multiaddr with their relevant information [`MultiaddrInfo`](multiaddr-info) |
 
 #### Example
 
 ```js
 peerStore.addressBook.get(peerId)
 // undefined
-peerStore.addressBook.set(peerId, discoveredMultiaddr)
+peerStore.addressBook.set(peerId, multiaddr)
 peerStore.addressBook.get(peerId)
 // [
 // {
@@ -585,7 +611,7 @@ peerStore.addressBook.get(peerId)
 
 ## peerStore.addressBook.getMultiaddrsForPeer
 
-Get the known `multiaddr` of a provided peer. All returned multiaddrs will include the encapsulated `PeerId` of the peer.
+Get the known `Multiaddr` of a provided peer. All returned multiaddrs will include the encapsulated `PeerId` of the peer.
 
 `peerStore.addressBook.getMultiaddrsForPeer(peerId)`
 
@@ -593,20 +619,20 @@ Get the known `multiaddr` of a provided peer. All returned multiaddrs will inclu
 
 | Name | Type | Description |
 |------|------|-------------|
-| peerId | `peerid` | peerId to get |
+| peerId | [`PeerId`][peer-id] | peerId to get |
 
 #### Returns
 
 | Type | Description |
 |------|-------------|
-| `Array<multiaddr>` | Array of peer's multiaddr |
+| `Array<Multiaddr>` | Array of peer's multiaddr |
 
 #### Example
 
 ```js
 peerStore.addressBook.getMultiaddrsForPeer(peerId)
 // undefined
-peerStore.addressBook.set(peerId, discoveredMultiaddr)
+peerStore.addressBook.set(peerId, multiaddr)
 peerStore.addressBook.getMultiaddrsForPeer(peerId)
 // [
 // /ip4/140.10.2.1/tcp/8000/p2p/QmW8rAgaaA6sRydK1k6vonShQME47aDxaFidbtMevWs73t
@@ -618,37 +644,50 @@ peerStore.addressBook.getMultiaddrsForPeer(peerId)
 
 Set known `multiaddrs` of a given peer.
 
-`peerStore.addressBook.set(peerId, multiaddrs, options)`
+`peerStore.addressBook.set(peerId, multiaddrs)`
 
 #### Parameters
 
 | Name | Type | Description |
 |------|------|-------------|
-| peerId | `peerid` | peerId to set |
-| multiaddrs | `multiaddr|Array<multiaddr>` | multiaddrs to store |
-| [options] | `object` | options to set |
-| [options.replace] | `Object` | replace stored data (if exists) or unique union (default: true) |
+| peerId | [`PeerId`][peer-id] | peerId to set |
+| multiaddrs | |`Array<Multiaddr>` | [`Multiaddrs`][multiaddr] to store |
 
 #### Returns
 
 | Type | Description |
 |------|-------------|
-| `Array<multiaddrInfo>` | Array of peer's multiaddr with their relevant information |
+| `Map<string, MultiaddrInfo>` | Map of known peers' string identifier with their relevant information [`MultiaddrInfo`](multiaddr-info) |
 
 #### Example
 
 ```js
-peerStore.addressBook.set(peerId, discoveredMultiaddr)
-// [
-// {
-//   multiaddr: /ip4/140.10.2.1/tcp/8000,
-//   ...
-// },
-// {
-//   multiaddr: /ip4/140.10.2.1/ws/8001
-//   ...
-// },
-// ]
+peerStore.addressBook.add(peerId, multiaddr)
+```
+
+### peerStore.protoBook.add
+
+Add known `protocols` of a given peer.
+
+`peerStore.protoBook.add(peerId, protocols)`
+
+#### Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| peerId | [`PeerId`][peer-id] | peerId to set |
+| protocols | `Array<string>` | protocols to add |
+
+#### Returns
+
+| Type | Description |
+|------|-------------|
+| `Map<string, string>` | Map of known peers' string identifier with their supported protocols |
+
+#### Example
+
+```js
+peerStore.protoBook.add(peerId, protocols)
 ```
 
 ### peerStore.protoBook.delete
@@ -661,7 +700,7 @@ Delete the provided peer from the book.
 
 | Name | Type | Description |
 |------|------|-------------|
-| peerId | `peerid` | peerId to remove |
+| peerId | [`PeerId`][peer-id] | peerId to remove |
 
 #### Returns
 
@@ -674,7 +713,7 @@ Delete the provided peer from the book.
 ```js
 peerStore.protoBook.delete(peerId)
 // false
-peerStore.protoBook.set(peerId, supportedProtocols)
+peerStore.protoBook.set(peerId, protocols)
 peerStore.protoBook.delete(peerId)
 // true
 ```
@@ -689,7 +728,7 @@ Get the known `protocols` of a provided peer.
 
 | Name | Type | Description |
 |------|------|-------------|
-| peerId | `peerid` | peerId to get |
+| peerId | [`PeerId`][peer-id] | peerId to get |
 
 #### Returns
 
@@ -702,7 +741,7 @@ Get the known `protocols` of a provided peer.
 ```js
 peerStore.protoBook.get(peerId)
 // undefined
-peerStore.protoBook.set(peerId, supportedProtocols)
+peerStore.protoBook.set(peerId, [ '/proto/1.0.0', '/proto/1.1.0' ])
 peerStore.protoBook.get(peerId)
 // [ '/proto/1.0.0', '/proto/1.1.0' ]
 ```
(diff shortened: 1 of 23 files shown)


vasco-santos:
@jacobheun I made a first pass through the PeerStore improvements milestone 1.

This includes a brief explanation on the first post of the PR with PeerStore current usage and data structures decision. Moreover, it includes a README file with the PeerStore goals and a brief writing on its implementation and API. 

I would say that for now, it would be great to have your feedback on the questions/decisions written on the PR and on the README (via PR review comments). You can also look into the structure and APIs for the `AddressBook` and `ProtoBook` but while we do not discuss what I previously mentioned things might need to be changed or completed.

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
...
+  validity: ,
+  confidence: 
+}
+```
+
+**Note:** except for multiaddr namings, the other properties are placeholders for now and might not be as described in the future milestones.
+
+- `addressBook.set()`
+- `addressBook.get()`
+- `getMultiaddrsForPeer()`
```
jacobheun: Are `getMultiaddrsForPeer` and `addressBook.get` any different? They seem to do the same thing, I think we should get rid of the former.

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
...
+- `addressBook.has()`
+- `addressBook.delete()`
+- `addressBook.peers()`
```
jacobheun: Would `addressBook.peers` ever return different peers than `peerStore.peers`? If so, what value do we get from this difference? If not, should we remove this from the books and have users get it from PeerStore instead?

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
...
+
+It is important pointing out that the API methods which return arrays of data (`set`, `get`, `getMultiaddrsForPeer`) shuld return the `multiaddr` property of the `multiaddrInfo` and not the entire `multiaddrInfo` as the remaining data should be used internally. Should we consider having two datastructure instead?
```
jacobheun:
```suggestion
It is important pointing out that the API methods which return arrays of data (`set`, `get`, `getMultiaddrsForPeer`) should return the `multiaddr` property of the `multiaddrInfo` and not the entire `multiaddrInfo` as the remaining data should be used internally. Should we consider having two datastructure instead?
```

jacobheun: Why should they only return the multiaddr? Is there a reason to not expose address confidence?

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
...
+
+`Map<string, Set<string>>`
+
+A `peerId.toString()` identifier mapping to a `Set` of protocol identifier strings.
+
+- `protoBook.set()`
+- `protoBook.get()`
+- `protoBook.has()`
+- `protoBook.delete()`
+- `protoBook.supports()`
```
jacobheun: Right now it looks like `has` is answering "do I have protocols for this peer?", but that doesn't seem actionable to me. I don't know why I'd ever ask that question. `supports` makes sense, because I will want to check protocol support before doing certain actions.

These Books aren't Map interface compliant so I don't see a need to add methods unnecessarily if they don't make sense.

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
 # Peerstore
 
-WIP
\ No newline at end of file
+Libp2p's Peerstore is responsible for keeping an updated register with the relevant information of the known peers. It should gather environment changes and be able to take decisions and notice interested parties of relevant changes. The Peerstore comprises four main components: `addressBook`, `keyBook`, `protocolBook` and `metadataBook`. These book components have similar characteristics with the `Javascript Map` implementation.
+
+The PeerStore needs to manage the high level operations on its inner books, have a job runner to trigger other books runners for data trimming or computations. Moreover, the peerStore should be responsible for noticing interested parties of relevant events, through its Event Emitter.
```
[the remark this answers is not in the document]
vasco-santos: It is a future component for when we have `ttl` for multiaddrs, with multiaddr confidence, scores... I will move it to future considerations!

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
...
+
+## Peers Environment
```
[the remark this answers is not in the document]
vasco-santos: Yeah, still thinking on a good naming. 

The goal in this section is to describe the picture of our network environment. What happens behind the scene to have peers finding out data of other peers (protocols, new multiaddrs, ...) and how they can act with this data via the PeerStore.

The current naming is:

Peers Environment
- Sense
- Act

We can probably move to something like:

Network environment:
- Data Gathering
- Data Consumption

Do you have suggestions?

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
...
+
+#### Sense
```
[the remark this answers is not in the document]
vasco-santos: fwiw, we should not use Peer Discovery here, as we use the PeerDiscovery naming for finding new peers on the network (with their multiaddr), and not for sensing the network / peer operations to get further information.

Check my comment above. I agree that the name here is terrible, but we can get to a good one.

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
...
+
+In the background, the Identify Service is also waiting for new protocols to be started by the peer. If a new protocol is started, the `identify-push` message is sent to all the connected peers, so that their PeerStore can be updated with the new protocol and relevant parties are noticed. As the `identify-push` also sends complete and updated information, the data in the PeerStore is replaced.
+
+On different context, it is also possible to gather relevant information for the peers. For instance, in `dht` operations, nodes can exchanges data of peers they know as part of the `dht` operation. In this case, we can get information from a peer that we already know. As a consequence, the `PeerStore` should act accordingly and not replace the data it owns, but just try to merge it the discovered data is new. For example, discovered a new address of a peer.
+
+#### Act
```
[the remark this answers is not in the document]
vasco-santos: I agree with this. Regarding the section naming, we should decide according to the 2 above. Let's centralize the naming discussion on the main section naming above

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
...
+
+Taking into account a different scenario, a peer might perform/receive a dial request to/from a unkwown peer. In such a scenario, the `PeerStore` must store the peer's multiaddr once a connection is established. 
+
+(NOTE: this should be removed later)
+(currently we silently put it to the peerStore, without emitting events, as this logic exists in the `onConnected` callback from the upgrader. This way, we are never emitting the `peer` event when inbound connections happen, or a unkwown peer is dialed. Should we differentiate this?)
+
+After a connection is established with a peer, the Identify Service will act on this connection. A stream is created and peers exchange their information (listenMuldiaddrs and running protocols). Once this information is obtained, the PeerStore can collect the new data. In this specific case, we have a guarantee that this data is complete and updated, so the data stored in the PeerStore should be replaced (older and outdated data should disappear). However, if the recorded `multiaddrs` or `protocols` have changed, interested parties must be informed via `change:multiaddrs` or `change:protocols` events.
```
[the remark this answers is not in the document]
vasco-santos: Yes, I agree that we should consider the signed records spec + address origins (provenance). I will look at them deeply.

I thought that the identify should provide all the information (and updated), that's why I  think the replace should be the best option in this circumstance. This way, "If we've successfully dialed addresses of that peer we should probably avoid replacing them" if this multiaddr is not sent to us anymore, the peer should not be listening on it anymore...
Is there anything that I am not foreseeing?

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
...
+#### API:
+
+Access to its underlying books:
+
+- `peerStore.protoBook.*`
+- `peerStore.addressBook.*`
+
+High level operations:
+
+- `peerStore.set(peerId, data, options)` or `events`
```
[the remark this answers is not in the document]
vasco-santos: My main concern about this was when we have discovery protocols that might offer a peer more than known multiaddrs, but also known protocols. This would make necessary to change `libp2p` discovery events consumer to check what data it receives and act accordingly. However, I don't think this is a strong argument to have the `.set()` method.

Should we drop it then?

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
...
+
+One aspect that we need to consider is wether we should add information to every book, even if we don't have any relevant information for it. For instance, if we just discover a peer via a typical discovery service, we will have the `peerId` and an array of `multiaddr`. When we do `peerStore.set()`, should we also do `protoBook.set()` with an empty list of protocols? I don't see any advantage on adding to the remaining ones.
+
+**IMPORTANT:** This is one of the biggest design decisions to make (set vs events). The programmatic API is the easiest solution but it can provide users an API that they sould not need. If we go on an event-based approach, the `peerStore` should receive all the relevant subsystems (discovery, identifyService, ...) and sense the environment (via events) to gather the information that would need to be sent via the API. Thile the latest seems the best solution, it is the more complex one to implement, as we would ideally have an interface that those subsystems would need to implement and each time we have a new subsystem that needs to add data to the peerStore, we might need to update the `peer-store` codebase (or have a good set of abstractions).
```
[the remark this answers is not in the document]
vasco-santos: The thought behind consuming events was to not allow libp2p users to put data directly in the `PeerStore`, as only libp2p should be responsible for it afaik. This was also where the private properties were getting in, so that users would not be incentivised to change the books and eventually lead to non consistent data.
 But I totally agree that it seems over complex and unecessary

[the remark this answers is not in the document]
vasco-santos: drop this?

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
...
+```js
+{
+  address: true,
+  proto: true,
+  key: true,
+  metadata: true
+}
+```
+
+- `peerStore.delete(peerId, [data])`
```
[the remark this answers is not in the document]
vasco-santos: I wanted to be consistent with map, which uses `delete` to remove the peer and the data. I also was thinking about using `clear()` to remove everything from the map (if we need it).

Any other idea to circumvent this?

src/peer-store/README.md
```diff
@@ -1,3 +1,156 @@
...
+The `addressBook` keeps the known multiaddrs of a peer. The multiaddrs of each peer are not a constant and the Address book must have this into consideration.
+
+`Map<string, multiaddrInfo>`
+
+A `peerId.toString()` identifier mapping to a `multiaddrInfo` object, which should have the following structure:
+
+```js
+{
+  multiaddr: ,
+  validity: ,
```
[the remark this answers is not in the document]
vasco-santos: not sure yet! `go-libp2p` uses `ttl` for each multiaddr and I was naming validity for it.

Anyway. until we get to the confidence part of things, I just added this as a placeholder, as I am not sure if we will need something like this or not.

src/peer-store/README.md
vasco-santos: `getMultiaddrsForPeer()` adds `/p2p/{id}`. This method already exists in the current PeerStore and I kept it.

Any reason to not add that to the `get` and get rid of the former?

src/peer-store/README.md
vasco-santos: the `peerStore.peers` would return a vision of things through the properties provided, `addr`, `protocols`, while the `addressBook.peers` would get the ones from `addr`. If we get rid of the combined methods from peerStore, should we also just provided the `addressBook.peers`, `protoBook.peers`?

src/peer-store/README.md
vasco-santos: hummm, I was thinking on the opposite, is there any reason to expose the address confidence?

What do you think on keeping `getMultiaddrsForPeer` where we would just return the `multiaddr` (with `/p2p/id`) and the `get` would return the complete object?

src/peer-store/README.md
vasco-santos: `has(peerId)` would check if we have protocols for that peer, while `supports(peerId, protocol)` would say if it supports the peer. So, it is a supplement from has. Anyway, we can achieve the `has` with `get`, so we can drop it if you feel that there is no use case where this is useful.

src/peer-store/README.md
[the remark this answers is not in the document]
jacobheun: Yeah I'd drop it and stick with an api approach.

src/peer-store/README.md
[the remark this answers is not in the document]
jacobheun: I think we should simplify this for now and just have `peerStore.delete(peerId)`, which would delete the peer and all of its data from each of the books. If the need exists to delete specific data, this could be done with the books, until there's a good reason to add a convenience method. If we keep the api small and targeted we can always add more convenience methods in the future.

src/peer-store/README.md
jacobheun: `getMultiaddrsForPeer` is mainly there as a stop gap because we didn't have the address book. I would expect to just be able to do the `addressBook.get`. We shouldn't append `/p2p/{PeerId}` in the address book though. We can just assemble those when dialing, or announcing an address.

src/peer-store/README.md
jacobheun: I'm not sure I am following. It would be helpful to see the API params and return value since this isn't stubbed in the code yet. I'm not clear on what this is supposed to do and why I wouldn't just use the PeerStore peers accessor.

src/peer-store/README.md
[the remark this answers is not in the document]
jacobheun: It might be best to just drop the `Peers Environment` header altogether and hoist the subsections to header level 2. We can look at renaming them later. I think what you have listed is better than `sense`and `act` so we can go with that for now.

src/peer-store/README.md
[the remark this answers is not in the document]
jacobheun: I would drop it. The PeerStore could assemble any needed data before emitting if that's a concern.

src/peer-store/README.md
jacobheun: Yeah I think we should drop it. I expect users to either want to look at all the protocols for a peer `.get`, or (and the likely thing) check for specific support for a set of protocols `.supports`.

src/peer-store/README.md
vasco-santos: what would you expect to return from `peerStore.peers()`?

My initial idea was to do `peerStore.peers({ protocols, addresses })` and it would return all the peers we know with the "required" information. This would do internally `addressBook.peers()` + `protoBook.peers()` and assemble the data together. However, if we are dropping the other  `peerStore` API methods (like set, get, ...) and do atomic operations, I think we should also just support the underlying `peers()` method, which would return an array with all the map content.

src/peer-store/README.md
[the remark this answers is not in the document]
jacobheun: In theory, yes the peer should be advertising that address but there may be instances where that isn't happening. I think this will get better over time as we add support for announce/listen/noAnnounce addresses, but we don't support those at the moment so we shouldn't remove known good addresses unless they're actually bad. Keep in mind that when we deploy the updates, even if we have better support for that, older nodes won't have it yet so we risk removing good addresses. Addresses dialed via things like `ipfs swarm connect` come to mind as potential areas that we might end up removing good addresses for.

src/peer-store/README.md
jacobheun: If you wanted to manually prune addresses, confidence could be a way of going about this. This may be a small use case, but what I am mainly trying to get at is avoiding scenarios where we're "hiding" data. There shouldn't be a reason to not expose the data, however, the usability of `addressBook.get()` might make the most sense to just have it return multiaddrs for each of use and then a secondary method could eventually be added to get the other metadata.

> What do you think on keeping getMultiaddrsForPeer where we would just return the multiaddr (with /p2p/id) and the get would return the complete object?

I think this is a good work around for now, and avoids us breaking the current API the peerstore has.

src/peer-store/README.md
jacobheun: I expect `.peers` to give me a list of all the peers. What you're describing to me sounds like a `find` method which I do think is useful. Right now `peers` is just a Map property, it's an accessor, and I think it should stay an accessor. We should add find methods for searching.

vasco-santos:
@jacobheun this is now ready to review

It did not introduce any breaking change yet, as I left the older API of `peerStore` with adapters for the new `addressBook` and `protoBook`, since other modules (like DHT, ...) are still using them. Anyway, all the internal usage of the `peerStore`, including the tests now use the new API.

I will do a follow up PR with the breaking change for the peerStore API with the PRs of the modules using it. This will precede the breaking change PRs to remove `peer-info` usages from internal code and API

src/identify/index.js
```diff
@@ -241,6 +206,9 @@ class IdentifyService {
       protocols: Array.from(this._protocols.keys())
     })
 
+    // TODO: should we add to peerStore.addressBook.set() here?
+    // We can have an inbound connection from an unkwown peer
```
jacobheun, line 210: No, this function is us responding to a request, identify will run for every connection regardless of direction so we'll get their addresses when it is called.

src/index.js
```diff
@@ -289,7 +289,11 @@ class Libp2p extends EventEmitter {
     const dialable = Dialer.getDialable(peer)
     let connection
     if (PeerInfo.isPeerInfo(dialable)) {
```
jacobheun, line 291: This technically works because PeerInfo is duck typing it's check, but that's not the case everywhere which had me second guessing this.

The whole `getDialable` logic is giving me pause, especially with the PeerStore changes. I feel like we should just be giving an array of multiaddrs to the dialer to dial. This would tie in nicely with adding multiaddr confidence in the future, as the dialer could just be given a prioritized array.

I feel like the logic here should change a bit:
1. Ensure the peer/multiaddr `dialProtocol` was given is in the PeerStore.
2. Get the addresses from the AddressBook for this peer.
3. If the provided `peer` was an address, make it first in the multiaddr array that will be dialed.
4. Call dial with the multiaddr array

I think we should add 2 utilities to PeerStore until we fully deprecate PeerInfo.
1. A method to take a PeerInfo instance and store that data in the PeerStore
2. A method to aggregate from all the PeerStore books for a PeerId into a PeerInfo and return that
This could simplify a lot of the logic scattered around the repo that has to account for PeerInfo eventually going away.

src/peer-store/README.md
```diff
@@ -1,3 +1,120 @@
 # Peerstore
 
-WIP
\ No newline at end of file
+Libp2p's Peerstore is responsible for keeping an updated register with the relevant information of the known peers. It should gather environment changes, be able to take decisions and notice interested parties of relevant changes. The Peerstore comprises four main components: `addressBook`, `keyBook`, `protocolBook` and `metadataBook`. These book components have similar characteristics with the `Javascript Map` implementation.
+
+The PeerStore manages the high level operations on its inner books. Moreover, the peerStore should be responsible for notifying interested parties of relevant events, through its Event Emitter.
```
jacobheun, line 5: It's being referred to as `PeerStore` and `Peerstore`, we should be consistent.

src/peer-store/README.md
```diff
@@ -1,3 +1,120 @@
 # Peerstore
 
-WIP
\ No newline at end of file
+Libp2p's Peerstore is responsible for keeping an updated register with the relevant information of the known peers. It should gather environment changes, be able to take decisions and notice interested parties of relevant changes. The Peerstore comprises four main components: `addressBook`, `keyBook`, `protocolBook` and `metadataBook`. These book components have similar characteristics with the `Javascript Map` implementation.
```
jacobheun, line 3: I wouldn't mention the Map piece. It's not compliant with a Map and we can't iterate on them, so I would avoid comparing the two. 

> It should gather environment changes, be able to take decisions and notice interested parties of relevant changes.

Can you clarify this? It's sounds like the PeerStore tries to be smart ("decisions"), it should **not** be smart. It should be the single source of truth for all peer data.
- If a subsystem learns about peer data it needs to go directly to the peer store. 
- If any system/application/person wants to know about peers or listen for updates, they need to go to the peerstore.

src/peer-store/README.md
```diff
@@ -1,3 +1,120 @@
...
+
+The PeerStore manages the high level operations on its inner books. Moreover, the peerStore should be responsible for notifying interested parties of relevant events, through its Event Emitter.
+
+(Future considerations: Peerstore should manage a job runner to trigger books runners for data trimming or computations)
```
jacobheun, line 7: This doesn't provide enough information for future devs (including us). We should make a dedicated `Future considerations` header and be a bit more descriptive about what the PeerStore might wish to do and why. What kind of data trimmings? What computations? These can be concise, but should be clear. For example:

```
## Future Considerations
- If multiaddr TTLs are added, the PeerStore may schedule jobs to delete all addresses that exceed the TTL to prevent AddressBook bloating
```

doc/API.md
```diff
@@ -507,6 +518,345 @@ const key = '/key'
...
+peerStore.addressBook.delete(peerId)
+// false
+peerStore.addressBook.set(peerId, discoveredMultiaddr)
+peerStore.addressBook.delete(peerId)
+// true
+```
+
+### peerStore.addressBook.get
+
+Get the known `multiaddrInfos` of a provided peer.
```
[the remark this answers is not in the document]
vasco-santos: I agree that we should document it, but I am not sure where. Any suggestion?
Regarding capitalize, I agree for cases such as `PeerInfo`, which was a mistake followed by copy paste... In this case, I usually follow [this](https://jsdoc.app/tags-type.html), where this type should be `multiaddrInfo` and I use it like that in the code.  What do you think?

doc/API.md
```diff
@@ -507,6 +518,345 @@ const key = '/key'
...
+`peerStore.addressBook.set(peerId, multiaddrs, options)`
+
+#### Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| peerId | `peerid` | peerId to set |
+| multiaddrs | `multiaddr|Array<multiaddr>` | multiaddrs to store |
+| [options] | `object` | options to set |
+| [options.replace] | `Object` | replace stored data (if exists) or unique union (default: true) |
```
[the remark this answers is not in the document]
vasco-santos: Agreed, I will change this!

doc/API.md
```diff
@@ -507,6 +518,345 @@ const key = '/key'
...
+#### Example
+
+```js
+peerStore.protoBook.set(peerId, supportedProtocols)
+// [ '/proto/1.0.0', '/proto/1.1.0' ]
+```
+
+### peerStore.protoBook.supports
+
+Verify if the provided peer supports the given `protocols`.
```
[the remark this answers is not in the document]
vasco-santos: This is just a convenience method. Given its ambiguity, we should probably remove it

doc/API.md
```diff
@@ -507,6 +518,345 @@ const key = '/key'
...
+//   multiaddrInfos: [...],
+//   protocols: [...]
+// }
+```
+
+### peerStore.peers
+
+Get all the stored information of every peer.
+
+`peerStore.peers()`
```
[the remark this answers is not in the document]
vasco-santos: Oh damn, this is a breaking change!

It is not a Map, but a getter now... I can make it without a breaking change by adding some adapter code that still populates this map, but I am not sure about being a valuable effort

doc/API.md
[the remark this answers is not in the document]
jacobheun: There is a Types section at the bottom of this doc, it should go there.

The JSDoc docs are inconsistent with their usage and don't have specific recommendations on casing, we should be consistent in libp2p. `@type {PropertiesHash}` and `@type {MyClass}` are also referenced on that page. In the codebase we create typedefs of 4 other objects, all of which are capitalized.

doc/API.md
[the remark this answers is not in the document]
jacobheun: It's fine if it's a getter, it will work the same, we just need to correct the docs.

doc/API.md
```diff
@@ -507,6 +517,311 @@ const key = '/key'
...
+peerStore.addressBook.set(peerId2, discoveredMultiaddrs)
+peerStore.delete(peerId2)
+// true
+```
+
+### peerStore.find
+
+Find the stored information of a given peer.
+
+`peerStore.find(peerId)`
```
jacobheun, line 775: Should we just call this `peerStore.get(peerId)`? I'm worried users might misinterpret this as a network search for the peer.

vasco-santos: Yes, I agree! The question is, should I do it now, or in the next PR? (this will break the return value)

src/index.js
vasco-santos: > The whole getDialable logic is giving me pause, especially with the PeerStore changes. I feel like we should just be giving an array of multiaddrs to the dialer to dial. This would tie in nicely with adding multiaddr confidence in the future, as the dialer could just be given a prioritized array.

I also want to reformulate the dialer in the `getDialable` logic. I was postponing this change to go together with `peer-info`, as deprecating `peer-info` will create the need for that change.

> I feel like the logic here should c hange a bit:
> 
> Ensure the peer/multiaddr dialProtocol was given is in the PeerStore.
> Get the addresses from the AddressBook for this peer.
> If the provided peer was an address, make it first in the multiaddr array that will be dialed.
> Call dial with the multiaddr array

Yes, I was thinking on something like this! 🚀 

> I think we should add 2 utilities to PeerStore until we fully deprecate PeerInfo.
>
> A method to take a PeerInfo instance and store that data in the PeerStore

I currently have `peerStore.put()` which does that. It is an adapter for the breaking change.

> A method to aggregate from all the PeerStore books for a PeerId into a PeerInfo and return that
> This could simplify a lot of the logic scattered around the repo that has to account for PeerInfo eventually going away.

`peerStore.peers` getter currently does that, until we do the breaking change.

vasco-santos: So, should we do this together with `peer-info` deprecation?

The idea is to get it all together in the same final release, so I think it would make the development easier

jacobheun: Yeah, if the plan is to release everything in 0.28, we can add this change to that PR.

src/peer-store/address-book.js
```diff
@@ -0,0 +1,220 @@
...
+   * @returns {Array<multiaddrInfo>}
+   */
+  _add (peerId, multiaddrInfos) {
+    const id = peerId.toString()
+    const rec = this.data.get(id)
+
+    // Add recorded uniquely to the new array (Union)
+    rec && rec.forEach((mi) => {
+      if (!multiaddrInfos.find(r => r.multiaddr.equals(mi.multiaddr))) {
+        multiaddrInfos.push(mi)
```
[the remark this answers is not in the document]
vasco-santos: I am not sure how this was before I moving code around to get `set` and `add`. However, now the `multiaddrInfos` is created by `this._toMultiaddrInfos()` function and it is "pushed" with data from the current record, and then set as the new data and not returned.

So, the only possible scenario of second effects here that I can see is if someone did get before and then changes the data after the `add`. So, I would say that for being safe we only need to destruct the record `m`?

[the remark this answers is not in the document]
vasco-santos: Other scenario is if someone changes the multiaddr provided! So yes, it might be better to create a new array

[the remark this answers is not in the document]
vasco-santos: but if we just create a new array, I think we will not solve any of these scenarios

src/peer-store/book.js
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Get the known data of a provided peer.
+   * @param {PeerId} peerId
+   * @returns {Array<Data>}
+   */
+  get (peerId) {
+    if (!PeerId.isPeerId(peerId)) {
+      throw errcode(new Error('peerId must be an instance of peer-id'), ERR_INVALID_PARAMETERS)
+    }
```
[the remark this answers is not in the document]
vasco-santos: I personally don't like to see `cannot access `toString()` of undefined as an error, as someone needs to go into the code/docs and see what is the issue. This way, I think the error is useful here. if we do not try to access properties of it, I would agree.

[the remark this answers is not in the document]
jacobheun: We can table this discussion for outside of this PR as we're already doing similar checks elsewhere. In short, I'd rather see us move to support static type checking for users that want it instead of doing runtime checks. While they're not big, doing these type checks everywhere slowly adds to bundle size as well.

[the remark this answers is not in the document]
vasco-santos: Should we open a general issue to discuss this?

src/peer-store/README.md
```diff
@@ -1,3 +1,122 @@
...
+
+While it is currently not supported in js-libp2p, future iterations may also support the [IdentifyDelta protocol](https://github.com/libp2p/specs/pull/176).
+
+It is also possible to gather relevant information for peers from other protocols / subsystems. For instance, in `DHT` operations, nodes can exchange peer data as part of the `DHT` operation. In this case, we can learn additional information about a peer we already know. In this scenario the PeerStore should not replace the existing data it has, just add it.
+
+## Data Consumption
+
+When the PeerStore data is updated, this information might be important for different parties.
+
+Every time a peer needs to dial another peer, it is essential that it knows the multiaddrs used by the peer, in order to perform a successful dial to it. The same is true for pinging a peer. While the `AddressBook` is going to keep its data updated, it will also emit `change:multiaddrs` events so that subsystems/users interested in knowing these changes can be notifyied instead of pooling the `AddressBook`.
```
jacobheun, line 29:
```suggestion
Every time a peer needs to dial another peer, it is essential that it knows the multiaddrs used by the peer, in order to perform a successful dial to it. The same is true for pinging a peer. While the `AddressBook` is going to keep its data updated, it will also emit `change:multiaddrs` events so that subsystems/users interested in knowing these changes can be notified instead of polling the `AddressBook`.
```

src/peer-store/README.md
```diff
@@ -1,3 +1,122 @@
...
+
+Everytime a peer starts/stops supporting a protocol, libp2p subsystems or users might need to act accordingly. `js-libp2p` registrar orchestrates known peers, established connections and protocol topologies. This way, once a protocol is supported for a peer, the topology of that protocol should be informed that a new peer may be used and the subsystem can decide if it should open a new stream with that peer or not. For these situations, the `ProtoBook` will emit `change:protocols` events whenever supported protocols of a peer change.
+
+## PeerStore implementation
+
+The PeerStore wraps four main components: `addressBook`, `keyBook`, `protocolBook` and `metadataBook`. Moreover, it provides a high level API for those components, as well as data events.
+
+### API
```
jacobheun, line 37: We shouldn't duplicate this, it's already in the `API.md`, just link there.

src/peer-store/README.md
```diff
@@ -1,3 +1,122 @@
-# Peerstore
+# PeerStore
 
-WIP
\ No newline at end of file
+Libp2p's PeerStore is responsible for keeping an updated register with the relevant information of the known peers. It should be the single source of truth for all peer data, where a subsystem can learn about peers' data and where someone can listen for updates. The PeerStore comprises four main components: `addressBook`, `keyBook`, `protocolBook` and `metadataBook`.
+
+The PeerStore manages the high level operations on its inner books. Moreover, the PeerStore should be responsible for notifying interested parties of relevant events, through its Event Emitter.
+
+## Data gathering
```
jacobheun, line 7: I think we can do it in a followup PR but I think we should simplify the data gathering and consumption sections. Reading through this it was hard to figure out where data comes from and who/what/why it's being consumed. I think it would be good to be very direct about what should actually be putting data into the peer store so it's clear, and how someone can get data out of the peer store.

IE:

## Submitting records to the PeerStore

### Identify
- The Identify protocol automatically runs on every connection when multiplexing is enabled. The protocol will put the multiaddrs and protocols provided by the peer to the PeerStore. As these records are directly from the peer, they should be considered the source of truth and weighted accordingly.

...

## Retrieving records from the PeerStore

### Protocols
- *how to get existing protocols*
- *how to listen for protocol changes*

...

src/peer-store/address-book.js
```diff
@@ -0,0 +1,214 @@
...
+     */
+    this.data = new Map()
+  }
+
+  /**
+   * Set known addresses of a provided peer.
+   * @override
+   * @param {PeerId} peerId
+   * @param {Array<Multiaddr>} addresses
+   * @returns {Map<string, Array<MultiaddrInfo>>}
```
jacobheun, line 53: As mentioned in the api doc, this should return the AddressBook itself.

src/peer-store/address-book.js
```diff
@@ -0,0 +1,214 @@
...
+    return this.data
+  }
+
+  /**
+   * Add known addresses of a provided peer.
+   * If the peer is not known, it is set with the given addresses.
+   * @override
+   * @param {PeerId} peerId
+   * @param {Array<Multiaddr>} addresses
+   * @returns {Map<string, Array<MultiaddrInfo>>}
```
jacobheun, line 110: Return the AddressBook

src/peer-store/proto-book.js
```diff
@@ -0,0 +1,137 @@
...
+    this.data = new Map()
+  }
+
+  /**
+   * Set known protocols of a provided peer.
+   * If the peer was not known before, it will be added.
+   * @override
+   * @param {PeerId} peerId
+   * @param {Array<string>} protocols
+   * @returns {Map<string, Set<string>>}
```
jacobheun, line 47: Return the ProtoBook

src/peer-store/proto-book.js
```diff
@@ -0,0 +1,137 @@
...
+    return this.data
+  }
+
+  /**
+   * Adds known protocols of a provided peer.
+   * If the peer was not known before, it will be added.
+   * @override
+   * @param {PeerId} peerId
+   * @param {Array<string>} protocols
+   * @returns {Map<string, Set<string>>}
```
jacobheun, line 95: Return the ProtoBook

src/registrar.js
```diff
@@ -74,9 +74,11 @@ class Registrar {
    * @returns {void}
    */
   onConnect (peerInfo, conn) {
-    if (!PeerInfo.isPeerInfo(peerInfo)) {
-      throw errcode(new Error('peerInfo must be an instance of peer-info'), ERR_INVALID_PARAMETERS)
-    }
+    // TODO: This is not a `peer-info` instance anymore, but an object with the data.
```
jacobheun, line 77: Perhaps to avoid confusion, when we get rid of PeerInfo we should start calling variables `peerData`?

src/peer-store/README.md
vasco-santos: Yeah, I totally agree to use that text format. I will PR with that on a followup

src/registrar.js
vasco-santos: Yes, I was having a hard time with this naming confusion. Good suggestion! 👍

PR #595, opened 2020-03-30
Make export explicitly `default` so that rollup can consume libp2p

svdo:
I'm trying to use js-libp2p in a (browser library) project that is using Rollup.js for bundling. When just using JavaScript, I get a warning from Rollup:

```
(!) Import of non-existent export
index.js
1: import Libp2p from 'libp2p'
```

When using TypeScript, this becomes an error:

```
[!] Error: 'default' is not exported by node_modules/src/index.js, imported by src/Main.ts
https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module
src/Main.ts (4:7)
4: import Libp2p from 'libp2p'
          ^
Error: 'default' is not exported by node_modules/src/index.js, imported by src/Main.ts
```

This error can be fixed by the simple change in this pull request. Thanks!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -504,7 +504,7 @@ class Libp2p extends EventEmitter {
   }
 }
 
-module.exports = Libp2p
+module.exports.default = Libp2p
 /**
  * Like `new Libp2p(options)` except it will create a `PeerInfo`
  * instance if one is not provided in options.
--- a/src/index.js
+++ b/src/index.js
@@ -504,6 +504,7 @@ class Libp2p extends EventEmitter {
   }
 }
 
+module.exports = Libp2p
 module.exports.default = Libp2p
 /**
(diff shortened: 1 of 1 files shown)


svdo:
By the way, I'd prefer this change to be covered by a test somehow, but I'm not sure what the best approach would be. Do you feel this is necessary? If yes: any suggestions on how to do this?

jacobheun:
I believe the issue here is the multiple export statements, we should just extend `Libp2p` with `create` and export only Libp2p. Can you check if this commit, https://github.com/libp2p/js-libp2p/commit/012beec8e84ec05e12e4265d947b6ccea3a437a6, resolves your issue?

svdo:
Yes that also solves my compilation errors, and I like it better, so go for it :) I'll close this PR.

PR #597, opened 2020-03-31
fix: pass libp2p to discovery services

jacobheun:
This adds support for passing libp2p to discovery services when creating them which makes it easier for discovery services to extend the behavior of other libp2p systems. This is value for discovery systems that may wish to use systems such as pubsub.

**Additional fix**
There was also an issue where a peers relay addresses were not being properly returned. This was due to the fact that `PeerStore.multiaddrsForPeer` was not checking the PeerId contained in the address. When we are listening on a relay we will have the PeerId of that relay in the address, so we should check to verify that the peerid in the multiaddr is not ours before returning.

### DIFF
--- a/test/peer-store/peer-store.spec.js
+++ b/test/peer-store/peer-store.spec.js
@@ -161,19 +161,22 @@ describe('peer-store', () => {
     expect(peerStore.peers.size).to.equal(0)
   })
 
-  it('should be able to remove a peer from store through its b58str id', async () => {
-    const [peerInfo] = await peerUtils.createPeerInfo()
+  it('should be able to get the multiaddrs for a peer', async () => {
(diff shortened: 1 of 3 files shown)


test/peer-store/peer-store.spec.js
```diff
@@ -161,19 +161,22 @@ describe('peer-store', () => {
     expect(peerStore.peers.size).to.equal(0)
   })
 
-  it('should be able to remove a peer from store through its b58str id', async () => {
-    const [peerInfo] = await peerUtils.createPeerInfo()
+  it('should be able to get the multiaddrs for a peer', async () => {
```
jacobheun: This test was improperly named and was just a copy+paste from the previous test name.

PR #598, opened 2020-04-01
chore: deprecate old peer store api

vasco-santos:
This PR is a follow up of #590 on the context of the PeerStore improvements described on #582 

It removes the older peerStore API and adds a temporary record of peer id's. This is needed because the dht needs the peerStore to keep track of peers' public key. Once we have the key book, it will be removed.

Needs:

- [x] [libp2p/js-libp2p#590](https://github.com/libp2p/js-libp2p/pull/590) (needs rebase when this merged)
- [x] [libp2p/js-libp2p-kad-dht#179](https://github.com/libp2p/js-libp2p-kad-dht/pull/179)

Unblocks:

- [ ] [libp2p/js-libp2p-daemon#38](https://github.com/libp2p/js-libp2p-daemon/pull/38)
- [ ] [libp2p/interop#36](https://github.com/libp2p/interop/pull/36) (also uses this here to get the tests green)

BREAKING CHANGE: the peer-store api changed. Check the API docs for the new specification.


vasco-santos:
This should be ready to review @jacobheun !

FYI. the interop tests are failing here, but I can run them locally with any problems. Not exactly sure yet about the issue, but I think you can start reviewing this, as well as the dht PR.

They depend on from one another, so I would say that we should probably make a beta release in the dht

src/peer-store/address-book.js
```diff
@@ -183,6 +185,12 @@ class AddressBook extends Book {
     return multiaddrInfos
   }
 
+  _setPeerId (peerId) {
+    if (!this._ps.peerIds.get(peerId)) {
+      this._ps.peerIds.set(peerId.toString(), peerId)
```
jacobheun, line 190: Use B58 string for consistency.
```suggestion
      this._ps.peerIds.set(peerId.toB58String(), peerId)
```

src/peer-store/address-book.js
```diff
@@ -183,6 +185,12 @@ class AddressBook extends Book {
     return multiaddrInfos
   }
 
+  _setPeerId (peerId) {
```
jacobheun, line 188: Add this to the base Book class?

src/peer-store/index.js
```diff
@@ -195,15 +111,16 @@ class PeerStore extends EventEmitter {
...
    * @param {PeerId} peerId
    * @returns {peerInfo}
    */
-  find (peerId) {
+  get (peerId) {
     if (!PeerId.isPeerId(peerId)) {
       throw errcode(new Error('peerId must be an instance of peer-id'), ERR_INVALID_PARAMETERS)
     }
 
+    const id = this.peerIds.get(peerId.toString())
```
jacobheun, line 123: B58. We need to switch everything together, otherwise we'll likely end up with a bug where we have the same peer stored with the wrong string id.
```suggestion
    const id = this.peerIds.get(peerId.toB58String())
```

src/peer-store/proto-book.js
```diff
@@ -132,6 +134,12 @@ class ProtoBook extends Book {
 
     return this
   }
+
+  _setPeerId (peerId) {
```
jacobheun, line 138: Same as above, add to Book class.

test/peer-store/proto-book.spec.js
```diff
@@ -96,7 +95,7 @@ describe('protoBook', () => {
       const supportedProtocols = ['protocol1', 'protocol2']
 
       let changeCounter = 0
-      ee.on('change:protocols', () => {
+      peerStore.on('change:protocols', () => {
```
jacobheun, line 98: I think we should add the events to the API docs at https://github.com/libp2p/js-libp2p/blob/0.28.x/doc/API.md#events. We could split that into subsections (`libp2p`, `libp2p.peerStore`) to link to from the TOC. WDYT?

vasco-santos: Yes, that makes sense.  Good call

vasco-santos:
Addressed you review, the only tests failing are related to the delegated routing issue that is being fixed.

I also made a pre release for the dht, we just need to figure out how we should merge this, with the interop dependency that needs the daemon updated.

jacobheun:
> I also made a pre release for the dht, we just need to figure out how we should merge this, with the interop dependency that needs the daemon updated.

We could leave it as a branch for now. The other option could be to cherry pick out the daemon specification updates in https://github.com/libp2p/interop/pull/32. That would enable us to do a pre release of the daemon and just install it and use it in the interop tests here on CI.

PR #604, opened 2020-04-09
fix: make circuit relay listening addresses more forgiving

jacobheun:
* Makes the delegate tests less restrictive, we don't need to mock the query params, validation should be done in the modules themselves. This should make libp2p tests more resilient to non breaking, additive changes.

This makes setting a relay address a bit more lenient. You can now use:
```
/p2p-circuit/ip4/0.0.0.0/tcp/9090/p2p/QmRelay
// or
/ip4/0.0.0.0/tcp/9090/p2p/QmRelay/p2p-circuit
```
You could previously only do the former. Both will result in the same announced address: `/ip4/0.0.0.0/tcp/9090/p2p/QmRelay/p2p-circuit/p2p/QmOurId`

### DIFF
--- a/src/circuit/listener.js
+++ b/src/circuit/listener.js
@@ -22,7 +22,7 @@ module.exports = (circuit) => {
    * @return {void}
    */
   listener.listen = async (addr) => {
-    const [addrString] = String(addr).split('/p2p-circuit').slice(-1)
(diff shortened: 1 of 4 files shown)


jacobheun:
windows test failures are due to an unrelated bug with the ipfs http client that's being fixed

PR #609, opened 2020-04-15
chore: update delegate routing deps to fix branches

jacobheun:
This updates the delegated routing dependencies for failing tests due to an issue with the latest ipfs http client

Depends on:
- [x] https://github.com/libp2p/js-libp2p-delegated-peer-routing/pull/30
- [x] https://github.com/libp2p/js-libp2p-delegated-content-routing/pull/37


jacobheun:
cc @achingbrain All the delegate routing tests are now passing with the latest ipfs-utils and http-client RC.

Issue #589, opened 2020-03-19
Deprecate peer-info and consolidate peers' information on PeerStore


vasco-santos:
As part of the [PeerStore improvements epic](https://github.com/libp2p/js-libp2p/issues/582), we intend to deprecate `peer-info` and consolidate all peer's information on the PeerStore.

## Background

On `libp2p@0.27` we introduced the PeerStore on the `js-libp2p` codebase replacing the `peer-book` module. The PeerStore is responsible for managing known peers, as well as their addresses and metadata.

The initial version of this PeerStore is similar to the previous `peer-book`. It maps peer identifier strings to their `peer-info` object and has the common operations of a store, such as `put`, `get`, `has`, `remove` and `replace`. 

## Problem

Taking into account that peers' information might vary as the time passes and network operations are performed, we introduced two events in the initial version of the PeerStore: `change:protocols` and `change:multiaddrs`, but some other should be added in the future.

Since we currently rely on a PeerInfo instance stored in the PeerStore and a PeerInfo instance, we are prone to some data inconsistencies. For example, if a peer discovery protocol discovers a peer, it will create a PeerInfo with that peer and its discovered multiaddr and add it to the PeerStore. In the case that we already know the peer (other multiaddr or not, and possibly some protocols), we currently need to destruct the old information and merge it to the new one. This also creates complexity on finding the changes in `multiaddr` / `protocols` / ...

Therefore, `peer-info` adds an extra layer of complexity without benefits, both in the `PeerStore` store and on moving data around `libp2p` subsystems. In addition, it is better to move around only the needed data, instead of creating unnecessary instances.

## PeerInfo usage

| Module | Details | PR |
|----------|:-------------:|------:|
| `libp2p` |  `.create()` |[libp2p/js-libp2p#610](https://github.com/libp2p/js-libp2p/pull/610) |
| `ipfs` | used to access id, or to access data from `PeerStore` | TODO |
| `libp2p-kad-dht` | `peer-discovery` / `dht.findPeer`/`dht.findProvs` return PeerInfo | [libp2p/js-libp2p-kad-dht#180](https://github.com/libp2p/js-libp2p-kad-dht/pull/180) /[libp2p/js-libp2p-kad-dht#181](https://github.com/libp2p/js-libp2p-kad-dht/pull/181) |
[...]

Issue #202, opened 2018-06-05
Add support for Addrs {listen, announce, noAnnounce}


daviddias:
With https://github.com/libp2p/js-libp2p/pull/166/files, the libp2p constructor will be fully refactored and with it comes the libp2p config.

We want to give the user the ability to(with example): 

```
addresses: {                  // Multiaddrs
  listen: []                  // To start transport listener
  announce: []                // To announce/share with the network (i.e DNS addr)
  noAnnounce: []              // To keep private
},
```

Proposed in: https://github.com/libp2p/js-libp2p/pull/166#discussion_r193074930

JGAntunes:
Hello! :wave:

Just found this through - https://github.com/ipfs/js-ipfs/issues/1865 - and thought maybe I could help out here?

What's the plan for these addresses vs the ones coming for the `PeerInfo` instance?

jacobheun:
The go-ipfs docs are a helpful reference here, https://github.com/ipfs/go-ipfs/blob/v0.4.18/docs/config.md#addresses, I think js-ipfs will eventually need to be updated to support `Announce` and `NoAnnounce` once libp2p has support for this.

Here are my thoughts for each piece (with untested pseudo code):
### listen
These addresses will be passed to the transports for listening on startup. Any existing `multiaddrs` in `PeerInfo` will also be listened to.
```js
// Pre startup:
const listenAddrs = [...options.addresses.listen, ...peerInfo.multiaddrs.toArray()]
```

### announce
These addresses would be added to `PeerInfo` once listening is completed. They are an amendment to the `multiaddrs` already in `PeerInfo` and as such, would not replace any existing addresses.

**Note**: On listening, a transport may replace multiaddrs in peerInfo with improved addresses. For example tcp may replace `/ip4/0.0.0.0/tcp/0` with something like `/ip4/0.0.0.0/tcp/8989`.

```js
// Post listening
options.addresses.announce.forEach(addr => peerInfo.multiaddrs.add(addr))
```

### noAnnounce
This is a blacklist of addresses that should be removed from `PeerInfo`, as any address in `PeerInfo` will be advertised to peers.

**Note**: noAnnounce should take into account encapsulated addresses. For example, if I add `/ip4/127.0.0.1` to noAnnounce, I expect addresses like `/ip4/127.0.0.1/tcp/8080/ipfs/Qm` to be filtered out.

```js
// Post announce additions
peerInfo.multiaddrs.forEach(addr => {
[...]

PR #610, opened 2020-04-15
chore: remove peer info usage

vasco-santos:
In the context of deprecating `peer-info` as described on [libp2p/js-libp2p#589](https://github.com/libp2p/js-libp2p/issues/589), this PR removes the `peer-info` usage on the API, internal code usage and tests. Moreover, all the modules were updated for the new versions also without using `peer-info` and the docs were also updated!

It is important point out that with these modifications:

- we could simplify our logic inherent to the Dialer, as discussed in #590.
- we added supported for specifying `{ addresses: { listen } }` in the context of #202 and #612 adds a AddressManager component with support for announce and noAnnounce, as well as docs for usage.

BREAKING CHANGE: all API methods with peer-info parameters or return values were changed. You can check the API.md document, in order to check the new values to use


doc/API.md
```diff
@@ -163,6 +164,10 @@ const libp2p = await Libp2p.create(options)
 await libp2p.stop()
 ```
 
+### addresses
+
+TODO with `address-manager`.
```
vasco-santos, line 169: This will come in a follow up PR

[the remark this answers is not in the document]
vasco-santos: #612

src/get-peer-id.js
```diff
@@ -0,0 +1,41 @@
...
+    } catch (err) {
+      throw errCode(
+        new Error(`${peer} is not a valid peer type`),
+        codes.ERR_INVALID_MULTIADDR
+      )
+    }
+  }
+
+  if (addr && peerStore) {
+    peerStore.addressBook.add(peer, [addr])
```
jacobheun, line 35: Since we're redoing this function it might be a good time to remove the update from it as it's a bit error prone (doing a get that includes a write). Perhaps this should be `getPeer` and we return `{ id, multiaddrs }`? Then the calling function can choose whether or not it wants to store the address.

src/identify/index.js
```diff
@@ -47,17 +47,20 @@ class IdentifyService {
    * @param {object} options
    * @param {Registrar} options.registrar
    * @param {Map<string, handler>} options.protocols A reference to the protocols we support
-   * @param {PeerInfo} options.peerInfo The peer running the identify service
+   * @param {PeerId} options.peerId The peer running the identify service
+   * @param {{ listen: Array<Multiaddr>}} options.addresses The peer aaddresses
```
jacobheun, line 51:
```suggestion
   * @param {{ listen: Array<Multiaddr>}} options.addresses The peer addresses
```

src/identify/index.js
```diff
@@ -76,7 +79,7 @@ class IdentifyService {
 
         await pipe(
           [{
-            listenAddrs: this.peerInfo.multiaddrs.toArray().map((ma) => ma.buffer),
+            listenAddrs: this.addresses.listen.map((ma) => ma.buffer),
```
jacobheun, line 82: We should make a note to update this when the Address Manager is in place. We really shouldn't be using `addresses.listen` here, but should instead be using `addresses.announce`

src/index.js
```diff
@@ -56,29 +59,31 @@ class Libp2p extends EventEmitter {
...
       metrics: this.metrics,
       onConnection: (connection) => {
-        const peerInfo = new PeerInfo(connection.remotePeer)
-        this.registrar.onConnect(peerInfo, connection)
+        const peerId = connection.remotePeer
+
+        this.registrar.onConnect(peerId, connection)
         this.connectionManager.onConnect(connection)
-        this.emit('peer:connect', peerInfo)
+        this.emit('peer:connect', peerId)
```
jacobheun, line 69: Something to think about since we need to break the API here anyway, should we emit the actual connection here? I think this could be useful. We'll have access to the PeerId via `connection.remotePeer`. It will also make it easy to act on the specific connection of the event, since we can potentially have multiple connections to a peer. Currently, when we get this event there is no way to map this back to a specific connection if we have more than 1 to that peer.

src/index.js
```diff
@@ -133,7 +138,8 @@ class Libp2p extends EventEmitter {
       // Add the identify service since we can multiplex
       this.identifyService = new IdentifyService({
         registrar: this.registrar,
-        peerInfo: this.peerInfo,
+        peerId: this.peerId,
+        addresses: this.addresses,
```
jacobheun, line 142: We'll eventually want to use announce addresses, so this should probably get passed the Address Manager once it's ready, so that we can get the most up to date announce addresses when responding to identify requests.

src/index.js
```diff
@@ -485,7 +477,11 @@ class Libp2p extends EventEmitter {
         let discoveryService
 
         if (typeof DiscoveryService === 'function') {
-          discoveryService = new DiscoveryService(Object.assign({}, config, { peerInfo: this.peerInfo, libp2p: this }))
+          discoveryService = new DiscoveryService(Object.assign({}, config, {
+            peerId: this.peerId,
+            multiaddrs: this.addresses.listen,
```
jacobheun, line 482: When we have an `AddressManager` the discovery service should just use that from `libp2p.addressManager`. While they don't right now, in the future our announce addresses will be able to change over time (AutoNat, AutoRelay, etc) and we should be sharing the most recent ones via the discovery services.

src/registrar.js
```diff
@@ -69,22 +71,20 @@ class Registrar {
...
    * @param {Connection} conn
    * @returns {void}
    */
-  onConnect (peerInfo, conn) {
-    // TODO: This is not a `peer-info` instance anymore, but an object with the data.
-    // This can be modified to `peer-id` though, once `peer-info` is deprecated.
-    // if (!PeerInfo.isPeerInfo(peerInfo)) {
-    //   throw errcode(new Error('peerInfo must be an instance of peer-info'), ERR_INVALID_PARAMETERS)
-    // }
+  onConnect (peerId, conn) {
```
jacobheun, line 78: `Connection` contains the PeerId `connection.remotePeer`, the params here can be consolidated.

src/registrar.js
```diff
@@ -97,46 +97,42 @@ class Registrar {
...
    */
-  onDisconnect (peerInfo, connection, error) {
-    // TODO: This is not a `peer-info` instance anymore, but an object with the data.
-    // This can be modified to `peer-id` though, once `peer-info` is deprecated.
-    // if (!PeerInfo.isPeerInfo(peerInfo)) {
-    //   throw errcode(new Error('peerInfo must be an instance of peer-info'), ERR_INVALID_PARAMETERS)
-    // }
-
-    const id = peerInfo.id.toB58String()
+  onDisconnect (peerId, connection, error) {
```
jacobheun, line 105: Consolidate as mentioned above?

src/identify/index.js
vasco-santos: yep, already being done on #612 :)

src/index.js
vasco-santos: yep, already being done on #612 :)

doc/API.md
```diff
@@ -823,9 +832,7 @@ Get the stored information of a given peer.
 
 | Type | Description |
 |------|-------------|
-| [`PeerInfo`][peer-info] | Peer information of the provided peer |
-
-TODO: change when `peer-info` is deprecated to new pointer
+| `{ id: PeerId, multiaddrInfos: Array<MultiaddrInfo>, protocols: Array<string> }` | Peer information of the provided peer |
```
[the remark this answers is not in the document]
vasco-santos: yeah, I was not totally happy with the `MultiaddrInfo` naming. we are already using that naming in the dht already, but I will do a follow up PR here and in the DHT to move into `Address`

doc/API.md
```diff
@@ -851,15 +859,13 @@ Get all the stored information of every peer.
...
-
-TODO: change when `peer-info` is deprecated to new pointer (breaking change)
+| `Map<string, { id: PeerId, multiaddrInfos: Array<MultiaddrInfo>, protocols: Array<string> }>` | Peer data of every peer known |
 
 #### Example
 
 ```js
-for (let [peerIdString, peerInfo] of peerStore.peers.entries()) {
-  // peerInfo instance
+for (let [peerIdString, peerData] of peerStore.peers.entries()) {
```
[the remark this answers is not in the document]
vasco-santos: Yeah, that's a good point! So, my proposal here is to change to `peer` in `js-libp2p` code and docs, as well as up the stack (daemon, daemon-client and ipfs). We are already using this name in a ton of modules down the stack, like discovery, dht...These we should change over time. What do you think?

src/index.js
vasco-santos: yeah, I got that issue on #611 . I will add this comment there and change it there :)

vasco-santos: https://github.com/libp2p/js-libp2p/pull/611#discussion_r414504819

src/registrar.js
vasco-santos: #611 is on the rescue for this 😛

src/registrar.js
vasco-santos: #611 is on the rescue for this 😛

doc/API.md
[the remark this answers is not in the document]
jacobheun: Yeah that sounds reasonable to me

src/index.js
vasco-santos: Will include in #612  to remove this

PR #611, opened 2020-04-19
chore: refactor conn mgr and registrar

vasco-santos:
This PR refactors the `connectionManager` and `registrar`, as both were keeping a record of the connection without a need for it.

Briefly, except for having connection kept only in the `connectionManager`, this PR moves the connection based events to into the `connectionManager` and adds the `connectionManager` API to the docs.

Previously, `js-libp2p` was emitting `peer:connect` and `peer:disconnect` with false positives. We could get cases where we have 2 connections to the same peer and both events where emitted as well for those cases. Moreover, as this PRs moves the events to the `connectionManager`, we can simplify a lot the `upgrader` handlers.

Instead of `registrar`, `identify`, `metrics` (and others in the future), need to keep track of the connections on the upgraders handler, now they can just rely on listening events of the `connectionManager`, making thinks simpler and more clear. In addition, when we support protocols such as QUIC (with a built in "upgrader"), the connections should need to be added to the `connectionManager`, in the same way as the upgrader does, without any need to provide the connection to the other components/subsystems (`registrar`. `identify`, `metrics`, ...).

Needs

- [x]  #610 
- [x]  #613


src/index.js
```diff
@@ -61,38 +61,33 @@ class Libp2p extends EventEmitter {
     this.upgrader = new Upgrader({
       localPeer: this.peerId,
       metrics: this.metrics,
+      // TODO: Identify should listen on connection Manager event
+      // and onConnection would only map to the connectionManager handler
```
vasco-santos, line 65: Still need to finish this

src/index.js
```diff
@@ -61,38 +61,33 @@ class Libp2p extends EventEmitter {
...
-        this.emit('peer:connect', peerId)
 
         // Run identify for every connection
         if (this.identifyService) {
           this.identifyService.identify(connection, peerId)
             .catch(log.error)
         }
       },
+      // TODO: metrics should listen on connection Manager event
+      // and onConnectionEnd would only map to the connectionManager handler
```
vasco-santos, line 78: Still need to finish this

doc/API.md
```diff
@@ -1118,23 +1160,25 @@ Once you have a libp2p instance, you can listen to several events it emits, so t
...
+### libp2p.connectionManager
 
 #### A new connection to a peer has been opened
 
 This event will be triggered anytime a new Connection is established to another peer.
 
 `libp2p.on('peer:connect', (peer) => {})`
 
-- `peer`: instance of [`PeerInfo`][peer-info]
+- `peer`: instance of [`PeerId`][peer-id]
```
vasco-santos, line 1173: We should change this to connection per https://github.com/libp2p/js-libp2p/pull/610#discussion_r414466296

vasco-santos: Done!

doc/API.md
```diff
@@ -1126,21 +1162,23 @@ unless they are performing a specific action. See [peer discovery and auto dial]
...
 - `peer`: instance of [`PeerId`][peer-id]
 
+### libp2p.connectionManager
+
 #### A new connection to a peer has been opened
 
 This event will be triggered anytime a new Connection is established to another peer.
 
-`libp2p.on('peer:connect', (peer) => {})`
+`libp2p.on('peer:connect', (connection) => {})`
```
jacobheun, line 1171:
```suggestion
`libp2p.connectionManager.on('peer:connect', (connection) => {})`
```

doc/API.md
```diff
@@ -1126,21 +1162,23 @@ unless they are performing a specific action. See [peer discovery and auto dial]
...
 
-- `peer`: instance of [`PeerId`][peer-id]
+- `connection`: instance of [`Connection`][connection]
 
 #### An existing connection to a peer has been closed
 
 This event will be triggered anytime we are disconnected from another peer, regardless of the circumstances of that disconnection. If we happen to have multiple connections to a peer, this event will **only** be triggered when the last connection is closed.
 
-`libp2p.on('peer:disconnect', (peer) => {})`
+`libp2p.on('peer:disconnect', (connection) => {})`
```
jacobheun, line 1179:
```suggestion
`libp2p.connectionManager.on('peer:disconnect', (connection) => {})`
```

src/connection-manager/index.js
```diff
@@ -122,21 +172,63 @@ class ConnectionManager {
    * @param {Connection} connection
    */
   onConnect (connection) {
+    if (!Connection.isConnection(connection)) {
+      throw errcode(new Error('conn must be an instance of interface-connection'), ERR_INVALID_PARAMETERS)
+    }
```
jacobheun, line 177: `onConnect` is an internal handler, we should avoid throwing errors as this is going to be fatal. Log and return instead.

src/identify/index.js
```diff
@@ -45,16 +45,28 @@ class IdentifyService {
   /**
    * @constructor
    * @param {object} options
-   * @param {Registrar} options.registrar
+   * @param {PeerStore} options.peerStore
+   * @param {ConnectionManager} options.connectionManager
```
jacobheun, line 49: This should probably just take `libp2p`, similar to the direction we're going with other services

vasco-santos: I am doing it in #612 . Can we wait for it instead?

PR #612, opened 2020-04-20
feat: address management

### LINKED ISSUE #202: Add support for Addrs {listen, announce, noAnnounce}

vasco-santos:
This PR adds an Address Management component. This component aims to allow users to add to the libp2p configuration `{ listen, announce, noAnnounce }` addresses.

These addresses will be used by each subsystem according to their needs. The details for the reasoning for these addresses are detailed on `/src/address-manager/README.md`, which should be read in this branch (last commit)

Needs:

- [x]  Address Manager tests
- [x] Remove multiaddrs from discovery
- [x]  Consider Listen addresses update on `noAnnounce` addresses filtering
- [x]  Improve configuration doc

Closes #202 

Depends on:
- [x]  #610 
- [x]  #613 
- [x]  #611 
- [x] [libp2p/js-libp2p-mdns#91](https://github.com/libp2p/js-libp2p-mdns/pull/91)

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -366,10 +366,9 @@ const latency = await libp2p.ping(otherPeerId)
 
 ## multiaddrs
 
-Get peer advertising multiaddrs. This computes the advertising multiaddrs of the peer by
-joining the multiaddrs that libp2p transports are listening on with the announce multiaddrs
-provided in hte libp2p config. No announce multiaddrs will be filtered out, even when
-using random ports in the provided multiaddrs.
+Gets the multiaddrs the libp2p node announces to the network. This computes the advertising multiaddrs 
+of the peer by joining the multiaddrs that libp2p transports are listening on with the announce multiaddrs
+provided in the libp2p config. Configured no announce multiaddrs will be filtered out of the advertised addresses.
 
 `libp2p.multiaddrs`
 
@@ -465,6 +464,7 @@ Get the multiaddrs that libp2p transports are using to listen on.
 // ...
 const listenMa = libp2p.transportManager.getAddrs()
 // [ <Multiaddr 047f00000106f9ba - /ip4/127.0.0.1/tcp/63930> ]
+```
 
 ### contentRouting.findProviders
 
--- a/doc/API.md
+++ b/doc/API.md
@@ -11,9 +11,10 @@
   * [`handle`](#handle)
   * [`unhandle`](#unhandle)
   * [`ping`](#ping)
-  * [`addressManager.listen`](#addressManagerlisten)
-  * [`addressManager.announce`](#addressManagerannounce)
-  * [`addressManager.noAnnounce`](#addressManagernoannounce)
+  * [`getAdvertisingMultiaddrs`](#getadvertisingmultiaddrs)
+  * [`addressManager.getListenMultiaddrs`](#addressmanagergetlistenmultiaddrs)
+  * [`addressmger.getAnnounceMultiaddrs`](#addressmanagergetannouncemultiaddrs)
+  * [`addressManager.getNoAnnounceMultiaddrs`](#addressmanagergetnoannouncemultiaddrs)
   * [`contentRouting.findProviders`](#contentroutingfindproviders)
   * [`contentRouting.provide`](#contentroutingprovide)
   * [`contentRouting.put`](#contentroutingput)
@@ -66,7 +67,7 @@ Creates an instance of Libp2p.
 |------|------|-------------|
 | options | `object` | libp2p options |
 | options.modules | `Array<object>` | libp2p modules to use |
-| [options.addresses] | `{ listen: Array<Multiaddr> }` | Addresses to use for transport listening and to announce to the network |
+| [options.addresses] | `{ listen: Array<string>, announce: Array<string>, noAnnounce: Array<string> }` | Addresses for transport listening and to advertise to the network |
 | [options.config] | `object` | libp2p modules configuration and core configuration |
 | [options.connectionManager] | `object` | libp2p Connection Manager configuration |
 | [options.datastore] | `object` | must implement [ipfs/interface-datastore](https://github.com/ipfs/interface-datastore) (in memory datastore will be used if not provided) |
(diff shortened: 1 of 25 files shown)


vasco-santos:
cc @jacobheun 

Once you have time, please check this out and let me know your thoughs. Go to the `feat: address manager` commit and look into:
- Address Manager Readme
- API Readme
- `src/address-manager/index.js`: there is one TODO comment

src/address-manager/index.js
```diff
@@ -0,0 +1,73 @@
+'use strict'
+
+const debug = require('debug')
+const log = debug('libp2p:addresses')
+log.error = debug('libp2p:addresses:error')
+
+const MultiaddrSet = require('./multiaddr-set')
+
+/**
+ * Responsible for managing known peers, as well as their addresses, protocols and metadata.
```
jacobheun, line 10: This isn't correct, maybe copied from AddressBook?

src/address-manager/index.js
```diff
@@ -0,0 +1,73 @@
...
+class AddressManager {
+  /**
+   * @constructor
+   * @param {object} [options]
+   * @param {Array<string>} [options.listen = []] list of multiaddrs string representation to listen.
+   * @param {Array<string>} [options.announce = []] list of multiaddrs string representation to announce.
+   * @param {Array<string>} [options.noAnnounce = []] list of multiaddrs string representation to not announce.
+   */
+  constructor ({ listen = [], announce = [], noAnnounce = [] } = {}) {
+    this._listen = new MultiaddrSet(listen)
```
jacobheun, line 21: We could potentially just simplify this and store everything as strings internally, and return `Multiaddr`s whenever the data is requested. Then this could just be a `Set`.

src/address-manager/index.js
```diff
@@ -0,0 +1,73 @@
...
+    return this._noAnnounce.toArray()
+  }
+
+  /**
+   * Replace listen multiaddrs after trying to listen the original ones by the transports.
+   * This should also forward the replace side effects to the NoAnnounce sets.
+   * @param {Array<Multiaddr>} newMultiaddrs
+   * @return {Array<Multiaddr>}
+   */
+  _replaceListen (newMultiaddrs) {
```
jacobheun, line 58: I don't think we should do this, it makes tracking things more difficult and since we're not tracking everything in `PeerInfo` this is a bit easier to manage. We should keep the `listen`, `announce`, and `noAnnounce` untouched. They could be added to, but we shouldn't internally replace them, these should be user controlled.

We should have a method that computes everything for advertised addresses, this could live on libp2p itself. `libp2p.getAddrs()` or similar. This method would do something like:
- Get the TransportManager addresses, `libp2p.transportManager.getAddrs()`
- Add in the announce addresses, `[...libp2p.transportManager.getAddrs(), ...libp2p.addressManager.announce]`
- And then filter out matching noAnnounce and any duplicate addresses

The big advantage of this is that we are able to update addresses over time and will always be able to reference our original addresses. Things like `Identify` will just be able to call this method when they need the latest addresses.

jacobheun: Also with this, I think we can simplify this class; get rid of the getters and just expose the properties.

vasco-santos: This solution seems good to me and I will try it out. My main concern here is the same issue that I still have to solve: `noAnnounce` addresses using random ports for filtering. Unless the match criteria does not consider the port, we would need to make a comparison between the provided `listenAddr` and what is in `libp2p.transportManager.getAddrs()`

vasco-santos: changed accordingly

src/address-manager/index.js
vasco-santos: Totally agree, I went with this at first because I was already thinking ahead on when we will want to allow users to modify them. But no need to go with this for nothing now!

src/address-manager/index.js
jacobheun: > This solution seems good to me and I will try it out. My main concern here is the same issue that I still have to solve: noAnnounce addresses using random ports for filtering

Not sure I follow. Are you saying concerned that a listening address of `/ip4/0.0.0.0/tcp/0` can't be filtered with a noAnnounce of `/ip4/0.0.0.0/tcp/0`? If so, I don't think this matters and users shouldn't be trying to do this. While this is helpful for testing, listening on random ports is not recommended for production usage. The only noAnnounce "wildcard" I think would be nice to have is filtering loopback IP addresses, but that _could_ be checked by the app before configuring libp2p.

vasco-santos: Yes. I already added support for that now, but it is a bit messy 😅 

> The only noAnnounce "wildcard" I think would be nice to have is filtering loopback IP addresses, but that could be checked by the app before configuring libp2p.

👍

src/address-manager/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Future Considerations
+
+### Dynamic address modifications 
+
+In a future iteration, we can enable these addresses to be modified in runtime. For this, the Address Manager should be responsible for notifying interested subsystems of these changes, through an Event Emitter.
+
+#### Modify Listen Addresses
+
+While adding new addresses to listen on runtime is a feasible operation, removing one listen address might have bad implications for the node, since all the connections using that listen address will be closed. With this in mind and taking also into consideration the lack of good use cases for removing listen addresses, the Address Manager API only allows libp2p users to add new Listen Addresses on runtime.
```
jacobheun, line 43: A future use case for removing listening addresses would be Relay and Rendezvous. If we asked used Rendezvous to find relays to listen on, we could add them to our list. If the relay we found was at its maximum listener capacity, it could potentially tell us to `go away`, at which point we would likely remove it from our listen addresses.

I don't think we need to speculate on the operations here though. Mentioning the potential of them changing in runtime and that it should provide a mechanism for listening to those changes would be sufficient here.

src/address-manager/index.js
```diff
@@ -0,0 +1,55 @@
+'use strict'
+
+const debug = require('debug')
+const log = debug('libp2p:addresses')
+log.error = debug('libp2p:addresses:error')
+
+const multiaddr = require('multiaddr')
+
+/**
+ * Responsible for managing the peer addresses.
```
jacobheun, line 10: Clarification of whos addresses
```suggestion
 * Responsible for managing this peers addresses.
```

src/identify/index.js
```diff
@@ -45,22 +45,20 @@ class IdentifyService {
...
    * @constructor
    * @param {object} options
-   * @param {PeerStore} options.peerStore
-   * @param {ConnectionManager} options.connectionManager
+   * @param {Libp2p} options.libp2p
    * @param {Map<string, handler>} options.protocols A reference to the protocols we support
-   * @param {PeerId} options.peerId The peer running the identify service
-   * @param {{ listen: Array<Multiaddr>}} options.addresses The peer addresses
    */
   constructor (options) {
```
jacobheun, line 51: Not that there are fewer params this could help simplify the `options` accessor calls in the constructor
```suggestion
  constructor ({ libp2p, protocols }) {
```

src/index.js
```diff
@@ -292,6 +293,54 @@ class Libp2p extends EventEmitter {
...
+        if (array.findIndex((otherMa) => otherMa.equals(ma)) !== index) {
+          return false
+        }
+
+        // Filter out if in noAnnounceMultiaddrs
+        if (filterMa.find((fm) => fm.equals(ma))) {
+          return false
+        }
+
+        // Filter out if in the special filter
```
jacobheun, line 330: I would remove support for this right now and we can look at doing a followup PR for this with verbose testing. I think this logic isn't quite right and non TCP addresses might behave unexpectedly here.

vasco-santos:
@jacobheun everything should be addresses now, and the dependent PR, as well as the pubsub discovery module were updated according to these changes

src/circuit/index.js
```diff
@@ -122,7 +122,7 @@ class Circuit {
           type: CircuitPB.Type.HOP,
           srcPeer: {
             id: this.peerId.toBytes(),
-            addrs: this.addresses.listen.map(addr => addr.buffer)
+            addrs: this.addressManager.getListenAddrs().map(addr => addr.buffer)
```
jacobheun, line 125: This should be `libp2p.multiaddrs` not the listen addrs.

src/index.js
```diff
@@ -189,6 +188,8 @@ class Libp2p extends EventEmitter {
    */
   async start () {
     log('libp2p is starting')
+    // TODO: consider validate listen addresses on start?
+    // depend on transports?
```
jacobheun, line 192: What do you mean? When the TransportManager listens it results in each Transport filtering out all addresses that don't match it. We're not logging any information around "hey, you gave us an address and there is no transport that's valid for it", unless non of the addresses match any transport.

doc/API.md
```diff
@@ -360,33 +364,108 @@ Pings a given peer and get the operation's latency.
...
 ```
 
-### peerRouting.findPeer
+## multiaddrs
 
-Iterates over all peer routers in series to find the given peer. If the DHT is enabled, it will be tried first.
+Get peer advertising multiaddrs. This computes the advertising multiaddrs of the peer by
+joining the multiaddrs that libp2p transports are listening on with the announce multiaddrs
+provided in hte libp2p config. No announce multiaddrs will be filtered out, even when
+using random ports in the provided multiaddrs.
```
jacobheun, line 372:
```suggestion
Gets the multiaddrs the libp2p node announces to the network. This computes the advertising multiaddrs 
of the peer by joining the multiaddrs that libp2p transports are listening on with the announce multiaddrs
provided in the libp2p config. Configured no announce multiaddrs will be filtered out of the advertised addresses.
```

doc/API.md
```diff
@@ -360,33 +364,108 @@ Pings a given peer and get the operation's latency.
...
+| Type | Description |
+|------|-------------|
+| `Array<Multiaddr>` | listening multiaddrs |
+
+#### Example
+
+```js
+// ...
+const listenMa = libp2p.transportManager.getAddrs()
+// [ <Multiaddr 047f00000106f9ba - /ip4/127.0.0.1/tcp/63930> ]
```
jacobheun, line 467: Missing the closing "```" which is why the formatting below is off

src/index.js
vasco-santos: This is not valid anymore. Old thought

PR #616, opened 2020-04-24
Adding browser example with communication between peers

svdo:
I'm trying to use libp2p-js to communicate between browser windows. I'm having some issues, so I thought I'd add an example of what I'm trying to accomplish.

This example is NOT working. You can only send one single message to the other peer, then it stops working.

Can somebody help me by telling me what I'm doing wrong? Thanks!!!


examples/chat-in-the-browser/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+  // Listen for new connections to peers
+  let dialedStream = null
+  libp2p.on('peer:connect', peerInfo => {
+    log(`Connected to ${peerInfo.id.toB58String()}`)
+    libp2p.dialProtocol(peerInfo, ['/chat']).then(({ stream }) => {
+      log('dialed a stream', stream)
+      dialedStream = stream
+    })
+  })
```
jacobheun, line 67: Thanks for sharing your use case here. In `0.28` we are planning on deprecating PeerInfo and as such will be returning the connection here. This will let you simplify this whole function to:

```js
libp2p.on('peer:connect', async connection => {
  log(`Connected to ${connection.remotePeer.toB58String()}`)
  dialedStream = await connection.newStream('/chat')
})
```

examples/chat-in-the-browser/index.js
```diff
@@ -0,0 +1,97 @@
...
+    pipe(handledStream, async function (source) {
+      for await (const msg of source) {
+        log(`Received message: ${msg}`)
+      }
+    })
+  })
+
+  function send () {
+    const value = txtSend.value
+    pipe([value], dialedStream)
```
jacobheun, line 91: The problem you're having is here, `pipe([value], dialedStream)`. What this does is write `value` and then it closes the write side of this stream, so you **cant** send anymore data. Since the write side of the stream is closing, your `for await` log loop on the receiver side will close the read side of the connection, but not the write. So you have half open streams right now.

You can do 1 of two things.
1. Keep the stream open and pass chats back and forth over it using something like `it-pushable`, or an async generator that yields your messages, so the stream doesn't close unless you explicitly close it or it errors.
2. Just create a new `/chat` stream and send the data when it needs to be sent.

Option 2 is likely going to be safest. There is some minor cost overhead creating the streams, but the connection is still open so it's pretty low. You can also 'ack' messages by sending back an empty value, `pipe([], dialedStream)`, which will fully close the stream.

I'd recommend updating `send` to take a peer of who you want to send to. 

```js
async function send (peerInfo, message) {
  try {
    const { stream } = await libp2p.dialProtocol(peerInfo, ['/chat'])
    await pipe([message], stream, consume) // consume from `streaming-iterables` will wait for the empty 'ack'
  } catch (err) {
    console.log('Could send the message', err)
  }
}
```

svdo: Yes that confirms my suspicion. I was thinking about a way to accomplish (1) using an async generator, but I wasn't sure how to do that. I'll try with the `it-pushable`.

As for your suggestion (2), I'm not sure what you mean. What API call were you thinking of to create that new stream?

(Thanks for responding so fast by the way!)

alexcastillo:
Hey @svdo,

Thanks for putting this example together. I had the same question.

Assuming we get past this issue:

If you deploy this example to a remote server (e.g. https://example.com), are you able to get the peers to connect? I tried with a very similar example and the peers discover each other but they don't connect.

@jacobheun, any ideas?

examples/chat-in-the-browser/index.js
jacobheun: The `send` function above will do this. When you want to send a message to a certain peer just pass their PeerInfo and the message to the send function. If there is already a connection `dialProtocol` will just create a new stream on the existing connection. It will send the data and then wait for the ack.

Your incoming message handler could then look something like:
```js
await libp2p.handle(['/chat'], async ({ connection, stream }) => {
  log(`handle chat from ${connection.remotePeer.toB58String()}`)
  // Read in all messages directly from the streams `source` async iterable
  for await (const msg of stream.source) {
    log(`Received message: ${msg}`)
  }
  // Write the ack (an empty reply) message to close the stream. You can call sink to do this directly
  await stream.sink([])
  log('Ack sent, stream is closed')
})
```

Note, you can use `pipe` instead of calling the stream methods, but this might be a bit clearer for a read and respond approach.

jacobheun:
> I tried with a very similar example and the peers discover each other but they don't connect.

@alexcastillo Do you have code or logs to look at? There could be a number of reasons causing the issue. The quickest thing to maybe check is increasing the `minConnections` configuration for the connection manager, https://github.com/libp2p/js-libp2p/blob/v0.27.7/doc/CONFIGURATION.md#configuring-connection-manager. If it's 0 (default), you won't auto dial peers. Try with a min of 25 and a max of 50 for browser nodes to see if that helps.

If that doesn't help can you open a new issue to discuss. I want to avoid splintering the discussion here.

alexcastillo:
Thanks @jacobheun. Your suggestion didn't work so I'll go ahead an open a new issue later today. Thank you!

examples/chat-in-the-browser/index.js
svdo: Thanks for clarifying @jacobheun! I made a few more commits. One where the example runs with `it-pushable`, which works. Then I moved on to your option 2 as mentioned above.

> Option 2 is likely going to be safest. There is some minor cost overhead creating the streams, but the connection is still open so it's pretty low. You can also 'ack' messages by sending back an empty value, `pipe([], dialedStream)`, which will fully close the stream.

I updated the example to use the `send` method you suggested, and it works as well. This raises two questions for me:

* You mentioned the `pipe([], dialedStream)` call to close the stream. I assume this was meant to be on the "handle" side of things, so `pipe([], handledStream)`, right? (This is how I now put it in the example.)
* Can you clarify what you meant by saying that this option is "safest"? Safe in what way? What kind of problems have you seen that are solved by this?

Based on your answers I will add some more comments to the example. Thanks again!!

svdo:
By the way, I've been doing my work in TypeScript, so as a bonus I converted this example to TypeScript as well; may it help others :)

examples/chat-in-the-browser/index.js
jacobheun: > You mentioned the pipe([], dialedStream) call to close the stream. I assume this was meant to be on the "handle" side of things, so pipe([], handledStream), right?

Yes, that was a mistake on my part

> Can you clarify what you meant by saying that this option is "safest"? Safe in what way?

There are 2 things you'll want to make sure to account for.
1. `it-pushable` has no flow control, which means you need to deal with write overflows backoff on the writer side. The nice thing about async iterables is allowing the stream to read in the messages rather than pushing them.
2. Keeping a long lived stream means you need to take better care about closing it when you no longer need it. The nice thing about option 2 is that the stream only lives as long as the message send and ack, so there's less risk in accidentally leaking streams by not closing them.

examples/chat-in-the-browser/index.ts
```diff
@@ -0,0 +1,131 @@
...
+  status.innerText = 'libp2p started!'
+  log(`libp2p id is ${libp2p.peerInfo.id.toB58String()}`)
+
+  const handleChat: ProtocolHandler = async ({ connection, stream }) => {
+    log(`handle chat from ${connection?.remotePeer.toB58String()}`)
+    const handledStream = stream
+    pipe(handledStream, async function (source: AsyncGenerator<any, any, any>) {
+      for await (const msg of source) {
+        log(`Received message: ${msg}`)
+        pipe([], handledStream)
```
jacobheun, line 103: This isn't causing issues now because you're only sending 1 message, but if you send multiples this will fail. The pipe needs to be outside of the for loop. If you want to ack individual messages you would likely use the `it-pushable` approach as well, or a [transform](https://github.com/libp2p/js-libp2p/blob/master/doc/STREAMING_ITERABLES.md#transform) to send the response for each message.

examples/chat-in-the-browser/index.js
svdo: That’s super helpful, thanks Jacob!

examples/chat-in-the-browser/index.ts
svdo: Yes, that makes sense,I’ll update the example.

svdo:
I think this example is pretty much done. Do you want to merge it @jacobheun?

svdo:
(I think the build errors and coverage change are not due to this PR?)

jacobheun:
@svdo our plan is to create a repository for registering/cataloging community created examples / projects like this, it's being tracked at https://github.com/libp2p/js-libp2p/issues/512. 

As that's not yet ready, what I'd like to do is have you add this example to a standalone repo, and then add a link and brief summary of it to the examples readme here. Once we create the community repo/site, I can then add yours and any other examples that get added in the mean time there.

svdo:
Hmm I was hoping that your suggestion was going to end with something like "since we're not there yet we will merge this example and move it later" or something like that. I thought maybe this example adds enough value to justify being here. Reason I'm saying this is that I don't see myself converting and maintaining this code in the future, so it would be "instant legacy".

jacobheun:
The project I mentioned is still going to just reference projects/examples, not include them directly. I appreciate that you likely won't be able to maintain the example with new releases, but there is still value to having it and we can add metadata to the version it does support. We will eventually integrate some examples and projects into CI testing and help to keep those updated, but we unfortunately don't have the bandwidth to directly support updating an increasing list of them.

If this still isn't something you're interested in considering we could look at amending the existing browser example to include dialing and sending data.

svdo:
Right, I completely understand that you're also mindful of the "bandwidth thing" :) I'll go ahead and post it to a repo of my own. I'll add the link here and close the pull request afterwards.

svdo:
Ok I moved the example to its own repo, here you go: https://github.com/svdo/js-libp2p-example-chat-in-the-browser.

Issue #591, opened 2020-03-25
PeerStore Persistence


vasco-santos:
# PeerStore Persistence

As part of the [PeerStore improvements epic](https://github.com/libp2p/js-libp2p/issues/582), we intend to back the PeerStore with a [datastore](https://github.com/ipfs/interface-datastore/).

## Sub-milestones

- Add persitence back end to the books
- Configurable persistence

## Overview

Centralizing all the information a peer has about its environment enables us to easily persist this data. This is particularly useful when we need to restart a node as we will be able to start establishing connections with peers that we know (other than the bootstrap nodes).

This should have a big impact on browser nodes, since they have less ways of discovering nodes and these discovery services sometimes take a longer times. Moreover, the nodes will have a bigger set of connected nodes considerably faster.

Other than faster connectivity, a persisted peerStore enables us to rely less in the bootstrap nodes (via configuration), so that we reduce the load on them.

#### What to persist

The PeerStore is composed by 4 different components, `addressBook`, `KeyBook`, `metadataBook` and `protoBook`. While some of the content of these books are super relevant to persist, others might not have a clear value.

The `addressBook` contains a list of `multiaddr` for each peer, as well as some relevant data for each `multiaddr`, including their validity, degree and confidence, .... While the `multiaddr` are super valuable, the remaining two are discussable. Considering that we will need to dial the peers when the peer restarts, the validity will be updated, so it is not that relevant to store. On the other side, the degree of confidence can have a good impact if we have multiple multiaddresses for the peer, since we will dial multiple ones in parallel.

The `keyBook` content must be stored as it is crucial for the correct work of the system.

The `metadataBook` may contain previously set metadata about each peer. As a result, this information should still exist when the peer is restarted.

The `protoBook` contains the list of protocols supported by each peer. As we will need to establish connections with the peers, we can run the identify service to get the updated list of protocols they support and multiaddrs they are listening on. Therefore, this information does not seem crucial to store. However, it could potentially be important if we had a large number of peers. In this context, we could choose the best peers to connect based on the protocols they run (but the information can also be outdated).

#### How to persist

A [datastore](https://github.com/ipfs/interface-datastore) stores the data in a key-value fashion. As a result, we need coherent keys so that we do not overwrite data.

A datastore allows us to query it through a key prefix. This way, we can find all the information if we define a consistent namespace that allow us to find the content without having any information.

The namespaces was defined as follows:

**AddressBook**

[...]

PR #619, opened 2020-04-27
feat: peerstore persistence

vasco-santos:
This PR is part of the [PeerStore improvements EPIC](https://github.com/libp2p/js-libp2p/issues/582) and tackles the milestone 3 - [PeerStore Persistence](https://github.com/libp2p/js-libp2p/issues/591).

As a follow up of the work in #590, #598 and #610, this PR adds a datastore to backup the data stored in the AddressBook and ProtoBoo..

Taking into account that most of our books (or all) will leverage this feature, it was added to the Book base class. However, since different books store data structured in different way and might also intend to store them in the datastore differently (or a subset of it), the Book allows the implementer class to transform the data through function parameters.

## Datastore

As stated in #591, the keys used to store this data should have common namespaces to allow it to loaded without prior knowledge on node startup. In addition, this was based on the `go-libp2p` naming as speaking one language is better than two. The namespaces naming is as follows

#### AddressBook

All the knownw peer addresses are stored with a key pattern as follows:

`/peers/addrs/<b32 peer id no padding>`

#### ProtoBook

All the knownw peer protocols are stored with a key pattern as follows:

`/peers/protos/<b32 peer id no padding>`

#### Other books

The other books naming is described in #591, but it is important pointing out that keybook will need a key suffix for specifying the type of key. This was already taken into account while implementing this.

#### Enabled by default

While the goal of persisting the PeerStore is for users to use it, if we default to use the Memory datastore the data will not be available on a browser page refresh, as an example. While storing data in the Memory datastore does not provide us real value, it is the only available datastore which works for every runtime. Accordingly, if users need to specify the datastore they want to use, it might be a good option to also require them to enable the peerStore persistence. As a consequence, there will be no misunderstandings on users thinking that the PeerStore data is being stored, but since it is only stored in Memory, it will not be available in the future.

If we go with not enabling it by default, we should do it once we work on improving the configuration. Since we intend to work on pre built configurations for the main runtimes, we would leverage the proper datastores in them.

#### Store protocols

As far as I Know, `go-libp2p` is persisting the AddressBook, KeyBook and MetadataBook, but not the ProtoBook. Firstly, we will run the identify protocol after establishing a connection with other peers and gather their updated protocols. So, as far as we want to connect to the stored peers blindly we do not need to store the protocols as they will be collected. However, there might be scenarios where this data could be relevant (even with no guarantess of being updated). For instance, if we already have a large number of peers stored once a libp2p node starts, it can potentially select a subset of the peers to connect according to the protocols it is running and the known protocols potentially have running.

Considering this, I am leaning towards allowing this to be configured  on user land. Each book should be able to have persistence enabled in the node configuration.

**EDIT**: in the libp2p sync call I got to know that `go-libp2p` is storing the protocols as well, but under the `metadata` namespace, with a `protocols` property:  https://github.com/libp2p/go-libp2p-peerstore/blob/master/pstoreds/protobook.go#L55 where `meta` is the metadata book: https://github.com/libp2p/go-libp2p-peerstore/blob/master/pstoreds/metadata.go

#### Consider storing multiaddrs metadata

Since we do not have multiaddrs metadata yet we will not store it for now. However, if we intend to store it in the future we should already consider it in terms of the protobuf definition. Potential validity and multiaddr confidence might be relevant in the present, but I am not sure on wether they will have value in the future.

Looking into [Browser storage limits references for IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Browser_storage_limits_and_eviction_criteria), I think that it will not be a problem to store this metadata in terms of storage capacity. However, this data might also be modified frequently, which will result in lots of operations with the DB.

Needs:

- [x] #611 
- [x] #612 

As a follow up, we should create an issue to benchmark the changes in the PeerStore datastore that are not relevant, since we are "updating" all data from a dirty peer, while only the data of one book might be modified.


src/peer-store/book.js
```diff
@@ -75,9 +197,17 @@ class Book {
       [this.eventProperty]: []
     })
 
+    // Update Persistence datastore
+    this._persistData(peerId, [])
```
vasco-santos, line 201: Maybe we should remove it instead 🤔

jacobheun: Yes, persist the delete action. Otherwise we'll bloat the datastore with peers that we "deleted"

vasco-santos:
@jacobheun you can have an initial look at this. Please read the main message of the PR, since there are some open questions that might influence this implementation.

I am currently assuming this enabled by default, no configuration for type of book to persist, persist protocols as their own thing and not considering multiaddrs metadata. I will need to change what we decide from  this, but from my initial thoughts, I would go with:

- Disabled by default
- Enabled Persistence configurable by book
- Persist protocols as its own thing (like it currently is)
- Do not consider multiaddrs metadata (like it currently is)

jacobheun:
> If we go with not enabling it by default, we should do it once we work on improving the configuration. Since we intend to work on pre built configurations for the main runtimes, we would leverage the proper datastores in them.

I think it makes sense to not enable it by default, as it's a waste of processing time since we can't actually persist anything through restarts/refreshes with the memory store.

> Considering this, I am leaning towards allowing this to be configured on user land. Each book should be able to have persistence enabled in the node configuration.

As mentioned, go does store this, and I think it makes sense for us to store it by default (assuming persistence is enabled). As we improve Topology usage with the Connection Manager this should help with bootstrapping the peer on restarts, assuming the peers are stable.

> Looking into Browser storage limits references for IndexedDB, I think that it will not be a problem to store this metadata in terms of storage capacity. However, this data might also be modified frequently, which will result in lots of operations with the DB.

We can evaluate storing the multiaddr metadata when that work starts. Maintaining TTLs and Confidence can be valuable when restarts are within the TTL expiration time.

re: lots of operations with the DB, we can look at performing flushes to avoid every metadata change causing a write.


> `/peers/protos/<b32 peer id no padding>`

fyi, The proto format description currently includes the PeerId, not the protocol

vasco-santos:
> We can evaluate storing the multiaddr metadata when that work starts. Maintaining TTLs and Confidence can be valuable when restarts are within the TTL expiration time.

My main concern about this will be the breaking change in the datastore. f you think maintaining ttls and confidence can be valuable, I feel that it might be a good idea to just take this into account in the protobuf definition:

```js
const message = `
message Address {
  required bytes addr = 1;
}

message Addresses {
  repeated Address addrs = 1;
}`
```
vs

```js
const message = `
message Addresses {
  repeated Add addrs = 1;
}`
```

jacobheun:
Yes, I think we'll want to keep track of Address records and not just multiaddrs, so I would account for that here.

vasco-santos:
This should now be good to a review @jacobheun

.aegir.js
```diff
@@ -45,7 +45,7 @@ const after = async () => {
 }
 
 module.exports = {
-  bundlesize: { maxSize: '179kB' },
+  bundlesize: { maxSize: '185kB' },
```
vasco-santos, line 48: This was updated meanwhile in master from 220 to 179, which had a minimal interval. I rebased `0.28.x` with master and this was immediately an issue, so I bumped it for now. We can analyse later on what the ideal value is

doc/API.md
```diff
@@ -74,6 +74,7 @@ Creates an instance of Libp2p.
 | [options.dialer] | `object` | libp2p Dialer configuration
 | [options.metrics] | `object` | libp2p Metrics configuration
 | [options.peerId] | [`PeerId`][peer-id] | peerId instance (it will be created if not provided) |
+| [options.peerStore] | [`PeerId`][peer-id] | libp2p PeerStore configuration |
```
jacobheun, line 77: Type is incorrect for the `peerStore` options

doc/CONFIGURATION.md
```diff
@@ -505,6 +505,32 @@ const node = await Libp2p.create({
 })
 ```
 
+#### Configuring PeerStore
+
+PeerStore persistence is disabled in libp2p by default. You can enable and configure it as follows. Aside from enabled being `false` by default, it will need an implementation of a [datastore](https://github.com/ipfs/interface-datastore).
```
jacobheun, line 510: It might be good to just add a brief note about not using the memory datastore as that will be ineffective for persistence.

doc/CONFIGURATION.md
```diff
@@ -505,6 +505,32 @@ const node = await Libp2p.create({
...
+  modules: {
+    transport: [TCP],
+    streamMuxer: [MPLEX],
+    connEncryption: [SECIO]
+  },
+  datastore: new LevelStore('path/to/store'),
+  peerStore: {
+    persistence: true,
+    threshold: 5
+  }
```
jacobheun, line 530: We need to make sure to document the options somewhere so it's clear what each one does

src/index.js
```diff
@@ -43,9 +44,15 @@ class Libp2p extends EventEmitter {
...
     this.peerId = this._options.peerId
-    this.peerStore = new PeerStore()
+    this.datastore = this._options.datastore
+
+    this.peerStore = !(this.datastore && this._options.peerStore.persistence)
+      ? new PeerStore()
+      : new PersistentPeerStore({
+        datastore: this.datastore,
+        ...this._options.peerStore
+      })
```
jacobheun, line 55: nit: flip this to avoid inverting the conditional, it's harder to follow that way

```suggestion
    this.peerStore = (this.datastore && this._options.peerStore.persistence)
      ? new PersistentPeerStore({
        datastore: this.datastore,
        ...this._options.peerStore
      })
      : new PeerStore()
```

src/peer-store/README.md
```diff
@@ -83,7 +83,47 @@ Access to its underlying books:
 - `change:multiaadrs` - emitted when a known peer has a different set of multiaddrs.
 - `change:protocols` - emitted when a known peer supports a different set of protocols.
 
+## Data Persistence
+
+The data stored in the PeerStore can be persisted if configured appropriately. Keeping a record of the peers already discovered by the peer, as well as their known data aims to improve the efficiency of peers joining the network after being offline.
+
+The libp2p node will need to receive a [datastore](https://github.com/ipfs/interface-datastore), in order to store this data in a persistent way. A [datastore](https://github.com/ipfs/interface-datastore) stores its data in a key-value fashion. As a result, we need coherent keys so that we do not overwrite data.
```
jacobheun, line 90:
```suggestion
The libp2p node will need to receive a [datastore](https://github.com/ipfs/interface-datastore), in order to persist this data across restarts. A [datastore](https://github.com/ipfs/interface-datastore) stores its data in a key-value fashion. As a result, we need coherent keys so that we do not overwrite data.
```

src/peer-store/README.md
```diff
@@ -83,7 +83,47 @@ Access to its underlying books:
...
+
+The PeerStore should not be continuously updating the datastore with the new data observed. Accordingly, it should only store new data after reaching a certain threshold of "dirty" peers, as well as when the node is stopped.
```
jacobheun, line 92:
```suggestion
The PeerStore should not continuously update the datastore whenever data is changed. Instead, it should only store new data after reaching a certain threshold of "dirty" peers, as well as when the node is stopped, in order to batch writes to the datastore.
```

src/peer-store/README.md
```diff
@@ -83,7 +83,47 @@ Access to its underlying books:
...
+
+Taking into account that a datastore allows queries using a key prefix, we can find all the information if we define a consistent namespace that allow us to find the content without having any information. The namespaces were defined as follows:
```
jacobheun, line 94: This is a bit hard to follow, maybe simplify to something like:
```suggestion
The peer id will be appended to the datastore key for each data namespace. The namespaces were defined as follows:
```

src/peer-store/README.md
```diff
@@ -83,7 +83,47 @@ Access to its underlying books:
...
+
+**AddressBook**
+
+All the knownw peer addresses are stored with a key pattern as follows:
```
jacobheun, line 98:
```suggestion
All the known peer addresses are stored with a key pattern as follows:
```

src/peer-store/README.md
```diff
@@ -83,7 +83,47 @@ Access to its underlying books:
...
+
+`/peers/addrs/<b32 peer id no padding>`
+
+**ProtoBook**
+
+All the knownw peer protocols are stored with a key pattern as follows:
```
jacobheun, line 104:
```suggestion
All the known peer protocols are stored with a key pattern as follows:
```

src/peer-store/persistent/index.js
```diff
@@ -0,0 +1,225 @@
...
+    this.threshold = threshold
+    this._addDirtyPeer = this._addDirtyPeer.bind(this)
+  }
+
+  /**
+   * Start Persistent PeerStore.
+   * @return {Promise<void>}
+   */
+  async start () {
+    log('Persistent PeerStore is starting')
```
jacobheun, line 54: Since the debug namespace is `libp2p:persistent-peer-store` I think you can drop `Persistent` in these logs.

src/peer-store/persistent/index.js
```diff
@@ -0,0 +1,225 @@
...
+const { Key } = require('interface-datastore')
+const multiaddr = require('multiaddr')
+const PeerId = require('peer-id')
+
+const PeerStore = require('..')
+
+const {
+  ADDRESS_NAMESPACE,
+  COMMON_NAMESPACE,
+  PROTOCOL_NAMESPACE
```
jacobheun, line 16: nit: consider calling these `NAMESPACE_<>`, it can make the values easier to auto suggest in IDEs

src/peer-store/persistent/index.js
```diff
@@ -0,0 +1,225 @@
...
+  /**
+   * Process datastore entry and add its data to the correct book.
+   * @private
+   * @param {Object} params
+   * @param {string} params.key datastore key
+   * @param {Buffer} params.value datastore value stored
+   */
+  _processDatastoreEntry ({ key, value }) {
+    try {
+      const keyParts = key.toString().split('/')
```
jacobheun, line 193: key is already a string, is `toString` necessary?

src/peer-store/persistent/index.js
```diff
@@ -0,0 +1,225 @@
...
+      const peerId = PeerId.createFromCID(keyParts[3])
```
jacobheun, line 194: I don't think this creates a problem because we compute the data for `peerStore.peers`, but unmarshalling from the datastore doesn't set the peer id in the peerstore itself. So on restart we'd have no ids in the peerstore, but data in each of the books which sounds like a recipe for some edge cases in the future.

test/peer-store/address-book.spec.js
```diff
@@ -41,21 +41,33 @@ describe('addressBook', () => {
...
     it('throwns invalid parameters error if invalid PeerId is provided', () => {
-      expect(() => {
+      try {
         ab.set('invalid peerId')
-      }).to.throw(ERR_INVALID_PARAMETERS)
+      } catch (err) {
+        expect(err.code).to.equal(ERR_INVALID_PARAMETERS)
+        return
+      }
+      throw new Error('invalid peerId should throw error')
```
jacobheun, line 50: Any reason for changing these over to try/catches? If you were having errors running this file by itself it's likely due to missing `chai.use(require('chai-as-promised'))` in the require statements.

test/peer-store/persisted-peer-store.spec.js
```diff
@@ -0,0 +1,374 @@
...
+
+    beforeEach(async () => {
+      memoryDatastore = new MemoryDatastore()
+      ;[libp2p] = await peerUtils.createPeer({
+        started: false,
+        config: {
+          datastore: memoryDatastore,
+          peerStore: {
+            persistence: true,
+            threshold: 2 // trigger on  second peer changed
```
jacobheun, line 300: Just something to think about, we should probably do/recommend a threshold of 1 in browsers. There is no interval and browser nodes won't "stop" properly out of the box, so we could end up with unflushed records when the window is closed

doc/CONFIGURATION.md
vasco-santos: I agree with this. Where do you think we should do it?

We are also not doing it for the dialer, metrics, connectionManager... We should probably add all of them in the `API.md`, after the create? I should probably do a follow up PR with all of them?

src/peer-store/persistent/index.js
vasco-santos: The key will be a https://github.com/ipfs/interface-datastore/blob/master/src/key.js, not a string

src/peer-store/persistent/index.js
vasco-santos: Yeah, this does not create an issue now, but the keyBook will be populated with the PeerId :)

doc/CONFIGURATION.md
jacobheun: I think at least having comments in the Configuration.md examples for each section that show the defaults would be helpful. We have it in some of the other sections, just not everywhere. A followup PR works for me

src/peer-store/persistent/index.js
jacobheun: jsdocs need updated then

test/peer-store/persisted-peer-store.spec.js
vasco-santos: We should probably discuss the addition of an interval for persist it. However, this will also not be a bullet proof solution. I can create an issue for this as an action item.

vasco-santos: Meanwhile, I added a note for that in the CONFIGURATION.md

test/peer-store/address-book.spec.js
vasco-santos: These tests were working correctly before because the error was not correctly required

If I try to use the approach with `chai-as-promised`, I got he error: `TypeError: [Function] is not a thenable.`, which makes sense because this is not a promise.

doc/CONFIGURATION.md
vasco-santos: I added comments as you recommended for the PeerStore now. But, I will make sure to add proper docs as a follow up PR to all the components

src/peer-store/persistent/index.js
vasco-santos: yeah, I did not realize I add a `string` there.

src/peer-store/persistent/index.js
jacobheun: Mostly what I mean is that `PeerStore.peerIds` is an unreliable Map. It might not contain all the peers we know about. `PeerStore.peers` is computed, so it will be reliable.

PR #626, opened 2020-05-04
feat: keybook

vasco-santos:
This PR is part of the [PeerStore improvements EPIC](https://github.com/libp2p/js-libp2p/issues/582) and tackles part of the milestone 4 - [KeyBook](https://github.com/libp2p/js-libp2p/issues/591).

As a follow up of the work in #590, #598, #610 and #619 , this PR adds they KeyBook and integrates it with the `PeerStorePersistence`.

Needs:

- [x] #619

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -814,8 +814,6 @@ Add known `protocols` of a given peer.
 peerStore.protoBook.add(peerId, protocols)
 ```
 
-* [`peerStore.keyBook.get`](#peerstorekeybookget)
-* [`peerStore.keyBook.set`](#peerstorekeybookset)
 
 ### peerStore.keyBook.delete
 
@@ -840,7 +838,7 @@ Delete the provided peer from the book.
 ```js
 peerStore.keyBook.delete(peerId)
 // false
-peerStore.keyBook.set(peerId)
+peerStore.keyBook.set(peerId, publicKey)
 peerStore.keyBook.delete(peerId)
 // true
 ```
@@ -868,7 +866,7 @@ Get the known `PublicKey` of a provided peer.
 ```js
 peerStore.keyBook.get(peerId)
 // undefined
-peerStore.keyBook.set(peerId) // with inline public key
+peerStore.keyBook.set(peerId, publicKey)
 peerStore.keyBook.get(peerId)
 // PublicKey
 ```
@@ -877,13 +875,14 @@ peerStore.keyBook.get(peerId)
 
 Set known `peerId`. This can include its Public Key.
 
-`peerStore.keyBook.set(peerId)`
+`peerStore.keyBook.set(peerId, publicKey)`
 
 #### Parameters
 
 | Name | Type | Description |
 |------|------|-------------|
 | peerId | [`PeerId`][peer-id] | peerId to set |
+| publicKey | [`RsaPublicKey|Ed25519PublicKey|Secp256k1PublicKey`][keys] | peer's public key |
 
 #### Returns
 
@@ -894,7 +893,8 @@ Set known `peerId`. This can include its Public Key.
 #### Example
 
 ```js
-peerStore.keyBook.set(peerId)
+const publicKey = peerId.pubKey
+peerStore.keyBook.set(peerId, publicKey)
 ```
 
 ### peerStore.protoBook.delete
@@ -1420,3 +1420,4 @@ This event will be triggered anytime we are disconnected from another peer, rega
 [connection]: https://github.com/libp2p/js-interfaces/tree/master/src/connection
 [multiaddr]: https://github.com/multiformats/js-multiaddr
 [peer-id]: https://github.com/libp2p/js-peer-id
+[keys]: https://github.com/libp2p/js-libp2p-crypto/tree/master/src/keys
--- a/doc/API.md
+++ b/doc/API.md
@@ -26,6 +26,9 @@
   * [`peerStore.addressBook.get`](#peerstoreaddressbookget)
   * [`peerStore.addressBook.getMultiaddrsForPeer`](#peerstoreaddressbookgetmultiaddrsforpeer)
   * [`peerStore.addressBook.set`](#peerstoreaddressbookset)
+  * [`peerStore.keyBook.delete`](#peerstorekeybookdelete)
+  * [`peerStore.keyBook.get`](#peerstorekeybookget)
+  * [`peerStore.keyBook.set`](#peerstorekeybookset)
   * [`peerStore.protoBook.add`](#peerstoreprotobookadd)
   * [`peerStore.protoBook.delete`](#peerstoreprotobookdelete)
   * [`peerStore.protoBook.get`](#peerstoreprotobookget)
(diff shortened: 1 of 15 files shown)


src/peer-store/key-book.js
```diff
@@ -0,0 +1,83 @@
...
+    const recPeerId = this.data.get(id)
+
+    !recPeerId && this._ps.emit('peer', peerId)
+    // If no record available, or it is incomplete
+    if (!recPeerId || (peerId.pubKey && !recPeerId.pubKey)) {
+      this._setData(peerId, peerId, {
+        emit: Boolean(peerId.pubKey) // No persistence if no public key
+      })
+      log(`stored provided public key for ${id}`)
+    }
```
jacobheun, line 61: This block is a bit confusing. I think it's trying to cover too many use cases?
1. we have no record of the peer
1.1 we have no record of the peer, and we dont have a public key to add so dont emit (why are we calling set on the keybook if this is the case?)
2. we have an existing record and a new public key but no existing public key (why do we have a record with no public key in the keystore?)

I think the keystore is also trying to keep track of peer ids, we should just avoid this. Let the peer store do that and just make it a computed property across the books, because if we have no addresses/protocols/keys/etc associated to a peer id, it really shouldn't exist for us as it has 0 value.

doc/API.md
```diff
@@ -811,6 +814,89 @@ Add known `protocols` of a given peer.
...
+peerStore.keyBook.set(peerId) // with inline public key
+peerStore.keyBook.get(peerId)
+// PublicKey
+```
+
+### peerStore.keyBook.set
+
+Set known `peerId`. This can include its Public Key.
+
+`peerStore.keyBook.set(peerId)`
```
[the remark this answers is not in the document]
vasco-santos: Just keeps the PeerId in it, and might update in the future. I would see the `peer-id` existing when we discover a new peer

src/peer-store/key-book.js
```diff
@@ -0,0 +1,83 @@
...
+  set (peerId) {
+    if (!PeerId.isPeerId(peerId)) {
+      log.error('peerId must be an instance of peer-id to store data')
+      throw errcode(new Error('peerId must be an instance of peer-id'), ERR_INVALID_PARAMETERS)
+    }
+
+    const id = peerId.toB58String()
+    const recPeerId = this.data.get(id)
+
+    !recPeerId && this._ps.emit('peer', peerId)
```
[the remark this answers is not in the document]
vasco-santos: > We should avoid other things triggering events on the peer store and instead rely on the peer store to determine when it should be emitting things. If we don't isolate we'll run the risk of having what we used to have, where several things are emitting peer events which can result in a lot of unnecessary events.

I did not understand your point here. Currently, the `Book.js` base class is emitting the events for changed peers, while this is the previous event emitted when the peer was discovered (in the addressBook), but now that we have the keyBook keeping the peerIds, it feels that this is the place this should be.

> It would probably be good to take a step back and think about the flow of the peers data into the peer store. I dont think that updating the keybook anytime we learn about the peer is the intuitive action. Updating the keybook after crypto negotiation makes sense.

> It might be good to revisit the peer store readme and make sure that is up to date and makes sense with these flows.

I added it to update the keyBook because I did not want to keep track of the peerIds in two different data structures. If we move into only storing the `peerId` in the keybook if we have a public key for it, we will need to go back with the `peer` event into the addressBook.

Updating the keybook after crypto negotiation makes sense, and I forgot that flow. However, subsystems like the dht will also need to be able to set the public key for a peer.

All things considered, should we go into only storing the PeerId when it has a public key (we can reconstruct it into the getters with `createFromCID` if we do not store it? With that, we do not update the keybook anymore on the other books set and we emit the `peer` event from the addressBook? This way, the upgrader and the dht would call the `keyBook.set()`...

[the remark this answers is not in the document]
vasco-santos: If we go this way, should we throw an error if there is no public key provided in the peerId to set? Probably we should change the API into `keyBook.set(peerId, publicKey)`. While the parameter is redundant, it will sign in an easier way to the user that it is expected

src/peer-store/README.md
```diff
@@ -10,7 +10,9 @@ Several libp2p subsystems will perform operations, which will gather relevant in
 
 In a libp2p node's life, it will discover peers through its discovery protocols. In a typical discovery protocol, addresses of the peer are discovered along with its peer id. Once this happens, the PeerStore should collect this information for future (or immediate) usage by other subsystems. When the information is stored, the PeerStore should inform interested parties of the peer discovered (`peer` event).
 
-Taking into account a different scenario, a peer might perform/receive a dial request to/from a unkwown peer. In such a scenario, the PeerStore must store the peer's multiaddr once a connection is established. 
+Taking into account a different scenario, a peer might perform/receive a dial request to/from a unkwown peer. In such a scenario, the PeerStore must store the peer's multiaddr once a connection is established.
+
+When a connection is being upgraded, more precisely after its encryption, or even in a discovery protocol, a libp2p node can get to know other parties public keys. In this scenario, libp2p will add the peer's public key to its `KeyBook`.
```
vasco-santos, line 15: @jacobheun I could not do this. 

I think that it makes sense, but afaik noise uses static public keys, which I would not expect to have here. In addition, we will need to provide libp2p to the crypto module for adding the key to the keyBook or change the crypto interface to also require the public key, so that we can update this in the `upgrader.js`.

We should probably give some thought on this and follow up with a good solution.

jacobheun: Noise has its own keycache and I dont think we need to worry about persistence of the crypto transport keys, just the libp2p id keys. I think this section is achieved by updating the peer after we've established a connection, because the crypto handshake will result in libp2p public key exchange and verification.

vasco-santos: For the inbound connection, we do have the public key after the connection being established. The same is not true for the outbound connection. Both for `libp2p-secio` and `libp2p-noise`

I went deeper on secio to figure out why this is happening and I found this: https://github.com/libp2p/js-libp2p-secio/blob/master/src/handshake/crypto.js#L66-L73

When an inbound connection, we go to the else statement while in an outbound connection we go to the if side of things. In the `if` part we basically do not add the public key. If I just add `state.id.remote = remoteId` inside the if and after the validation condition, it works. Is this the expected?

vasco-santos: In noise, I could not follow the flow so easily yet

jacobheun: Note: this is fixed in secio now, I'll look into it in noise as part of my work there.

vasco-santos: what do you recommend here? change the test configuration to use secio for now and create an issue to track this?

src/peer-store/persistent/index.js
```diff
@@ -109,12 +111,15 @@ class PersistentPeerStore extends PeerStore {
     log('create batch commit')
     const batch = this._datastore.batch()
     for (const peerIdStr of commitPeers) {
-      // PeerId (replace by keyBook)
-      const peerId = this.peerIds.get(peerIdStr)
+      // PeerId
+      const peerId = this.keyBook.data.get(peerIdStr) || PeerId.createFromB58String(peerIdStr)
```
jacobheun, line 115:
```suggestion
      const peerId = this.keyBook.data.get(peerIdStr) || PeerId.createFromCID(peerIdStr)
```

test/peer-store/key-book.spec.js
```diff
@@ -0,0 +1,64 @@
...
+describe('keyBook', () => {
+  let peerId, peerStore, kb
+
+  beforeEach(async () => {
+    [peerId] = await peerUtils.createPeerId()
+    peerStore = new PeerStore()
+    kb = peerStore.keyBook
+  })
+
+  it('throwns invalid parameters error if invalid PeerId is provided in set', () => {
```
jacobheun, line 26:
```suggestion
  it('throws invalid parameters error if invalid PeerId is provided in set', () => {
```

test/peer-store/key-book.spec.js
```diff
@@ -0,0 +1,64 @@
...
+    try {
+      kb.set('invalid peerId')
+    } catch (err) {
+      expect(err.code).to.equal(ERR_INVALID_PARAMETERS)
+      return
+    }
+    throw new Error('invalid peerId should throw error')
+  })
+
+  it('throwns invalid parameters error if invalid PeerId is provided in get', () => {
```
jacobheun, line 36:
```suggestion
  it('throws invalid parameters error if invalid PeerId is provided in get', () => {
```

test/peer-store/key-book.spec.js
```diff
@@ -0,0 +1,64 @@
...
+  })
+
+  it('throwns invalid parameters error if invalid PeerId is provided in set', () => {
+    try {
+      kb.set('invalid peerId')
+    } catch (err) {
+      expect(err.code).to.equal(ERR_INVALID_PARAMETERS)
+      return
+    }
+    throw new Error('invalid peerId should throw error')
```
jacobheun, line 33: No need to change but you could instead do something like:
```js
expect(() => kb.set('invalid peerId')).to.throw()
      .that.has.property('code', ERR_INVALID_PARAMETERS)
```
which avoids potentially forgetting the return/throw statements when a failure occurs.

test/peer-store/peer-store.node.js
```diff
@@ -0,0 +1,51 @@
...
+    const remoteMultiaddr = `${remoteLibp2p.multiaddrs[0]}/p2p/${remoteIdStr}`
+    const conn = await libp2p.dial(remoteMultiaddr)
+
+    expect(conn).to.exist()
+    expect(spyAddressBook).to.have.property('called', true)
+    expect(spyKeyBook).to.have.property('called', true)
+
+    const localPeers = libp2p.peerStore.peers
+    expect(localPeers.size).to.equal(1)
+    // const publicKeyInLocalPeer = localPeers.get(remoteIdStr).id.pubKey
```
jacobheun, line 42: Is there an issue to track this? Otherwise just add a TODO in the comments here.

src/peer-store/README.md
jacobheun: Leave it for now, I'll work on getting a patch for noise to correct this. We can add a TODO item to the peer store epic to track getting this in.

test/peer-store/peer-store.node.js
vasco-santos: Adding a TODO with an issue in libp2p-noise

test/peer-store/key-book.spec.js
vasco-santos: I tried this out and I got:

```
AssertionError: expected [Function] to have property 'code'
```

I will skip this for now, we can revisit later

PR #630, opened 2020-05-06
docs: libp2p components options specified

vasco-santos:
This adds a specification for the properties of each libp2p component options

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -80,16 +80,16 @@ Creates an instance of Libp2p.
 | Name | Type | Description |
 |------|------|-------------|
 | options | `object` | libp2p options |
-| options.modules | `Array<object>` | libp2p modules to use |
+| options.modules | [`Array<object>`](./CONFIGURATION.md#modules) | libp2p modules to use |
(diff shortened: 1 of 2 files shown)


doc/API.md
```diff
@@ -66,15 +66,15 @@ Creates an instance of Libp2p.
 | Name | Type | Description |
 |------|------|-------------|
 | options | `object` | libp2p options |
-| options.modules | `Array<object>` | libp2p modules to use |
+| options.modules | [`Array<object>`](./CONFIGURATION.md#modules) | libp2p modules to use |
 | [options.addresses] | `{ listen: Array<string>, announce: Array<string>, noAnnounce: Array<string> }` | Addresses for transport listening and to advertise to the network |
 | [options.config] | `object` | libp2p modules configuration and core configuration |
-| [options.connectionManager] | `object` | libp2p Connection Manager configuration |
+| [options.connectionManager] | [`object`](./CONFIGURATION.md#configuring-connection-manager) | libp2p Connection Manager configuration |
```
vasco-santos, line 72: The properties for the connection manager now live within the `CONNECTION_MANAGER.md`. I think we should move them here to be coherent with the remaining ones. What do you think of this approach and the tables?

PR #631, opened 2020-05-07
feat: keychain

vasco-santos:
This PR is part of the [PeerStore improvements EPIC](https://github.com/libp2p/js-libp2p/issues/582) and tackles part of the milestone 4 - [KeyChain](https://github.com/libp2p/js-libp2p/issues/591).

Points to note:
- The entire libp2p-keychain history has been added
- Additional commits were made to move keychain under `./src/keychain` and its tests under `./test/keychain`.
- Some of the unneeded files were removed (like LICENSE) since libp2p already has a version of that file
- libp2p-keychain is now created on `Libp2p.create` and follows the same approach as `js-ipfs` did https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/src/core/components/init.js#L178-L257 (This will not be needed anymore in ipfs, you can also see this temporary commit with the new usage: https://github.com/ipfs/js-ipfs/commit/1a342c5dbc481596c214348f044d0ac09164d3fe)

## **Note on Merging**: 

Right now the plan (obtained from `libp2p-keychain` PR) is to:

- squash the merge to minimize the commit pollution
- update the readme to point to the original repo
- append the Co-Author list to the squashed commit to maintain author credit
  - This co author list was generated from the libp2p-keychain package.json which is auto generated from the commit history there.

```
Co-authored-by: Alan Shaw <alan.shaw@protocol.ai>
Co-authored-by: Alberto Elias <hi@albertoelias.me>
Co-authored-by: Alex Potsides <alex@achingbrain.net>
Co-authored-by: David Dias <daviddias.p@gmail.com>
Co-authored-by: Hugo Dias <mail@hugodias.me>
Co-authored-by: Jacob Heun <jacobheun@gmail.com>
Co-authored-by: Maciej Krüger <mkg20001@gmail.com>
Co-authored-by: Masahiro Saito <camelmasa@gmail.com>
Co-authored-by: Richard Schneider <makaretu@gmail.com>
Co-authored-by: Vasco Santos <vasco.santos@moxy.studio>
Co-authored-by: Vasco Santos <vasco.santos@ua.pt>
Co-authored-by: Victor Bjelkholm <victorbjelkholm@gmail.com>
```

This replaces #525 because the other PR was super outdated and complex to rebase at this point.

PR #634, opened 2020-05-13
chore: integrate libp2p-keychain into js-libp2p (#633)

vasco-santos:
Integrates the libp2p-keychain codebase into libp2p.keychain.

This is the second part of decoupling #631 and replaces it.


src/keychain/index.js
```diff
@@ -1,3 +1,469 @@
+/* eslint max-nested-callbacks: ["error", 5] */
 'use strict'
 
-module.exports = require('./keychain')
```
vasco-santos, line 3: content of `src/keychain/keychain` was moved here to be consistent with the rest of the libp2p codebase

doc/CONFIGURATION.md
```diff
@@ -422,6 +423,37 @@ const node = await Libp2p.create({
 })
 ```
 
+#### Setup with Keychain
+
+Libp2p allows you to setup a secure key chain to manage your keys. The keychain configuration object should have the following properties:
```
jacobheun, line 428:
```suggestion
Libp2p allows you to setup a secure keychain to manage your keys. The keychain configuration object should have the following properties:
```

src/config.js
```diff
@@ -17,6 +18,9 @@ const DefaultConfig = {
     maxDialsPerPeer: Constants.MAX_PER_PEER_DIALS,
     dialTimeout: Constants.DIAL_TIMEOUT
   },
+  keychain: {
+    datastore: new MemoryDatastore()
+  },
```
jacobheun, line 23: This is going to cause us to carry around the MemoryDatastore even if we don't use the keychain. I think we should just default this to null and check the datastore before creating the keychain. There's no real value in a memory datastore for the keychain.

vasco-santos: you are right! just removed it and added a validation. In the docs it is already defined as required

PR #638, opened 2020-05-15
feat: metadata book

### LINKED ISSUE #627: MetadataBook
## Overview

With the PeerStore improvements work for `libp2p@0.28` #582 , we currently have the AddressBook, ProtoBook and KeyBook instead of the old mapping between peer identifiers and `PeerInfo`. In this context, we will also be working on the **MetadataBook**.

The **MetadataBook** will allow libp2p users to associate data to a specific peer. This data should also be persisted, so that users can rely on them after a node restart. Examples of this data can be peer nicknames, roles in the network, associated apps, etc

## Implementation Design

Taking into account the generic purpose use of this book, we need to think about how this data will be stored. Since one peer can have n metadata entries, we would need to have the following type of data structure in the MetadataBook:

`Map<string, Map<string, *>>`

This is, a map of peer identifier strings to their metadata, which should also be a Map indexed by a string identifier of the metadata and the content associated to that metadata to store.

In the datastore side of things, we will be able to use the following as key:

`/peers/metadata/<b32 peer id no padding>/<key>`

The metadata data will be stored as ... [TO BE COMPLETED, SEE CONCERNS]

## API

[TO BE COMPLETED, SEE CONCERNS]

### Stored Data

While the "in-memory" book and the datastore key are straightforward, the data we can store might rise some issues. Regarding the "in-memory" book, users would be able to just add any type of content to the store and get it, which would be completely transparent to the users. However, considering that we will need to persist this data to offer a bigger value to the users, we need to know how this data is structured, in order to properly store it in the datastore by marshalling and unmarshalling it on retrieve.

`go-libp2p` PeerStore MetadataBook accepts any type of object by using the [encoding/gob package](https://golang.org/pkg/encoding/gob). This allows them to just marshal any kind of received data for storing it, and unmarshal it for the users.

In JS land, we can potentially use a module like [surrial](https://github.com/nicojs/node-surrial) to offer the same capabilities. This module adds [1.3kB](https://bundlephobia.com/result?p=surrial@2.0.2) to the bundle being minified and gzipped.

Since most users will probably not use this feature, I am reluctant on adding a module for this specific use case. So, other alternatives might be to require users to provide a Buffer to store the data (and users can use this type of module for the marshalling without being part of the libp2p bundle), or to simply accept strings, which we can convert into Buffer easily. Users would also be able to simply convert JS objects into strings and provide it (same for buffers).

### API

While for the AddressBook, ProtoBook and KeyBook are similar to JS Maps regarding their API, the MetadataBook will probably not be the case.

The metadata book will have 3 inputs: `PeerId`, `MetadataKey`, `MetadataValue`, instead of 2 like the remaining ones.

- `peerStore.metadataBook.set(peerId: PeerId, key: string, value: *)`
- `peerStore.metadataBook.get(peerId: PeerId)` => Get all metadata for the Peer
- `peerStore.metadataBook.getValue(peerId: PeerId, key: string)` => Get a specific part of the metadata
- `peerStore.metadataBook.delete(peerId: PeerId)` => Delete all peer metadata from the book
- `peerStore.metadataBook.deleteValue(peerId: PeerId, key: string)` => Delete part of the metadata of a peer from the book.

vasco-santos:
This PR adds the metadataBook to the PeerStore as described in #627 

Closes #627

### DIFF
--- a/src/peer-store/index.js
+++ b/src/peer-store/index.js
@@ -76,27 +76,12 @@ class PeerStore extends EventEmitter {
    * @returns {Map<string, Peer>}
    */
   get peers () {
-    const storedPeers = new Set()
-
-    // AddressBook
-    for (const [idStr] of this.addressBook.data.entries()) {
-      storedPeers.add(idStr)
-    }
-
-    // ProtoBook
-    for (const [idStr] of this.protoBook.data.entries()) {
-      storedPeers.add(idStr)
-    }
-
-    // KeyBook
-    for (const [idStr] of this.protoBook.data.entries()) {
-      storedPeers.add(idStr)
-    }
-
-    // MetadataBook
-    for (const [idStr] of this.metadataBook.data.entries()) {
-      storedPeers.add(idStr)
-    }
+    const storedPeers = new Set([
+      ...this.addressBook.data.keys(),
+      ...this.keyBook.data.keys(),
+      ...this.protoBook.data.keys(),
+      ...this.metadataBook.data.keys()
+    ])
 
     const peersData = new Map()
     storedPeers.forEach((idStr) => {
@@ -135,7 +120,7 @@ class PeerStore extends EventEmitter {
     const metadata = this.metadataBook.get(peerId)
     const protocols = this.protoBook.get(peerId)
 
-    if (!addresses && !metadata && !protocols) {
+    if (!id && !addresses && !metadata && !protocols) {
       return undefined
     }
 
--- a/src/peer-store/index.js
+++ b/src/peer-store/index.js
@@ -76,61 +76,32 @@ class PeerStore extends EventEmitter {
    * @returns {Map<string, Peer>}
    */
   get peers () {
-    const peersData = new Map()
+    const storedPeers = new Set()
 
     // AddressBook
-    for (const [idStr, addresses] of this.addressBook.data.entries()) {
-      const id = this.keyBook.data.get(idStr) || PeerId.createFromCID(idStr)
-      peersData.set(idStr, {
-        id,
-        addresses,
-        metadata: this.metadataBook.get(id),
-        protocols: this.protoBook.get(id) || []
-      })
+    for (const [idStr] of this.addressBook.data.entries()) {
+      storedPeers.add(idStr)
     }
 
     // ProtoBook
-    for (const [idStr, protocols] of this.protoBook.data.entries()) {
-      const pData = peersData.get(idStr)
-      const id = this.keyBook.data.get(idStr) || PeerId.createFromCID(idStr)
-
-      if (!pData) {
-        peersData.set(idStr, {
-          id,
-          addresses: [],
-          metadata: this.metadataBook.get(id),
-          protocols: Array.from(protocols)
-        })
-      }
+    for (const [idStr] of this.protoBook.data.entries()) {
+      storedPeers.add(idStr)
     }
 
     // KeyBook
-    for (const [idStr, peerId] of this.protoBook.data.entries()) {
-      const pData = peersData.get(idStr)
-
-      if (!pData) {
-        peersData.set(idStr, {
-          id: peerId,
-          addresses: [],
-          metadata: this.metadataBook.get(peerId),
-          protocols: []
-        })
-      }
+    for (const [idStr] of this.protoBook.data.entries()) {
+      storedPeers.add(idStr)
     }
 
     // MetadataBook
-    for (const [idStr, metadata] of this.metadataBook.data.entries()) {
-      const pData = peersData.get(idStr)
-
-      if (!pData) {
-        peersData.set(idStr, {
-          id: PeerId.createFromCID(idStr),
-          addresses: [],
-          metadata,
-          protocols: []
-        })
-      }
+    for (const [idStr] of this.metadataBook.data.entries()) {
+      storedPeers.add(idStr)
     }
+
+    const peersData = new Map()
+    storedPeers.forEach((idStr) => {
+      peersData.set(idStr, this.get(PeerId.createFromCID(idStr)))
+    })
 
     return peersData
   }
--- a/src/peer-store/index.js
+++ b/src/peer-store/index.js
@@ -10,6 +10,7 @@ const PeerId = require('peer-id')
 
 const AddressBook = require('./address-book')
 const KeyBook = require('./key-book')
+const MetadataBook = require('./metadata-book')
 const ProtoBook = require('./proto-book')
 
 const {
@@ -21,6 +22,8 @@ const {
  * @fires PeerStore#peer Emitted when a new peer is added.
  * @fires PeerStore#change:protocols Emitted when a known peer supports a different set of protocols.
  * @fires PeerStore#change:multiaddrs Emitted when a known peer has a different set of multiaddrs.
+ * @fires PeerStore#change:pubkey Emitted emitted when a peer's public key is known.
+ * @fires PeerStore#change:metadata Emitted when the known metadata of a peer change.
  */
 class PeerStore extends EventEmitter {
   /**
@@ -46,6 +49,11 @@ class PeerStore extends EventEmitter {
      * KeyBook containing a map of peerIdStr to their PeerId with public keys.
      */
     this.keyBook = new KeyBook(this)
+
+    /**
+     * MetadataBook containing a map of peerIdStr to their metadata Map.
+     */
+    this.metadataBook = new MetadataBook(this)
 
     /**
      * ProtoBook containing a map of peerIdStr to supported protocols.
@@ -76,6 +84,7 @@ class PeerStore extends EventEmitter {
       peersData.set(idStr, {
         id,
         addresses,
+        metadata: this.metadataBook.get(id),
         protocols: this.protoBook.get(id) || []
       })
     }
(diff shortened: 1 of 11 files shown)


src/peer-store/metadata-book.js
```diff
@@ -0,0 +1,165 @@
...
+   * @param {Buffer} value metadata value
+   * @returns {ProtoBook}
+   */
+  set (peerId, key, value) {
+    if (!PeerId.isPeerId(peerId)) {
+      log.error('peerId must be an instance of peer-id to store data')
+      throw errcode(new Error('peerId must be an instance of peer-id'), ERR_INVALID_PARAMETERS)
+    }
+
+    if (!key || !value || !(typeof key === 'string') || !Buffer.isBuffer(value)) {
```
jacobheun, line 60: This can be simplified. We could potentially just do a `Buffer.from` try/catch to support strings/arraybuffer/etc out of the box as well.
```suggestion
    if (typeof key !== 'string') || !Buffer.isBuffer(value)) {
```

src/peer-store/metadata-book.js
```diff
@@ -0,0 +1,165 @@
...
+   * @returns {Map<string, Buffer>}
+   */
+  get (peerId) {
+    if (!PeerId.isPeerId(peerId)) {
+      throw errcode(new Error('peerId must be an instance of peer-id'), ERR_INVALID_PARAMETERS)
+    }
+
+    const rec = this.data.get(peerId.toB58String())
+
+    return rec ? new Map(rec) : undefined
```
jacobheun, line 103: `rec` should already be a Map or undefined correct?
```suggestion
    return this.data.get(peerId.toB58String())
```

src/peer-store/metadata-book.js
```diff
@@ -0,0 +1,165 @@
...
+   */
+  getValue (peerId, key) {
+    if (!PeerId.isPeerId(peerId)) {
+      throw errcode(new Error('peerId must be an instance of peer-id'), ERR_INVALID_PARAMETERS)
+    }
+
+    const rec = this.data.get(peerId.toB58String())
+    const value = rec && rec.get(key)
+
+    return value
```
jacobheun, line 120:
```suggestion
    return rec && rec.get(key)
```

src/peer-store/persistent/index.js
```diff
@@ -58,6 +65,7 @@ class PersistentPeerStore extends PeerStore {
     this.on('change:protocols', this._addDirtyPeer)
     this.on('change:multiaddrs', this._addDirtyPeer)
     this.on('change:pubkey', this._addDirtyPeer)
+    this.on('change:metadata', this._addDirtyPeerMetadata)
```
jacobheun, line 68: Why don't we do the same thing as the other events and just mark the peer itself as dirty, then get the latest when we go to commit?

test/peer-store/metadata-book.spec.js
```diff
@@ -0,0 +1,380 @@
...
+    beforeEach(() => {
+      peerStore = new PeerStore()
+      mb = peerStore.metadataBook
+    })
+
+    afterEach(() => {
+      peerStore.removeAllListeners()
+    })
+
+    it('throwns invalid parameters error if invalid PeerId is provided', () => {
```
jacobheun, line 36:
```suggestion
    it('throws invalid parameters error if invalid PeerId is provided', () => {
```

test/peer-store/metadata-book.spec.js
```diff
@@ -0,0 +1,380 @@
...
+      try {
+        mb.set('invalid peerId')
+      } catch (err) {
+        expect(err.code).to.equal(ERR_INVALID_PARAMETERS)
+        return
+      }
+      throw new Error('invalid peerId should throw error')
+    })
+
+    it('throwns invalid parameters error if no key provided', () => {
```
jacobheun, line 46:
```suggestion
    it('throws invalid parameters error if no key provided', () => {
```

test/peer-store/metadata-book.spec.js
```diff
@@ -0,0 +1,380 @@
...
+      try {
+        mb.set(peerId)
+      } catch (err) {
+        expect(err.code).to.equal(ERR_INVALID_PARAMETERS)
+        return
+      }
+      throw new Error('no key provided should throw error')
+    })
+
+    it('throwns invalid parameters error if no value provided', () => {
```
jacobheun, line 56:
```suggestion
    it('throws invalid parameters error if no value provided', () => {
```

test/peer-store/metadata-book.spec.js
```diff
@@ -0,0 +1,380 @@
...
+      try {
+        mb.set(peerId, 'location')
+      } catch (err) {
+        expect(err.code).to.equal(ERR_INVALID_PARAMETERS)
+        return
+      }
+      throw new Error('no value provided should throw error')
+    })
+
+    it('throwns invalid parameters error if no value provided', () => {
```
jacobheun, line 66:
```suggestion
    it('throws invalid parameters error if value is not a buffer', () => {
```

test/peer-store/metadata-book.spec.js
```diff
@@ -0,0 +1,380 @@
...
+
+  describe('metadataBook.get', () => {
+    let peerStore, mb
+
+    beforeEach(() => {
+      peerStore = new PeerStore()
+      mb = peerStore.metadataBook
+    })
+
+    it('throwns invalid parameters error if invalid PeerId is provided', () => {
```
jacobheun, line 165:
```suggestion
    it('throws invalid parameters error if invalid PeerId is provided', () => {
```

test/peer-store/metadata-book.spec.js
```diff
@@ -0,0 +1,380 @@
...
+
+  describe('metadataBook.getValue', () => {
+    let peerStore, mb
+
+    beforeEach(() => {
+      peerStore = new PeerStore()
+      mb = peerStore.metadataBook
+    })
+
+    it('throwns invalid parameters error if invalid PeerId is provided', () => {
```
jacobheun, line 201:
```suggestion
    it('throws invalid parameters error if invalid PeerId is provided', () => {
```

test/peer-store/metadata-book.spec.js
```diff
@@ -0,0 +1,380 @@
...
+
+  describe('metadataBook.deleteValue', () => {
+    let peerStore, mb
+
+    beforeEach(() => {
+      peerStore = new PeerStore()
+      mb = peerStore.metadataBook
+    })
+
+    it('throwns invalid parameters error if invalid PeerId is provided', () => {
```
jacobheun, line 307:
```suggestion
    it('throws invalid parameters error if invalid PeerId is provided', () => {
```

test/peer-store/peer-store.spec.js
```diff
@@ -158,4 +158,57 @@ describe('peer-store', () => {
...
   })
+
+  describe('peerStore.peers', () => {
+    let peerStore
+
+    beforeEach(() => {
+      peerStore = new PeerStore()
+    })
+
+    it('returns peers if only have addresses', () => {
```
jacobheun, line 169:
```suggestion
    it('returns peers if only addresses are known', () => {
```

test/peer-store/peer-store.spec.js
```diff
@@ -158,4 +158,57 @@ describe('peer-store', () => {
...
+
+      const peerData = peers.get(peerIds[0].toB58String())
+      expect(peerData).to.exist()
+      expect(peerData.id).to.exist()
+      expect(peerData.addresses).to.have.lengthOf(1)
+      expect(peerData.protocols).to.have.lengthOf(0)
+      expect(peerData.metadata).to.not.exist()
+    })
+
+    it('returns peers if only have protocols', () => {
```
jacobheun, line 183: This test is an interesting case to consider for the future. If we don't have addresses for a peer we may want to do some datastore cleanup, however there is a case for multiaddrs expiring and then we search peer routing to find the latest record for the peer because it previously supported a certain protocol. We should create an issue to address datastore cleanup strategies for future iterations.

```suggestion
    it('returns peers if only protocols are known', () => {
```

test/peer-store/peer-store.spec.js
```diff
@@ -158,4 +158,57 @@ describe('peer-store', () => {
...
+
+      const peerData = peers.get(peerIds[0].toB58String())
+      expect(peerData).to.exist()
+      expect(peerData.id).to.exist()
+      expect(peerData.addresses).to.have.lengthOf(0)
+      expect(peerData.protocols).to.have.lengthOf(1)
+      expect(peerData.metadata).to.not.exist()
+    })
+
+    it('returns peers if only have metadata', () => {
```
jacobheun, line 197:
```suggestion
    it('returns peers if only metadata is known', () => {
```

src/peer-store/metadata-book.js
vasco-santos: I wanted to go with the Buffer only support because if a user sets metadata with a string, it will work, and then on get a. buffer will be returned... The code suggestion seems good to me

src/peer-store/persistent/index.js
vasco-santos: There is one issue with that approach, that I cannot circumvent unless I add the extra dirty record. 

While in the other books we know that if something changed and it is not in the book anymore, it is deleted, in the metadataBook is not that simple. So, for adding it is simple, but for deleting we would not know the metadata keys that were deleted (because they were simply gone from the book), which would result in metadata being persisted and not removed anymore. When we do the restart, that metadata would be around.

test/peer-store/peer-store.spec.js
vasco-santos: I agree and I will create that issue

src/peer-store/persistent/index.js
jacobheun: Fair enough, we could check the datastore keys and then delete the ones that we no longer have, but this might have some drawbacks to shutdown storage time (extra datastore round trips). We can punt this decision for now, as there's no user facing impact with this.

src/peer-store/index.js
```diff
@@ -89,11 +98,40 @@ class PeerStore extends EventEmitter {
...
+
+      if (!pData) {
+        peersData.set(idStr, {
+          id: PeerId.createFromCID(idStr),
+          addresses: [],
+          metadata,
+          protocols: []
+        })
+      }
+    }
```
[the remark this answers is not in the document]
vasco-santos: This got definitely cleaner! thanks :)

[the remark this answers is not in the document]
jacobheun: These are all maps so in theory you could simplify it further to:
```js
const storedPeers = new Set([
  ...this.addressBook.data.keys(),
  ...this.keyBook.data.keys(),
  ...this.protoBook.data.keys(),
  ...this.metadataBook.data.keys(),  
])
```

Also, protobook is listed twice and keybook is missing.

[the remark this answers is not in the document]
vasco-santos: That's true! Looking into it

PR #643, opened 2020-05-19
feat: support dial only on transport manager to tolerate errors

### LINKED ISSUE #575: Add configuration option for address listen error tolerance
## Problem
Currently if libp2p is provided a list of multiaddrs and there are no valid transports configured for those multiaddrs, libp2p will throw an error on startup and shutdown.

Additionally, if mulitaddrs for an existing transport are provided but listening fails for all multiaddrs for that transport, such as a webrtc-star/etc server being down, libp2p will thrown an error and shutdown.

## Solution
We should support the configuration of this tolerance. For some applications it could be perfectly acceptable for nodes to start in dial only mode if all multiaddrs failed for a given transport, or all transports. While libp2p is currently more aggressive with its errors to favor transparency of issues, it shouldn't be restricted to this.

Refs:
- https://github.com/libp2p/js-libp2p-websocket-star/issues/61
- https://github.com/ipfs/js-ipfs/issues/2779

vasco-santos:
This PR allows libp2p users to support dial only mode, which will make transport manager tolerate errors, if none of the multiaddrs provided were successful used for listening on the configured transports.

We can also have this in the `addresses` configuration, but I preferred to have it assigned to the correct component. Moreover, I default this to the current behaviour but I do not have a strong opinion if we should go this way, or invert the logic. Let me know your thoughts

Closes #575

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -528,6 +528,8 @@ const TCP = require('libp2p-tcp')
 const MPLEX = require('libp2p-mplex')
 const SECIO = require('libp2p-secio')
 
+const { FaultTolerance } = require('libp2p/src/transport-manager')}
+
 const node = await Libp2p.create({
   modules: {
     transport: [TCP],
@@ -535,7 +537,7 @@ const node = await Libp2p.create({
     connEncryption: [SECIO]
   },
   transportManager: {
-    supportDialOnly: true
+    faultTolerance: FaultTolerance.NO_FATAL
   }
 })
 ```
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -23,6 +23,7 @@
       - [Setup with Keychain](#setup-with-keychain)
       - [Configuring Dialing](#configuring-dialing)
       - [Configuring Connection Manager](#configuring-connection-manager)
+      - [Configuring Transport Manager](#configuring-transport-manager)
       - [Configuring Metrics](#configuring-metrics)
       - [Configuring PeerStore](#configuring-peerstore)
       - [Customizing Transports](#customizing-transports)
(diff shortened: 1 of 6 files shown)


doc/CONFIGURATION.md
```diff
@@ -517,6 +518,28 @@ const node = await Libp2p.create({
...
+const SECIO = require('libp2p-secio')
+
+const node = await Libp2p.create({
+  modules: {
+    transport: [TCP],
+    streamMuxer: [MPLEX],
+    connEncryption: [SECIO]
+  },
+  transportManager: {
+    supportDialOnly: true
```
jacobheun, line 538: We originally talked about this being a bit more flexible, https://github.com/libp2p/js-libp2p-websocket-star/issues/61#issuecomment-437425776. Perhaps we could change this to something like `faultTolerance`, and we could export an enum of values that could be used. For now we could just support `FATAL_ALL` (no transport was able to listen) and `NO_FATAL` (no listening is required).

vasco-santos: I didn't see this discussion. I agree with adding more flexibility :)
Should we go with `NO_FATAL` as the default?

jacobheun: `FATAL_ALL`, as that's the current behavior. In reality this should be different for browsers/node, but we can update that when we add base configuration support.

src/transport-manager.js
```diff
@@ -212,4 +218,11 @@ class TransportManager {
   }
 }
 
+const FAULT_TOLERANCE = {
```
jacobheun, line 221: We should add some jsdocs to describe each of these as they're not specified anywhere.

vasco-santos: I did not find any standard to document the Enum values. I added them in the main body description for now. If we increase this list, we should rethink this.

PR #645, opened 2020-05-23
fix(example): rename misleading variable

revolunet:


examples/pnet/libp2p-node.js
```diff
@@ -13,7 +13,7 @@ const Protector = require('libp2p/src/pnet')
  * @param {Buffer} swarmKey
```
vasco-santos, line 13: There is one more reference to `swarmKeyPath` in the docs above. Can you change it?

revolunet: oops, fixed

PR #646, opened 2020-05-26
fix: expose the muxed stream interface on inbound streams

jacobheun:
Inbound streams aren't exposing the muxed stream interface (id, timeline, close, etc). Outbound streams are already doing this.


vasco-santos:
Restarted CI, the branch was using interop cached tests

PR #660, opened 2020-06-09
feat: add ConnectionManager#getAll

wemeetagain:
The implementation is _almost_ identical to `ConnectionManager#get` except instead of returning the first open connection or null, `getAll` returns an array of _all_ open connections (or an empty array).

Perhaps `get` could be refactored to use `getAll` to avoid code duplication?

Open to any suggestions, naming, implementation, otherwise

See https://github.com/ChainSafe/js-libp2p-gossipsub/pull/83#discussion_r437279222

### DIFF
--- a/src/connection-manager/index.js
+++ b/src/connection-manager/index.js
@@ -230,6 +230,26 @@ class ConnectionManager extends EventEmitter {
   }
 
   /**
+   * Get all open connections with a peer.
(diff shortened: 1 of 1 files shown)


jacobheun:
This sounds good to me, switching `get` over to leverage `getAll` would be great.

PR #665, opened 2020-06-11
fix: throw if no conn encryption module provided

### LINKED ISSUE #664: no peer:connect event with autoDial
With libp2p V0.28,  peer:connect events seems not to be fired
```
const peerid = ...;
const libp2p = await Libp2p.create({
        peerId: peerid,
        addresses: {
          listen: ['/ip4/0.0.0.0/tcp/0'],
        },
        modules: {
          transport: [TCP],
          streamMuxer: [Mplex],
          peerDiscovery: [MulticastDNS],
        },
        config: {
          peerDiscovery: {
            autoDial: true,
            [MulticastDNS.tag]: { // or mdns: ?
              interval: 1000,
              enabled: true,
            },
          },
        },
      });

      libp2p.on('peer:discovery', (peerId) => {
        debug('Discovered: %O', peerId);
        // libp2p.dial(peerId);
      });

      libp2p.connectionManager.on('peer:connect', (connection) => { 
        debug('peer connected %O',connection);
      });

      libp2p.start();
```

With the same setup, if I launch `libp2p.start()` from another node, the `peer:discovery` is well logged, but not `peer:connect`. I thought the `autoDial:true` should make a peer dials other discovered peers automatically.
I've tried also to call `libp2p.dial(peerId)` into the peer:discovery event, but `peerId` has no `multiaddrs` value, and have not found how to dial a discovered peer.

The documentation is sometimes confusing with this discovery process (for example, some explains to use `mdns`, other tells to use `[MulticastDNS.tag]`in config)

vasco-santos:
The main focus of this PR is throwing an error if a connEncryption module is not provided.
With that in mind, it also improves the configuration docs visibility from the `Libp2p.create` API doc, as well as that transports and connEncryption modules are required.

It also changes the peerDiscovery example to use the discovery module tag and be coherent with the docs.

Fixes #664


test/insecure/configuration.spec.js
```diff
@@ -0,0 +1,57 @@
+'use strict'
+/* eslint-env mocha */
+
+const chai = require('chai')
+chai.use(require('dirty-chai'))
+chai.use(require('chai-as-promised'))
+const { expect } = chai
+
+const Transport = require('libp2p-websockets')
+const { NOISE: Crypto } = require('libp2p-noise')
```
jacobheun, line 10: This is in the `insecure` folder but is using noise. Did you intend to use plaintext for this? If not this should just get moved into core for now

src/errors.js
```diff
@@ -2,13 +2,15 @@
 
 exports.messages = {
   NOT_STARTED_YET: 'The libp2p node is not started yet',
-  DHT_DISABLED: 'DHT is not available'
+  DHT_DISABLED: 'DHT is not available',
+  CONN_ENCRYPTION_REQUIRED: 'The connection encryption module is required'
```
jacobheun, line 6:
```suggestion
  CONN_ENCRYPTION_REQUIRED: 'At least one connection encryption module is required'
```

test/insecure/configuration.spec.js
vasco-santos: I am moving it to core instead

achingbrain:
\*cough\* \*cough\* breaking change \*cough\* \*cough\*

jacobheun:
You're right, this is an api change. While this shouldn't break production applications as it is required for libp2p to function properly, it can break testing environments for people. I'll do a revert commit for this and move this to an 0.29 branch.

jacobheun:
Reverted in v0.28.2

jacobheun:
Branch [0.29.x](https://github.com/libp2p/js-libp2p/tree/0.29.x) has been created for the next minor release. The change has been reapplied for that branch.

PR #669, opened 2020-06-12
chore: use ed25519 keys in tests

vasco-santos:
Needs:

- [x] [libp2p/js-libp2p-crypto#145](https://github.com/libp2p/js-libp2p-crypto/issues/145)
- [x] [libp2p/js-peer-id#132](https://github.com/libp2p/js-peer-id/pull/132)


jacobheun:
This should now be unblocked in master.

vasco-santos:
I will work on rebase this and get it finished

vasco-santos:
There were some tests failing now, due to the persistent peer store, more precisely as a consequence of the following lines:

- https://github.com/libp2p/js-libp2p/blob/master/src/peer-store/persistent/index.js#L151
- https://github.com/libp2p/js-libp2p/blob/master/src/peer-store/persistent/index.js#L213

With ed25519 keys, if we create a peer from the CID, we can get its pubKey through the peerId.pubKey getter. As a consequence, we will have always pubKey. The cleanest solution here is to get from the keyBook again in: https://github.com/libp2p/js-libp2p/blob/master/src/peer-store/persistent/index.js#L213

vasco-santos:
On another note, this PR changes all the tests that use our fixtures (the majority of the tests atm) to use `ed25519` keys. For the remaining ones, the easiest solution is to change the `peer-id` create to use `ed25519` as a breaking change release. We don't need to do it now, as well, but I think it is easier (and perhaps more correct for future changes) to do that than manually change all the tests do use it instead of RSA

jacobheun:
> With ed25519 keys, if we create a peer from the CID, we can get its pubKey through the peerId.pubKey getter. As a consequence, we will have always pubKey. The cleanest solution here is to get from the keyBook again in

For inlined public keys there's no reason to store their public key, we should avoid doing this it's just a waste of space/datastore transactions. We may want to add an easy way of checking this in `peer-id`. Currently only rsa keys aren't inline.

vasco-santos:
Inline public keys check was added and tests updated accordingly

test/identify/index.spec.js
```diff
@@ -364,7 +364,7 @@ describe('Identify', () => {
       expect(libp2p.identifyService.identify.callCount).to.equal(1)
 
       // The connection should have no open streams
-      expect(connection.streams).to.have.length(0)
+      await pWaitFor(() => connection.streams.length === 0)
```
vasco-santos, line 367: Flakey test in Firefox where sometimes the stream was still closing at this moment

jacobheun (approved):

PR #673, opened 2020-06-16
Fix dialer.connectToPeer to throw a clean error

### LINKED ISSUE #672: Dialer.connectToPeer fails badly if no addrs
Version: package.json version or the commit you have installed.
Platform: output of `uname -a` (UNIX), or version and 32 or 64-bit (Windows). If using in a Browser, please share the browser version as well.
Subsystem: if known, please specify affected core module name (e.g Transports, SECIO, etc).

If possible, please provide code that demonstrates the problem, keeping it as
simple and free of external dependencies as you are able.
-->

- **Version**: 0.28.2
- **Platform**: Darwin Kernel Version 17.7.0: Wed May 27 17:00:02 PDT 2020; root:xnu-4570.71.80.1~1/RELEASE_X86_64 x86_64
- **Subsystem**: libp2p/dialer

#### Description:
calling Dialer.connectToPeer with a peer that has no addresses fails with 
> Unhandled Rejection (TypeError): Cannot read property 'length' of undefined

```
const peer = new PeerId(Buffer.from("QmfEa1LrSt2R7KiZ31yPo3TT8ipfyDfrMRkaQix3rcXXXX"))
await dialer.connectToPeer(peer)
```

dthorpe:
when dialTarget.addrs is undefined

Fixes #672

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,7 @@
     "libp2p-tcp": "^0.14.1",
     "libp2p-webrtc-star": "^0.18.0",
     "libp2p-websockets": "^0.13.1",
+    "mocha": "^8.0.1",
     "multihashes": "^0.4.19",
     "nock": "^12.0.3",
     "p-defer": "^3.0.0",
--- a/src/dialer/index.js
+++ b/src/dialer/index.js
@@ -70,7 +70,7 @@ class Dialer {
   async connectToPeer (peer, options = {}) {
     const dialTarget = this._createDialTarget(peer)
 
-    if (!dialTarget.addrs.length) {
+    if (!dialTarget.addrs || !dialTarget.addrs.length) {
       throw errCode(new Error('The dial request has no addresses'), codes.ERR_NO_VALID_ADDRESSES)
     }
     const pendingDial = this._pendingDials.get(dialTarget.id) || this._createPendingDial(dialTarget, options)
--- a/test/dialing/direct.node.js
+++ b/test/dialing/direct.node.js
@@ -96,6 +96,14 @@ describe('Dialing (direct, TCP)', () => {
       .and.to.have.nested.property('._errors[0].code', ErrorCodes.ERR_TRANSPORT_UNAVAILABLE)
   })
 
+  it('should fail appropriately if peer has no addresses', async () => {
+    const dialer = new Dialer({ transportManager: localTM, peerStore })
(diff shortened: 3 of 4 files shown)


test/dialing/direct.node.js
```diff
@@ -96,6 +96,14 @@ describe('Dialing (direct, TCP)', () => {
       .and.to.have.nested.property('._errors[0].code', ErrorCodes.ERR_TRANSPORT_UNAVAILABLE)
   })
 
+  it('should fail appropriately if peer has no addresses', async () => {
+    const dialer = new Dialer({ transportManager: localTM, peerStore })
+    const peer = new PeerId(Buffer.from('QmfEa1LrSt2R7KiZ31yPo3TT8ipfyDfrMRkaQix3rcXeTo'))
```
vasco-santos, line 101: Can you use the `createPeerId` function already required to create the `PeerId`? We try to use that because it creates a PeerId from a JSON which already have everything computed and the tests will be faster to run.

package.json
```diff
@@ -111,6 +111,7 @@
     "libp2p-tcp": "^0.14.1",
     "libp2p-webrtc-star": "^0.18.0",
     "libp2p-websockets": "^0.13.1",
+    "mocha": "^8.0.1",
```
vasco-santos, line 114: What is the reason to require mocha here?

src/dialer/index.js
```diff
@@ -70,7 +70,7 @@ class Dialer {
   async connectToPeer (peer, options = {}) {
     const dialTarget = this._createDialTarget(peer)
 
-    if (!dialTarget.addrs.length) {
+    if (!dialTarget.addrs || !dialTarget.addrs.length) {
```
vasco-santos, line 73: What do you think on doing the following in the `this._createDialTarget`?

```js
let addrs = this.peerStore.addressBook.getMultiaddrsForPeer(id) || []
```

I think it would be cleaner.

test/dialing/direct.node.js
dthorpe: k

package.json
dthorpe: seemed to be required to run the unit tests in intellij.  🤷

vasco-santos: This is strange. I do not have intellij to test this out, but running the tests in the CLI does not require `mocha` as we are not using it

vasco-santos:
@dthorpe will you be able to finish this review, or should I take it?

Issue #653, opened 2020-06-02
Signed Peer Records


vasco-santos:
https://github.com/libp2p/specs/pull/217

## Background

A libp2p node shares its data publicly (e.g through the DHT) or relies on potentially untrustworthy intermediaries to relay information. This might have consequences as the peer data can be tempered by a third party.

A Signed Peer Record is an all-purpose data container that includes a signature of the data, so we can verify that the data came from a specific peer and that it hasn't been tampered with.

## Milestones

| Milestone | Issue | PR |
|----------|:-------------:|------:|
| `1) Implementation of Signed Peer Records` |  N/A | #681 |
| `2) Identify protocols use Signed Peer Records` |  N/A | #682  |
| `3) Certified Address Book` |  N/A | #683 |
| `4) Pubsub routers use Signed Peer Records` |  N/A | TODO |

jacobheun:
Done and released in 0.29!

PR #681, opened 2020-06-21
feat: signed peer records data types

vasco-santos:
This PR adds the Libp2p Envelope and Peer Record data type for #653 .

## Needs:

- [x] [libp2p/js-libp2p-interfaces#52](https://github.com/libp2p/js-libp2p-interfaces/pull/52)
- [x] [libp2p/js-libp2p-utils#8](https://github.com/libp2p/js-libp2p-utils/pull/8)

## Solution design 

A `Record` interface was created [libp2p/js-libp2p-interfaces#52](https://github.com/libp2p/js-libp2p-interfaces/pull/52) was the base for any `Record` compatible with a Libp2p `Envelope`. A `record` implementation should run the interface tests to guarantee its compatibility with the Envelope.

A Libp2p `Envelope` will be a container for a record implementing the `Record` interface. It will include the record marshalled data as a `payload` property, as well as a signature and a payloadType for easing the decoding in the future, when we have multiple records differently serialised.

A `PeerRecord` was created implementing the `Record` interface. This record will be using for many libp2p protocols that exchange peers' listen multiaddrs, such as `Identify`, `Gossipsub1.1`, `DHT`, `Rendezvous`, etc. The protocols should support the legacy protocol for the time being.

## What is missing?

The envelope signature generate still is being used as a placeholder and not compliant with the spec. This should be solved for the creation of the module. There is no js implementation for the uvarint spec and I will work on it (there is one that I need to cross check with the spec).

## What's next?

I expect to create the follow up PRs

1. A libp2p node should exchange a signed peer record in the Identify protocol (identify + identify-push), so that we find needed changes for the RecordManager/Envelope implementations.
2. Create certified AddressBook / Adapt AddressBook to store signed peer records data. Update API.md
3. Interop tests for the identify protocol using the new protocol.
4. Integrate signed Peer Records in Pubsub
5. Integrate signed Peer Records in the DHT
6. Integrate signed Peer Records in the Rendezvous

## References

- [libp2p/specs#217](https://github.com/libp2p/specs/pull/217)
- https://github.com/libp2p/go-libp2p-core/blob/master/record/record.go
- https://github.com/libp2p/go-libp2p-core/blob/master/peer/record.go
- https://github.com/libp2p/go-libp2p-core/blob/master/record/envelope.go


vasco-santos:
@jacobheun can I have an initial feedback on this implementation design?

Once we are aligned, I plan to:
- Create `peer-record` module and move the code there
- Create `libp2p-envelope` module, move the code there and make the signature compliant with the spec
- Get this PR using the modules and ready to be merged

Follow up draft PRs for refactoring the Identify Protocol and CertifiedAddressBook for initial feedback should land during the next couple of days

vasco-santos:
> This PR is for the support of Peer Routing Records. With that in mind I'd like to clarify the flow, the API specifically, of how these records are created and fetched, as this isn't very clear.

Yes, this is an important thing to discuss. I created #682 so that you could see it in practise.
I started by creating the Envelope and the PeerRecord inside js-libp2p to avoid the mountain of extra modules we have. But then, I gave more thought to it and I did not like how other subsystems would need to use this we keep this in here.

While the `identify` protocol it is straightforward, as it also lives in this repo, other modules like gossipsub/dht would need to require the modules as follows:

```js
const Envelope = require('libp2p/src/...')
const PeerRecord = require('libp2p/src/...')
```

or libp2p can provide some API methods to access them without going this way.
The other solution would be to have them published as their own modules.

For now, the open PRs are going on the first solution, but I personally don't like it.

----

Regarding its usability, I expect the following:

### Create own peer record

Once the listen addresses are known, it should create a PeerRecord instance, and use `Envelope.seal()` to create an envelope wrapping the record. This enveloped should be cached to be sent anytime, and should be updated when necessary.

### Send own peer record

Get a cached signed peer record though `libp2p.x()` and send it through the wire according to the protocol being used.

### Receive peer record

A protocol receives the envelope from the wire protocol and it should use `Envelope.openAndCertify()` to open the envelope and verify the signatures. If the envelope is valid, the envelope should be added to a Certified AddressBook (still looking on this part).

### Send other party peer record

Get it from the AddressBook

src/record-manager/envelope/envelope.proto.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
+
+const protons = require('protons')
+
+const message = `
+message Envelope {
```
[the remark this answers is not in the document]
vasco-santos: Maybe also report in the spec? I think the `SignedEnvelope` naming is redundant, as the Envelope purpose is to have the signature

vasco-santos:
Me and @jacobheun synced yesterday and we will do as follows:

- Envelope and Libp2p Record will live inside the `js-libp2p` repo
- Identify will be responsible for creating the self record
  - Self record stored in the PeerStore AddressBook
  - Identify gets it from the PeerStore (if it exists, otherwise identify creates it and stores it in the PeerStore)
  - Identify verifies if it is updated (do we really need this? let's discuss further in the identify PR)
  - Identify sends the self peer record
  - FUTURE: Identify listens for address changes and creates the new PeerRecord on identify.push / delta

--

Since we do not need the RecordManager anymore, at least until having more records in libp2p, the scope of this PR will change to have the implementations of the Envelope and Peer Record. #682 will create self and exchange it in the Identify Protocol. Follow up PR will handle the peer records storage in the PeerStore.

src/record-manager/envelope/index.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+* Seal marshals the given Record, places the marshaled bytes inside an Envelope
+* and signs with the given private key.
+* @async
+* @param {Record} record
+* @param {PeerId} peerId
+* @return {Envelope}
+*/
+exports.seal = async (record, peerId) => {
```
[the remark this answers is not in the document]
vasco-santos: We are currently doing the same in the `peer-id`. Anyway, I like that we move this like you suggest and also in the Record, as in the record we could also add it to the interface validations (`createFromProtobuf`)

src/record/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+Libp2p provides an all-purpose data container called **envelope**. It was created to enable the distribution of verifiable records, which we can prove originated from the addressed peer itself. The envelope includes a signature of the data, so that its authenticity is verified.
+
+This envelope stores a marshaled record implementing the [interface-record](https://github.com/libp2p/js-libp2p-interfaces/tree/master/src/record). These Records are designed to be serialized to bytes and placed inside of the envelopes before being shared with other peers.
+
+You can read further about the envelope in [libp2p/specs#217](https://github.com/libp2p/specs/pull/217).
+
+### Usage
+
+- create an envelope with an instance of an `interface-record` implementation and prepare it for being exchanged:
```
vasco-santos, line 15: TODO: link `interface-record` once merged

src/record-manager/envelope/envelope.proto.js
```diff
@@ -0,0 +1,25 @@
...
+  // deterministically.
+  bytes payload_type = 2;
+
+  // payload is the actual payload carried inside this envelope.
+  bytes payload = 3;
+
+  // signature is the signature produced by the private key corresponding to
+  // the enclosed public key, over the payload, prefixing a domain string for
+  // additional security.
+  bytes signature = 5;
```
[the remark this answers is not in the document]
vasco-santos: Should we already change it, or wait for go?

src/record/peer-record/index.js
```diff
@@ -0,0 +1,98 @@
...
+const PeerId = require('peer-id')
+const Record = require('libp2p-interfaces/src/record')
+
+const Protobuf = require('./peer-record.proto')
+const {
+  ENVELOPE_DOMAIN_PEER_RECORD,
+  ENVELOPE_PAYLOAD_TYPE_PEER_RECORD
+} = require('./consts')
+
+const arraysAreEqual = (a, b) => a.length === b.length && a.sort().every((item, index) => b[index].equals(item))
```
jacobheun, line 13: This might be worth adding to a utility file in multiaddr or libp2p-utils. Equality, unique, etc are used in other places.

src/record/peer-record/index.js
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * The PeerRecord is responsible for TODOTODOTRDO
```
jacobheun, line 16: Probably don't need to define responsibility here. What it is would be helpful though.

src/record/peer-record/index.js
```diff
@@ -0,0 +1,98 @@
...
+ */
+class PeerRecord extends Record {
+  /**
+   * @constructor
+   * @param {object} params
+   * @param {PeerId} params.peerId
+   * @param {Array<multiaddr>} params.multiaddrs public addresses of the peer this record pertains to.
```
jacobheun, line 23: These are not necessarily public. LAN peers could use this with local addresses.
```suggestion
   * @param {Array<multiaddr>} params.multiaddrs addresses of the associated peer
```

src/record/peer-record/index.js
```diff
@@ -0,0 +1,98 @@
...
+
+    return this._marshal
+  }
+
+  /**
+   * Verifies if the other PeerRecord is identical to this one.
+   * @param {Record} other
+   * @return {boolean}
+   */
+  isEqual (other) {
```
jacobheun, line 62: Change to `equals`

src/record/peer-record/index.js
```diff
@@ -0,0 +1,98 @@
...
+      addresses: this.multiaddrs.map((m) => ({
+        multiaddr: m.buffer
+      }))
+    })
+
+    return this._marshal
+  }
+
+  /**
+   * Verifies if the other PeerRecord is identical to this one.
```
jacobheun, line 58:
```suggestion
   * Returns true if `this` record equals the `other`
```

src/record/peer-record/index.js
```diff
@@ -0,0 +1,98 @@
...
+      return false
+    }
+
+    // Validate seqNumber
+    if (this.seqNumber !== other.seqNumber) {
+      return false
+    }
+
+    // Validate multiaddrs
+    if (this.multiaddrs.length !== other.multiaddrs.length || !arraysAreEqual(this.multiaddrs, other.multiaddrs)) {
```
jacobheun, line 74: The utility function already checks length.

```suggestion
    if (!arraysAreEqual(this.multiaddrs, other.multiaddrs)) {
```

src/record/README.md
```diff
@@ -0,0 +1,114 @@
...
+} catch (err) {}
+
+const rec = Record.createFromProtobuf(e.payload)
+```
+
+## Peer Record
+
+All libp2p nodes keep a `PeerStore`, that among other information stores a set of known addresses for each peer, which can come from a variety of sources.
+
+Libp2p peer records were created to enable the distribution of verifiable address records, which we can prove originated from the addressed peer itself. With such guarantees, libp2p can prioritize addresses based on their authenticity, with the most strict strategy being to only dial certified addresses.
```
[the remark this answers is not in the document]
vasco-santos: Yeah, we still do not support any. I added: "(no strategies implemented at the time of writing)"

src/record/envelope/index.js
```diff
@@ -0,0 +1,174 @@
...
+}
+
+/**
+ * Helper function that prepares a buffer to sign or verify a signature.
+ * @param {string} domain
+ * @param {Buffer} payloadType
+ * @param {Buffer} payload
+ * @return {Buffer}
+ */
+const createSignData = (domain, payloadType, payload) => {
```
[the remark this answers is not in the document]
vasco-santos: Yeah, I did not like this naming at the time, but forgot to get a better one! Thanks for the suggestion!

src/record/envelope/index.js
```diff
@@ -0,0 +1,174 @@
...
+  return new Envelope({
+    peerId,
+    payloadType,
+    payload,
+    signature
+  })
+}
+
+/**
+ * Open and certify a given marshalled envelope.
```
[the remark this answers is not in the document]
vasco-santos: This will unmarshal the envelope. The function receives a marshalled envelope

src/record/peer-record/index.js
vasco-santos: https://github.com/libp2p/js-libp2p-utils/pull/8

I will add the others ones on a better opportunity to not get out of the scope of the PR

src/record/README.md
```diff
@@ -0,0 +1,129 @@
...
+} catch (err) {}
+
+const rec = ExampleRecord.createFromProtobuf(e.payload)
+```
+
+## Peer Record
+
+All libp2p nodes keep a `PeerStore`, that among other information stores a set of known addresses for each peer, which can come from a variety of sources.
+
+Libp2p peer records were created to enable the distribution of verifiable address records, which we can prove originated from the addressed peer itself. With such guarantees, libp2p is able to prioritize addresses based on their authenticity, with the most strict strategy being to only dial certified addresses (no strategies implemented at the time of writing).
```
jacobheun, line 63:
```suggestion
Libp2p peer records were created to enable the distribution of verifiable address records, which we can prove originated from the addressed peer itself. With such guarantees, libp2p is able to prioritize addresses based on their authenticity, with the most strict strategy being to only dial certified addresses (no strategies have been implemented at the time of writing).
```

PR #682, opened 2020-06-22
feat: exchange signed peer records in identify

vasco-santos:
This PR adds the exchange of signed peer records in identify as part of #653 

Needs:

- [x] #681 

In the context of using signed peer records on identify, the protocols were updated

For the time being, `js-libp2p` will fallback to the legacy protocols if the negotiated stream is using them, so that we stil have compatibility with peers using older versions of libp2p.

This PR will have a follow up PR to store the certified multiaddrs properly, as the multiaddrs are still being stored in the AddressBook as a regular multiaddr discovered.

### DIFF
--- a/src/identify/consts.js
+++ b/src/identify/consts.js
@@ -2,5 +2,9 @@
 
 module.exports.PROTOCOL_VERSION = 'ipfs/0.1.0'
 module.exports.AGENT_VERSION = 'js-libp2p/0.1.0'
-module.exports.MULTICODEC_IDENTIFY = '/ipfs/id/1.0.0'
-module.exports.MULTICODEC_IDENTIFY_PUSH = '/ipfs/id/push/1.0.0'
+module.exports.MULTICODEC_IDENTIFY = '/p2p/id/1.1.0'
+module.exports.MULTICODEC_IDENTIFY_PUSH = '/p2p/id/push/1.1.0'
+
+// Legacy
+module.exports.MULTICODEC_IDENTIFY_LEGACY = '/ipfs/id/1.0.0'
+module.exports.MULTICODEC_IDENTIFY_PUSH_LEGACY = '/ipfs/id/push/1.0.0'
--- a/src/identify/consts.js
+++ b/src/identify/consts.js
@@ -2,9 +2,5 @@
 
 module.exports.PROTOCOL_VERSION = 'ipfs/0.1.0'
 module.exports.AGENT_VERSION = 'js-libp2p/0.1.0'
-module.exports.MULTICODEC_IDENTIFY = '/p2p/id/1.1.0'
-module.exports.MULTICODEC_IDENTIFY_PUSH = '/p2p/id/push/1.1.0'
-
-// Legacy
-module.exports.MULTICODEC_IDENTIFY_1_0_0 = '/ipfs/id/1.0.0'
-module.exports.MULTICODEC_IDENTIFY_PUSH_1_0_0 = '/ipfs/id/push/1.0.0'
+module.exports.MULTICODEC_IDENTIFY = '/ipfs/id/1.0.0'
+module.exports.MULTICODEC_IDENTIFY_PUSH = '/ipfs/id/push/1.0.0'
--- a/src/identify/consts.js
+++ b/src/identify/consts.js
@@ -6,5 +6,5 @@ module.exports.MULTICODEC_IDENTIFY = '/p2p/id/1.1.0'
 module.exports.MULTICODEC_IDENTIFY_PUSH = '/p2p/id/push/1.1.0'
 
 // Legacy
-module.exports.MULTICODEC_IDENTIFY_LEGACY = '/ipfs/id/1.0.0'
-module.exports.MULTICODEC_IDENTIFY_PUSH_LEGACY = '/ipfs/id/push/1.0.0'
+module.exports.MULTICODEC_IDENTIFY_1_0_0 = '/ipfs/id/1.0.0'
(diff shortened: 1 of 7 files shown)


src/identify/consts.js
```diff
@@ -2,5 +2,9 @@
 
 module.exports.PROTOCOL_VERSION = 'ipfs/0.1.0'
 module.exports.AGENT_VERSION = 'js-libp2p/0.1.0'
-module.exports.MULTICODEC_IDENTIFY = '/ipfs/id/1.0.0'
-module.exports.MULTICODEC_IDENTIFY_PUSH = '/ipfs/id/push/1.0.0'
+module.exports.MULTICODEC_IDENTIFY = '/p2p/id/1.1.0'
+module.exports.MULTICODEC_IDENTIFY_PUSH = '/p2p/id/push/1.1.0'
```
jacobheun, line 6: Where are the versions listed, I was having trouble finding the updates. Is this in the specs repo or in go-libp2p?

src/identify/index.js
```diff
@@ -172,8 +183,40 @@ class IdentifyService {
     // Get the observedAddr if there is one
     observedAddr = IdentifyService.getCleanMultiaddr(observedAddr)
 
+    // LEGACY: differentiate message with SignedPeerRecord
+    if (protocol === MULTICODEC_IDENTIFY_1_0_0) {
```
jacobheun, line 187: No need to check the protocol, I'd just check the existence of `signedPeerRecord`. I think it might be better to catch and log the errors unmarshalling the signed peer record and then fallback to adding the `listenAddrs`. If for some reason the payload is malformed this will allow us to easily fallback.

src/identify/index.js
```diff
@@ -258,17 +306,73 @@ class IdentifyService {
...
+        this.peerStore.addressBook.set(id, message.listenAddrs.map((addr) => multiaddr(addr)))
+      } catch (err) {
+        return log.error('received invalid listen addrs', err)
+      }
+
+      // Update the protocols
+      this.peerStore.protoBook.set(id, message.protocols)
+
+      return
+    }
```
jacobheun, line 323: Again, I'd try the record first to allow fallback to happen if it exists but can't be unmarshalled.

src/identify/index.js
```diff
@@ -258,17 +306,73 @@ class IdentifyService {
...
     // Update the protocols
     this.peerStore.protoBook.set(id, message.protocols)
   }
+
+  /**
+   * Get self signed peer record raw envelope.
+   * @return {Buffer}
+   */
+  async _getSelfPeerRecord () {
+    // TODO: Verify if updated
```
jacobheun, line 361: Make it easier to search the codebase for this later?
```suggestion
    // TODO: support invalidation when dynamic multiaddrs are supported
```

src/identify/index.js
```diff
@@ -258,17 +306,73 @@ class IdentifyService {
...
 
     // Update the protocols
     this.peerStore.protoBook.set(id, message.protocols)
   }
+
+  /**
+   * Get self signed peer record raw envelope.
+   * @return {Buffer}
+   */
+  async _getSelfPeerRecord () {
```
jacobheun, line 360: If this throws we're not catching it anywhere. We should probably catch and log internally here and return null if that happens. This **shouldn't** happen, but its an uncaught exception if it does somehow.

src/identify/consts.js
vasco-santos: This commit added them: https://github.com/libp2p/go-libp2p/commit/077a81814f4357379d11dc93d40f48788e515fa0#diff-98be6dba669aa4123b7ef190fe7113e8

However, now that I looked further, the same protocol is now being used and go reverted this change per https://github.com/libp2p/go-libp2p/pull/907#discussion_r419878500

I am doing the same

src/identify/index.js
```diff
@@ -90,10 +96,12 @@ class IdentifyService {
     const pushes = connections.map(async connection => {
       try {
         const { stream } = await connection.newStream(MULTICODEC_IDENTIFY_PUSH)
+        const signedPeerRecord = await this._getSelfPeerRecord()
```
jacobheun, line 99: Get this once instead of in the map function. Also, we should do the same thing with the listenAddrs and protocols to avoid doing that for X connections.

src/identify/index.js
```diff
@@ -172,8 +181,26 @@ class IdentifyService {
...
 
+    let addresses
+
+    try {
+      const envelope = await Envelope.openAndCertify(signedPeerRecord, PeerRecord.DOMAIN)
+      const peerRecord = await PeerRecord.createFromProtobuf(envelope.payload)
+
+      addresses = peerRecord.multiaddrs
+    } catch (err) {
+      log('received invalid envelope, discard it and fallback to listenAddrs is available')
```
jacobheun, line 192: While it's not a strict failure we should still log the error.
```suggestion
      log('received invalid envelope, discard it and fallback to listenAddrs is available', err)
```

src/identify/index.js
```diff
@@ -258,17 +288,56 @@ class IdentifyService {
...
+
+    let addresses
+
+    try {
+      const envelope = await Envelope.openAndCertify(message.signedPeerRecord, PeerRecord.DOMAIN)
+      const peerRecord = await PeerRecord.createFromProtobuf(envelope.payload)
+
+      addresses = peerRecord.multiaddrs
+    } catch (err) {
+      log('received invalid envelope, discard it and fallback to listenAddrs is available')
```
jacobheun, line 301:
```suggestion
      log('received invalid envelope, discard it and fallback to listenAddrs is available', err)
```

PR #683, opened 2020-06-25
feat: certified addressbook

vasco-santos:
This PR adds the signed peer records in the AddressBook as part of #653 

Remaining:

- [x] Identify service should use these new features to store self and fetch it

Needs:

- [x] #681 
- [x] #682

The protobuf definition for persistence was inspired (copied) from go-libp2p, and the API to consume and get the records was inspired on go-libp2p CertifiedAddressBook [interface](https://github.com/libp2p/go-libp2p-core/blob/master/peerstore/peerstore.go#L147).

---

The AddressBook was previously holding `Map<string, Array<Address>>`, which mapped a peer id string identifier to an Array of Address (object with multiaddr property). It was changed to `Map<string, Array<Entry>>`, where the Entry is an object containing `{ addresses: Array<Address>, record: { raw: Buffer, seqNumber: number} }`. The Address now also includes a `isCertified` property that aims to allow subsystems to filter known multiaddrs of a peer that are certified without a need to unmarshall the record. The `raw` envelope is stored for being exchanged easily with other peers.

There are 3 remaining questions to be answered:

1. When the AddressBook receives a new peer record, should it replace the previously stored multiaddrs (not certified) by the new ones, or should it merge? If it merges, when the AddressBook receives a new peer record with different addresses from a previously stored record, what happens to the previously addresses marked as certified?
2. We emit `change:multiaddrs` when a peer has new multiaddrs. The question is, what if we already know a multiaddr but it is now certified? From the persistence layer standpoint, it is important to receive the event, but is this expected from a regular listener of the events?
3. (this we talked before, and we should just delay a decision and add it to Future Work section): When loading from the datastore on restart, should we ignore the multiaddrs stored that are not in the signed peer record?

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -816,6 +816,7 @@ peerStore.addressBook.getMultiaddrsForPeer(peerId)
 
 Set known `multiaddrs` of a given peer. This will replace previously stored multiaddrs, if available.
 Replacing stored multiaddrs might result in losing obtained certified addresses, which is not desirable.
+Consider using `addressBook.add()` if you're not sure this is what you want to do.
 
 `peerStore.addressBook.set(peerId, multiaddrs)`
 
--- a/doc/API.md
+++ b/doc/API.md
@@ -814,7 +814,8 @@ peerStore.addressBook.getMultiaddrsForPeer(peerId)
 
 ### peerStore.addressBook.set
 
-Set known `multiaddrs` of a given peer.
+Set known `multiaddrs` of a given peer. This will replace previously stored multiaddrs, if available.
+Replacing stored multiaddrs might result in losing obtained certified addresses, which is not desirable.
 
 `peerStore.addressBook.set(peerId, multiaddrs)`
 
--- a/src/peer-store/address-book.js
+++ b/src/peer-store/address-book.js
@@ -125,9 +125,10 @@ class AddressBook extends Book {
   }
 
   /**
-   * Get a peer raw envelope.
+   * Get the raw Envelope for a peer. Returns
+   * undefined if no Envelope is found.
    * @param {PeerId} peerId
-   * @return {Buffer}
+   * @return {Buffer|undefined}
    */
   getRawEnvelope (peerId) {
     const entry = this.data.get(peerId.toB58String())
@@ -141,8 +142,9 @@ class AddressBook extends Book {
 
   /**
    * Get an Envelope containing a PeerRecord for the given peer.
+   * Returns undefined if no record exists.
    * @param {PeerId} peerId
-   * @return {Promise<Envelope>}
+   * @return {Promise<Envelope|void>}
    */
   getPeerRecord (peerId) {
     const raw = this.getRawEnvelope(peerId)
--- a/src/peer-store/address-book.js
+++ b/src/peer-store/address-book.js
@@ -158,6 +158,8 @@ class AddressBook extends Book {
 
   /**
    * Set known multiaddrs of a provided peer.
+   * This will replace previously stored multiaddrs, if available.
+   * Replacing stored multiaddrs might result in losing obtained certified addresses.
    * @override
    * @param {PeerId} peerId
    * @param {Array<Multiaddr>} multiaddrs
--- a/src/peer-store/address-book.js
+++ b/src/peer-store/address-book.js
@@ -88,7 +88,7 @@ class AddressBook extends Book {
     }
 
     // Verify peerId
-    if (peerRecord.peerId.toB58String() !== envelope.peerId.toB58String()) {
+    if (!peerRecord.peerId.equals(envelope.peerId)) {
       log('signing key does not match PeerId in the PeerRecord')
       return false
     }
@@ -220,10 +220,10 @@ class AddressBook extends Book {
     const id = peerId.toB58String()
 
     const entry = this.data.get(id) || {}
-    const rec = entry.addresses
+    const rec = entry.addresses || []
 
     // Add recorded uniquely to the new array (Union)
-    rec && rec.forEach((addr) => {
+    rec.forEach((addr) => {
       if (!addresses.find(r => r.multiaddr.equals(addr.multiaddr))) {
         addresses.push(addr)
       }
@@ -244,7 +244,7 @@ class AddressBook extends Book {
     log(`added provided multiaddrs for ${id}`)
 
     // Notify the existance of a new peer
-    if (!rec) {
+    if (!entry.addresses) {
       this._ps.emit('peer', peerId)
     }
 
--- a/src/peer-store/address-book.js
+++ b/src/peer-store/address-book.js
@@ -9,10 +9,12 @@ const multiaddr = require('multiaddr')
 const PeerId = require('peer-id')
 
 const Book = require('./book')
+const PeerRecord = require('../record/peer-record')
 
 const {
   codes: { ERR_INVALID_PARAMETERS }
 } = require('../errors')
+const Envelope = require('../record/envelope')
 
 /**
  * The AddressBook is responsible for keeping the known multiaddrs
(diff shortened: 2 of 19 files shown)


jacobheun:
> 1. When the AddressBook receives a new peer record, should it replace the previously stored multiaddrs (not certified) by the new ones, or should it merge? If it merges, when the AddressBook receives a new peer record with different addresses from a previously stored record, what happens to the previously addresses marked as certified?

Non certified addresses should be expired by TTL. When we receive a new Peer Record, the record with the highest sequence wins, the other certified addresses should be removed.

> 2. We emit change:multiaddrs when a peer has new multiaddrs. The question is, what if we already know a multiaddr but it is now certified? From the persistence layer standpoint, it is important to receive the event, but is this expected from a regular listener of the events?

We should trigger the change. While the underlying multiaddr didn't change, our state did. This is helpful for apps that may want to take action when addresses become certified.

> 3. (this we talked before, and we should just delay a decision and add it to Future Work section): When loading from the datastore on restart, should we ignore the multiaddrs stored that are not in the signed peer record?

We should check the TTL on load for these and only keep them if it hasn't expired. 



I haven't checked the address book code yet, let me know if you want me to give that a pass.

vasco-santos:
Since we now have self being stored in the AddressBook, should we filter out `self` in `libp2p.peerStore.peers`?
Note: the interop tests are breaking because the daemon is breaking because of this (trying to get a connection from self to self as a result of being present in the `peerStore.peers`)

wemeetagain:
Just noticed that gossipsub v1.1 will need to import from this repo at runtime to access `Envelope.openAndCertify`. Not just a devDependency, as before.
Is this intended?

vasco-santos:
> Just noticed that gossipsub v1.1 will need to import from this repo at runtime to access Envelope.openAndCertify. Not just a devDependency, as before.
Is this intended?

Yes! I am not a fan of requiring stuff as `libp2p/src/...`, but the alternative is getting a mountain of repos. We are moving forward on having all modules tight with libp2p inside the libp2p repo. We can move it to its own repo/module in the future if we think it is valuable.

src/peer-store/index.js
```diff
@@ -83,6 +85,9 @@ class PeerStore extends EventEmitter {
       ...this.metadataBook.data.keys()
     ])
 
+    // Remove self peer if present
```
vasco-santos, line 88: What do you think on this?
For several reasons, we should not return self on peers. We use it widely, for listing peers in IPFS/daemon, for autoDial, ... Users should also be using it, so I think this should not be returned in peers.

jacobheun: Yeah we should exclude it when not getting it specifically. Self is not a peer.

src/peer-store/address-book.js
```diff
@@ -109,12 +218,14 @@ class AddressBook extends Book {
 
     const addresses = this._toAddresses(multiaddrs)
     const id = peerId.toB58String()
-    const rec = this.data.get(id)
+
+    const entry = this.data.get(id) || {}
+    const rec = entry.addresses
```
jacobheun, line 223:
```suggestion
    const rec = entry.addresses || []
```

src/peer-store/address-book.js
```diff
@@ -109,12 +218,14 @@ class AddressBook extends Book {
...
 
     // Add recorded uniquely to the new array (Union)
-    rec && rec.forEach((mi) => {
-      if (!addresses.find(r => r.multiaddr.equals(mi.multiaddr))) {
-        addresses.push(mi)
+    rec && rec.forEach((addr) => {
```
jacobheun, line 226: Combine with the empty array set above.
```suggestion
    rec.forEach((addr) => {
```

src/peer-store/index.js
```diff
@@ -37,9 +37,11 @@ class PeerStore extends EventEmitter {
   /**
    * @constructor
    */
-  constructor () {
+  constructor ({ peerId } = {}) {
```
jacobheun, line 40: Why is this optional? I don't see a use case for it and it just adds unnecessary checks when filtering self out.

src/peer-store/persistent/index.js
```diff
@@ -296,9 +300,15 @@ class PersistentPeerStore extends PeerStore {
 
           this.addressBook._setData(
```
jacobheun, line 301: Isn't this going to lose the isCertified flag on addresses?

src/record/envelope/index.js
```diff
@@ -129,6 +124,13 @@ const unmarshalEnvelope = async (data) => {
   })
 }
 
+/**
+ * Unmarshal a serialized Envelope protobuf message.
+ * @param {Buffer} data
+ * @return {Promise<Envelope>}
+ */
+Envelope.createFromProtobuf = unmarshalEnvelope
```
jacobheun, line 132: Why not just set the function here and avoid two names for the same method?

test/peer-store/address-book.spec.js
```diff
@@ -396,4 +401,266 @@ describe('addressBook', () => {
       return defer.promise
     })
   })
+
+  describe('certified records', () => {
+    let peerStore, ab
+
+    describe('consumes successfully a valid peer record and stores its data', () => {
```
jacobheun, line 408:
```suggestion
    describe('consumes a valid peer record and stores its data', () => {
```

src/peer-store/address-book.js
```diff
@@ -39,16 +56,104 @@ class AddressBook extends Book {
...
+
+    return true
+  }
+
+  /**
+   * Get a peer raw envelope.
+   * @param {PeerId} peerId
+   * @return {Buffer}
+   */
+  getRawEnvelope (peerId) {
```
vasco-santos, line 132: Should we add a convenience method for `getRawSelfEnvelope`?
We are using this providing own peerId in the identify protocol, and I am also using it now in rendezvous

src/peer-store/address-book.js
```diff
@@ -83,7 +189,10 @@ class AddressBook extends Book {
       }
     }
 
-    this._setData(peerId, addresses)
+    this._setData(peerId, {
+      addresses,
+      record: entry.record
+    })
```
jacobheun, line 195: This is a bit concerning. Where is `set` being called? If we already have a signed peer record and `set` gets called it will override certified addresses. If we keep this behavior we **really** need to be clear in the docs about what set does. It would be pretty easy for a user to make a mistake and override signed addresses. I don't think we should prohibit this, but it's not a method that should be used regularly.

src/peer-store/address-book.js
vasco-santos: It is not needed, but could potentially help. I did not have access to the self peerId in the rendezvous server (in the previous PR state), and I needed to change everything to provide libp2p because of that. But yeah, it was mostly a question, we don't need it, it could just help in some cases

src/peer-store/address-book.js
vasco-santos: > gets called it will override certified addresses.
Yes, that is true. Like map operations, set will replace what existed before.

In the libp2p codebase, `addressBook.set` was used in identify, which is not anymore as a result of identify using the signedPeerRecords now. It is used a lot in the tests.

I can add an attention notice to this in the `API.md` + jsdoc. I was thinking that we could deprecate set in the future. But, we still use `set` in ProtoBook and MetadataBook and we will use it as well.
If we decide to deprecate it, it should only be deprecated for the AddressBook, which would create a difference in their APIs

src/peer-store/address-book.js
```diff
@@ -39,18 +56,110 @@ class AddressBook extends Book {
...
+      return undefined
+    }
+
+    return Envelope.createFromProtobuf(raw)
+  }
+
   /**
    * Set known multiaddrs of a provided peer.
+   * This will replace previously stored multiaddrs, if available.
+   * Replacing stored multiaddrs might result in losing obtained certified addresses.
```
jacobheun, line 162: I think it would be good to encourage people to use add unless they know this is what they want.
```suggestion
   * Replacing stored multiaddrs might result in losing obtained certified addresses.
   * If you are not sure, it's recommended to use `add` instead.
```

doc/API.md
```diff
@@ -814,7 +814,8 @@ peerStore.addressBook.getMultiaddrsForPeer(peerId)
 
 ### peerStore.addressBook.set
 
-Set known `multiaddrs` of a given peer.
+Set known `multiaddrs` of a given peer. This will replace previously stored multiaddrs, if available.
+Replacing stored multiaddrs might result in losing obtained certified addresses, which is not desirable.
```
jacobheun, line 818:
```suggestion
Replacing stored multiaddrs might result in losing obtained certified addresses, which is not desirable.
Consider using `addressBook.add()` if you're not sure this is what you want to do.
```

PR #693, opened 2020-07-07
test: add pubsub reconnect test

jacobheun:
This aims to test and resolve issues with reconnects in pubsub.

If a node quickly reconnects, such as a browser refreshing and reconnecting to known peers, and the original connection doesn't immediately hang up, it's possible for us to establish a new connection to that peer before pubsub cleans up the old connection. This causes the remote pubsub peer to eventually lose the old, bad connection, but not form a new outbound stream on the new connection. While the refreshed node would be able to send messages, it won't be able to receive them.

We need to improve the pubsub peer logic to account for this, without creating duplicate outbound streams.

- [x] requires https://github.com/libp2p/js-libp2p-pubsub/pull/57


package.json
```diff
@@ -102,11 +102,12 @@
     "libp2p-delegated-content-routing": "^0.5.0",
     "libp2p-delegated-peer-routing": "^0.5.0",
     "libp2p-floodsub": "^0.21.0",
-    "libp2p-gossipsub": "^0.4.0",
+    "libp2p-gossipsub": "^0.4.6",
     "libp2p-kad-dht": "^0.19.1",
     "libp2p-mdns": "^0.14.1",
     "libp2p-mplex": "^0.9.5",
     "libp2p-noise": "^1.1.1",
+    "libp2p-pubsub": "github:libp2p/js-libp2p-pubsub#ffe8cd1ef4e2eaf2dfb661a4f37c79c63a1f805b",
```
vasco-santos, line 110: Needs to be removed on release

Issue #694, opened 2020-07-07
[Pubsub] Cannot read property 'catch' of undefined in pubsub.attachConnection


geolffreym:
when a peer is connected a get this :( something happen with webrtc-star using ipfs@0.46.0

my conf!!
```
module.exports = (opts) => {
	// Set convenience variables to clearly showcase some of the useful things that are available
	// const peerId = opts.peerId;
	
	opts.libp2pOptions.modules.peerDiscovery = [Bootstrap];
	opts.libp2pOptions.modules.transport.push(WebrtcStar);
	opts.libp2pOptions.config.transport = {
		[WebrtcStar.prototype[Symbol.toStringTag]]: {
			wrtc
		}
	}
	
	// Build and return our libp2p node
	return new Libp2p(opts.libp2pOptions)
}
```

` peerInfo.protocols.forEach((p) => newPeerInfo.protocols.add(p))`

```
TypeError: Cannot read property 'forEach' of undefined
    at PeerStore.add (/node_modules/libp2p/src/peer-store/index.js:83:24)
    at PeerStore.put (/node_modules/libp2p/src/peer-store/index.js:61:19)
    at Libp2p._onDiscoveryPeer (/node_modules/libp2p/src/index.js:443:20)
    at EventEmitter.emit (events.js:315:20)
    at ClassIsWrapper._peerDiscovered (/node_modules/libp2p-webrtc-star/src/index.js:237:20)
    at Socket.Emitter.emit (/node_modules/component-emitter/index.js:133:20)
    at Socket.onevent (/socket.io-client/lib/socket.js:278:10)
    at Socket.onpacket (/socket.io-client/lib/socket.js:236:12)
    at Manager.<anonymous> (/node_modules/component-bind/index.js:21:15)
    at Manager.Emitter.emit (/node_modules/component-emitter/index.js:133:20)

```

CC @jacobheun @vasco-santos

vasco-santos:
@geolffreym 
you need to bump the `webrtc-star` in this case. You can see what dependencies you need to bump here [https://github.com/libp2p/js-libp2p/pull/649/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R101](https://github.com/libp2p/js-libp2p/pull/649/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R101). 

I will make sure this gets more visible for people migrating

geolffreym:
[...]

PR #695, opened 2020-07-07
chore: add modules to update in 0.28.x migration

vasco-santos:
Adding modules to update in the `0.28.x` release per #694


vasco-santos:
@jacobheun I was thinking about creating a `MIGRATION_TEMPLATE.md` for the migration guide, like we have for the releases. It would have a fixed structure to be consistent for users migrating and would also guarantee that we include all the information needed, like the modules that need to be updated from now on. What do you think?

jacobheun:
@vasco-santos 💯 yes please!

Issue #655, opened 2020-06-02
⚡️ 0.30 RELEASE 🚀


vasco-santos:
> Libp2p is becoming self sufficient on discoverability and connectivity - The first step

# 🔦 Highlights

## 📝 Type definitions

`js-libp2p` is now distributed with type definitions for TypeScript in its core API. We are using JSDoc to type the code inline and [aegir](https://github.com/ipfs/aegir) for type checking and generate type definitions on release.

Upstream typescript projects have been leveraging projects like https://github.com/ChainSafe/libp2p-ts to have definitions for libp2p. With `0.30` typescript projects can start using the built in libp2p types.

This will also enable us to provide automated (and updated) docs, without the need to manually keep them.

## 🌐 Auto relay

AutoRelay enables libp2p nodes to dynamically find and bind to relays on the network. Once binding (listening) is done, the node can and should advertise its addresses to the network, allowing any other node to dial it over its bound relay(s). While direct connections to nodes are preferable, it's not always possible to do so due to NATs or browser limitations.

This is a crucial feature in terms of connectivity for some peers in the network, which will now be able to connect to other peers in a more independent fashion.

## 🗺 Improved Peer advertising and Dialer mechanics

Libp2p announces all the addresses a peer is listening on. However, most of the time these addresses include local private addresses, which as expected are not reachable. With peer store persistence, other peers will store (and even exchange) these local private addresses that will not be useful for others. Libp2p now supports a `customAnnounceFilter` function that enables via config to not announce any private addresses for instance. For simplifying this even more, `libp2p-utils` already has some filters that can just be imported and used.

`libp2p-websockets` has allowed TCP and DNS addresses, both with ws or wss, to be used for dialling. This should not be the default scenario due to security and browser policies. Now, `libp2p-websockets` will only support DNS based multiaddrs, as well as supporting `wss` only in the browser. And yes, you still need other types of multiaddrs for testing/development and we got you covered. You can override the websockets filter config easily via transport config options.

On more general dialer mechanics enhancements, the dialer will now sort the known addresses of a peer on a public and certified first logic. This way, the dialer will try to dial first the addresses that in theory will be more likely to work. This behaviour can also be configured via a custom function in the config.

Last but not least, in order for other nodes on the network to easily find a given libp2p node, this node will now recurrently find its closest peers on the network, connect to them and exchange its address records. This will enable other peers on the network to find a peer in an easier and faster fashion.

## 📂 Production Guides and Examples improvements

[...]

PR #698, opened 2020-07-09
fix: not dial all known peers in parallel on startup

vasco-santos:
This PR aims to block large numbers of dials on a peer restart, if this peer already new a large number of peers in the past.

It checks the `minPeers` as the `maybeConnect` function used by the discovery event handlers. This was also using the `minPeers`, but as all the dials were put in parallel, the `connectionManager.size` would not be updated to avoid further dials.

When we work on connectionManager, we should work on better strategies for this, and likely have them as configurable

### DIFF
--- a/src/connection-manager/index.js
+++ b/src/connection-manager/index.js
@@ -25,6 +25,7 @@ const defaultOptions = {
   maxReceivedData: Infinity,
   maxEventLoopDelay: Infinity,
   pollInterval: 2000,
+  maybeConnectInterval: 10000,
   movingAverageInterval: 60000,
   defaultPeerValue: 1
 }
@@ -48,6 +49,7 @@ class ConnectionManager extends EventEmitter {
    * @param {Number} options.pollInterval How often, in milliseconds, metrics and latency should be checked. Default=2000
    * @param {Number} options.movingAverageInterval How often, in milliseconds, to compute averages. Default=60000
    * @param {Number} options.defaultPeerValue The value of the peer. Default=1
+   * @param {Number} options.maybeConnectInterval How often, in milliseconds, it should preemptively guarantee connections are above the low watermark. Default=10000
    */
   constructor (libp2p, options) {
     super()
@@ -76,8 +78,11 @@ class ConnectionManager extends EventEmitter {
      */
     this.connections = new Map()
 
+    this._started = false
     this._timer = null
+    this._maybeConnectTimeout = null
     this._checkMetrics = this._checkMetrics.bind(this)
+    this._maybeConnectN = this._maybeConnectN.bind(this)
   }
 
   /**
@@ -104,8 +109,11 @@ class ConnectionManager extends EventEmitter {
     })
     this._onLatencyMeasure = this._onLatencyMeasure.bind(this)
     this._latencyMonitor.on('data', this._onLatencyMeasure)
-    this._maybeConnectN()
+
+    this._started = true
     log('started')
+
+    this._maybeConnectN()
   }
 
   /**
@@ -113,9 +121,12 @@ class ConnectionManager extends EventEmitter {
    * @async
    */
   async stop () {
+    clearTimeout(this._maybeConnectTimeout)
+
     this._timer && this._timer.clear()
     this._latencyMonitor && this._latencyMonitor.removeListener('data', this._onLatencyMeasure)
 
+    this._started = false
     await this._close()
     log('stopped')
   }
@@ -210,7 +221,6 @@ class ConnectionManager extends EventEmitter {
       this.connections.delete(peerId)
       this._peerValues.delete(connection.remotePeer.toB58String())
       this.emit('peer:disconnect', connection)
-      this._maybeConnectN()
     }
   }
 
@@ -279,6 +289,8 @@ class ConnectionManager extends EventEmitter {
    * @private
    */
   async _maybeConnectN () {
+    this._isTryingToConnect = true
+
     const minConnections = this._options.minConnections
 
     // Already has enough connections
@@ -302,11 +314,18 @@ class ConnectionManager extends EventEmitter {
         log('connecting to a peerStore stored peer %s', peers[i].id.toB58String())
         try {
           await this._libp2p.dialer.connectToPeer(peers[i].id)
+
+          // Connection Manager was stopped
+          if (!this._started) {
+            return
+          }
         } catch (err) {
           log.error('could not connect to peerStore stored peer', err)
         }
       }
     }
+
+    this._maybeConnectTimeout = setTimeout(this._maybeConnectN, this._options.maybeConnectInterval)
   }
 
   /**
--- a/src/connection-manager/index.js
+++ b/src/connection-manager/index.js
@@ -1,9 +1,12 @@
 'use strict'
 
+const debug = require('debug')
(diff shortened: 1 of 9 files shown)


src/index.js
```diff
@@ -492,6 +489,33 @@ class Libp2p extends EventEmitter {
...
+   */
+  _maybeConnectPeerStorePeers () {
+    if (!this._config.peerDiscovery.autoDial) {
+      return
+    }
+
+    const minPeers = this._options.connectionManager.minPeers || 0
+
+    // TODO: this should be ordered by score on PeerStoreV2
+    this.peerStore.peers.forEach(async (peer) => {
```
vasco-santos, line 505: @jacobheun I started by doing the dials in serial as the token based dial will already use parallel dials for several multiaddrs the peer we are targeting might have. Also, the discovery services kicked in, and their new discoveries might be more important than the older ones (this is where having scores for the older peers would help a lot too).

Another solution I thought is to do the `minPeers` parallel dials and then do the rest in serial (to compensate possible offline peers). Let me know what do you think about the pros and cons of both approaches.

src/index.js
```diff
@@ -492,6 +489,33 @@ class Libp2p extends EventEmitter {
     peer.protocols && this.peerStore.protoBook.set(peer.id, peer.protocols)
   }
 
+  /**
+   * Will dial a substet of the peers already known
+   * @private
+   * @returns {void}
+   */
+  _maybeConnectPeerStorePeers () {
+    if (!this._config.peerDiscovery.autoDial) {
```
vasco-santos, line 498: Maybe this should deserve its own config in the peerStore instead? 🤔

jacobheun:
I think the more general case that would be helpful here is going to the PeerStore when our connection count is below the low watermark. Right now `autoDial` is only ever triggered on peer discovery. I don't think we should be mass emitting peer discovered events on boot for persisted peers. Application code can check the peer store. 

I think we should add a counterpart of the `ConnectionManager_.maybeDisconnectOne` that deals with proactive connecting. If our connection count is below our low watermark, we should attempt to dial peers from our peerstore (ideally ranked, but we can do that later) that we are not connected to, until we are at or above the low watermark.
On startup, we can just start this recursive call. We could batch the requests to reduce initial connection time. This should also get called when a peer disconnects and we drop below the low watermark.

vasco-santos:
> I don't think we should be mass emitting peer discovered events on boot for persisted peers. Application code can check the peer store.

Totally agree, but I don't think this is the correct timing for that. We are doing that in `0.27` and `0.28` and this will be a breaking change for who expects this! I agree we include this in [0.30](https://github.com/libp2p/js-libp2p/issues/655). I can open an issue for it.

> I think the more general case that would be helpful here is going to the PeerStore when our connection count is below the low watermark.

This PR does that for boot and I agree with this, but with some other things in consideration. When having a more intelligent PeerStore/ConnectionManager, we should select peers based on their score and the protocols we are running.

> I think we should add a counterpart of the ConnectionManager_.maybeDisconnectOne that deals with proactive connecting. If our connection count is below our low watermark, we should attempt to dial peers from our peerstore (ideally ranked, but we can do that later) that we are not connected to, until we are at or above the low watermark.

It seems a good idea to do it on disconnect, but just iterating blindly in the PeerStore and connect will not likely bring much value. But I think we can include it here as it is quite simple.

> On startup, we can just start this recursive call. We could batch the requests to reduce initial connection time. This should also get called when a peer disconnects and we drop below the low watermark.

This seems the way to go for me, but this should happen for the `0.28.x` line for me.

jacobheun:
> Totally agree, but I don't think this is the correct timing for that. We are doing that in 0.27 and 0.28 and this will be a breaking change for who expects this!

We can talk about this separately, it doesn't need to change for this, but it should change soon because we're going to hit scaling issues. If I know thousands of peers when I boot, that's going to get nasty real quick.

> It seems a good idea to do it on disconnect, but just iterating blindly in the PeerStore and connect will not likely bring much value.

Actually, we do have some metrics we could use. Instead of looking at ALL peers in the peer store, for now we could prioritize peers we have keys/protocols for. Both should indicate a previous direct exchange, although keys _might_ have come indirectly (i cant recall which subsystems we might have added this to in JS).

src/connection-manager/index.js
```diff
@@ -206,6 +210,7 @@ class ConnectionManager extends EventEmitter {
       this.connections.delete(peerId)
       this._peerValues.delete(connection.remotePeer.toB58String())
       this.emit('peer:disconnect', connection)
+      this._maybeConnectN()
```
jacobheun, line 213: If a disconnect occurs and the previous call to `_maybeConnectN` happens we're going to run two calls in parallel, we should avoid this.

vasco-santos: you mean the `_maybeConnectN` in start?

vasco-santos: We can potentially have it running from the start call while trying to dial a bunch of peers and one of them meanwhile disconnected. I can add a verification for wether we are currently running it

src/connection-manager/index.js
```diff
@@ -257,15 +272,62 @@ class ConnectionManager extends EventEmitter {
...
+  /**
+   * Proactively tries to connect to known peers stored in the PeerStore.
+   * It will keep the number of connections below the upper limit and sort
+   * the peers to connect based on wether we know their keys and protocols.
+   * @async
+   * @private
+   */
+  async _maybeConnectN () {
+    this._isTryingToConnect = true
+
```
jacobheun, line 293: This can go away
```suggestion
```

src/connection-manager/index.js
```diff
@@ -257,15 +272,62 @@ class ConnectionManager extends EventEmitter {
...
+    const minConnections = this._options.minConnections
+
+    // Already has enough connections
+    if (this.size >= minConnections) {
+      return
```
jacobheun, line 298: The timeout should be set again, if we lose connections we'll never call again, or we need to call the function on disconnect.

src/connection-manager/index.js
```diff
@@ -257,15 +272,62 @@ class ConnectionManager extends EventEmitter {
...
+          if (!this._started) {
+            return
+          }
+        } catch (err) {
+          log.error('could not connect to peerStore stored peer', err)
+        }
+      }
+    }
+
+    this._maybeConnectTimeout = setTimeout(this._maybeConnectN, this._options.maybeConnectInterval)
```
jacobheun, line 328: Use `retimer` for this. We're already using it and it will cut down on multiple timers in the background.

test/connection-manager/index.node.js
```diff
@@ -112,4 +115,148 @@ describe('libp2p.connections', () => {
...
+
+      const [libp2p] = await peerUtils.createPeer({
+        fixture: false,
+        config: {
+          addresses: {
+            listen: ['/ip4/127.0.0.1/tcp/0/ws']
+          },
+          connectionManager: {
+            minConnections,
+            maybeConnectInterval
```
jacobheun, line 239: This should be `autoDialInterval`. Since it's the wrong param, the default should be 10s, it's odd this test is passing. I assume it's actively in the process of connecting when the first check is done.

vasco-santos: Yeah, I debugged it earlier. It was failing sometimes before I add the interval param. I will fix it now

Issue #696, opened 2020-07-08
Production Infrastructure docs


vasco-santos:
## Background

IPFS/libp2p projects provide own infrastructure for enabling their users to learn the introductory concepts first and experiment with its functionalities.

These infrastructure pieces should not be used in production and only for experimentation. This issue should track the creation of several documents to guide users on the road from experimenting to production.

## Proposal

Create an `INFRASTRUCTURE.md` document with guides for setting up own infrastructure for each type of deployment needs, such as:

- [ ] `webrtc-star` signalling servers
- [ ] delegate routers
- [ ] relay server nodes
- [ ] rendezvous server nodes

References:

- https://github.com/libp2p/js-libp2p/pull/507
- https://github.com/libp2p/js-libp2p/pull/706
- https://discuss.libp2p.io/t/cant-connect-to-webrtc-star-with-ssl/522
- https://github.com/ipfs/js-ipfs/issues/2779

vasco-santos:
cc @jacobheun

maschad:
The components mentioned in this stack have been deprecated or are discouraged from use. Please explore:

- https://github.com/libp2p/js-libp2p/tree/master/packages/transport-webrtc
- https://cid.contact/

as substitutes

PR #706, opened 2020-07-13
Update README.md

Strernd:
This might be too specific, but I think it's helpful. I'm using docker-compose to run my node setup locally and ran into this issue and I haven't found this information anywhere else.


jacobheun:
> This might be too specific, but I think it's helpful.

Yes to both of these. I think along the lines of https://github.com/libp2p/js-libp2p/issues/696, this could be included in an example of running the various libp2p services in a single docker compose setup. cc @vasco-santos

vasco-santos:
Yeah, thanks for this PR @Strernd 

I am planning to write a markdown document including all types of setups regarding infrastructure and deployment. This seems a good point to add there. I will open an Issue to track it and will link this PR there for now. Once I have some time to work on it, I will move this PR to that

vasco-santos:
I am closing this issues as this will be included in #706 

Thanks for this @Strernd

PR #710, opened 2020-07-14
fix: retimer reschedule does not work as interval

vasco-santos:
`retimer` reschedule is not an `interval` feature but a "delay" of a previously defined interval. This PR fixes it

### DIFF
--- a/src/connection-manager/index.js
+++ b/src/connection-manager/index.js
@@ -178,7 +178,7 @@ class ConnectionManager extends EventEmitter {
     const total = received + sent
     this._checkMaxLimit('maxData', total)
     log('metrics update', total)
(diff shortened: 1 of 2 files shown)


src/connection-manager/index.js
```diff
@@ -178,7 +178,7 @@ class ConnectionManager extends EventEmitter {
     const total = received + sent
     this._checkMaxLimit('maxData', total)
     log('metrics update', total)
-    this._timer.reschedule(this._options.pollInterval)
+    retimer(this._checkMetrics, this._options.pollInterval)
```
jacobheun, line 181: Need to set this, otherwise we can't clear on stop.
```suggestion
    this._timer = retimer(this._checkMetrics, this._options.pollInterval)
```

PR #711, opened 2020-07-16
chore: add migration guide template

vasco-santos:
Per suggestion on https://github.com/libp2p/js-libp2p/pull/695#issuecomment-655482342 this PR adds a migration guide template.

Its goal is to create a consistent structure for this document to ease the life of users migrating and also to guarantee that we do not miss important things like the modules to update. Other points might be added in the future

### DIFF
--- a/MIGRATION_TEMPLATE.md
+++ b/MIGRATION_TEMPLATE.md
@@ -35,7 +35,10 @@ __Describe__
 
 With this release you should update the following libp2p modules if you are relying on them:
 
-<!--Specify module versions in JSON for migration below-->
+<!--Specify module versions in JSON for migration below.
+It's recommended to check package.json changes for this: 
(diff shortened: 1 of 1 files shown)


MIGRATION_TEMPLATE.md
```diff
@@ -0,0 +1,42 @@
...
+
+```
+
+-->
+
+## Module Updates
+
+With this release you should update the following libp2p modules if you are relying on them:
+
+<!--Specify module versions in JSON for migration below-->
```
jacobheun, line 38: Should we add a recommendation for checking this easily?
```suggestion
<!--Specify module versions in JSON for migration below.
It's recommended to check package.json changes for this: 
`git diff <release> <prev> -- package.json`
-->
```

PR #714, opened 2020-07-20
fix: return empty array when no multiaddrs are known

achingbrain:
Returning `undefined` makes the address length check in [dialler/index.js](https://github.com/libp2p/js-libp2p/blob/master/src/dialer/index.js#L73-L75) fail with `cannot read property length of undefined` as it has no null-guard, so the change here is to always return an array, but it might be empty if we don't know any multiaddrs for the given peer.

### DIFF
--- a/src/peer-store/address-book.js
+++ b/src/peer-store/address-book.js
@@ -179,7 +179,7 @@ class AddressBook extends Book {
     const record = this.data.get(peerId.toB58String())
 
     if (!record) {
-      return undefined
+      return []
     }
 
     return record.map((address) => {
--- a/test/peer-store/address-book.spec.js
+++ b/test/peer-store/address-book.spec.js
@@ -323,10 +323,10 @@ describe('addressBook', () => {
       throw new Error('invalid peerId should throw error')
     })
 
-    it('returns undefined if no multiaddrs are known for the provided peer', () => {
+    it('returns empty array if no multiaddrs are known for the provided peer', () => {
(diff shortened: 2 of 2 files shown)


jacobheun:
We should check in dialer to keep the API consistent with Maps. This is the fix proposed on https://github.com/libp2p/js-libp2p/pull/673, @vasco-santos can you submit an updated PR with your suggestion from that.

vasco-santos:
> We should check in dialer to keep the API consistent with Maps. This is the fix proposed on #673, @vasco-santos can you submit an updated PR with your suggestion from that.

On it

achingbrain:
`.getMultiaddrsForPeer` doesn't read as a map-like operation.  It seems like there's some clear intent in the method naming.

It should be made clear in the docs that it might return an empty list, or it might return `undefined` (should be `null`, really).

I find @vasco-santos's [own comment](https://github.com/libp2p/js-libp2p/pull/673#discussion_r441369535) on that PR to be a bit of a flag around this API.

vasco-santos:
> .getMultiaddrsForPeer doesn't read as a map-like operation. It seems like there's some clear intent in the method naming.

Not this specific one, but the other methods of the API are: https://github.com/libp2p/js-libp2p/blob/master/doc/API.md#peerstoreaddressbookdelete

The `getMultiaddrsForPeer` is for consistency with the rest. But I agree that we can document the return values as @achingbrain suggests

jacobheun:
This is resolved in #715 and #716. The api docs have been updated to reflect the undefined value.

PR #720, opened 2020-07-27
add example filter subscription

felipemarts:
I made this example to show how to put an additional filter on pubsub. So each node can verify that the message is valid before sharing it with the other nodes. This helps not to overload the network in cases where the node can check the validity of the messages by itself.


examples/pubsub-filter/1.js
```diff
@@ -0,0 +1,82 @@
...
+
+  const validateFruit = (msgTopic, peer, msg) => {
+    const fruit = msg.data.toString();
+    const validFruit = ['banana', 'apple', 'orange']
+    const valid = validFruit.includes(fruit);
+    return valid;
+  }
+
+  //validate fruit
+  node1.pubsub._pubsub.topicValidators.set(topic, validateFruit);
```
vasco-santos, line 66: @wemeetagain @jacobheun we should provide a friendly API for this.

felipemarts:
Do I place each example as two different folders or two examples in the same folder?

vasco-santos:
> Do I place each example as two different folders or two examples in the same folder?

I think it should live under the pubsub examples folder. But, let's wait for @jacobheun feedback

jacobheun:
We can just move it into `./examples/pubsub/filter` for now. We can revisit the root readme/example in `./examples/pubsub` later (when we revisit external examples and the docs site).

examples/pubsub-filter/1.js
jacobheun: @vasco-santos can you open a new issue to discuss the pubsub abstraction we currently have in place here? I think we should look at getting rid of it or finding a cleaner way to support direct access to various modules. The API is unnecessarily restrictive atm.

vasco-santos: Yes, of course

felipemarts:
Done

jacobheun:
I cleaned up some language and merged this via a331b84f134a83192c9779486fe274eb53bcdeaf. Thanks for the PR!

Issue #699, opened 2020-07-10
[discovery] Add support for AutoRelay


jacobheun:
AutoRelay enables libp2p nodes to dynamically find and bind to relays on the network. Once binding (listening) is done, the node can and should advertise its addresses to the network, allowing any other node to dial it over its bound relay(s). While direct connections to nodes are preferable, it's not always possible to do so due to NATs.

##### A connected relay
The Circuit Relay module has the ability to ask connected nodes if they support the relay `HOP` protocol. If they do, the libp2p node can bind to that relay as a listener. A node **should** record a list of relays and use only a subset of those to avoid binding to a large number of relays, as this is not considerate to those relays.

1. When a node connects to a peer that announces support for the relay protocol, `/libp2p/circuit/relay/0.1.0`, it may ask that node if it support `HOP`.
1. If the peer supports `HOP`, libp2p should add it to its list of known relays.
1. When a known relay is added, libp2p should check to see if it needs a/another relay (to meet some min threshold). If below that threshold, libp2p should bind to that relay and add it to its announced addresses.
   1. When adding announced addresses, libp2p **SHOULD** avoid adding private addresses of the relay, such as `/ip4/127.0.0.1/tcp/4001`.
   1. If libp2p does not have enough known relays, it should [query the network](#remote-relays) if possible.
1. When libp2p updates its addresses, it should update the network.

##### Remote Relays

When a node with AutoRelay enabled boots, it should search the network for the relay protocol, `/libp2p/relay`. This search may be done via one of several options leveraging `libp2p.contentRouting.findProviders()`. Depending on what features are turned on this may include: the DHT, Delegate Routing, and/or the Rendezvous Protocol (WIP).

If a node does not have enough known relays connected, and is unable to connect to other known relays, it should query the network for more relays. This should be done with a backoff to avoid spamming queries.

**References**:
- [Cant connect to /p2p-circuit peers](https://github.com/ipfs/go-ipfs/issues/7433#issuecomment-656410330)
- [go-libp2p AutoRelay](https://github.com/libp2p/go-libp2p/blob/v0.10.0/p2p/host/relay/autorelay.go)

vasco-santos:
After discussing with @jacobheun I am planning on dividing this effort in a set of PRs as follows:

| Milestone | PR | State |
|----------|:-------------:|------:|
| `1) Relay topology - initial implementation` |  #723  | WIP |
| `2) Avoid announcing private addresses` |  #723  | WIP |
[...]

Issue #703, opened 2020-07-13
[Meta] Improving libp2p discoverability and connectivity


jacobheun:
This is a Meta issue for tracking several sub systems involved in improving discoverability and connectivity in js-libp2p. This list is not an exhaustive and will evolve over time.

## Connectivity
- [ ] [AutoRelay support](https://github.com/libp2p/js-libp2p/issues/699)
- [ ] AutoNAT to determine dialability

## Discoverability
- [ ] [Connect to closest peers](https://github.com/libp2p/js-libp2p/issues/704)
- [ ] Implement the DHT changes from go-ipfs 0.5 to improve performance
- [ ] Implement the [rendezvous protocol](https://github.com/libp2p/js-libp2p/issues/651)
- [X] Delegate routing

wemeetagain:
closing as stale. most or all of these items have been implemented

PR #723, opened 2020-07-31
feat: auto relay

vasco-santos:
This PR is part of the #699 initiative to add support for AutoRelay.

It includes the initial points described in the issue for connecting with peers with a relay protocol, more precisely, it includes this points:

1. When a node connects to a peer that announces support for the relay protocol, /libp2p/circuit/relay/0.1.0, it may ask that node if it support HOP.
2. If the peer supports HOP, libp2p should add it to its list of known relays.
3. When a known relay is added, libp2p should check to see if it needs a/another relay (to meet some min threshold). If below that threshold, libp2p should bind to that relay and add it to its announced addresses.

Follow up PRs:
1. When libp2p updates its addresses, it should update the network - Identify push on add/remove listen + self peer record update on addr change
2. Query the network if it needs more relays
3. New connectionManager integration for disconnect improvement

It is important to mention the following implementation details:

- After a connection is established, the identify protocol will kick in. As a result of the identify protocols, peers will exchange the protocols they support. The auto-relay relies on the `change:protocols` event from the PeerStore, in order to understand what peers to ask for HOP support
- When a peer supports HOP, this information is stored in the metadataBook
- When a peer disconnects (or does not support relay protocol anymore - this should not be possible at the moment), remaining open connections with peers supporting HOP will be used to replace the disconnected one
  - This should be refactored to leverage the new connectionManager as a follow up PR
- Already relayed connections, will not be used for auto relay
- Regarding 3,  `maxListeners` was added (happy to reconsider the name) instead the `minThreshold`. It limits the number of relay addresses to listen on. Each time a node is connected, it will ask if it supports HOP (if peer has the relay protocol). This should be polished with the connectionManagement improvements to have a general approach for the `autoDial`


src/circuit/auto-relay.js
```diff
@@ -0,0 +1,170 @@
...
+    //   // TODO: should it avoid to try the peer in question?
+    //   if (peer.id.equals(peerId) || this._listenRelays.has(idB58Str)) {
+    //     continue
+    //   }
+    // }
+  }
+}
+
+// TODO: be careful about relay connect to relay peer that might create a double relayed conn
+// TODO: trigger identify to share new signed peer record
```
vasco-santos, line 168: This should be part of a follow up PR

src/circuit/listener.js
```diff
@@ -24,7 +29,7 @@ module.exports = (circuit) => {
   listener.listen = async (addr) => {
     const addrString = String(addr).split('/p2p-circuit').find(a => a !== '')
 
-    const relayConn = await circuit._dialer.connectToPeer(multiaddr(addrString))
+    const relayConn = await libp2p.dial(multiaddr(addrString))
```
vasco-santos, line 32: I changed this to reuse connections if already exist, instead of creating a new one. What do you think?

vasco-santos:
@jacobheun your thoughts regarding the current direction of this work are welcome :) 
I think this first iteration is almost ready

src/circuit/auto-relay.js
```diff
@@ -0,0 +1,170 @@
...
+   * Registrar notifies a connection successfully with circuit protocol.
+   * @private
+   * @param {PeerId} peerId remote peer-id
+   * @param {Connection} connection connection to the peer
+   */
+  async _onPeerConnected (peerId, connection) {
+    // Check if already listening on enough relays
+    if (this._listenRelays.size >= this.maxListeners) {
+      return
+    }
```
jacobheun, line 63: This is not really taking advantage of the topology, and we might not want to use multicodec for this. Right now, every peer that supports the relay protocol is going to get added as a peer to the topology, which isn't good because nearly all peers support the relay protocol (both go and js default to this for STOP support).

We still want to listen on protocol updates to see who supports relay, and then check `canHop` on those. If they do support HOP we should add them to a list of known HOP relays. This could be as simple as adding metadata to the peer store (hop_relay=true or something) so we can persist this knowledge. I think separating the check from the listen would be good. This would enable us to more easily check for more relays without listening on them. For example, I really only need 1 relay to listen on, but Ideally I'd like to know about a couple in case the first has issues.

> Should we keep a backup set of connected peers that have the relay protocol, so that we don't need to iterate all the connections when a peer disconnects?

I think tracking the metadata would resolve this. We could ditch the topology altogether for now and just listen on the peerstore protocol changes for the checks. I'd definitely try to keep this as simple as possible for now.

test/relay/auto-relay.node.js
```diff
@@ -0,0 +1,286 @@
...
+    let relayLibp2p3
+    let autoRelay1
+
+    beforeEach(async () => {
+      const peerIds = await createPeerId({ number: 4 })
+      // Create 4 nodes, and turn HOP on for the relay
+      ;[libp2p, relayLibp2p1, relayLibp2p2, relayLibp2p3] = peerIds.map((peerId, index) => {
+        const opts = baseOptions
+
+        if (index !== 0) {
```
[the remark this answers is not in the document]
vasco-santos: Yes, this is used to test the maximum number of relays to listen on

src/circuit/auto-relay.js
vasco-santos: I will switch to listen on protocol event changes. The metadata seems nice to have this information for persistence on restart. But, (as mentioned earlier in a PR review), it feels strange to me to mix the metadata that the application layer might insert and libp2p metadata. It would be nice to think of a way of at least flagging the origin/context of the metadata insertion

vasco-santos:
@jacobheun this is ready for review now. Should we create a base branch for `0.30.x` or for the work included in https://github.com/libp2p/js-libp2p/issues/703 ?

jacobheun:
> Should we create a base branch for 0.30.x or for the work included in #703?

@vasco-santos Yeah let's do that, there are a few things we want to land in conjunction with this so it would be good to gather those outside of master.

vasco-santos:
I changed it to 0.30.x branch as we might not include all of #703 in the next release

src/circuit/auto-relay.js
```diff
@@ -0,0 +1,190 @@
...
+    this._listenRelays.add(id)
+
+    // Create relay listen addr
+    const remoteMultiaddr = connection.remoteAddr
+    let listenAddr
+
+    if (!remoteMultiaddr.protoNames().includes('p2p')) {
+      listenAddr = `${remoteMultiaddr.toString()}/p2p/${connection.remotePeer.toB58String()}/p2p-circuit/p2p/${this._peerId.toB58String()}`
+    } else {
+      listenAddr = `${remoteMultiaddr.toString()}/p2p-circuit/p2p/${this._peerId.toB58String()}`
```
jacobheun, line 129: We shouldn't need our peer id in the listen address, it should just end with `/p2p-circuit`. The circuit listener logic will need to be fixed because it expects our id at the end.

src/circuit/auto-relay.js
```diff
@@ -0,0 +1,190 @@
...
+    this._listenRelays.delete(id)
+    // TODO: this should be responsibility of the connMgr
+    this._listenOnAvailableHopRelays()
+  }
+
+  /**
+   * Try to listen on available hop relay connections.
+   * @return {Promise<void>}
+   */
+  async _listenOnAvailableHopRelays () {
```
jacobheun, line 157: This function should change, really what we want to do is `find` relays if we don't have enough. This would involve:
1. Check the metadata store for known relays, try them first if we're not connected
1. If we dont have enough, find relays on the network. (https://github.com/libp2p/go-libp2p/blob/fb3179e617bdf049734f58aa8ced2bb42de7ea5d/p2p/host/relay/autorelay.go#L250)

Autorelay should also start with a network search for relays. Ideally this would happen once we determined ourselves to be private (via autoNat) so we don't bind to relays when we're publicly dialable, but we dont have autonat yet.

If we find multiple relays we should eventually prioritize by lowest latency, but for now we can focus on the use case of users configuring their node to listen to a specific relay, and allow ambient canhop checks to supplement it.

src/circuit/circuit/hop.js
```diff
@@ -117,6 +117,31 @@ module.exports.hop = async function hop ({
...
+  streamHandler.write({
+    type: CircuitPB.Type.CAN_HOP
+  })
+
+  const response = await streamHandler.read()
+  await streamHandler.close()
+
+  if (response.code !== CircuitPB.Status.SUCCESS) {
+    throw errCode(new Error('Relay is not able to hop'), Errors.ERR_CANNOT_HOP)
+  }
```
jacobheun, line 142: This method should return a boolean instead of throwing. This gives us the clearer distinction of no support vs something bad happened on the protocol. Right now this is going to log a **lot** of errors that aren't errors.

src/circuit/listener.js
```diff
@@ -8,13 +8,18 @@ const log = debug('libp2p:circuit:listener')
...
  * @returns {Listener} a transport listener
  */
-module.exports = (circuit) => {
+module.exports = (libp2p) => {
   const listener = new EventEmitter()
   const listeningAddrs = new Map()
 
+  // Remove listeningAddrs when a peer disconnects
+  libp2p.connectionManager.on('peer:disconnect', (connection) => {
+    listeningAddrs.delete(connection.remotePeer.toB58String())
```
jacobheun, line 20: If this actually causes a change in our addresses we need to perform an identify push .

src/circuit/auto-relay.js
```diff
@@ -0,0 +1,190 @@
...
+  async _addListenRelay (connection, id) {
+    // Check if already listening on enough relays
+    if (this._listenRelays.size >= this.maxListeners) {
+      return
+    }
+
+    this._listenRelays.add(id)
+
+    // Create relay listen addr
+    const remoteMultiaddr = connection.remoteAddr
```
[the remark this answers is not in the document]
vasco-santos: That is a good point! The problem is what address to use from the announced ones. We will need to do something like just get the first addr (that is certified)

src/circuit/auto-relay.js
vasco-santos: > The circuit listener logic will need to be fixed because it expects our id at the end.
It does not seem necessary. I just checked and the Circuit listener does not seem to use the id at all. Also removed it and everything seems to work as expected.

src/circuit/listener.js
vasco-santos: I will call the identify push there if deleted. But, once we support a dynamic transport manager with runtime transport add and remove, we should probably have the IdentifyService to listen on updates

vasco-santos: I was planning on doing this in a follow up PR (point 4 in the issue), as I was also not doing the push when the new listen addr is added.
Anyway, I can do both now

src/circuit/auto-relay.js
vasco-santos: I mentioned in the description that this was the initial PR for the issue. The follow up PR would include that part to ease the implementation/review. But, if we start by a simple version, I can also include it here:

I will change the logic a bit, it is important to notice that we can be connected to a peer that supports hop, but we discarded it before and we should leverage it after if we need to. 

Regarding the start part, I was also expecting as a follow up. We can start it when the circuit listener start, but we need to guarantee that the peerStore is loaded first.

src/circuit/auto-relay.js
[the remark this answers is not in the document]
jacobheun: > We will already have the connection to the peer, so it will just be reused

The problem is our advertised address for this. The actual address we're connected to the peer doesn't matter, the address of the relay we advertise to other peers does.

src/circuit/listener.js
jacobheun: Follow up PRs are fine for this, but it would be good to at least denote some todo comments so we don't forget code paths in the subsequent PRs.

src/circuit/auto-relay.js
vasco-santos: Reconsidering this, I think we should try to leverage the connections we already have before dialling other peers. I think we should start by doing what I had:

1. iterate the available connections, if we are connected to a peer which supports hop but we are not listening on it, we should add it as a listener. 
2. if we still need more peers, go to the PeerStore
3. Finally (if we need more), we find relays on the network

I am hitting an "issue" with this approach, which also happens as you suggested. Considering that we call this if a disconnect happens and we remove a listen relay, if we iterate on the metadataBook, we will just try to reconnect to that node. This flow is odd and should be avoided. We would sort peers by score and eventually not dial it in the future, but for now we cannot do that. Or we "accept" this behaviour for now, or `_listenOnAvailableHopRelays` should receive an optional list of peers to ignore (where we would add the disconnected peer, if we are call `_listenOnAvailableHopRelays` as a consequence of it.

jacobheun: Why not prioritize known relays by whether or not they're connected?
* Check the Peer Store for relay peers
* If we're connected and they're not already a relay, bind to them
* If we don't have enough, perform a search
* Once the search is done, attempt to bind to known relays (which may include a dial)

If we're already connected to them we should have already checked their HOP status, so prioritizing all known relays by their connectedness makes sense, instead fo checking each connected peer which would be less efficient.

The immediate redial problem should really be solved with backoffs (connection gating, plus maybe a ttl tag in the peer store), especially if they disconnected from us.

vasco-santos: If we have already connections that we can use to bind, I think they should be used first instead of establishing new connections with other peers. Specially as we head towards having more meaningful connections to the peer. We will be establishing a new connection that we will want to maintain (auto relay binds will be important connections to the peer), while we could be using a connection that might be already important for the peer (as it already exists). It will also be faster to replace the listen address.

>  instead fo checking each connected peer which would be less efficient.

Yes, we will need to iterate on the connections and do a metadataBook.get.
But, we can also iterate the metadataBook and check if we have a connection to the peers that support hop. Listen on them if we have, and follow up on the ones we were not connected after. We will get some more complex logic and additional gets to the connections map, but isn't this worth than establishing new connections if we can use the ones we already have?

vasco-santos: > The immediate redial problem should really be solved with backoffs (connection gating, plus maybe a ttl tag in the peer store), especially if they disconnected from us.

Yeah, so I will tackle that together with the connection manager stuff.

src/circuit/listener.js
vasco-santos: I will have the comments and do it as a follow up after all. I found a bug on connect + disconnect because the addr were not certified anymore. Then I remembered that we still did not support invalidating the self record on multiaddr change: https://github.com/libp2p/js-libp2p/blob/v0.29.0/src/identify/index.js#L319

I will do that in the follow up PR with the identify push updates

vasco-santos:
This is ready for review again @vasco-santos , the expected follow up PRs are described in the main post of the PR:

1. When libp2p updates its addresses, it should update the network - Identify push on add/remove listen + self peer record update on addr change
2. Query the network if it needs more relays
3. New connectionManager integration for disconnect improvement

Note: codecov jobs are not working properly atm. I rebuild everything, but it is failing the `gcov command`. So, some of the annotations in the PR diff are not valid as there are actually tests for that flows, but were added in the last commit

src/circuit/auto-relay.js
```diff
@@ -0,0 +1,230 @@
...
+    if (this._listenRelays.size >= this.maxListeners) {
+      return
+    }
+
+    // Create relay listen addr
+    let listenAddr, remoteMultiaddr
+
+    try {
+      const remoteAddrs = this._peerStore.addressBook.get(connection.remotePeer)
+      remoteMultiaddr = remoteAddrs.find(a => a.isCertified).multiaddr // Get first announced address certified
```
jacobheun, line 128: This isn't going to be sufficient, we'll need to try and prioritize public addresses for this. It's still pretty common for peers to advertise private addresses. We can make a note to do this in a followup PR though.

HOP relays should really avoid advertising private addresses (we should document this in a "setting up relays" section of the production guides), but we can't rely on this behavior.

src/circuit/auto-relay.js
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    if (!remoteMultiaddr.protoNames().includes('p2p')) {
+      listenAddr = `${remoteMultiaddr.toString()}/p2p/${connection.remotePeer.toB58String()}/p2p-circuit`
+    } else {
+      listenAddr = `${remoteMultiaddr.toString()}/p2p-circuit`
+    }
+
+    // Attempt to listen on relay
+    this._listenRelays.add(id)
```
jacobheun, line 141: Why not just do the add after the listen call in the try? Then the delete isn't required if it fails.

src/circuit/listener.js
```diff
@@ -24,7 +34,7 @@ module.exports = (circuit) => {
   listener.listen = async (addr) => {
     const addrString = String(addr).split('/p2p-circuit').find(a => a !== '')
 
-    const relayConn = await circuit._dialer.connectToPeer(multiaddr(addrString))
+    const relayConn = await libp2p.dial(multiaddr(addrString))
     const relayedAddr = relayConn.remoteAddr.encapsulate('/p2p-circuit')
```
jacobheun, line 38: Something to consider here is that in AutoRelay we are creating a listen addr and then calling `transportManager.listen` to "connect". We will already have a connection per the AutoRelay logic, so that address is going to get thrown out and replaced with whatever this address is. We need to do some reconciliation here, I would find it odd to call listen with one address, and then have my actual address end up being something different.

I think this likely won't be a huge issue for initial relay connections, but if we reconnect to known relays, this address could change. The provided address **should** be the address we end up with, but if it fails for some reason we will dial other known addresses for the peer.

src/circuit/auto-relay.js
vasco-santos: Yes, we have a milestone for it, but I will add a comment: https://github.com/libp2p/js-libp2p/issues/699#issuecomment-667122074

I am not sure yet on the best approach for this, but it is being tracked

src/circuit/auto-relay.js
vasco-santos: Well, we will block inside the following `try ... catch` block, waiting for the `transportManager.listen`.
If we get multiple calls at the same time, all of them will not be blocked from reaching the `transportManager.listen` and we will probably end up with more listenRelays than needed.
Perhaps, I can use a `p-queue` instead? Otherwise, I might also cancel others that fail (should not expect `transportManager.listen` to fail, but it can)

src/circuit/listener.js
vasco-santos: > We need to do some reconciliation here, I would find it odd to call listen with one address, and then have my actual address end up being something different.

That is true! I changed this to avoid creating multiple connections to the same peer, but I agree with your point.
So you suggest that we go back and use the `connectToPeer` with a fallback to the dial if we fail?

src/circuit/auto-relay.js
jacobheun: I think we can leave it like this for now. It really shouldn't fail, we're already connected. If it does fail that should mean the connection was dropped during the listen attempt, which should trigger us to connect to another known relay if it exists.

src/circuit/listener.js
jacobheun: > So you suggest that we go back and use the connectToPeer with a fallback to the dial if we fail?

No. We shouldn't care how we're connected to the peer, really, but we need to be careful about the address we're advertising for this. Something like: Prioritize public addresses, if the connected address matches one of those use it, otherwise pick one of the others.

Again, I think we can do a follow PR for this, so we can focus on clear tests for that.

jacobheun:
CI is failing, but otherwise I think we can merge once that's fixed and continue work in subsequent PRs

vasco-santos:
Done @jacobheun ! CI test are good, don't know what happened before, but just removed cache and re-run and everything got good again. However, the codecov CI got broken in this PR and it never ran again 🤷‍♂️

jacobheun (approved):

PR #724, opened 2020-08-04
fix(identify): make agentversion dynamic and add it to the peerstore

jacobheun:
The AgentVersion in identify shouldn't be static. In go-libp2p the AgentVersion is generated by the application name and version, so for go-ipfs this looks like `go-ipfs/<version>`. Ideally we would replicate this, however this is a trickier to handle in JS as we need to account for browser usage. At the very least we need to default to the js-libp2p version, so this adds that basic level of support.

This is extremely helpful for understanding what versions are being used on the network, which helps us when determining upgrade strategies and change rollouts on the network.

go-libp2p also tracks the AgentVersion of peers in the metadata book, so I added that here as well, which is necessary for actually checking this data (we were previously ignoring this).

### DIFF
--- a/src/identify/index.js
+++ b/src/identify/index.js
@@ -175,6 +175,7 @@ class IdentifyService {
     // Update peers data in PeerStore
     this.peerStore.addressBook.set(id, listenAddrs.map((addr) => multiaddr(addr)))
     this.peerStore.protoBook.set(id, protocols)
+    this.peerStore.metadataBook.set(id, 'AgentVersion', Buffer.from(message.agentVersion))
 
     // TODO: Track our observed address so that we can score it
     log('received observed address of %s', observedAddr)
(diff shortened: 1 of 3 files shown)


src/identify/index.js
```diff
@@ -175,6 +175,7 @@ class IdentifyService {
     // Update peers data in PeerStore
     this.peerStore.addressBook.set(id, listenAddrs.map((addr) => multiaddr(addr)))
     this.peerStore.protoBook.set(id, protocols)
+    this.peerStore.metadataBook.set(id, 'AgentVersion', Buffer.from(message.agentVersion))
```
vasco-santos, line 178: I was thinking that `metadataBook` would only be used by the application layer.
I think it is useful to store the agentVersion, but I wonder if we should have a way of identifying who stored data in the metadataBook

jacobheun: If there is a need to separate them in the future we could make an internal book for libp2p but I'd prefer to avoid that unless a problem arises. Worst case scenario users/apps can override values, but if we document what libp2p stores there I think it should be fine.

PR #725, opened 2020-08-04
feat(keychain): add support for ed25519 and secp keys

jacobheun:
Adds support in the keychain for ed25519 and secp256k1 keys.

- [x] requires https://github.com/libp2p/js-libp2p-crypto/pull/179

unblocks https://github.com/libp2p/js-libp2p/issues/669
unblocks usage of these keys in js-ipfs and any other apps leveraging the keychain

### DIFF
--- a/src/keychain/index.js
+++ b/src/keychain/index.js
@@ -7,6 +7,7 @@ const crypto = require('libp2p-crypto')
 const DS = require('interface-datastore')
 const CMS = require('./cms')
 const errcode = require('err-code')
+const { Number } = require('ipfs-utils/src/globalthis')
 
 require('node-forge/lib/sha512')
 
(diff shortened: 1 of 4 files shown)


src/keychain/index.js
```diff
@@ -174,7 +176,7 @@ class Keychain {
    * @param {int} size - The key size in bits.
```
vasco-santos, line 176: default value and optional?

jacobheun: I updated this to be optional and updated the logic to remove the default. 0 doesn't really make sense, so now this just checks the size when it matters (when type is rsa). Also updated the api docs to reflect this.

vasco-santos: That makes sense, cool!

PR #726, opened 2020-08-04
fix: allow certain keychain operations without a password

achingbrain:
Listing, removing, renaming etc keys do not require a password so
the user should not be required to provide one.

This means we don't have to prompt the user to create a password
when they aren't going to do any operations that require a password.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -82,7 +82,7 @@ class Libp2p extends EventEmitter {
     }
 
     // Create keychain
-    if (this._options.keychain && this._options.keychain.pass && this._options.keychain.datastore) {
+    if (this._options.keychain && this._options.keychain.datastore) {
       log('creating keychain')
 
       const keychainOpts = Keychain.generateOptions()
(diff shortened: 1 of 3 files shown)


jacobheun:
I've update this to not require a password. I'll create an issue for supporting a password rotate feature. We should be able to set new passwords on they keychain. This will give us the ability to start without one and then "rotate" to one later.

vasco-santos:
@achingbrain is this enough for `js-ipfs` to use libp2p's keychain without a password?

I believe that this needs to be changed: https://github.com/libp2p/js-libp2p/blob/master/src/index.js#L85

jacobheun:
> I believe that this needs to be changed: https://github.com/libp2p/js-libp2p/blob/master/src/index.js#L85

Good catch, I rebased with master and added a test for libp2p creation without a keychain pass.

jacobheun:
FYI I verified it fixes js-ipfs, https://github.com/ipfs/js-ipfs/pull/3212. I'm working on bubbling it up there.

achingbrain:
Thanks for pulling this across the line!

PR #730, opened 2020-08-12
fix: replace node buffers with uint8arrays

achingbrain:
Upgrades all deps and replaces all use of node Buffers with Uint8Arrays

Depends on:

- [x] https://github.com/NodeFactoryIo/js-libp2p-noise/pull/70
- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/62
- [x] https://github.com/libp2p/interop/pull/42

BREAKING CHANGES:

- All deps used by this module now use Uint8Arrays in place of node Buffers


achingbrain:
This is using a temporary fork of libp2p-noise that has the changes from NodeFactoryIo/js-libp2p-noise#70 applied and the typescript compiled to js.

jacobheun:
Looks like the browser tests are failing

test/keychain/cms-interop.spec.js
```diff
@@ -58,7 +50,7 @@ cn4oisOvxCprs4aM9UVjtZTCjfyNpX8UWwT1W3rySV+KQNhxuMy3RzmL
     expect(key.id).to.equal(aliceKid)
   })
 
-  it('decrypts node-forge example', async () => {
+  it.only('decrypts node-forge example', async () => {
```
vasco-santos, line 53:
```suggestion
  it('decrypts node-forge example', async () => {
```

achingbrain: No wonder the tests were passing!

vasco-santos (approved):
LGTM!
pending the interface release

jacobheun:
Ci is failing due to interop, this is a release order issue with the daemon. We can resolve this before release.

PR #733, opened 2020-08-14
feat: gossipsub 1.1

vasco-santos:
This PR main purpose is to update gossipsub 1.1

With this initiative, we included a refactor of the pubsub abstraction API included in `js-libp2p` module per https://github.com/libp2p/js-libp2p-pubsub/issues/68 , as well as a pubsub base implementation ([libp2p/js-libp2p-pubsub](https://github.com/libp2p/js-libp2p-pubsub)) full refactor that also included moving it into libp2p interfaces per https://github.com/libp2p/js-libp2p-interfaces/issues/53

With the above changes, `libp2p-floodsub` was also updated accordingly.

Needs:

- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/60 (with https://github.com/libp2p/js-libp2p-pubsub/tree/v0.6.x)
- [x] [libp2p/js-libp2p-floodsub#108](https://github.com/libp2p/js-libp2p-floodsub/pull/108)
- [ ] [ChainSafe/js-libp2p-gossipsub#82](https://github.com/ChainSafe/js-libp2p-gossipsub/pull/82)

- [ ] Interop circular dependency https://github.com/libp2p/interop/pull/43


doc/API.md
```diff
@@ -1382,7 +1383,129 @@ const handler = (msg) => {
...
+  const validInputs = ['a', 'b', 'c']
+
+  if (!validInputs.includes(input)) {
+    throw new Error('no valid input received')
+  }
+}
+libp2p.pubsub.topicValidators.set(topic, validateMessage)
+```
+
+## pubsub.topicValidators.delete
```
vasco-santos, line 1486: Needs to be added to the toc

doc/API.md
```diff
@@ -1382,7 +1383,129 @@ const handler = (msg) => {
...
+  // msg.data - pubsub data received
+}
+
+libp2p.pubsub.removeListener(topic handler)
+libp2p.pubsub.unsubscribe(topic)
+```
+
+## pubsub.topicValidators.set
+
+Pubsub routers support message validators per topic, which will validate the message before its propagations. Set is used to add a validator for a topic.
```
jacobheun, line 1454: 'add' makes it sound like multiple validators can be added per topic which is not the case
```suggestion
Pubsub routers support message validators per topic, which will validate the message before its propagations. Set is used to specify a validator for a topic.
```

package.json
```diff
@@ -59,7 +59,7 @@
     "it-pipe": "^1.1.0",
     "it-protocol-buffers": "^0.2.0",
     "libp2p-crypto": "^0.18.0",
-    "libp2p-interfaces": "^0.5.0",
+    "libp2p-interfaces": "libp2p/js-libp2p-interfaces#feat/interface-pubsub",
```
jacobheun, line 62:
```suggestion
    "libp2p-interfaces": "^0.5.1",
```

src/index.js
```diff
@@ -185,9 +185,12 @@ class Libp2p extends EventEmitter {
...
     }
 
-    // start pubsub
+    // Create pubsub if provided
     if (this._modules.pubsub) {
-      this.pubsub = pubsub(this, this._modules.pubsub, this._config.pubsub)
+      const Pubsub = this._modules.pubsub
+      // using pubsub adapter with *DEPRECATED* handlers functionality
+      this.pubsub = PubsubAdapter(Pubsub, this, this._config.pubsub)
+      // this.pubsub = new Pubsub(this, this._config.pubsub)
```
jacobheun, line 193:
```suggestion
```

src/pubsub-adapter.js
```diff
@@ -0,0 +1,39 @@
...
+     * @returns {void}
+     */
+    unsubscribe (topic, handler) {
+      if (!handler) {
+        this.removeAllListeners(topic)
+      } else {
+        this.removeListener(topic, handler)
+      }
+
+      super.unsubscriber(topic)
```
jacobheun, line 34: For the legacy api we need to check the listener count.
```suggestion
      if (this.listenerCount(topic) === 0) {
        super.unsubscribe(topic)
      }
```

src/pubsub-adapter.js
```diff
@@ -0,0 +1,39 @@
...
+     * @override
+     * @param {string} topic
+     * @param {function(msg: InMessage)} [handler]
+     * @returns {void}
+     */
+    subscribe (topic, handler) {
+      super.subscribe(topic)
+
+      // Bind provided handler
+      handler && this.on(topic, handler)
```
jacobheun, line 17: It shouldn't matter, but we should bind the listener first. This is also consistent with the examples for the new api.

```suggestion      
      // Bind provided handler
      handler && this.on(topic, handler)
      super.subscribe(topic)
```

jacobheun (commented):
Just a couple things. I think once those are fixed we can do an RC since only the dev dependencies are still referencing branches. This make it easier for those to release their beta versions against the RC.

jacobheun (approved):
lgtm

jacobheun:
merging this and will do an RC, we can update deps prior to release

PR #736, opened 2020-08-24
docs: migration 0.28 to 0.29

vasco-santos:
This PR adds the migration guide from 0.28 to 0.29


doc/migrations/v0.28-v.29.md
```diff
@@ -0,0 +1,271 @@
...
+  const from = msg.from
+  // process message
+  throw new Error('not a valid message')
+}
+libp2p.pubsub.topicValidators.set(topic, validator)
+```
+
+### Uint8Arrays replace node Buffers
+
+Aiming to improve libp2p browser support, we are moving away from node core modules unless we can guarantee that the code we are writing will not run in a browser. It is worth to mention that modern JavaScript runtimes have TypedArrays such as Uint8Array backed by ArrayBuffers. All libp2p dependencies were also updated to use Uint8Array.
```
jacobheun, line 141:
```suggestion
Aiming to improve libp2p browser support, we are moving away from node core modules unless we can guarantee that the code we are writing will not run in a browser. It is worth mentioning that modern JavaScript runtimes have TypedArrays such as Uint8Array backed by ArrayBuffers. All libp2p dependencies were also updated to use Uint8Array.
```

doc/migrations/v0.28-v.29.md
```diff
@@ -0,0 +1,271 @@
...
+```js
+const validator = (msgTopic, msg) => {
+  const from = msg.from
+  // process message
+  throw new Error('not a valid message')
+}
+libp2p.pubsub.topicValidators.set(topic, validator)
+```
+
+### Uint8Arrays replace node Buffers
```
jacobheun, line 139: We're using several utility functions below. I would add a short ref to the uint8arrays module and show the various imports that are being used below.

jacobheun (approved):
This looks good, although both file paths are wrong. Should be:

```
doc/migrations/v0.27-v0.28.md
doc/migrations/v0.28-v0.29.md 
```

instead of 

```
doc/migrations/v0.27-v.0.28.md 
doc/migrations/v0.28-v.29.md 
```

PR #739, opened 2020-08-26
fix: peer record interop with go

jacobheun:
The peer record was using its string representation as the Record codec, which is not accurate, it should be a Uint8Array `Uint8Array.from([3, 1])` (0x0301) instead of `"libp2p-peer-record"`, this caused the record to be improperly marshaled.


jacobheun:
Interop is now passing in CI with this change using https://github.com/libp2p/interop/pull/43

vasco-santos (approved):
LGTM! 🎉

PR #745, opened 2020-09-08
chore: update docs after secio deprecation announcement

vasco-santos:
As we announced that we're removing support for the SECIO security transport, I removed its usage from the documentation, so that new users do not use it.

When we official deprecate it, we should remove it from the `CONFIGURATION.md` + Packages table and remove it as a fallback in the examples.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -2,7 +2,7 @@
 
 * [Static Functions](#static-functions)
   * [`create`](#create)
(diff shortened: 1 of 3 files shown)


jacobheun (approved):
❤️  LGTM, we should also update the examples, but that can be done as a followup.

vasco-santos:
Got it, I will also update the examples!

PR #746, opened 2020-09-08
chore: add libp2p-examples repo to release checklist

vasco-santos:
We did not update `libp2p-examples` with the uint8array refactor.

I suggest that we add that task in our release process to avoid having inconsistencies between the latest `js-libp2p` and what is in `libp2p-examples#master`. Meanwhile, I will update the `libp2p-examples` with the latest release.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -26,6 +26,7 @@
 - Documentation
   - [ ] Ensure that README.md is up to date
   - [ ] Ensure that all the examples run
+  - [ ] Ensure [libp2p/js-libp2p-examples](https://github.com/libp2p/js-libp2p-examples) is updated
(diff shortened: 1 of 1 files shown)


vasco-santos:
I just noticed that you already did it in a branch @jacobheun https://github.com/libp2p/js-libp2p-examples/commit/a266e9370db587fb7c339bd6ffaaa56fc3143ef1

Should we get that in and a reference of `0.28` in the example readme?

jacobheun:
> Should we get that in and a reference of 0.28 in the example readme?

Yes, it still needs the browser chapters updated though, I only got through the node version.

vasco-santos:
> Yes, it still needs the browser chapters updated though, I only got through the node version.

I can update the browser side of things in the branch then

jacobheun (approved):

PR #747, opened 2020-09-08
chore: examples not using secio

vasco-santos:
As we announced that we're removing support for the SECIO security transport, this PR removes its usages from the examples per https://github.com/libp2p/js-libp2p/pull/745#pullrequestreview-484066297


jacobheun (approved):
LGTM. fyi I rebased with master and pushed a couple small commits to correct some things I noticed while going through the examples and docs related to secio/noise.

PR #748, opened 2020-09-10
chore: auto relay multiaddr update push

vasco-santos:
This PR is a follow up of the [main implementation](https://github.com/libp2p/js-libp2p/pull/723) of auto relay to inform the connected peers of the new addresses to announce, by sending the new signed peer record via `identify-push`.

The self peer record update logic was added to create the new record, when the multiaddr changes. Ping unmount needed to be modified to before the transports close, otherwise when the ping unhandle protocol is triggered, the `identify-push` will try to access the addresses when they do not exist, causing unhandled exceptions, such as https://github.com/libp2p/js-libp2p-tcp/blob/v0.15.1/src/listener.js#L97

When we support dynamic transport updates, we should automate the self peer record update based on events for updated addresses. This will enable us to not verify if the addresses are outdated each time we want to send the self peer record envelope.

Needs:

- [x] #723

### DIFF
--- a/src/circuit/listener.js
+++ b/src/circuit/listener.js
@@ -20,8 +20,8 @@ module.exports = (libp2p) => {
     const deleted = listeningAddrs.delete(connection.remotePeer.toB58String())
 
     if (deleted) {
-      // TODO push announce multiaddrs update
-      // libp2p.identifyService.pushToPeerStore()
+      // Announce multiaddrs update on listen success
+      libp2p.identifyService.pushToPeerStore()
     }
   })
 
--- a/src/circuit/listener.js
+++ b/src/circuit/listener.js
@@ -20,8 +20,8 @@ module.exports = (libp2p) => {
     const deleted = listeningAddrs.delete(connection.remotePeer.toB58String())
 
     if (deleted) {
-      // Announce multiaddrs update on listen success
-      libp2p.identifyService.pushToPeerStore()
+      // Announce listen addresses change
+      listener.emit('listening')
     }
   })
 
--- a/src/circuit/listener.js
+++ b/src/circuit/listener.js
@@ -21,7 +21,7 @@ module.exports = (libp2p) => {
 
     if (deleted) {
       // Announce listen addresses change
-      listener.emit('listening')
+      listener.emit('close')
     }
   })
 
--- a/src/identify/index.js
+++ b/src/identify/index.js
@@ -314,13 +314,30 @@ class IdentifyService {
    * @return {Uint8Array}
    */
   async _getSelfPeerRecord () {
-    const selfSignedPeerRecord = this.peerStore.addressBook.getRawEnvelope(this.peerId)
+    // Update self peer record if needed
+    await this._createOrUpdateSelfPeerRecord()
 
-    // TODO: support invalidation when dynamic multiaddrs are supported
-    if (selfSignedPeerRecord) {
-      return selfSignedPeerRecord
+    return this.peerStore.addressBook.getRawEnvelope(this.peerId)
+  }
+
+  /**
+   * Creates or updates the self peer record if it exists and is outdated.
+   * @return {Promise<void>}
+   */
+  async _createOrUpdateSelfPeerRecord () {
+    const selfPeerRecordEnvelope = await this.peerStore.addressBook.getPeerRecord(this.peerId)
+
+    if (selfPeerRecordEnvelope) {
+      const peerRecord = PeerRecord.createFromProtobuf(selfPeerRecordEnvelope.payload)
+
+      const mIntersection = peerRecord.multiaddrs.filter((m) => this._libp2p.multiaddrs.some((newM) => m.equals(newM)))
+      if (mIntersection.length === this._libp2p.multiaddrs.length) {
+        // Same multiaddrs as already existing in the record, no need to proceed
+        return
+      }
     }
 
+    // Create / Update Peer record
     try {
       const peerRecord = new PeerRecord({
         peerId: this.peerId,
@@ -328,12 +345,9 @@ class IdentifyService {
       })
       const envelope = await Envelope.seal(peerRecord, this.peerId)
       this.peerStore.addressBook.consumePeerRecord(envelope)
-
-      return this.peerStore.addressBook.getRawEnvelope(this.peerId)
     } catch (err) {
-      log.error('failed to get self peer record')
+      log.error('failed to create self peer record')
     }
-    return null
   }
 }
 
--- a/src/identify/index.js
+++ b/src/identify/index.js
@@ -64,11 +64,6 @@ class IdentifyService {
     this.connectionManager = libp2p.connectionManager
 
     /**
-     * @property {TransportManager}
-     */
-    this.transportManager = libp2p.transportManager
-
-    /**
      * @property {PeerId}
      */
     this.peerId = libp2p.peerId
@@ -88,10 +83,11 @@ class IdentifyService {
       this.identify(connection, peerId).catch(log.error)
     })
 
-    // When new addresses are used for listening, update self peer record
-    this.transportManager.on('listening', async () => {
-      await this._createSelfPeerRecord()
-      this.pushToPeerStore()
+    // When self multiaddrs change, trigger identify-push
+    this.peerStore.on('change:multiaddrs', ({ peerId }) => {
+      if (peerId.toString() === this.peerId.toString()) {
+        this.pushToPeerStore()
+      }
     })
   }
 
@@ -101,7 +97,7 @@ class IdentifyService {
    * @returns {Promise<void>}
    */
   async push (connections) {
-    const signedPeerRecord = await this._getSelfPeerRecord()
+    const signedPeerRecord = await this.peerStore.addressBook.getRawEnvelope(this.peerId)
     const listenAddrs = this._libp2p.multiaddrs.map((ma) => ma.bytes)
     const protocols = Array.from(this._protocols.keys())
 
@@ -250,7 +246,7 @@ class IdentifyService {
       publicKey = this.peerId.pubKey.bytes
     }
 
-    const signedPeerRecord = await this._getSelfPeerRecord()
+    const signedPeerRecord = await this.peerStore.addressBook.getRawEnvelope(this.peerId)
 
     const message = Message.encode({
       protocolVersion: PROTOCOL_VERSION,
@@ -318,40 +314,6 @@ class IdentifyService {
 
     // Update the protocols
     this.peerStore.protoBook.set(id, message.protocols)
-  }
-
-  /**
-   * Get self signed peer record raw envelope.
-   * @return {Promise<Uint8Array>}
-   */
-  _getSelfPeerRecord () {
-    const selfSignedPeerRecord = this.peerStore.addressBook.getRawEnvelope(this.peerId)
-
-    if (selfSignedPeerRecord) {
-      return selfSignedPeerRecord
-    }
-
-    return this._createSelfPeerRecord()
-  }
-
-  /**
-   * Create self signed peer record raw envelope.
-   * @return {Uint8Array}
-   */
-  async _createSelfPeerRecord () {
-    try {
-      const peerRecord = new PeerRecord({
-        peerId: this.peerId,
-        multiaddrs: this._libp2p.multiaddrs
-      })
-      const envelope = await Envelope.seal(peerRecord, this.peerId)
-      this.peerStore.addressBook.consumePeerRecord(envelope)
-
-      return this.peerStore.addressBook.getRawEnvelope(this.peerId)
-    } catch (err) {
-      log.error('failed to get self peer record')
-    }
-    return null
   }
 }
 
--- a/src/identify/index.js
+++ b/src/identify/index.js
@@ -63,11 +63,10 @@ class IdentifyService {
      */
     this.connectionManager = libp2p.connectionManager
 
-    this.connectionManager.on('peer:connect', (connection) => {
-      const peerId = connection.remotePeer
-
-      this.identify(connection, peerId).catch(log.error)
-    })
+    /**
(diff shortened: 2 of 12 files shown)


src/index.js
```diff
@@ -258,6 +258,7 @@ class Libp2p extends EventEmitter {
       await this.peerStore.stop()
       await this.connectionManager.stop()
 
+      ping.unmount(this)
```
vasco-santos, line 261: ping unhandle protocol is triggered, the identify-push will try to access the addresses when they do not exist

jacobheun: This sounds like a bug. After the TransportManager is closed, we should have no active listeners, which should just yield an empty array of addresses.

This approach is also not good because we should not be doing an identify push when we are shutting down and this change is going to cause that to happen.

src/identify/index.js
```diff
@@ -314,26 +314,40 @@ class IdentifyService {
...
-    if (selfSignedPeerRecord) {
-      return selfSignedPeerRecord
+    return this.peerStore.addressBook.getRawEnvelope(this.peerId)
+  }
+
+  /**
+   * Creates or updates the self peer record if it exists and is outdated.
+   * @return {Promise<void>}
+   */
+  async _createOrUpdateSelfPeerRecord () {
```
jacobheun, line 327: This method addition seems unnecessary? This just makes `_getSelfPeerRecord` pretty much useless.

Also, we should not be doing update checks in here, we should just update the record when an actual change happens. This is going to cause a lot of unnecessary address checks.

src/circuit/listener.js
```diff
@@ -20,8 +20,8 @@ module.exports = (libp2p) => {
     const deleted = listeningAddrs.delete(connection.remotePeer.toB58String())
 
     if (deleted) {
-      // TODO push announce multiaddrs update
-      // libp2p.identifyService.pushToPeerStore()
+      // Announce multiaddrs update on listen success
+      libp2p.identifyService.pushToPeerStore()
```
jacobheun, line 24: This is being called in a couple places, I think we shouldn't be doing this here.

Perhaps instead we should be updating our addresses in the PeerStore? We don't currently store them there, but it has an event system for changes, which we could just listen to for triggering a self record update and identify push. This enables the identify system to subscribe to the changes and manage everything itself, instead of other systems like this one needing to know these specifics.

src/index.js
vasco-santos: If we do `await this.transportManager.close()` we will stop the transport listeners. With this, we have an error on `getAddrs()`, at list from `libp2p-tcp`: https://github.com/libp2p/js-libp2p-tcp/blob/master/src/listener.js#L96

So, when we do `ping.unmount`, it will call the `unhandle` which will try to do the `identify-push` and use the `libp2p.multiaddrs` getter, which uses the transport `getAddrs`, which fail as noted above. 

Or we unmount before, which in my opinion is correct as we are also unmounting the other protocols before, or we need to change the `transportManager.getAddrs()` to catch the transport.getAddrs error

vasco-santos: > This approach is also not good because we should not be doing an identify push when we are shutting down and this change is going to cause that to happen.

We are doing that for all other protocols. I think we should go with a more central fix for that. Before identify push, we should check that libp2p is not being stopped. What do you think?

vasco-santos: We have a check for `if (this.isStarted() && this.identifyService) {`. We can have a `._isStopping` property, or just turn false for isStarted before stopping the protocols

src/circuit/listener.js
vasco-santos: We can do that, but it will be tricky.

To enable this, we need to be able to track transportManager `getAddrs()` changes.  With this, once these addresses change, who is responsible to tracking this (perhaps the transportManager while it tracks the transports addrs) needs to update the addressBook for self addresses (with addressBook.set or addressBook.add).

This will cause the event to be triggered as you mentioned, but it will also cause an issue. So, the `identifyService` would listen on address changes for the self peer id and it would create the new self signed peer record, storing it in the peerStore. But, using `addressBook.consumePeerRecord()` will currently trigger a new `address:change` event and get in a loop. 

Regarding the second point, we can use Book `_setData` emit option and put it false if we had the same multiaddrs as before. But, I think we are complicating this a lot. 
I would go with when a self address change, we just create the new peer record and do `addressBook.consumePeerRecord()`, which triggers the event that identify would listen on and Identify does the Push. We can do the same to add own protocol to the protoBook with `add` on handle and remove on unhandle. With this approach, we would be removing the self peer record creation from Identify. We can go back the my first solution using a Record Manager, or we can figure out a way of making this part of the Address Manager.

I would like to have in the future to support dynamic address announce/noAnnounce changes via an API.  So improving the AddressManager to both be responsible for listening for addresses change in the TransportManager, or to the announce/noAnnounce and create/store the self signed peer record would be a good solution. The first point will be difficult to change though.

src/index.js
jacobheun: > With this, we have an error on getAddrs(), at list from libp2p-tcp: https://github.com/libp2p/js-libp2p-tcp/blob/master/src/listener.js#L96

`close` sets all the listeners to empty, https://github.com/libp2p/js-libp2p/blob/v0.29.0/src/transport-manager.js#L72, and then `getAddrs` iterates over that list, which should immediately exit. The transport should **never** get called because it should already be cleaned up. I think there is a teardown race here.

src/circuit/listener.js
jacobheun: Let's table this for now, it's not a huge deal since it's all internal and I don't think we need to conflate this prematurely.

jacobheun: Actually, I think we need to solve this to avoid the unnecessary change checks in `_getSelfPeerRecord` which is going to run on every identify request.

src/index.js
vasco-santos: Ok, so I already understood the issue. In the test that causes this error we do:

```js
await libp2p.stop()
await libp2p.start()
```

When we do the libp2p stop, the `ping.unmount()` is not async. So the node will stop and going to restart while at the same time `ping.unmount` will call the unhandle while tcp server.listen has already kicked in from start. 

The best solution in my opinion is to add the `_isStopping` property. More than fixing this issue, it will block the identify-push from pubsub.stop, dht.stop from being triggered.

vasco-santos: For now, I just changed the `the._isStarted = false` for the beginning of stop. It seems not problematic to do this, as even if the stop code throws it would be assigned to false either way. We might get in a weird state if an error happens, but it would be problematic with the previous logic as well. A restart will fix the weird state

src/circuit/listener.js
vasco-santos: Yeah, checking if is different anytime is not a good call. So this is my proposal:

- transportManager will listen for transport [listening events](https://github.com/libp2p/js-libp2p-interfaces/tree/master/src/transport#api).
- when it triggers, transportManager will emit a `listening` event with `getAddrs()` result. It might be also useful for application layer, for example ipfs could use this for the daemon stdout of addresses listening on.
- AddressManager/IdentifyService tracks this event and is responsible for creating the self peer record and add it to the AddressBook. Self addresses will also be updated.

For not complicating too much now, I think we can keep this on the IdentifyService and move it to the Address Manager when we support dynamic API changes. Both need `start` and `stop` methods for the transportManager event listener. The event listener should only be setup after all the initial listeners so that we do not create signed peer records per listener in startup. This way, we will only track updates and keep the current behaviour of creating if it does not exist in the first identify call. If we went with the AddressManager, we would need to have extra logic for this initial creation, or the identifyService would need to call a `addressManager.createSelfPeerRecord` if there was none in the peerStore.

src/circuit/listener.js
```diff
@@ -20,8 +20,8 @@ module.exports = (libp2p) => {
     const deleted = listeningAddrs.delete(connection.remotePeer.toB58String())
 
     if (deleted) {
-      // TODO push announce multiaddrs update
-      // libp2p.identifyService.pushToPeerStore()
+      // Announce listen addresses change
+      listener.emit('listening')
```
vasco-santos, line 24: Not a fan of using the `listening` event here. But, the other option is to have a new event for the not listening or trigger a manual update of the peerRecord/addressBook self. I was trying to be compliant with the `interface-transport`. 
What do you think @jacobheun ?

vasco-santos:
@jacobheun 
I added two commits two fix the above review

- assign `._isStarted` to false earlier on stop
  - it would be false either way, if any error is thrown
  - it avoids doing identify-push for pubsub+dht that was happening before
  - it does not cause a race condition on ping unmount + fast node.stop() and node.start()
- use listening transport events to create self peer record on updates
  - there is one concern on removing listen addresses commented on the review,

src/circuit/listener.js
vasco-santos: I guess I can use the `close` event from the listeners and have a `transportManager.emit('change')`. But, circuit would need to emit close here and transportManager would track `close` and `listening`

src/transport-manager.js
```diff
@@ -150,6 +158,9 @@ class TransportManager {
         const listener = transport.createListener({}, this.onConnection)
         this._listeners.get(key).push(listener)
 
+        // Track listen events
+        listener.on('listening', () => this.emit('listening'))
```
jacobheun, line 162: I don't think the TransportManager needs to emit events. 

If we watch the `listening` and `close` events for each transport listener, instead of bubbling those events we can have the TransportManager itself create and seal the updated Peer Record. The TransportManager should always know when a listeners status changes, so pushing updates from here makes sense.

1. A transport begins/stops listening on an address
1. TransportManager sees the change and generates a new peer record and stores it the address book
1. Storing the new record triggers the multiaddr changes
1. Identify sees the multiaddr change event from the peer store for self and sends out a push

If we then store our protocols in the peer store instead of the upgrader, the identify service can also subscribe to those which keeps the api consistent.

vasco-santos: > If we then store our protocols in the peer store instead of the upgrader, the identify service can also subscribe to those which keeps the api consistent.

That I totally agree, and I will follow up a PR to store the protocols in protocol and subscribe to changes

jacobheun: The advantage to internalizing the address change to the transport manager is that it doesn't affect the API on any system at this point so making that change is low risk. I think the Address Manager owning this makes sense, but I'd like to see a design before we change those APIs, to avoid needing to change them in the future.

The Address Manager should also own our observed addresses, tracking them and scoring them accordingly. How should observed addresses flow into the Address Manager? Does it affect the API?

vasco-santos: Ok, so I went with your suggestion for now. I was not expecting to break the API adding this to the AddressManager as it would be basically the same as now in the TransportManager  + TransportManager being an event emitter.

Regarding the observed addresses I agree. So, I will start for creating an issue and we can discuss this when it is relevant

test/identify/index.spec.js
```diff
@@ -404,5 +412,53 @@ describe('Identify', () => {
...
+const _createSelfPeerRecord = async (libp2p) => {
+  try {
+    const peerRecord = new PeerRecord({
+      peerId: libp2p.peerId,
+      multiaddrs: libp2p.multiaddrs
+    })
+    const envelope = await Envelope.seal(peerRecord, libp2p.peerId)
+    libp2p.peerStore.addressBook.consumePeerRecord(envelope)
+  } catch (_) {}
+}
```
jacobheun, line 464: Move this to a utility file?

test/transports/transport-manager.node.js
```diff
@@ -40,10 +50,16 @@ describe('Transport Manager (TCP)', () => {
...
   it('should be able to listen', async () => {
+    sinon.spy(tm, '_createSelfPeerRecord')
+
     tm.add(Transport.prototype[Symbol.toStringTag], Transport)
     await tm.listen(addrs)
     expect(tm._listeners).to.have.key(Transport.prototype[Symbol.toStringTag])
     expect(tm._listeners.get(Transport.prototype[Symbol.toStringTag])).to.have.length(addrs.length)
+
+    // Created Self Peer record on new listen address
+    expect(tm._createSelfPeerRecord.callCount).to.equal(addrs.length)
```
jacobheun, line 61: Might be better to validate the record is being stored in the address book. Validating the correct record is being stored would also be useful.

src/transport-manager.js
```diff
@@ -215,6 +226,26 @@ class TransportManager {
...
+  async _createSelfPeerRecord () {
+    try {
+      const peerRecord = new PeerRecord({
+        peerId: this.libp2p.peerId,
+        multiaddrs: this.libp2p.multiaddrs
+      })
+      const envelope = await Envelope.seal(peerRecord, this.libp2p.peerId)
+      this.libp2p.peerStore.addressBook.consumePeerRecord(envelope)
+
+      return this.libp2p.peerStore.addressBook.getRawEnvelope(this.libp2p.peerId)
```
jacobheun, line 243: There's no need to return anything, nothing is consuming the result of this, so we should always return null.

nit: Calling this function `_updateSelfPeerRecord` might be a bit clearer about what it's doing.

test/identify/index.spec.js
```diff
@@ -404,5 +412,41 @@ describe('Identify', () => {
...
+
+      await libp2p.start()
+
+      sinon.spy(libp2p.identifyService, 'identify')
+      sinon.spy(libp2p.identifyService, 'push')
+
+      const connection = await libp2p.dialer.connectToPeer(remoteAddr)
+      expect(connection).to.exist()
+      // Wait for nextTick to trigger the identify call
+      await delay(1)
```
jacobheun, line 430: Is this needed? There is a wait immediately after this for identify.

jacobheun (approved):
LGTM just one small thing in the tests, feel free to merge when ready.

test/identify/index.spec.js
vasco-santos: No, it was a debugging remaining, just removing it

PR #749, opened 2020-09-11
feat: auto relay network query

vasco-santos:
This PR add to the auto-relay ambient discovery, the ability to query the network for providers of the hop relay service. This will be performed if there are not enough peers connected/known with that capability.

The concept of the libp2p discovery serviced is added here (still not formally, but will be once we integrate rendezvous).

The Circuit was decoupled into two parts, the Circuit Transport and the Relay Service. This decoupling is important, so that we have lifecycle methods (start and stop for the relay). These methods will be used for calling the `discovery.advertise` similarly to go-libp2p's relay implementation, as well as to future active relay service network query on start.

The advertise options were added with the [same](https://github.com/libp2p/go-libp2p/blob/fb3179e617bdf049734f58aa8ced2bb42de7ea5d/p2p/host/relay/relay.go#L12-L16) defaults as go

Needs:

- [x] https://github.com/libp2p/js-libp2p/pull/723
- [x]  https://github.com/libp2p/js-libp2p/pull/753 -- Needs 0.30.x rebase from master 
- [x] https://github.com/libp2p/js-libp2p/pull/748


test/relay/auto-relay.node.js
```diff
@@ -452,4 +460,125 @@ describe('auto-relay', () => {
...
+      // Wait for peer added as listen relay
+      await pWaitFor(() => libp2p.relay._autoRelay._addListenRelay.callCount === 1)
+      expect(libp2p.relay._autoRelay._listenRelays.size).to.equal(1)
+      await pWaitFor(() => libp2p.multiaddrs.length === originalMultiaddrsLength + 1)
+
+      const relayedAddr = libp2p.multiaddrs[libp2p.multiaddrs.length - 1]
+      libp2p2.peerStore.addressBook.set(libp2p2.peerId, [relayedAddr])
+
+      // Dial from peer 2 through the relayed address
+      const conn = await libp2p2.dial(libp2p2.peerId)
```
jacobheun, line 580: This is dialing itself, it should be failing. 

Also, I'd recommend changing the names of these nodes because `libp2p2` and `libp2p` is difficult to quickly distinguish. (local/remote, source/dest, etc)

src/circuit/auto-relay.js
```diff
@@ -224,7 +233,32 @@ class AutoRelay {
       }
     }
 
-    // TODO: Try to find relays to hop on the network
+    // Try to find relays to hop on the network
+    try {
+      const cid = await namespaceToCid(RELAY_RENDEZVOUS_NS)
+      for await (const provider of this._libp2p.contentRouting.findProviders(cid)) {
+        if (!provider || !provider.id || !provider.multiaddrs || !provider.multiaddrs.length) {
```
jacobheun, line 240: Are all these checks necessary? This feels like a bug in the API if so

src/circuit/auto-relay.js
```diff
@@ -224,7 +233,32 @@ class AutoRelay {
       }
     }
 
-    // TODO: Try to find relays to hop on the network
+    // Try to find relays to hop on the network
+    try {
+      const cid = await namespaceToCid(RELAY_RENDEZVOUS_NS)
+      for await (const provider of this._libp2p.contentRouting.findProviders(cid)) {
```
jacobheun, line 239: We don't have support to do this right now, but it would be great to actually prioritize the results of this based on latency of the connection. Worth adding an issue for a future enhancement here I think

src/circuit/constants.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+
+const minute = 60 * 1000
+
+module.exports = {
```
jacobheun, line 5: We should document these and clarify what each are for, some are clearer than others.

src/circuit/index.js
```diff
@@ -1,197 +1,75 @@
...
+  async _advertiseService () {
+    try {
+      const cid = await namespaceToCid(RELAY_RENDEZVOUS_NS)
+      await this._libp2p.contentRouting.provide(cid)
+    } catch (err) {
+      if (err.code === 'NO_ROUTERS_AVAILABLE') {
+        log('there are no routers configured to advertise hop relay service')
+      } else {
+        log.error(err)
+      }
```
jacobheun, line 65: There's no return so if this fails it's just going to keep starting the timeout, is that intended? I think calling `stop` and returning would be a clearer approach.

src/circuit/index.js
```diff
@@ -1,197 +1,75 @@
...
+    }
 
-    return multiaddrs.filter((ma) => {
-      return mafmt.Circuit.matches(ma)
-    })
+    // Restart timeout
+    this._timeout = setTimeout(() => {
+      this._advertiseService()
+    }, this._options.advertise.ttl || ADVERTISE_TTL)
```
jacobheun, line 71: We should just set the options in the constructor so we don't have to keep checking the ttl existence.

src/circuit/index.js
```diff
@@ -1,197 +1,75 @@
...
-      disconnectOnFailure && await relayConnection.close()
-      throw err
+  start () {
+    // Advertise service if HOP enabled
+    const canHop = this._options.hop.enabled
+
+    if (canHop) {
+      this._timeout = setTimeout(() => {
+        this._advertiseService()
+      }, this._options.advertise.bootDelay || ADVERTISE_BOOT_DELAY)
```
jacobheun, line 40: Advertising should be configurable. I may not want to advertise my hop support.

jacobheun (commented):
I like the separation of the circuit transport file 👍

test/relay/auto-relay.node.js
vasco-santos: > This is dialing itself, it should be failing.

lol yes, it works because of the mock response. I will change the names as you suggest

src/circuit/auto-relay.js
vasco-santos: nope, I will simplify to check if `!provider || !provider.multiaddrs.length`.

src/circuit/auto-relay.js
vasco-santos: Yeah, that would be good! I can create an issue for that. We should get that after the `libp2p.discovery` API, also with the rendezvous

vasco-santos: https://github.com/libp2p/js-libp2p/issues/758

src/circuit/constants.js
vasco-santos: I added comments and also updated the `CONFIGURATION.md` file.

src/circuit/auto-relay.js
jacobheun: Provider should always exist, if it doesn't that's an issue. We really **should** have addresses too, otherwise the remote end is doing bad things, but that check is more reasonable.

src/circuit/index.js
```diff
@@ -1,197 +1,87 @@
...
    */
-  filter (multiaddrs) {
-    multiaddrs = Array.isArray(multiaddrs) ? multiaddrs : [multiaddrs]
+  async _advertiseService () {
+    try {
+      const cid = await namespaceToCid(RELAY_RENDEZVOUS_NS)
+      await this._libp2p.contentRouting.provide(cid)
+    } catch (err) {
+      if (err.code === 'NO_ROUTERS_AVAILABLE') {
+        log('there are no routers configured to advertise hop relay service')
```
jacobheun, line 70: I think this should be an error since you can disable advertising. If I'm running a "private" relay, I should turn off advertising and shouldn't need to be running a DHT. I would also clarify the error message and log the error as well.
```suggestion
        log.error('a content router, such as a DHT, must be provided in order to advertise the relay service', err)
```

src/circuit/index.js
```diff
@@ -1,197 +1,87 @@
...
+      } else {
+        log.error(err)
+      }
+      // Stop the advertise
+      this.stop()
```
jacobheun, line 75: We shouldn't stop on normal errors. It's possible the provide could fail for some one off reason (internet connection drop, etc) even though everything is configured properly. Right now a single advertise failure will cause this to shut down. If we don't have routers we should stop trying, but otherwise we should let this retry.

src/config.js
```diff
@@ -51,6 +52,11 @@ const DefaultConfig = {
     },
     relay: {
       enabled: true,
+      advertise: {
+        bootDelay: RelayConstants.ADVERTISE_BOOT_DELAY,
+        enabled: true,
```
jacobheun, line 57: This should be false by default since we disable HOP by default
```suggestion
        enabled: false,
```

src/circuit/auto-relay.js
```diff
@@ -121,15 +124,21 @@ class AutoRelay {
...
     } catch (_) {
       log.error(`${id} does not have announced certified multiaddrs`)
-      return
+
+      // Attempt first if existing
+      if (!remoteAddrs || !remoteAddrs.length) {
+        return
+      }
+
+      remoteMultiaddr = remoteAddrs[0].multiaddr
```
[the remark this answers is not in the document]
vasco-santos: I added a sorter. In follow up PRs, I will add the support for custom sort function for dial and custom filter for announce. Then, it will be bubbled up to here. But I wanted to keep things simple in this PR

src/circuit/auto-relay.js
```diff
@@ -229,8 +244,53 @@ class AutoRelay {
...
+        await this._addListenRelay(connection, peerId.toB58String())
+
+        // Check if already listening on enough relays
+        if (this._listenRelays.size >= this.maxListeners) {
+          return
+        }
+      }
+    } catch (err) {
+      if (err.code !== 'NO_ROUTERS_AVAILABLE') {
+        throw err
```
jacobheun, line 268: Nothing is catching this error, just log it. The log below should also be an error and we should still log the actual error.

jacobheun (commented):
Last couple things otherwise this looks good

jacobheun (approved):
LGTM, I'll leave the rebase/squash in your hands in case there are other branches you have based on this one.

PR #755, opened 2020-09-16
chore: complement 0.29 migration for pubsub subscribe

vasco-santos:
Added a point in `pubsubs.subscribe` about subscribing to multiple topics.
While we did not have any documentation for it in the past, people can be using it, as it was happening on https://github.com/ipfs/js-ipfs/issues/3282

### DIFF
--- a/doc/migrations/v0.28-v0.29.md
+++ b/doc/migrations/v0.28-v0.29.md
@@ -90,7 +90,7 @@ libp2p.pubsub.on(topic, handler)
 libp2p.pubsub.subscribe(topic)
 ```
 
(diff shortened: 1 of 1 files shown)


doc/migrations/v0.28-v0.29.md
```diff
@@ -90,6 +90,36 @@ libp2p.pubsub.on(topic, handler)
 libp2p.pubsub.subscribe(topic)
 ```
 
+In the latest releases, despite not being documented in `libp2p` the underlying pubsub routers supported subscribing to multiple topics at the same time. We removed that code complexity, since this is easily achieved in the application layer if needed.
```
jacobheun, line 93:
```suggestion
In the latest release, despite not being documented in `libp2p` the underlying pubsub routers supported subscribing to multiple topics at the same time. We removed that code complexity, since this is easily achieved in the application layer if needed.
```

jacobheun (approved):
LGTM

Issue #744, opened 2020-09-07
Connection Manager Overhaul


vasco-santos:
# Connection Manager Overhaul

This Issue is an EPIC to track the work related to the Connection Manager Overhaul. Each milestone context and initial thoughts are described next.

## Background

As we land new features like the auto-relay and rendezvous as part of improving connectivity and discoverability in libp2p [libp2p/js-libp2p#703](https://github.com/libp2p/js-libp2p/issues/703), the connection manager overhaul becomes an important work stream to guarantee these protocols work as expected. In addition, this work will be important for some already implemented features/protocols like `webrtc-star` and `bootstrap`. Finally, this work is really important to enable the DHT work.

This overhaul should be an initial step towards the future [ConnMgr v2](https://github.com/libp2p/specs/pull/161).

## Milestones Overview

| Milestone | Issue | PR | State |
|----------|:-------------:|------:|------:|
| `0) Documentation - Baseline` |  NA | #757 | WIP |
| `1) Watermarks Observation - Proactive Dial` |  TODO | TODO | TODO |
| `2) Keep Alive` |  TODO | TODO | TODO |
| `3) Protect Connections - Connection Tags` |  TODO | TODO | TODO |
| `4) Protect Connections - Decaying Tags` |  TODO | TODO | TODO |
| `5) Watermarks Observation - Trimming` |  TODO | TODO | TODO |
| `6) Connection Gater` |  TODO | #1142 | Done |
| `7) Dial retry` |  TODO | TODO | TODO |
| `8) Disconnect message` |  TODO | TODO | TODO |

These milestones do not need to be worked on in the displayed sequence. For instance, Connection tags, Connection Gater and Keep Alive can be isolated and implemented.

## Context

The Connection manager is responsible for managing all the connections a peer has over time. It allows users to enforce an upper bound on the total number of open connections. To avoid possible service disruptions, connections can be tagged with metadata and optionally "protected" to guarantee that essential connections are kept alive.

### 0) Documentation - Connection flows

Create a `DISCOVERABILITY_AND_CONNECTIVITY.md` document to be a subsequent to the `GETTING_STARTED` document. After someone getting up to speed with how to configure and start libp2p on the getting started document, they should move into how to setup their peer/network according to their use case/environment, in order to enable peers to be discovered and connections with them to be established.

This will be divided in two categories:
- define a baseline of what is a desirable set of connections for each environment / use case
- improve current documentation to clarify some flows like the webrtc-star server
[...]

PR #757, opened 2020-09-18
docs: discoverability and connectivity

vasco-santos:
This PR addresses the base milestone for the Connection Manager overhaul #744 .

A DISCOVERABILITY_AND_CONNECTIVITY.md document to be a follow up to the GETTING_STARTED document. After someone getting up to speed with how to configure and start libp2p on the getting started document, they should move into how to setup their peer/network according to their use case/environment, in order to enable peers to be discovered and connections with them to be established.

Needs:
- [x] link follow up from getting started
- [x] link from README


achingbrain:
Closing due to staleness

PR #760, opened 2020-09-24
chore: store self protocols in protobook

vasco-santos:
This PR stores the self protocols in the Protobook on `handle`/`unhandle`.
This allows the identify service to be consistent, by tracking self protocols change in the same way as #748 does for multiaddrs.

`Protobook.remove` API method was added for the unhandle.

Needs:
- [x] #748

### DIFF
--- a/test/peer-store/proto-book.spec.js
+++ b/test/peer-store/proto-book.spec.js
@@ -5,7 +5,9 @@ const chai = require('chai')
 chai.use(require('dirty-chai'))
 const { expect } = chai
 
+const sinon = require('sinon')
 const pDefer = require('p-defer')
+const pWaitFor = require('p-wait-for')
 
 const PeerStore = require('../../src/peer-store')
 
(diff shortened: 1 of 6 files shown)


test/peer-store/proto-book.spec.js
```diff
@@ -224,6 +224,113 @@ describe('protoBook', () => {
...
+    beforeEach(() => {
+      peerStore = new PeerStore({ peerId })
+      pb = peerStore.protoBook
+    })
+
+    afterEach(() => {
+      peerStore.removeAllListeners()
+    })
+
+    it('throwns invalid parameters error if invalid PeerId is provided', () => {
```
jacobheun, line 239:
```suggestion
    it('throws invalid parameters error if invalid PeerId is provided', () => {
```

test/peer-store/proto-book.spec.js
```diff
@@ -224,6 +224,113 @@ describe('protoBook', () => {
...
+      expect(() => {
+        pb.remove('invalid peerId')
+      }).to.throw(ERR_INVALID_PARAMETERS)
+    })
+
+    it('throwns invalid parameters error if no protocols provided', () => {
```
jacobheun, line 245:
```suggestion
    it('throws invalid parameters error if no protocols provided', () => {
```

test/peer-store/proto-book.spec.js
```diff
@@ -224,6 +224,113 @@ describe('protoBook', () => {
...
+      const removedProtocols = ['protocol1']
+      const finalProtocols = supportedProtocols.filter(p => !removedProtocols.includes(p))
+
+      let changeTrigger = 2
+      peerStore.on('change:protocols', ({ protocols }) => {
+        changeTrigger--
+        if (changeTrigger === 0 && arraysAreEqual(protocols, finalProtocols)) {
+          defer.resolve()
+        }
+      })
```
jacobheun, line 264: just a nit, you could pass a sinon spy as the event handler, then you could validate the emitted data after each update. Then you dont have to deal with the trigger counter.

test/peer-store/proto-book.spec.js
```diff
@@ -224,6 +224,113 @@ describe('protoBook', () => {
...
+      // set
+      pb.set(peerId, supportedProtocols)
+
+      // remove
+      pb.remove(peerId, removedProtocols)
+
+      // Wait 50ms for incorrect second event
+      setTimeout(() => {
+        defer.resolve()
+      }, 50)
```
jacobheun, line 328: spy as the handler would also avoid you needing to use a timeout here

jacobheun (approved):
overall looks good, just some nits in the tests

PR #761, opened 2020-09-25
fix: event handlers error

### LINKED ISSUE #751: libp2p.on() silently eats errors/exceptions
Check out this [libp2p with preact](https://github.com/mcclure/ts-hello/tree/libp2p-preact) sample code to 0.29.0 demo (libp2p-preact branch). This is basically a rewrite of the libp2p-in-the-browser example. Run it with
```
npm install && npx webpack && (cd site && http-server -c-1)
```
Open this in a browser and you will see it makes lists of connected and discovered nodes. So far so good.

Okay, now edit the file src/index.tsx. Scroll down to line 163 and add a new first line of connectionManager.on(
```
node.connectionManager.on('peer:connect', (connection:any) => {
  let hfge:any         // New line, to make typescript happy
  hfge.waaf.oiew.fef() // New line
```
Obviously this is gibberish and will fail. But now try running the program. You will find
1. The program fails uncleanly; although the problem is Javascript throws an exception when you use the undefined "hfge" object, the visible symptom will be that it appears we only discover and never connect to them.
2. No error of any kind is printed in the Javascript console (tested on Chrome).

**Impact**

I would consider this a medium to high problem with the API. It is a problem because silent failures make it very difficult to debug what is failing or even realize something is failing. Exacerbating the problem is
1. it's really unclear to me what the type is of the connectionManager.on() object, what the type of the object passed to it is, or what the methods on it are— if this is documented **anywhere** I've not found it in the many months, they are just magic functions in the getting started examples
2. The signatures of these objects/functions routinely change from release to release. I got stuck on this because I was upgrading my code from 0.27.x to 0.29.x, the holdover code was bad because it used the old method signatures, and the lack of an error caused me to mis-diagnose the problem (ie: I thought the problem was that I was discovering peers but not connecting, but it was a simpler problem that I was connecting to peers but my on-connect handler was crashing).

This makes it unusually likely that users will make trivial errors in these exact functions that node.on will not catch.

**Template questions**

Version: 0.29.0, as documented in example repo package.lock
uname -a of test system: `Darwin Pearl.local 17.7.0 Darwin Kernel Version 17.7.0: Sun Dec  1 19:19:56 PST 2019; root:xnu-4570.71.63~1/RELEASE_X86_64 x86_64`
Test version of chrome: 85.0.4183.83 (Official Build) (64-bit)

vasco-santos:
For some scenarios, if a user error in a libp2p event handler is thrown, the handler execution fails silently and the user has no way of knowing about the error.

The best solution according to the EventEmitter docs is to throw all the side effects of the `emit` in the global context, as it happens in [node.js](https://github.com/nodejs/node/blob/e36ffb72bebae55091304da51837ca204367dc16/lib/events.js#L177-L179) out of the box. Several browser shims for the EventEmitter I looked into leave this to the user and do not throw in the global context. Examples:

- https://github.com/protobufjs/protobuf.js/blob/master/lib/eventemitter/index.js#L81
- https://github.com/browserify/events/blob/master/events.js#L158
 
As JS has a variety of shims for each building tool, it is not feasible to fix this upstream.  I created [proper-event-emitter](https://github.com/vasco-santos/proper-event-emitter) that extends Event Emitter and wraps the `super.emit` with a try catch and throw it in the global context using `setTimeout`

Closes #751

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "p-fifo": "^1.0.0",
     "p-settle": "^4.0.1",
     "peer-id": "^0.14.2",
+    "proper-event-emitter": "^1.0.0",
     "protons": "^2.0.0",
     "retimer": "^2.0.0",
     "sanitize-filename": "^1.6.3",
--- a/src/circuit/listener.js
+++ b/src/circuit/listener.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const EventEmitter = require('events')
+const EventEmitter = require('proper-event-emitter')
 const multiaddr = require('multiaddr')
 
 const debug = require('debug')
--- a/src/connection-manager/index.js
+++ b/src/connection-manager/index.js
@@ -9,7 +9,7 @@ const mergeOptions = require('merge-options')
 const LatencyMonitor = require('./latency-monitor')
 const retimer = require('retimer')
 
-const { EventEmitter } = require('events')
+const EventEmitter = require('proper-event-emitter')
 
 const PeerId = require('peer-id')
 
@@ -199,6 +199,7 @@ class ConnectionManager extends EventEmitter {
     const storedConn = this.connections.get(peerIdStr)
 
     this.emit('peer:connect', connection)
+
     if (storedConn) {
       storedConn.push(connection)
     } else {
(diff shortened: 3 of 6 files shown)


achingbrain:
Closing due to staleness

PR #770, opened 2020-10-03
Edited transport example

EthanTheMaster:
The transport example said to install `peer-info`but `npm` ended up complaining about the [package being deprecated](https://www.npmjs.com/package/peer-info). It turns out that the example never used this dependency and instead opted in to using the PeerStore. Thus this PR removes the instruction to install `peer-info`. Additionally, some of the code snippets in the document did not match the full solution. In particular, section 2 had code that printed the stream data in a way that causes the output to not match the final output given. For the sake of consistency, this PR updated the snippets to match the code in the corresponding solution files. The desynchronization also led to issues where I would follow the example but not have the code work properly. Notably, I had an issue in section 2 where the stream multiplexer was not available. The document never mentioned that `libp2p-mplex` should be installed, but `2.js` showed that this package was required. Therefore, this PR updates the code blocks, telling the reader to install more dependencies, by adding dependencies not mentioned but used in the solution files and removing those that are never used in the solution files. This PR also added some complementary text in section 2 that explains the code snippet and links to additional documentation.

### DIFF
--- a/examples/transports/README.md
+++ b/examples/transports/README.md
@@ -13,10 +13,10 @@ When using libp2p, you need properly configure it, that is, pick your set of mod
 You will need 4 dependencies total, so go ahead and install all of them with:
 
 ```bash
-> npm install libp2p libp2p-tcp libp2p-noise peer-info
+> npm install libp2p libp2p-tcp libp2p-noise
 ```
 
-Then, on your favorite text editor create a file with the `.js` extension. I've called mine `1.js`.
+Then, in your favorite text editor create a file with the `.js` extension. I've called mine `1.js`.
 
 First thing is to create our own libp2p node! Insert:
 
@@ -30,7 +30,7 @@ const { NOISE } = require('libp2p-noise')
 const createNode = async () => {
   const node = await Libp2p.create({
     addresses: {
-      // To signall the addresses we want to be available, we use
+      // To signal the addresses we want to be available, we use
       // the multiaddr format, a self describable address
       listen: ['/ip4/0.0.0.0/tcp/0']
     },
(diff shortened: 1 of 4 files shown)


vasco-santos (changes requested):
Thanks for improving this @EthanTheMaster 
Can you address my comment below?

EthanTheMaster:
Thanks @vasco-santos for pointing out the ambiguity. This PR was a result of me attempting to walk through the example as a learner and me trying to fill in the gaps of the document based on what I could find in the docs. It was somewhat difficult to reconstruct what was happening and so the explanation I gave was (not surprisingly) inaccurate. This new edit incorporated your comment and used the analogy given in [page on peer ids](https://docs.libp2p.io/concepts/peer-id/) to explain the function of the address book.

examples/transports/README.md
```diff
@@ -132,8 +150,15 @@ node1.peerStore.addressBook.set(node2.peerId, node2.multiaddrs)
...
...ted. Next, we want to add `node2` into `node1`'s "address book" so that `node1` can contact `node2` via `node2`'s peer id (this will be demonstrated later). Think of the address book like a phone book where the associated peer id is like a friend's name and the corresponding set of multiaddresses is like a list of various phone numbers we could use to contact the friend. We now want `node1` to open a connection to `node2`. In libp2p, [this is known as dialing](https://docs.libp2p.io/reference/glossary/#dial). [Dialing requires that we specify both the peer being connected to and the protocol](https://github.com/libp2p/js-libp2p/blob/master/doc/API.md#dialprotocol). In this case, we use `node2`'s `peerId` to specify `node2` as the peer and state that `node1` will be dialing on the `/print` protocol. Note that because we used the `peerId` to identify `node2`, `node2` must be in `node1`'s address book. Omitting the line where we add `node2` into `node1`'s address book will generate an error. Alternatively, you could instead supply one of `node2`'s multiaddresses to dial `node2` similar to how you might call a friend using their phone number as opposed to referring to your contact list
```
vasco-santos, line 153: While this explanation is nice, I think we should not go this way. We are mixing examples with tutorials/docs and adding a lot of content for an example that should be simple. I'd rather see links to [libp2p/js-libp2p/blob/master/src/peer-store/README.md](https://github.com/libp2p/js-libp2p/blob/master/src/peer-store/README.md) and docs.libp2p.io explaining the process, so that we avoid duplicating documentation on how things work. Otherwise, we will be prone to get outdated information when we have it in not common places.

vasco-santos (changes requested):

vasco-santos:
@EthanTheMaster sorry for the delay. I only saw your changes now. Can you please rebase the PR?

examples/transports/README.md
```diff
@@ -147,29 +166,29 @@ Hello p2p world!
...
-    multiaddrs = [multiaddrs]
+const createNode = async (transports, addresses = []) => {
+  if (!Array.isArray(addresses)) {
+    addresses = [addresses]
   }
 
   const node = await Libp2p.create({
     addresses: {
-      listen: multiaddrs.map((a) => multiaddr(a))
+      listen: addresses.map((a) => a)
```
vasco-santos, line 191:
```suggestion
      listen: addresses
```
This can also be changed in the example

vasco-santos (changes requested):
Just had a new look into this! Just found a redundant map that would be nice to fix both in the Readme and example. Can you change it?

vasco-santos (approved):
LGTM

Issue #768, opened 2020-10-02
Discovery API


vasco-santos:
# Discovery API

The discovery API aims to provide libp2p users (and other libp2p subsystems) a way for active peer discovery. Peers providing a service will use this API to advertise their presence in some namespace. On the other side, peers seeking a service use the API to discover peers that have previously advertised as service providers.

This issue aims to decide the following:

- API
- Libp2p Configuration
- Implementations
- Reprovider

## API

Based on [libp2p/go-libp2p-core/discovery/discovery.go](https://github.com/libp2p/go-libp2p-core/blob/master/discovery/discovery.go)

- `libp2p.discovery.advertise(namespace, options)`
- `libp2p.discovery.findPeers(namespace, options)`

A draft of it can be seen in [libp2p/js-libp2p/blob/feat/discovery-api/doc/API.md#discoveryadvertise](https://github.com/libp2p/js-libp2p/blob/feat/discovery-api/doc/API.md#discoveryadvertise).

The discovery implementations (for now `contentRouting` modules and `rendezvous`) have some differences that will need to be considered.

### Advertise

The advertise is trivial to implement, but the way it will be used over time to reprovide content needs to be considered. Rendezvous supports a given ttl for its records, but the DHT and the delegates currently have a fixed ttl.

🏁 For an initial solution, it will be fine to not support a ttl in advertise and default rendezvous to have the same to ease reprovide implementations. However, we should support this in the long run. I foresee application layer use cases interested in just announcing a service for a small interval and have application logic to understand if should advertise again or not. If we go with this solution, an issue to track an enhancement should be created.

Dropping this ttl for now would make the initial API for advertise: `libp2p.discovery.advertise(namespace)`

### FindPeers

The findPeers API might need additional options from the Rendezvous and content routing current implementations. It should support an option for specifying the max number of distinct peers needed.

If we only have setup one Discovery implementation, it will be trivial to implement. However, if we setup multiple implementations we need to consider how they will be leverage, namely in Series or Parallel.

In series (perhaps following an order configured) will be easy to implement, and once we get the maxNumber we can just return. However, having a parallel approach would result in faster results. In a parallel approach, the implementations will need to be synchronised regarding when there are enough distinct peers. One possible solution is to add Abort Controllers support for rendezvous and content routing modules discovery. Once libp2p found enough peers, both would be aborted.

[...]

PR #771, opened 2020-10-05
feat: rendezvous integration

vasco-santos:
In the context of Improving libp2p discoverability and connectivity (#703), this PR integrates the new rendezvous implementations. This is the initial step for the discovery API, which is described in #768 and will have follow up PRs.

This PR integrates rendezvous as described in https://github.com/libp2p/js-libp2p-rendezvous/blob/feat/rendezvous-protocol-full-implementation/LIBP2P.md. The rendezvous module is provided in the libp2p config, and properly started and stopped with the node. Moreover, docs on how to configure a libp2p node acting as a rendezvous server, as well as a rendezvous client were added.

Needs:

- [ ] https://github.com/libp2p/js-libp2p-rendezvous/pull/6
- [ ] Example

Follow up work:

In the context of #768 , libp2p will provide a service discovery API, which will leverage the rendezvous if provided, content routing, and possibly other modules in the future. Once we have the discovery work done, it will be important to create a new example with the discovery API and rendezvous. Moreover, production guides for the server should be created


src/index.js
```diff
@@ -192,6 +192,15 @@ class Libp2p extends EventEmitter {
       this.pubsub = PubsubAdapter(Pubsub, this, this._config.pubsub)
     }
 
+    // Create rendezvous if provided
+    if (this._modules.rendezvous) {
+      const Rendezvous = this._modules.rendezvous
+      this.rendezvous = new Rendezvous({
```
vasco-santos, line 198: Considering that we will provide the `libp2p.discovery` API, I am thinking wether we should provide `libp2p.rendezvous`, or just leave it as `libp2p._rendezvous`. Thoughts?
If we keep this as it is, this PR should include API docs for rendezvous

vasco-santos:
Replaced by #856

PR #773, opened 2020-10-06
chore: update aegir

vasco-santos:
This PR updates `aegir` to its latest versions. One of the highlights is the `eslint` related changes to support typescript definition generation from jsdocs.

Most of the fixes were made using the automatic lint fix feature.


jacobheun (approved):
lgtm

PR #783, opened 2020-10-15
feat: custom announce filter

### LINKED ISSUE #769: Improve announced addresses
This issue aims to find the best solution for announced addresses.

Libp2p is not controlling in any way the announced addresses. This means that users will need to properly configure them, in order to not announce not diable addresses. Moreover, we do not have any multiaddr confidence yet.

In this moment, libp2p announces its addresses as: Union between all listen addresses and provided announce addresses, filtering out provided noAnnounce addresses.

## Listen addresses filtering 

One of the common issues it to announce local network addresses. While uses cases that leverage these addresses exist, they are not the most common and they are easy to predict (MDNS discovery, or manual dials / bootstrap).

🏁 The best approach is defaulting to not announcing local network addresses unless explicitly added as announce addresses. This would result in: Union between all **non-local** listen addresses and provided announce addresses, filtering out provided noAnnounce addresses. This will result in identify not announcing local addresses that might get propagated from other peers.

Manual dials and bootstrap nodes rely on previously provided addresses. If a node needs to connect to other nodes in the local network, these will be added to the AddressBook anyway and there will be no issue. For MDNS, we can also add the local addresses in the listen addresses for the query responses.

## Dial improvements

As there will be older nodes on the network, we should also try to improve the dialer to avoid using undiable addresses provided from older nodes.

While we do not have multiaddr confidence, we can also slightly improve the dialer. There are two angles we can look. The dial request can sort the known multiaddrs and try the locals in the end. But, if enough tokens it might try locals in a first batch. We can probably do the dial in 2 steps, first try out all non local addresses and then the local ones.

🏁If a combination of non local and local addresses exist, the probability that the local addr work is not high. This way, I think we can probably take the 2 steps dial.

## Migration

This change should not have big consequences in working systems as they need to be relying on non local addresses. Anyway, people experimenting with local projects or local setups will probably break as a result of this change.

This should be well documented in the next migration guide.

vasco-santos:
This PR adds a custom announce multiaddrs filter. The default is kept, but users can easily change it to not announce private addresses via config.

This PR also changes the `chai` requires on all the codebase to leverage `aegir`

Closes #769


test/keychain/keychain.spec.js
```diff
@@ -73,9 +73,9 @@ describe('keychain', () => {
   it('can find a key without a password', async () => {
     const keychain = new Keychain(datastore2)
     const keychainWithPassword = new Keychain(datastore2, { passPhrase: `hello-${Date.now()}-${Date.now()}` })
-    const id = `key-${Math.random()}`
+    const name = `key-${Math.random()}`
 
-    await keychainWithPassword.createKey(id, 'ed25519')
+    const { id } = await keychainWithPassword.createKey(name, 'ed25519')
```
vasco-santos, line 78: This was not failing in the past, but the promise below would return undefined. I think that the reason behind it was not using `chai-as-promised`. Leveraging the aegir chai that uses if every time was causing this to fail.

src/index.js
```diff
@@ -364,11 +364,13 @@ class Libp2p extends EventEmitter {
...
    */
   get multiaddrs () {
+    const announceFilter = this._options.addresses.announceFilter || ((multiaddrs) => multiaddrs)
+
     // Filter noAnnounce multiaddrs
     const filterMa = this.addressManager.getNoAnnounceAddrs()
 
     // Create advertising list
-    return this.transportManager.getAddrs()
+    return announceFilter(this.transportManager.getAddrs()
```
jacobheun, line 373: Looking at this code I realized the announce behavior is not the same as go. Go doesn't concat the listen and announce addresses. If you provide announce addresses, they are the only addresses that get advertised. 

So the behavior should likely be:

- If `announce` addresses are provided, only they will be announced (no filters are applied)
- If `announce` is not provided, the transport addresses will be filtered
  - If `noAnnounce` is provided it will be applied as a filter
  - if the `announceFilter` is provide it will be run

I feel like with `announceFilter` we should just drop support for `noAnnounce`. Downstream projects like js-ipfs can use the filter to do achieve the same end, but with more control.

Additionally, something to think about for js-ipfs that we can use the announceFilter for, is that go-ipfs takes `noAnnounce` from the config and applies those as net masks. It leverages https://github.com/whyrusleeping/multiaddr-filter, so you can more easily filter out ip ranges.

vasco-santos: yap, I will change this accordingly. I agree on deprecating `noAnnounce`

vasco-santos: The multiaddr-filter with the net masks is Interesting. I will create an issue in libp2p-utils for us to discuss it better later on.

jacobheun (approved):
LGTM, looks like there is an empty file added though `test/dialing/utils.js`

PR #787, opened 2020-10-20
fix: ensure streams are closed on connection close

jacobheun:
This patch ensures that whenever a connection is closed, any remaining streams on the connection are aborted.

- [x] requires https://github.com/libp2p/js-libp2p-mplex/pull/116

### DIFF
--- a/src/upgrader.js
+++ b/src/upgrader.js
@@ -5,6 +5,7 @@ const log = debug('libp2p:upgrader')
 log.error = debug('libp2p:upgrader:error')
 const Multistream = require('multistream-select')
 const { Connection } = require('libp2p-interfaces/src/connection')
+const ConnectionStatus = require('libp2p-interfaces/src/connection/status')
 const PeerId = require('peer-id')
 const pipe = require('it-pipe')
 const errCode = require('err-code')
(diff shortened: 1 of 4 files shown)


jacobheun:
CI failures are currently due to the latest release of node 15 on current. We'll resolve those separately.

src/upgrader.js
```diff
@@ -268,8 +269,13 @@ class Upgrader {
     maConn.timeline = new Proxy(_timeline, {
       set: (...args) => {
         if (connection && args[1] === 'close' && args[2] && !_timeline.close) {
-          connection.stat.status = 'closed'
-          this.onConnectionEnd(connection)
+          // Wait for close to finish before notifying of the closure
+          (async () => {
```
achingbrain, line 273: Can this anonymous function throw?  The error needs to be handled if so.

achingbrain (changes requested):

src/upgrader.js
jacobheun: Yes, in theory the connection.close could throw although that should be rare. Good catch, a handler for this to log the error

jacobheun:
npm 7 broke how we're doing some dev testing, https://github.com/libp2p/js-libp2p/issues/790, so I've locked ci on Node 14 for now to avoid the CI failures.

achingbrain (approved):

PR #792, opened 2020-10-28
feat: custom dialer addr sorter

vasco-santos:
This PR adds a custom dialer addr sorter via config. It is used both by the dialer and the autoRelay. 

The default sorter function puts public addresses first as a first criterium and certified addresses as a second criterium. This comparator function should probably be moved into `libp2p-utils`.

Needs:

- [x] #783 
- [x] https://github.com/libp2p/js-libp2p-utils/pull/13


src/dialer/index.js
```diff
@@ -120,7 +124,16 @@ class Dialer {
       this.peerStore.addressBook.add(id, multiaddrs)
     }
 
-    let knownAddrs = this.peerStore.addressBook.getMultiaddrsForPeer(id) || []
+    let knownAddrs = this.addressSorter(
+      this.peerStore.addressBook.get(id) || []
```
jacobheun, line 128: Any reason you switched away from `getMultiaddrsForPeer`? The address sorter should be able to handle `/p2p` on a multiaddr and avoids the need for the map logic here.

src/circuit/auto-relay.js
```diff
@@ -129,28 +128,19 @@ class AutoRelay {
...
-      log.error(`${id} does not have announced certified multiaddrs`)
-
-      // Attempt first if existing
-      if (!remoteAddrs || !remoteAddrs.length) {
-        return
-      }
+    // Get peer known addresses and sort them per public addresses first
+    // Sorting addresses to public first
+    const remoteAddrs = this._addressSorter(
+      this._peerStore.addressBook.get(connection.remotePeer) || []
```
jacobheun, line 134: Use `getMultiaddrsForPeer`? This should also avoid the `if` check below for the `p2p` protoName.

src/dialer/utils.js
```diff
@@ -0,0 +1,45 @@
+'use strict'
+
+const isPrivate = require('libp2p-utils/src/multiaddr/is-private')
+
+// TODO: Move to libp2p-utils?
```
jacobheun, line 5: Yeah, I think this would be useful for other modules that would avoid needing to use libp2p as a dependency.

test/dialing/direct.spec.js
```diff
@@ -176,14 +177,36 @@ describe('Dialing (direct, WebSockets)', () => {
       .and.to.have.property('code', ErrorCodes.ERR_TIMEOUT)
   })
 
+  it('should sort addresses on dial', async () => {
+    sinon.spy(dialerUtils, 'sortPublicAddressesFirst')
```
jacobheun, line 181: It would be good to actually verify the dial order here, we dont check anything after `connectToPeer`. Just pass a stub to `addressSorter` and return a different order than addressBook returns. We should be able to spy on the transportManager and ensure the addresses are executed in the same order `addressSorter` returns.

src/dialer/index.js
vasco-santos: Yes! I want the sorter to get an `Address` and not a `multiaddr`. With this, we have the ability to sort with other data, like the `.isCertified` boolean.

vasco-santos: The downside is having to do the `/p2p` logic which is embedded in the `getMultiaddrsForPeer`. I suggest I move that logic to `libp2p-utils` and reuse it here. What do you think? Any naming suggestions?

jacobheun: Ah, right. What if we just extend `getMultiaddrsForPeer` to take an optional sorter? Then we can keep that logic isolated and allow different systems to use different sorters as needed.

vasco-santos: that seems good! I will change that as we will only use this with the `getMultiaddrsForPeer`

doc/CONFIGURATION.md
```diff
@@ -509,6 +510,7 @@ const MPLEX = require('libp2p-mplex')
 const { NOISE } = require('libp2p-noise')
 
 const { dnsaddrResolver } = require('multiaddr/src/resolvers')
+const { sortPublicAddressesFirst } = require('libp2p/src/dialer/utils')
```
jacobheun, line 513:
```suggestion
const { publicAddressesFirst } = require('libp2p-utils/src/address-sort')
```

doc/CONFIGURATION.md
```diff
@@ -522,7 +524,8 @@ const node = await Libp2p.create({
     dialTimeout: 30e3,
     resolvers: {
       dnsaddr: dnsaddrResolver
-    }
+    },
+    addressSorter: sortPublicAddressesFirst
```
jacobheun, line 528:
```suggestion
    addressSorter: publicAddressesFirst
```

jacobheun (commented):
Pending CI fixes this lgtm, just a couple fixes to the docs

vasco-santos:
CI is green and suggestions added! 🎉

jacobheun (approved):
LGTM

PR #795, opened 2020-11-02
chore: auto relay example

vasco-santos:
This PR adds an example to use libp2p's auto relay.

With it, I added a setup to run the examples in CI, inspired by `js-ipfs` older test setup pre lerna migration https://github.com/ipfs/js-ipfs/pull/2528

Needs:

- [x] guidelines for what's next?
- [x] tests


examples/auto-relay/README.md
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }
+})
+
+await node.start()
+console.log(`Node started: ${node.peerId.toB58String()}`)
+
+await node.dial(relayAddr)
+
+// Wait for connection and relay to be bind for the example purpose
```
jacobheun, line 114:
```suggestion
// Wait for connection and relay to bind for the purpose of this example
```

examples/auto-relay/README.md
```diff
@@ -0,0 +1,192 @@
...
+await pWaitFor(() => node.multiaddrs.length > 0)
+
+console.log('connected to the HOP relay')
+console.log('Listening on:')
+node.multiaddrs.forEach((ma) => console.log(`${ma.toString()}/p2p/${node.peerId.toB58String()}`))
+```
+
+As you can see in the code, we need to provide the `relayAddr` as a process argument. This node will dial the relay and automatically bind to the relay.
```
jacobheun, line 122:
```suggestion
As you can see in the code, we need to provide the relay address, `relayAddr`, as a process argument. This node will dial the provided relay address and automatically bind to it.
```

examples/auto-relay/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+You should now run the following to start the relay node:
```
jacobheun, line 124:
```suggestion
You should now run the following to start the node running Auto Relay:
```

examples/auto-relay/README.md
```diff
@@ -0,0 +1,192 @@
...
+node auto-relay.js /ip4/192.168.1.120/tcp/58941/ws/p2p/QmQKCBm87HQMbFqy14oqC85pMmnRrj6iD46ggM6reqNpsd
+```
+
+This should print out something similar to the following:
+
+```sh
+Node started: QmerrWofKF358JE6gv3z74cEAyL7z1KqhuUoVfGEynqjRm
+connected to the HOP relay
+Listening on:
+/ip4/192.168.1.120/tcp/61592/ws/p2p/QmWDn2LY8nannvSWJzruUYoLZ4vV83vfCBwd8DipvdgQc3/p2p-circuit/p2p/QmerrWofKF358JE6gv3z74cEAyL7z1KqhuUoVfGEynqjRm
```
jacobheun, line 136: Consider cleaning the logging up a bit to be a bit clearer, something like:
```suggestion
Node started with id QmerrWofKF358JE6gv3z74cEAyL7z1KqhuUoVfGEynqjRm
Connected to the HOP relay QmWDn2LY8nannvSWJzruUYoLZ4vV83vfCBwd8DipvdgQc3
Advertising with a relay address of /ip4/192.168.1.120/tcp/61592/ws/p2p/QmWDn2LY8nannvSWJzruUYoLZ4vV83vfCBwd8DipvdgQc3/p2p-circuit/p2p/QmerrWofKF358JE6gv3z74cEAyL7z1KqhuUoVfGEynqjRm
```

examples/auto-relay/auto-relay.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+  })
+
+  await node.start()
+  console.log(`Node started: ${node.peerId.toB58String()}`)
+
+  await node.dial(relayAddr)
+
+  // Wait for connection and relay to be bind for the example purpose
+  await pWaitFor(() => node.multiaddrs.length > 0)
```
jacobheun, line 39: Instead of `pWaitFor` perhaps it would be best to listen on the Peer Store for our own address change. This also ensures we're only using libp2p deps in this example.

examples/auto-relay/auto-relay.js
```diff
@@ -0,0 +1,44 @@
...
+
+  console.log('connected to the HOP relay')
+  console.log('Listening on:')
+  node.multiaddrs.forEach((ma) => console.log(`${ma.toString()}/p2p/${node.peerId.toB58String()}`))
+})()
```
jacobheun, line 44: nit: instead of a self executing anonymous function it might be helpful to put these in a function like `main` and then execute that. Same for the other files.

examples/auto-relay/other-node.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
```
jacobheun, line 1: Instead of `other-node` perhaps we should use the same format as other examples: `dialer`, `listener`?

examples/auto-relay/relay.js
```diff
@@ -0,0 +1,37 @@
...
+;(async () => {
+  const node = await Libp2p.create({
+    modules: {
+      transport: [Websockets],
+      connEncryption: [NOISE],
+      streamMuxer: [MPLEX]
+    },
+    addresses: {
+      listen: ['/ip4/0.0.0.0/tcp/0/ws']
+      // announceFilter: TODO check production section
```
jacobheun, line 17: What production section?

jacobheun (commented):
Once all the code is updated we should make sure to update the code shown in the readme so it matches properly.

examples/auto-relay/README.md
```diff
@@ -0,0 +1,192 @@
...
+This should print out something similar to the following:
+
+```sh
+Node started: QmWDn2LY8nannvSWJzruUYoLZ4vV83vfCBwd8DipvdgQc3
+Listening on:
+/ip4/127.0.0.1/tcp/61592/ws/p2p/QmWDn2LY8nannvSWJzruUYoLZ4vV83vfCBwd8DipvdgQc3
+/ip4/192.168.1.120/tcp/61592/ws/p2p/QmWDn2LY8nannvSWJzruUYoLZ4vV83vfCBwd8DipvdgQc3
+```
+
+TODO: Docker Image with a repo
```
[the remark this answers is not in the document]
vasco-santos: I think we should create a simple Docker Image with relay and HOP enabled to enable people to easily deploy Relays. I will remove this from now, while I think in the best solution to also include the announce addresses logic

examples/auto-relay/relay.js
vasco-santos: Changed, it was the `What's next section` where I mention that the announceAddrs should be changed for production

examples/auto-relay/listener.js
```diff
@@ -0,0 +1,50 @@
...
+  console.log(`Node started with id ${node.peerId.toB58String()}`)
+
+  const conn = await node.dial(relayAddr)
+
+  // Wait for connection and relay to be bind for the example purpose
+  await new Promise((resolve) => {
+    node.peerStore.on('change:multiaddrs', ({ peerId }) => {
+      // Updated self multiaddrs?
+      if (peerId.equals(node.peerId)) {
+        resolve()
```
jacobheun, line 41: Maybe get rid of the promise logic and just put the console logs here? This event listener never gets cleaned up and can throw an error if our addresses change again and call `resolve` on the already resolved promise.

vasco-santos: yes, that makes a lot of sense! thanks

jacobheun (approved):
LGTM

Issue #704, opened 2020-07-13
[discovery] Connect to closest peers in the network to share peer information


jacobheun:
In order for nodes on the network to be able to easily find a libp2p node, it needs to find its closest peers on the network and connect to them in order to exchange address records. This will enable other peers who don't yet know the address to search for the PeerId on the network (via the DHT, delegate routers, or other future modules). This query will result in discovering peers who know about our addresses, because we have already exchanged that information with them.

js-libp2p currently does not do this out of the box. As long as we have PeerRouting modules enabled, we should attempt to do this.

1. When libp2p boots it should search for its closest peers and attempt to connect to them (this may require relays with `Active HOP` support).
1. When libp2p changes announce addresses, it should search and connect again. This will help ensure the closest nodes on the network have the latest addresses.
  1. We should avoid querying too frequently to update. On boot we are likely to be more actively finding addresses dynamically, especially if AutoRelay/AutoNAT is enabled.
  1. We should attempt to maintain connections to our closest peers, as they will get push notifications whenever address changes occur.

Ref: https://github.com/ipfs/go-ipfs/issues/7433#issuecomment-656410330

vasco-santos:
This is partially implemented in `libp2p@0.30`.

maschad:
Closing as completed

PR #798, opened 2020-11-12
feat: discover and connect to closest peers

vasco-santos:
This PR adds a peer routing service to discover closest peers recurrently over time. The discovered addresses will be added to the AddressBook and the peers will be dialled if the autoDial is enabled and the connectionManager threshold is not reached. This will enable peers to better advertise themselves on the network as their closest peers will have their address records. 

Following up this PR, we should work on maintaining connections to our closest peers, as they will get push notifications whenever address changes occur. Further optimisations might be done, like trigger a find task when the announce addresses change. Keeping connections and more proactive connections should probably come together with connection manager overhaul as we need to implement dial strategies that can be used for different other subsystems. Once this PR is merged I will report the current state in #704 

Related to #704

Needs:

- [x] [vasco-santos/set-delayed-interval#1](https://github.com/vasco-santos/set-delayed-interval/pull/1)

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -35,6 +35,11 @@ const DefaultConfig = {
     persistence: false,
     threshold: 5
   },
+  peerRouting: {
+    enabled: true,
+    interval: 300e3,
+    bootDelay: 10e3
+  },
   config: {
     dht: {
       enabled: false,
--- a/src/config.js
+++ b/src/config.js
@@ -36,9 +36,11 @@ const DefaultConfig = {
     threshold: 5
   },
   peerRouting: {
-    enabled: true,
-    interval: 300e3,
-    bootDelay: 10e3
+    refreshManager: {
+      enabled: true,
+      interval: 6e5,
+      bootDelay: 10e3
+    }
   },
   config: {
     dht: {
--- a/src/peer-routing.js
+++ b/src/peer-routing.js
@@ -1,40 +1,131 @@
 'use strict'
 
 const errCode = require('err-code')
+const debug = require('debug')
+const log = debug('libp2p:peer-routing')
+log.error = debug('libp2p:peer-routing:error')
+
+const all = require('it-all')
 const pAny = require('p-any')
 
-module.exports = (node) => {
-  const routers = node._modules.peerRouting || []
+/**
+ * Responsible for managing the usage of the available Peer Routing modules.
+ */
+class PeerRouting {
+  /**
+   * @class
+   * @param {Libp2p} libp2p
+   */
+  constructor (libp2p) {
+    this._peerId = libp2p.peerId
+    this._peerStore = libp2p.peerStore
+    this._routers = libp2p._modules.peerRouting || []
 
-  // If we have the dht, make it first
-  if (node._dht) {
-    routers.unshift(node._dht)
+    // If we have the dht, make it first
+    if (libp2p._dht) {
+      this._routers.unshift(libp2p._dht)
+    }
+
+    this._options = libp2p._options.peerRouting
   }
 
-  return {
-    /**
-     * Iterates over all peer routers in series to find the given peer.
-     *
-     * @param {string} id - The id of the peer to find
-     * @param {object} [options]
-     * @param {number} [options.timeout] - How long the query should run
-     * @returns {Promise<{ id: PeerId, multiaddrs: Multiaddr[] }>}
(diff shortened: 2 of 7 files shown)


src/peer-routing.js
```diff
@@ -1,40 +1,126 @@
...
+const debug = require('debug')
+const log = debug('libp2p:peer-routing')
+log.error = debug('libp2p:peer-routing:error')
+
+const all = require('it-all')
 const pAny = require('p-any')
+const {
+  setDelayedInterval,
+  clearDelayedInterval
+} = require('set-delayed-interval')
```
vasco-santos, line 13: I created this module for us to use here, but also in:

- https://github.com/libp2p/js-libp2p/blob/0.30.x/src/circuit/index.js#L43-L88
- https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/random-walk.js#L46-L89

It is always tricky to implement a "setInterval" when the recurrent task is a promise. Other than that, if we need a delay beforehand, things get even more complex as we need to guarantee that the timer can be stopped for any of the pauses.

When looking at the `relay` logic I found a bug. If we clear the timeout when ` await this._libp2p.contentRouting.provide(cid)` is running, then the timeout will be assigned again afterward. We should be checking if we had timeout before creating the new one. This module aims to avoid introducing these issues when dealing with this use cases.

It would be great to have your 👀  on the module: https://github.com/vasco-santos/set-delayed-interval/pull/1

src/config.js
```diff
@@ -35,6 +35,11 @@ const DefaultConfig = {
     persistence: false,
     threshold: 5
   },
+  peerRouting: {
+    enabled: true,
+    interval: 300e3,
+    bootDelay: 10e3
+  },
```
jacobheun, line 42: This is a bit confusing as it's not quite peer routing. This is really the refresh manager. In go-libp2p since they just have the DHT right now they run the table refreshes when they search for their closest peers. This happens when DHT bootstrap is run at startup and then on an interval. Ideally we would not use a boot delay timer as this is finicky, and instead run once we've finished a bootstrap phase. Since we don't really have this concept atm, I think a delay is fine for now.

Go runs this every 10 minutes by default right now, we should just match that for consistency.

src/peer-routing.js
jacobheun: Yeah, I'd really like to get rid of this whole boot delay logic. In reality we should probably have some async boot function and when that finishes these other systems initiate, but I think we can revisit this when we update the config logic. I'll review the module

vasco-santos: It would be great to orchestrate subsystems like that. I will add this comment to the configuration issue for us to remember

src/config.js
vasco-santos: I am going to change this into 10 minutes yes. Regarding the namings, thinking on changing this to:
```
peerRouting: {
  refreshManager: { ... }
}
```

The peerRouting is the scope where we make these queries, so it seems that this configs should live bound to it. What do you think?

vasco-santos: Added as suggested for now

src/peer-routing.js
```diff
@@ -1,40 +1,126 @@
...
+   * @returns {AsyncIterable<{ id: PeerId, multiaddrs: Multiaddr[] }>}
+   */
+  async * getClosestPeers (key, options = { timeout: 30e3 }) {
+    if (!this._routers.length) {
+      throw errCode(new Error('No peer routers available'), 'NO_ROUTERS_AVAILABLE')
+    }
 
-        // If we don't have a result, we need to provide an error to keep trying
-        if (!result || Object.keys(result).length === 0) {
+    const result = await pAny(
```
jacobheun, line 109: Not going to block this PR on this (as we're doing it in other content/peer routing calls), but we should think about the parallel actions this is doing. Most people shouldn't be using two, but in the change they are, the parallel calls could cause some odd behavior. This is probably worth discussing in more depth when we look at the discovery api / configuration updates.

jacobheun (approved):
Looks good, just 1 comment to discuss later when we look at the disco api/config updates.

I'll let you merge this and will review the other delayed interval PR so you can combine them.

src/peer-routing.js
vasco-santos: Makes sense, I will cross post it here for us to discuss

PR #800, opened 2020-11-13
feat: custom and store self agent version + store self protocol version

### LINKED ISSUE #799: Store protocolVersion and agentVersion in the MetadataBook
We currently have a hardcoded value for `agentVersion` and `protocolVersion`. While these values should be used as defaults, libp2p needs to support receiving the `protocolVersion` value.

Both these values, should be stored in the MetadataBook, in the same way as when we receive this information from other peers in the identify protocol.

vasco-santos:
This PR adds the ability to custom the agentVersion of libp2p, as well as to store it and the protocol version self values in the metadataBook.

I added a host property to track this custom options. `go-libp2p` also has the concept of host and it seems a good naming. When we work on the configuration improvements, we should probably think about moving other options inside the host, such as the AddressBook options.

I followed the same logic of signedPeerRecords being created in the IdentifyService and added this into the Identify subsystem.

Closes #799

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -92,7 +92,7 @@ Creates an instance of Libp2p.
 | options.modules | [`Array<object>`](./CONFIGURATION.md#modules) | libp2p [modules](./CONFIGURATION.md#modules) to use |
 | [options.addresses] | `{ listen: Array<string>, announce: Array<string>, noAnnounce: Array<string> }` | Addresses for transport listening and to advertise to the network |
 | [options.config] | `object` | libp2p modules configuration and core configuration |
-| [options.host] | `{agentVersion: string, protocolVersion: string}` | libp2p host options |
+| [options.host] | `{ agentVersion: string }` | libp2p host options |
 | [options.connectionManager] | [`object`](./CONFIGURATION.md#configuring-connection-manager) | libp2p Connection Manager [configuration](./CONFIGURATION.md#configuring-connection-manager) |
 | [options.transportManager] | [`object`](./CONFIGURATION.md#configuring-transport-manager) | libp2p transport manager [configuration](./CONFIGURATION.md#configuring-transport-manager) |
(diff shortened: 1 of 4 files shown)


doc/API.md
```diff
@@ -92,6 +92,7 @@ Creates an instance of Libp2p.
 | options.modules | [`Array<object>`](./CONFIGURATION.md#modules) | libp2p [modules](./CONFIGURATION.md#modules) to use |
 | [options.addresses] | `{ listen: Array<string>, announce: Array<string>, noAnnounce: Array<string> }` | Addresses for transport listening and to advertise to the network |
 | [options.config] | `object` | libp2p modules configuration and core configuration |
+| [options.host] | `{agentVersion: string, protocolVersion: string}` | libp2p host options |
```
jacobheun, line 95: I'm not inclined to allow setting `protocolVersion`. `agentVersion` is fine it has no impact, but any change to  `protocolVersion` warrants a connection closure in the spec (although we dont enforce that here). We _could_ allow it, but we'd need to be very clear about not changing it unless you're intentionally creating an isolated network. ie: "DO NOT change unless you know what you're doing"

vasco-santos: Yes, that makes sense.

So, `js-ipfs` has supposedly `protocolVersion` assigned to `9000`. But in theory, this is not used anywhere I think.  I am not sure about the reasoning behind that number, but it would need to be changed to libp2p's default value. If so, we can drop the setting of `protocolVersion` unless a new use case appears

achingbrain: go-IPFS reports `protocolVersion` as `'ipfs/0.1.0'` so it should probably change to that?

jacobheun: Yeah I'm not sure why js-ipfs is trying to use that, but you are correct, it won't change anything we're actually sending over the wire. That should be corrected in js-ipfs. 

Let's leave out being able to configure `protocolVersion` for now. We may add it in the future, but that's a can of worms I'd rather not open right now.

jacobheun (approved):
lgtm

PR #801, opened 2020-11-13
chore: update pubsub

vasco-santos:
This PR updates the pubsub routers. The pubsub routers were updated with the newest `pubsub` interface with a new `signingPolicy` option to match the spec.

This PR also changes the default value of `emitSelf` to false. This was already the default value on the router level but was being overwritten here to not cause a breaking change upstream, but this is the time to do it. Related to: https://github.com/ChainSafe/js-libp2p-gossipsub/pull/144

BREAKING CHANGE: pubsub signing policy properties were changed according to libp2p-interfaces changes to a single property. The emitSelf option default value was also modified to match the routers value

Needs:
- [x] https://github.com/libp2p/js-libp2p-floodsub/pull/111


jacobheun (approved):
lgtm

PR #802, opened 2020-11-16
chore: add typedefs

vasco-santos:
This PR adds typedefs in most of the core codebase. Libp2p interfaces was updated to leverage its new types.

Github actions replaced travis, which also enables us to just use it for windows 🎉 

This brings a set of advantages:
- enable us to generate typescript declaration files automatically on release with aegir.
- Intellisense for libp2p users
  - Users will be able to see on the fly the libp2p API, expectations for each functions and associated docs/examples
  - [Example](https://code.visualstudio.com/docs/editor/intellisense#_types-of-completions)
- Intellisense for libp2p core team
  - Core team will be able to see internal types and expectations of each functions without the need to fly around the code. It also enables other editors features like jump to definition, or pop ups with implementation of a highlighted function
- Automated docs
  - As a follow up, we should be able to automatically generate the API docs

What are we missing / What is next?

- Event Emitters
  - @Gozala created https://www.npmjs.com/package/typed-events.ts for the EventEmitter typedefs ❤️ .
After syncing with him, we might want to get the declaration file from there in and use https://www.npmjs.com/package/proper-event-emitter per https://github.com/libp2p/js-libp2p/pull/761 (we also have emittery and some modules and we need to double check ig we should just use it). Accordingly, I think we can land the typedefs for EventEmitter in a new PR to not block this.
- DHT should have typedefs for its options, so that we can integrate them here
- Metrics and Latency monitor needs to be finished - currently a mess and not user level API
- Some Keychain options types (this needs the `interface-datastore` update and adding types to the keychain. I decided to postpone this).
- Replace API.md for docs with automated aegir
  - We need to move the examples in the JSDocs first
(once this is merged, I will create an issue to track the missing bits)

Needs:

- [ ] https://github.com/libp2p/js-libp2p-interfaces/pull/74
At the moment, it is using a branch from the PR below that has a prepare script to generate the build while we do not release it.


src/identify/index.js
```diff
@@ -290,7 +264,7 @@ class IdentifyService {
    *
    * @private
    * @param {object} options
-   * @param {*} options.stream
+   * @param {DuplexIterable} options.stream
```
jacobheun, line 267: We should probably track adding a type for `Stream` in the interfaces module, that should probably live with Connection. While any `DuplexIterable` should be fine, we really want these to be Stream's

jacobheun (commented):
> enable us to generate typescript declaration files automatically on release with aegir.

Will we get these automatically now, do we need to update anything for the release? 

> What are we missing / What is next?

For what's missing, has this been checked against any downstream TS projects? The intellisense is nice, but if we're going to be generating the types on release we should make sure to test this on a larger existing project. For example, can we swap out the types here in Lodestar, https://github.com/ChainSafe/lodestar/tree/master/packages/lodestar, for the types at https://github.com/ChainSafe/libp2p-ts.git that it's using in dev?

vasco-santos:
> Will we get these automatically now, do we need to update anything for the release?

Pending https://github.com/ipfs/aegir/pull/671 for support on release. But yes, I can do that test by that time

src/address-manager/index.js
```diff
@@ -27,7 +29,7 @@ class AddressManager {
   /**
    * Get peer listen multiaddrs.
    *
-   * @returns {Array<Multiaddr>}
+   * @returns {Array<multiaddr>}
```
Gozala, line 32: Nit: Using upper case first char in type names is an idiomatic pattern in TS & think it would be best to follow it. Assuming we do not want to rename everywhere you could define a type at the end of the file e.g.:

```js
/**
 * @typedef {import('multiaddr')} Multiaddr
 */

Which is also how we do it in js-ipfs. I think it would be also less confusing as in js-ipfs we use `Multiaddr`.

File: src/circuit/auto-relay.js [hunk 3]
```diff
@@ -21,14 +21,25 @@ const {
   RELAY_RENDEZVOUS_NS
 } = require('./constants')
 
+/**
+ * @typedef {import('libp2p-interfaces/src/connection').Connection} Connection
+ * @typedef {import('../peer-store/address-book').Address} Address
+ */
+
+/**
+ * @typedef {Object} AutoRelayProperties
```
Gozala, line 30: Nit: In js-ipfs we settled on calling injected deps with `Config` suffix e.g. AutoRelayConfig`. Might be worth following same convention across both code bases.

src/content-routing.js
```diff
@@ -24,7 +29,7 @@ module.exports = (node) => {
      * @param {object} [options]
      * @param {number} [options.timeout] - How long the query should run
      * @param {number} [options.maxNumProviders] - maximum number of providers to find
-     * @returns {AsyncIterable<{ id: PeerId, multiaddrs: Multiaddr[] }>}
+     * @returns {AsyncIterable<{ id: PeerId, multiaddrs: Array<multiaddr> }>}
```
Gozala, line 32: I believe `T[]` is more idiomatic style that `Array<T>`, only cases where later is recommended over the former are when otherwise it's ambiguous e.g: 

```
[T, U][] -> Array<[T, U]>
T[][] -> Array<T[]>
```

src/peer-store/book.js
```diff
@@ -30,7 +29,7 @@ class Book {
     /**
      * Map known peers to their data.
      *
-     * @type {Map<string, Array<Data>}
+     * @type {Map<string, Array<*>}
```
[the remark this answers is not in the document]
Gozala: You can make `Book` generic by adding `@template T` above class definition. That would also make `options.eventTransformer` bounded.

Gozala (commented):
Just added few nits & general feedback

src/address-manager/index.js
vasco-santos: Yes, I mentioned this topic in the main post of the PR. 
I would also prefer to use `Multiaddr`, but the drawback is doing the `@typedef` import in every single file that uses them, which are a lot. I don't have a strong opinion on this, and I am happy to change if we think that is a better solution.

src/circuit/auto-relay.js
vasco-santos: Probably mentioning in the one below, right?
The current approach is: `xProperties` to libp2p references needed and `xOptions` for user options, so that we can use `xOptions` on the top level config docs. We can definitely change `xOptions` to `xConfig` and I would be happy in having a better name for `xProperties` if you have any suggestions

Gozala: Bo I was referring to `xProperties` actually, maybe I’ve misread the code, but it looked like dependencies injected into `AutoRelay` and we called such things with `xConfig` e.g.:

https://github.com/ipfs/js-ipfs/blob/530767d85d85b5932abdb87721e395c0a214b148/packages/ipfs-core/src/components/dag/get.js#L8-13

src/address-manager/index.js
vasco-santos: changed to `Multiaddr` everywhere

src/circuit/auto-relay.js
vasco-santos: I would like to get a good way for here. So, in the libp2p stand point we have lots of places where we inject libp2p, or some libp2p components on other libp2p components. Moreover, these libp2p components have configurations that users can modify.
We need to differentiate between them. My initial approach was to make `ComponentNameProperties` for the libp2p components we provide, and `ComponentNameOptions` for user options. Naming config for dependencies that are not "configurable" does not seem the best naming to me. However, I am not totally happy with `xProperties`, so if we can find a better convention would be great

src/content-routing.js
vasco-santos: We are doing the `Array<>` everywhere in libp2p. But, I think we can change to `T[]` as it is also faster to type.

vasco-santos:
The first batch of review comments were addressed. I will test this with lodestar over the week.

Meanwhile, @Gozala created https://www.npmjs.com/package/typed-events.ts for the EventEmitter stuff.
After syncing with him, we might want to get the declaration file from there in and use https://www.npmjs.com/package/proper-event-emitter per https://github.com/libp2p/js-libp2p/pull/761 (we also have emittery and some modules and we need to double check ig we should just use it). Accordingly, I think we can land the typedefs for EventEmitter in a new PR to not block this.

src/circuit/auto-relay.js
Gozala: Maybe `AutoRelayComponents` then ? 

The reason I end up calling those config in js-ipfs is because it is configuration of the component. It is true that user can't modify those but who ever creates a component can provide alternative configuration e.g. in memory repo vs on disk repo etc...

Things that user can configure are referred as `PrefixOptions`.


I don't have strong feeling about naming convention itself, just would love to settle on the same convention across ipfs and libp2p

vasco-santos: Components seems a good alternative. Will wait for other people to weight in before renaming everywhere

vasco-santos:
@hugomrdias @Gozala @jacobheun 

This is ready and already tested successfully with lodestar build. 

Unfortunately, it seems that we will not be able to use templates in the PeerStore Book and I needed to add any to the data types for now (more info on: https://github.com/libp2p/js-libp2p/pull/821 -- I will create an issue in typescript as the aegir type check works ok with #821 templates, but then the `*.ts.d` are not correctly generated which breaks user builds)

src/ping/index.js
```diff
@@ -1,30 +1,39 @@
...
  * @returns {Promise<number>}
  */
 async function ping (node, peer) {
+  // @ts-ignore multiaddr might not have toB58String
   log('dialing %s to %s', PROTOCOL, peer.toB58String ? peer.toB58String() : peer)
 
   const { stream } = await node.dialProtocol(peer, PROTOCOL)
 
-  const start = new Date()
+  const start = new Date().getTime()
```
Gozala, line 36: Seems like following should do, if not please add a code comment explaining why this instead.

```suggestion
  const start = Date.now()
```

src/record/peer-record/index.js
```diff
@@ -58,8 +66,8 @@ class PeerRecord extends Record {
   /**
    * Returns true if `this` record equals the `other`.
    *
-   * @param {Record} other
-   * @returns {boolean}
+   * @param {PeerRecord} other
+   * @returns {other is Record}
```
Gozala, line 70: Since other is already required to be a `PeerRecord` instance which implements `Record` interface it is granted so this return type will have no effect. 

```suggestion
   * @returns {boolean}
```

Gozala (changes requested):
Thanks @vasco-santos this must have been a lot of work & I can’t wait for this to lang and propagate across code base 🤩

I added bunch of minor comments and called out few instances where use of generics seems more appropriate. I think it’s mostly there, I just would really like type-check action to get uncommented to ensure all of this type-checks.

src/address-manager/index.js
```diff
@@ -1,23 +1,25 @@
...
-log.error = debug('libp2p:addresses:error')
-
 const multiaddr = require('multiaddr')
 
 /**
- * Responsible for managing the peer addresses.
- * Peers can specify their listen and announce addresses.
- * The listen addresses will be used by the libp2p transports to listen for new connections,
- * while the announce addresses will be used for the peer addresses' to other peers in the network.
+ * @typedef {import('multiaddr')} Multiaddr
```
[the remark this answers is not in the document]
vasco-santos: Once https://github.com/multiformats/js-multiaddr/pull/159 is shipped, I intend to capitalize multiaddr. The reason is that the exported multiaddr is a factory and we do lowercase for factories and uppercase for classes. In the multiaddr PR, we are moving into exporting the class by default, which will just be what unblocks this.

Should we skip any comments on it and wait for the multiaddr update? Or should I add the comment?

src/circuit/auto-relay.js
vasco-santos: yeah I agree. But let's keep the discussion open until hugo/jacob comments

src/connection-manager/index.js
```diff
@@ -159,15 +169,13 @@ class ConnectionManager extends EventEmitter {
...
    */
   setPeerValue (peerId, value) {
     if (value < 0 || value > 1) {
       throw new Error('value should be a number between 0 and 1')
     }
-    if (peerId.toB58String) {
-      peerId = peerId.toB58String()
-    }
-    this._peerValues.set(peerId, value)
+    this._peerValues.set(peerId.toB58String(), value)
```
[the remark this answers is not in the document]
vasco-santos: Yes, it accepted string in the past. However, we already have in the [API docs](https://github.com/libp2p/js-libp2p/blob/master/doc/API.md#connectionmanagersetpeervalue) `PeerId` as the the type. So, this should probably be legacy support and it is better if we just let user take care of type coercion as you mentioned

src/index.js
```diff
@@ -29,22 +30,74 @@ const PubsubAdapter = require('./pubsub-adapter')
...
+ * @property {import('./circuit').RelayAdvertiseOptions} advertise
+ * @property {import('./circuit').HopOptions} hop
+ * @property {import('./circuit').AutoRelayOptions} autoRelay
+ *
+ * @typedef {Object} Libp2pConfig
+ * @property {any} [dht] dht module options
+ * @property {PeerDiscoveryOptions} [peerDiscovery]
+ * @property {Pubsub} [pubsub] pubsub module options
+ * @property {RelayOptions} [relay]
+ * @property {Object} [transport] transport options indexed by transport key
```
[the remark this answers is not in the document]
vasco-santos: TransportOptions might be anything, as it is highly coupled with the transport being configured. Simplifying this for now into `Record<string, Object>`.

We will look into improving the libp2p configuration setup in a couple of releases. We will probably be able to improve this by that time as this is outside the interface scope.

src/index.js
```diff
@@ -29,22 +30,74 @@ const PubsubAdapter = require('./pubsub-adapter')
...
+ * @property {boolean} autoDial
+ *
+ * @typedef {Object} RelayOptions
+ * @property {boolean} enabled
+ * @property {import('./circuit').RelayAdvertiseOptions} advertise
+ * @property {import('./circuit').HopOptions} hop
+ * @property {import('./circuit').AutoRelayOptions} autoRelay
+ *
+ * @typedef {Object} Libp2pConfig
+ * @property {any} [dht] dht module options
```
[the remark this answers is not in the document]
vasco-santos: I can change it to object. I have in the PR description a list of things that should come into follow up, where the DHT is included. I can also create the issue and already link it. What do you think?

src/index.js
```diff
@@ -147,6 +201,7 @@ class Libp2p extends EventEmitter {
     })
 
     if (this._config.relay.enabled) {
+      // @ts-ignore Circuit prototype
       this.transportManager.add(Circuit.prototype[Symbol.toStringTag], Circuit)
```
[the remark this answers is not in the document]
vasco-santos: We use `Circuit.prototype[Symbol.toStringTag]` as a way to identify transports, so that we provide in libp2p config transport options as:

```js
transports: {
  [Circuit.prototype[Symbol.toStringTag]]: { // .... }
}
```

Not sure if this is what you were looking for?

src/index.js
```diff
@@ -317,12 +376,12 @@ class Libp2p extends EventEmitter {
...
    *
    * @param {PeerId|Multiaddr|string} peer - The peer to dial
-   * @param {object} options
+   * @param {object} [options]
    * @param {AbortSignal} [options.signal]
    * @returns {Promise<Connection>}
    */
   dial (peer, options) {
-    return this.dialProtocol(peer, null, options)
+    return this.dialProtocol(peer, undefined, options)
```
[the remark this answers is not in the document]
vasco-santos: Changing to `null|string|string[]`. No protocols can now be `null`, `[]` or ``. We might remove null support further on, but I would not do another breaking change at the moment.

src/keychain/cms.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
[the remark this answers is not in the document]
vasco-santos: There are still a lot of issues with the keychain types + libp2p crypto + dependency on getting [ipfs/interface-datastore#56](https://github.com/ipfs/interface-datastore/pull/56). This is one of the items in the follow up work. Should I create the issue with all the follow up work and link it on the `ts-nocheck`?

This file itself has a few things that need to change

src/keychain/index.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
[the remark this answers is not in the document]
vasco-santos: There are still a lot of issues with the keychain types + libp2p crypto + dependency on getting [ipfs/interface-datastore#56](https://github.com/ipfs/interface-datastore/pull/56). This is one of the items in the follow up work. Should I create the issue with all the follow up work and link it on the `ts-nocheck`?

This file itself has a few things that need to change

src/keychain/util.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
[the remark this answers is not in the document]
vasco-santos: There are still a lot of issues with the keychain types + libp2p crypto + dependency on getting [ipfs/interface-datastore#56](https://github.com/ipfs/interface-datastore/pull/56). This is one of the items in the follow up work. Should I create the issue with all the follow up work and link it on the `ts-nocheck`?

This file itself has a few things that need to change

src/metrics/old-peers.js
```diff
@@ -6,9 +6,10 @@ const LRU = require('hashlru')
  * Creates and returns a Least Recently Used Cache
  *
  * @param {number} maxSize
- * @returns {LRUCache}
+ * @returns {any}
```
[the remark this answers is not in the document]
vasco-santos: The issue is that we change the `LRU` implementation, adding the `delete` function as the `remove` function.  This causes other issues with the return types. As this is only used internally, it is also one of the items for follow up work in the PR descruption.

src/metrics/index.js
```diff
@@ -1,7 +1,8 @@
+// @ts-nocheck
```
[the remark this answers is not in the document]
vasco-santos: As this is only used internally, it is also one of the items for follow up work in the PR description. There is a large set of type issues that need to be fixed here with a small value now. Should I create the follow up issue and reference it here?

src/metrics/stats.js
```diff
@@ -68,7 +71,7 @@ class Stats extends EventEmitter {
   /**
    * Returns a clone of the current stats.
    *
-   * @returns {Map<string, Stat>}
+   * @returns {Object}
```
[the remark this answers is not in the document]
vasco-santos: It is not. Metrics still have a lot of work to be done as follow up

src/peer-store/book.js
```diff
@@ -10,16 +10,19 @@ const {
 const passthrough = data => data
 
 /**
- * The Book is the skeleton for the PeerStore books.
+ * @typedef {import('./')} PeerStore
  */
+
 class Book {
   /**
+   * The Book is the skeleton for the PeerStore books.
```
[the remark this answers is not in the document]
vasco-santos: Yes, it should probably have a composed template. I started doing it and it was working good here. However, when I tried to integrate this in lodestar, I was getting issues on building their project. You have more information on this in https://github.com/libp2p/js-libp2p/pull/802#issuecomment-738128860 . I have it done in #821 and I already talked with @hugomrdias about it. It seems like a typescript issue (I need to open an issue there), and we should track it before changing it here.

src/record/peer-record/index.js
vasco-santos: It errors with:

```
Property 'equals' in type 'PeerRecord' is not assignable to the same property in base type 'Record'.
  Type '(other: PeerRecord) => boolean' is not assignable to type '(other: any) => other is Record'.
    Signature '(other: PeerRecord): boolean' must be a type predicate.
```

Per https://github.com/libp2p/js-libp2p-interfaces/blob/feat/add-types/src/record/types.ts#L20

Should I just change the param to any?

src/circuit/circuit/hop.js
```diff
@@ -1,22 +1,39 @@
...
-module.exports.handleHop = async function handleHop ({
+/**
+ * @typedef {import('libp2p-interfaces/src/connection').Connection} Connection
+ */
+
+/**
+ * @typedef {Object} HopRequest
+ * @property {Connection} connection
+ * @property {any} request
+ * @property {any} streamHandler
```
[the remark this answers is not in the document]
vasco-santos: Added template and Request definition to types.ts

Gozala:
> Unfortunately, it seems that we will not be able to use templates in the PeerStore Book and I needed to add any to the data types for now (more info on: #821 -- I will create an issue in typescript as the aegir type check works ok with #821 templates, but then the `*.ts.d` are not correctly generated which breaks user builds)

I think the issue you may be referring to is https://github.com/microsoft/TypeScript/issues/41258 which got fixed just yesterday. But in the meantime there is also a workaround that projects can use:
https://github.com/multiformats/js-multiformats#typescript-support

vasco-santos:
> I think the issue you may be referring to is microsoft/TypeScript#41258 which got fixed just yesterday.

Thanks! I think it is that! ❤️  Sadly, they did not release it yet.

> But in the meantime there is also a workaround that projects can use:
https://github.com/multiformats/js-multiformats#typescript-support

I feel it is better to add that as a follow up PR. Seems odd to me to release `0.30` with type definitions for TypeScript as one of the headlines, but then people need to `skipLibCheck` it. I would rather get it in when the new typescript is released as a patch. 

What do you think? Any concerns on this approach? From the user stand point, not having template specified in the book yet will not have influence as they will use the book implementations that export the correct types.

src/record/peer-record/index.js
Gozala: Ok so there are couple of things here:

1. That error occurs because interface says you can pass **anything** to the method, but then implementation contradicts it by saying by saying you only pass `PeerRecord`, implying that you can't pass `number` while interface says you can do that. In other words you can accept wider range of inputs (superset types) in implementations than type / interface requires and be compatible, but you can not do other way around.
2. I'm honestly confused by `equals(other: any): other is Record`. Using `other is Type` kind of annotations are only useful for predicates that are meant to be used for type refinements e.g. [example illustrating this](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCIFcEGtkG8CwAUMsgI7qJYAUAlAFzIBuA9sACZEC+RRoksiFKmYBzPERIAjOFGr0mrDoW6EiCZiADOYZME0ZsyALzIqjOABt0EOnBABPeuaso9aTDiMA+CU0vXkADJA5DB7AAcIZhg-FwA6ckpjIxMAchh0EAQwYA1UnkIIAA9w5igddS0dAHcyrFARAFEiuABbcIsUEzN-G2ERAB8DLBpjL3FiXRiqPWGelxpRgkmSAHpV5ABJEFCACzdJC2ZDMMjkaNiAtwADYevfEmdrBIpsWl8VX2Bpp5Rg0IiURivxeSRSJgARBksjkNBClg9kOsSCjUWj0SQAHrY7HIAAKUGYkXK9mQqUS2FSyDYzAgmmQIGYOmKeh0GgBZ1Sw2QAzQolScWQiJBFLkiORKJxuIJROgYTJoqpNLpDKZyBZ2nOO1OKC5Hh5fJEAo+XCAA)
   ```ts
    interface Duck {
      quack(): void
    }

    interface Dog {
      bark(): void
    }

    const isDuck = (value:any):value is Duck =>
      value && typeof value.quack === 'function'

    export const workingExample = (value:Dog|Duck) => {
      if (isDuck(value)) {
        // In this block type of value is `Duck`
        value.quack()
      }

      if (value && typeof value.quack === "function") {
        //                      ^^^^^ Property 'quack' does not exist on type 'Duck | Dog'. 
        value.quack()
        //    ^^^^^ Property 'quack' does not exist on type 'Duck | Dog'.
      }
    }
   ```
   Note that in some cases on might restrict type of value in predicate in some way (does not need to be `any`), but the point is that TS essentially treats such predicate functions to do type refinements.

   I am not sure why one would want to do type refinement with `equals` check e.g. because if type of `value` in `value.equals(other)` is known and it's equal to `other` than `value` could be used in it's place e.g.:

   ```ts
   const example = (duck:Duck, other:Duck|Dog) {
     if (duck.equals(other)) {
       duck.quack() // as opposed to other.quack()
     }
   }
   ```

3. I think you may want to use `Record` for parameter type, as in it does not matter if `PeerRecord` implementation is used or some compatible `OtherRecord` implementation.

4. I think generally it is best to avoid `any` unless you want to tell TS to just shut up about that **value**. That is because in that case TS will not complain no matter you do with it. I think in most cases actual intent is to say that value is something of any type (via generics `interface Record { equals <T>(other:T):boolean }` or more simply that value is of `unknown` type (via `interface Record { equals(other:unknown):boolean }`). This inspired me to write [(what I hope is fun) explainer](https://www.typescriptlang.org/play?#code/MYGwhgzhAEBCD2APAPAFQHzQN4FgBQ00AbmCAK4CmAXNKvocPAHYQAuATmcK-OwBQly1WgEps9QtFYALAJYQAdIMrQAvMVKUJAXwlkmAIyR8RVVOIKT2FVmXZMpcxcoo78uvPllNWFdgDMwYApoAFEARwBBJgBPC0IAegToAHl-XwcABwp4TJAQsggQgAMwWOLoMqZ4VjBWWWZHCjj4dIoHAFsKMsc66AAqftlWQehgHoMQspiZbwBzCSToAHc5fKkY7LHpCmAAaz8pMAOYGTqAGmhMsHZpMEyIS554aBBh63AQOLKAEylrOowQbDUaQSqxWZMOYKCQUcJkUgQPjDGjTUzQIzwfJldz4PF4UCQGAANQAjGhMBREBkfjAECgMNBZB08hQuj4YBFonFcJYltE-qt2kyHDIQszWezavVGsCRv0macAawYGCIPAupC5tAINlgLJ-LJgNBkaKnGNIK4+ckwI5vCthtJoMUFPojIgTBVNdJ4D8xKxNiFgDt9ocILJ8qwvtB4JMYo9Km12JUQCAmawVmCeCAfotkjx-t0M3LQaqIXIoZUM6Q079ehnvKBKLTKnbK8tHc7XYZjCIvTYfT8YZY4QiQEiUWj4lYbHYzfJu+6TGpVOphove248B58EsAArsWP5DqK+un6oZqlBKNxVZ9MBkGS8aBdMqsS7VaB+Q-sVXWaDsGA8gUH8fSMCyEZWks9RdJc3hsN0fxinEEB2IEwRVgB+gwSEgC8G4ARHv4EwFDLNAZJ8AArCICijoilEiLuySSMxAB6bHsextCBqE7A-jQ65ukg541K2-j6NwDRMPi+LeL4ARBCEEQAMoagO8zTksADqdwZtkuTrIUY7Rq+LBSC8EBgHE8hniWCrjA4kytoM6revMoytOCcQBtksLwoi0DICprlQugyKsFQQVqVC6KYtiUnbvihJQGRABMFJftS7QtvS6USse7QqmE4SRVqGnJAAkg4qxGk64xFBsWzBn4BzJh2qZjBqeRAVVOwOLZp4cEWIGVEC-QuVFczuf4eY6nqBpGg1FCXG8Bw6qppVnBmdwwMUgken2L4Dr6w6ELR46BetFZzKFKIlVdYi8sx1i2PYdqKHty6rmurAbvtEiJExzFA8DIPQBxLHQAeuR+AG0AAOR7XD0A-PAFAwBeGXyBmjTeSEcN3fMcMKHwKUAMykwAnAxlgeDuBLgMlxKk+lVI0nSSC5Sy+UckVBOVo90BaXIwYOu12KgQ4D5PsmBYWTElyBP+7TwGQczSJc5XQHMIkxCryaZPAskKNA5X+AMQzysJ1bm+6gwzbrZAWkwcMZi6H19pcBiPisvXmyCCrWbaxT0hUYoOIwZA5tADtXNi9Vuz2+3FAA-AAhL5Y4QBdwXXWFNB83MD3-QBs6vcMIpsGUwQefS0AAGR129v2fau6bN9ThC09JeDEaRTP0TRfnjvRgvJIE45Wl4Ph+OhinhAAqkwezwMsDgC0slXQOhGZ3rpfjquHKtR3cRAhBZsh-B0vBn3lXk6YtMZm-18iXMsFCppcgz6Mvq+DMbGf+Tzt-aoq8YrwCxN0eKdMkokgACws0ykwbKHNGR5TZAVTkC8l4rzXjNTe1URZ1RCLjbYzVDhtTTOBLq9ohR9QtqMayg06jDSzFdaAewQGiheCfK0ANDpPj+LtROnoTpfiHhOcK0BgE4KLpYQgz05xN3diub67di5LFBpo1iHFIaHmyOwWGCNE5IxRmjaAGMqRYxjKKQM8NpGrzhluaBDMSQUQQWzOAKDMBoKlJgxeP9cHWmgNpGqos0zi3BJUR8PoZbmUspcB2cN-zsFkGrascxupMifvQgOo1v44NGCfeYM1vS+i7O7Co9lzEiUchwnBxtYDewNH7S2gcbZIHckwaMYdXg2CkTtduxQAHDxRFIpenDZFPVLg4cu8FahMGrmbWuDclHCLEF9Nu7snH4l7mRCifAADs1EzpIiOaPLeiIrS7OJPso5g9M58F2fSQ5IgxDQU4BQIAA)

If I had to guess you either want `equals(other:unknown): boolean` or bit more constrained `equals(other:Record):boolean` in both interface and a class.

src/address-manager/index.js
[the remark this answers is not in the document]
Gozala: Thanks for explaining, given that things are well on way to change I don't think comment is needed.

Gozala (approved):
> What do you think? Any concerns on this approach? From the user stand point, not having template specified in the book yet will not have influence as they will use the book implementations that export the correct types.

src/circuit/transport.js
```diff
@@ -89,7 +101,7 @@ class Circuit {
         remoteAddr,
         localAddr
       })
-      const type = CircuitPB.Type === CircuitPB.Type.HOP ? 'relay' : 'inbound'
+      const type = request.Type === CircuitPB.Type.HOP ? 'relay' : 'inbound'
```
Gozala, line 104: I believe it should be following instead:

```suggestion
      const type = request.type === CircuitPB.Type.HOP ? 'relay' : 'inbound'
```

Gozala: And a reason TS failed to catch this is because `StreamHandler` is generic and on line 56 `streamHandler` was inferred as `StreamHandler<any>`.

src/keychain/cms.js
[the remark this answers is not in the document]
vasco-santos: Thanks, I will add a checklist there with all the remaining work for better tracking it, once this gets merged

src/peer-store/book.js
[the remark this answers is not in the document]
Gozala: I have looked into this & turns out this was different issue fixed in typescript@4.1.x. So, I have submitted #831 against #821, which addresses this problem and fixes few type miss matches that have came up.

I think it would be good to take a look at that pull as it also uncovered some issues that I have missed when reviewing this.

src/record/peer-record/index.js
vasco-santos: Thanks for the example and detailed go through. I am changing the interface + here to unknown and leverage the `instance of`

src/index.js
```diff
@@ -29,22 +30,74 @@ const PubsubAdapter = require('./pubsub-adapter')
...
+   * Libp2p node.
+   *
+   * @class
+   * @param {Libp2pOptions} _options
+   */
   constructor (_options) {
     super()
     // validateConfig will ensure the config is correct,
     // and add default values where appropriate
     this._options = validateConfig(_options)
```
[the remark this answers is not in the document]
vasco-santos: added this, but need to revert. Some libp2p components use it directly

src/index.js
```diff
@@ -29,22 +30,74 @@ const PubsubAdapter = require('./pubsub-adapter')
...
 
+    /** @type {PeerId} */
```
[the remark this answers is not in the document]
vasco-santos: In this case, the `peerId` is not inferred because of `validateConfig` uses the `merge-options` module and it is lost. I cannot also assume `this._options` as `Libp2pOptions` at the moment, because it will not include some stuff (like the DHT options) that are added by `validateConfig` as default values. I could add a return type to `validateConfig`, but it would overlap with follow up work like types for DHT, peer-discovery, ... (that should be imported from these modules)

My plan here is to improve this once we work on the configuration improvements, as things will probably change drastically.

src/circuit/transport.js
vasco-santos: good catch! I also added a type for line 56: `/** @type {import('./circuit/stream-handler')<CircuitRequest>} */`

PR #718, opened 2020-07-24
docs: production guides for delegated routing

vasco-santos:
This PR adds the main structure for production guides as part of #696  

This depends on #507

PR #804, opened 2020-11-19
docs: production guide base setup

vasco-santos:
These were started in #718, but doing all of them in the same PR will just get things blocked longer. So, this PR is the first of a set of PRs that should land over time. One per guide to ease reviews and prioritize  accordingly.

This PR adds the base README for the production guides.

Needs:
- [x] https://github.com/libp2p/js-libp2p-hop-relay-server/pull/2

### DIFF
--- a/doc/production/README.md
+++ b/doc/production/README.md
@@ -1,0 +1,38 @@
+# Production
+
+Nowadays, you can run JavaScript code in several different environments, some of them with their own particularities. Moreover, you can use `js-libp2p` for a wide range of use cases. Different environments and different use cases mean different configurations and challenges in the network.
+
+Libp2p nodes can vary from nodes behind an application, to infrastructure nodes that enable the network to operate and to be efficient. In this context, the Libp2p project provides public infrastructure to boost the network, enable nodes connectivity and improve constrained nodes performance. This public infrastructure should be leveraged for learning the concepts and experimenting. When an application on top of libp2p aims to move into production, its own infrastructure should be setup as the public nodes will be intensively used by others and its availability is not guaranteed.
+
+This guide aims to guide you from using the public infrastructure into setting up your own.
+
+## Table of Contents
+
+* [Production](#production)
+  * [Star servers](#star-servers)
+  * [Delegate nodes](#delegate-nodes)
+  * [Circuit Relay](#circuit-relay)
+
+## `webrtc-star` servers
+
+While the libp2p core codebase aims to work in multiple environments, there are some limitations that are not possible to overcome at the time of writing. Regarding `webRTC`, at the time of writing a set of star servers are needed to act as a rendezvous point, where peers can learn about other peers (`peer-discovery`), as well as exchange their SDP offers (signaling data).
+
+You can read on how to setup your own set of delegated nodes in [libp2p/js-libp2p-webrtc-star/DEPLOYMENT.md](https://github.com/libp2p/js-libp2p-webrtc-star/blob/master/DEPLOYMENT.md).
+
+It is worth pointing out that with new discovery protocols on the way, as well as support for distributed signaling, the star servers should be deprecated on the long run.
+
+## Delegate nodes
+
+Libp2p nodes in scenarios such as browser environment and constrained devices will not be an efficient node in the libp2p DHT overlay, as a consequence of their known limitations regarding connectivity and performance.
+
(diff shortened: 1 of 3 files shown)


doc/production/README.md
```diff
@@ -0,0 +1,38 @@
...
+Libp2p nodes can vary from nodes behind an application, to infrastructure nodes that enable the network to operate and to be efficient. In this context, the Libp2p project provides public infrastructure to boost the network, enable nodes connectivity and improve constrained nodes performance. This public infrastructure should be leveraged for learning the concepts and experimenting. When an application on top of libp2p aims to move into production, its own infrastructure should be setup as the public nodes will be intensively used by others and its availability is not guaranteed.
+
+This guide aims to guide you from using the public infrastructure into setting up your own.
+
+## Table of Contents
+
+* [Production](#production)
+  * [Star servers](#star-servers)
+  * [Delegate nodes](#delegate-nodes)
+  * [Circuit Relay](#circuit-relay)
```
vasco-santos, line 14: Add SSL per https://github.com/libp2p/js-libp2p-relay-server/pull/2#discussion_r533403300

doc/production/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## `webrtc-star` servers
+
+While the libp2p core codebase aims to work in multiple environments, there are some limitations that are not possible to overcome at the time of writing. Regarding `webRTC`, at the time of writing a set of star servers are needed to act as a rendezvous point, where peers can learn about other peers (`peer-discovery`), as well as exchange their SDP offers (signaling data).
+
+You can read on how to setup your own set of delegated nodes in [libp2p/js-libp2p-webrtc-star/DEPLOYMENT.md](https://github.com/libp2p/js-libp2p-webrtc-star/blob/master/DEPLOYMENT.md).
+
+It is worth pointing out that with new discovery protocols on the way, as well as support for distributed signaling, the star servers should be deprecated on the long run.
+
+## Delegate nodes
```
jacobheun, line 25: These are IPFS only, we should clearly note that. Our only supported delegates atm leverage the IPFS http api.

doc/production/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Table of Contents
+
+* [Production](#production)
+  * [Star servers](#star-servers)
+  * [Delegate nodes](#delegate-nodes)
+  * [Circuit Relay](#circuit-relay)
+  * [SSL](#ssl)
+
+## `webrtc-star` servers
```
jacobheun, line 17: I think it would be helpful to more clearly state what environments these are useful for, which is to dial a browser. If that's not really in my user flow, I don't need them.

Also listing alternatives would be helpful. In this instance, relays.

doc/production/README.md
```diff
@@ -0,0 +1,43 @@
+# Production
+
+Nowadays, you can run JavaScript code in several different environments, some of them with their own particularities. Moreover, you can use `js-libp2p` for a wide range of use cases. Different environments and different use cases mean different configurations and challenges in the network.
+
+Libp2p nodes can vary from nodes behind an application, to infrastructure nodes that enable the network to operate and to be efficient. In this context, the Libp2p project provides public infrastructure to boost the network, enable nodes connectivity and improve constrained nodes performance. This public infrastructure should be leveraged for learning the concepts and experimenting. When an application on top of libp2p aims to move into production, its own infrastructure should be setup as the public nodes will be intensively used by others and its availability is not guaranteed.
+
+This guide aims to guide you from using the public infrastructure into setting up your own.
+
+## Table of Contents
```
jacobheun, line 9: I would reframe this guide. Right now it is focused on the services and less on what the user actually needs. I would think about it in terms of flows. Something like:

- Joining the Network
  - How do I join the public network, how can I run my own bootstrap nodes? JS isn't currently great for being bootstrap nodes, because we need to improve the DHT first to be an effective server.
- Connecting to browser nodes
  - This is where webrtc-star and relays come in
- Querying the network from the browser
  - Delegates go here. In theory the DHT _can_ work, it's just slow and requires WSS or webrtc-star support on all dialable nodes, we just really don't recommend it and if you're using the public dht it's really not going to work.

vasco-santos: Yes, I think this makes sense

vasco-santos: I added a "Others" section where I included the SSL subsection.  SSL and other things will be transversal to the flows described

PR #806, opened 2020-11-23
chore: update websockets

vasco-santos:
This PR updates websockets version, which comes with a BREAKING CHANGE on the filter. Only DNS+WSS addresses are now returned on filter by default in the browser. The tests were update to pass a custom filter so that we can test this locally with loopback websocket addresses

Needs:
- [x] https://github.com/libp2p/js-libp2p-websockets/pull/116

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "libp2p-secio": "^0.13.1",
     "libp2p-tcp": "^0.15.1",
     "libp2p-webrtc-star": "^0.20.0",
-    "libp2p-websockets": "libp2p/js-libp2p-websockets#feat/custom-address-filter-for-transport",
+    "libp2p-websockets": "^0.15.0",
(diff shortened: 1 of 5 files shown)


jacobheun (approved):
LGTM, I just want to note though that while the Websocket transport behavior is a breaking change, it's a breaking change in behavior only, and there is no reason why someone cannot use `libp2p-websockets` 0.15 with js-libp2p 0.29.x.

It's not necessary for this to land in 0.30, but since it's testing only it doesn't matter to consuming users.

vasco-santos:
>  it's a breaking change in behavior only, and there is no reason why someone cannot use libp2p-websockets 0.15 with js-libp2p 0.29.x

That is true, but it might break local testing environments of users like in`js-libp2p`. I would like to just change it to latest once we have the migration guide and this information in there

PR #808, opened 2020-11-25
docs: migration 0.29 to 0.30

vasco-santos:
This PR adds the migration guide to `0.30`


doc/migrations/v0.29-v0.30.md
```diff
@@ -0,0 +1,191 @@
...
+    pubsub: {
+      globalSignaturePolicy: SignaturePolicy.StrictNoSign
+    }
+  }
+})
+```
+
+### Addresses
+
+Libp2p has supported `noAnnounce` addresses configuration for some time now. However, it did not came with the best developer experience. In this release, we dropped the `noAnnounce` configuration property in favor of an `announceFilter` property function.
```
vasco-santos, line 99:
```suggestion
Libp2p has supported `noAnnounce` addresses configuration for some time now. However, it did not provide the best developer experience. In this release, we dropped the `noAnnounce` configuration property in favor of an `announceFilter` property function.
```

doc/migrations/v0.29-v0.30.md
```diff
@@ -0,0 +1,191 @@
...
+const libp2p = await Libp2p.create({
+  modules: {
+    // ... Add required modules according to the Configuration docs
+  },
+  peerId,
+  addresses: {
+    listen: ['/ip4/127.0.0.1/tcp/8000/ws'],
+    noAnnounce: ['/ip4/127.0.0.1/tcp/8000/ws'],
+  }
+})
```
jacobheun, line 117: `peerID` isn't necessary in here, nor is modules. I would just truncate for simplicity:
```suggestion
const libp2p = await Libp2p.create({
  addresses: {
    listen: ['/ip4/127.0.0.1/tcp/8000/ws'],
    noAnnounce: ['/ip4/127.0.0.1/tcp/8000/ws'],
  },
  // ... additional configuration per the Configuration docs
})
```

doc/migrations/v0.29-v0.30.md
```diff
@@ -0,0 +1,191 @@
...
+  modules: {
+    // ... Add required modules according to the Configuration docs
+  },
+  peerId,
+  addresses: {
+    listen: ['/ip4/127.0.0.1/tcp/8000/ws'],
+    // Filter function: (ma: Array<multiaddr>) => Array<multiaddr>
+    announceFilter: (multiaddrs) => multiaddrs.filter(m => !isPrivate(m))
+  }
+})
```
jacobheun, line 140: same thing here

```suggestion

const libp2p = await Libp2p.create({
  addresses: {
    listen: ['/ip4/127.0.0.1/tcp/8000/ws'],
    // Filter function: (ma: Array<multiaddr>) => Array<multiaddr>
    announceFilter: (multiaddrs) => multiaddrs.filter(m => !isPrivate(m))
  },
  // ... additional configuration per the Configuration docs
})
```

doc/migrations/v0.29-v0.30.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+It is important pointing out another change regarding address advertising. This is not an API breaking change, but it might have influence on your libp2p setup.
+Previously, when using the addresses `announce` property, its multiaddrs were concatenated with the `listen` multiaddrs and then they were filtered out by the `noAnnounce` multiaddrs, in order to create the list of multiaddrs to advertise. 
+In `libp2p@0.30`, we are taking the same approach as `go-libp2p` and if `announce` addresses are provided, only them are used for advertising. If no addresses are provided, the advertised addresses will be the `listen` addresses filtered out by the `announceFilter` (if provided).
```
jacobheun, line 145: I think it might be helpful to also reword this section to itemize it for clarity:
```suggestion
In `libp2p@0.30` the logic now operates as follows:

- If `announce` addresses are provided, only they will be announced (no filters are applied)
- If `announce` is not provided, the transport addresses will be filtered (if a filter is provided)
  - if the `announceFilter` is provide it will be passed the transport addresses
```

jacobheun (approved):
A couple minor things, otherwise this LGTM

PR #809, opened 2020-11-25
chore: use set-delayed-interval module on circuit

vasco-santos:
This PR makes circuit to use `set-delayed-interval` module per https://github.com/libp2p/js-libp2p/pull/798#discussion_r521950167


jacobheun (approved):
LGTM

PR #810, opened 2020-11-25
feat: add UPnP NAT manager

achingbrain:
Adds a NAT manager similar to go-libp2p that attempts to use UPnP/PMP to punch a hole through your router for any IPV4 tcp addresses you have configured via the [@motrix/nat-api](https://www.npmjs.com/package/@motrix/nat-api)* module.  Only works under node.js, I've added a note to use WebRTC for browser dialability.

Adds any hole-punched addresses to the node's observed addresses list and merged that list into `libp2p.multiaddrs` so it'll be shared with connected nodes when `identify` is run.

Adds configuration options under `config.nat`

Hole punching is async to not affect start up time

Attempts to detect double NATing, when the external IP address your router reports is a private network address

The external port configured on your router is high and random to minimise the chance of port collisions/other nodes on your network receiving your traffic.

Attempts to remove mappings when the node stops but doesn't attempt to remove stale mappings from the router on startup as the TTL expiry should take care of this. Not all hole-punching protocols support listing existing port mappings so I'm not sure this is even possible to do reliably and transparently.

Uses the node's libp2p version and peer ID as the mapping description - on my ISP provided router this appears in the port mapping screen which is useful to see which 

\* = I have no strong preference for this module - I tried it, it worked


src/nat-manager.js
```diff
@@ -0,0 +1,137 @@
...
+        publicPort: port,
+        privatePort: port,
+        protocol: transport.toUpperCase()
+      })
+
+      this._addressManager.announce.add(Multiaddr.fromNodeAddress({
+        family: 'IPv4',
+        address: publicIp,
+        port
+      }, transport))
```
jacobheun, line 94: The mapping code will only ever run once per TCP address we have, correct? The NAT Manager has an autoUpdate option, so I'm assuming the client itself will automatically update addresses if they change. Is that the case? If so, we're not updating our nat addresses at all.

If this does need to update over time ( I think it does ), how we're setting the addresses will need to change, as an add isn't going to be sufficient. Also, this add won't trigger an Identify Push. The consideration, https://github.com/libp2p/js-libp2p/tree/master/src/address-manager#modify-announce-addresses, is something we need to account for with this. cc @vasco-santos. At the bare minimum we'll likely need to leverage the work done for 0.30 and update our address record to cause the trigger of events in the subsystems, https://github.com/libp2p/js-libp2p/blob/0.30.x/src/transport-manager.js#L164.

achingbrain: > The NAT Manager has an autoUpdate option

Maybe this is poorly named and should be `autoRefresh` or `keepAlive` or something - it means to periodically tell the router to keep the UPnP mapping alive, otherwise it'll be removed once the TTL is reached.

jacobheun: `keepAlive` sounds like the way to go.

For adding the address I think we should add this as an observed address, and add that base logic to the address manager. We'll eventually want to factor in confidence to these, but for now we can just add Nat Manager addresses to it. The announce address logic is being updated to reflect go-libp2p, so if any announce address is provided it will cause the other listen addresses to be hidden which is not what we want to happen. @vasco-santos  what are your thoughts on this.

vasco-santos: I agree on getting the observedAddresses logic into the AddressManager would be the best way to go. We should gather also the observed addresses from [identify](https://github.com/libp2p/js-libp2p/blob/v0.29.3/src/identify/index.js#L207).

For creating the announce addresses, I think we should keep things as `0.30` currently does. This is, if we have user provided announce addresses only these addresses should be announced. If we do not have announce addresses, we will return the listenAddr + observedAddr

Regarding updating the self peer record and doing identify push to update the connected peers about the self multiaddrs changes, I think that we can make `AddressManager` an event emitter to trigger `updateSelfPeerRecord`. This would also unblock the programatic API for the AddressManager.

achingbrain:
Not sure why codecov is complaining about `src/circuit/auto-relay.js`, it's not changed in this PR.

achingbrain:
@vasco-santos @jacobheun since 0.30.x is out of the door, can this be merged now?

test/addresses/address-manager.spec.js
```diff
@@ -45,6 +54,57 @@ describe('Address Manager', () => {
...
+    const am = new AddressManager(peerId)
+
+    expect(am.observed).to.be.empty()
+
+    am.addObservedAddr(ma)
+    am.addObservedAddr(ma)
+    am.addObservedAddr(ma)
+
+    expect(am.observed).to.have.property('size', 1)
+    expect(am.observed).to.include(ma)
```
vasco-santos, line 79: Maybe worth to check that only one event was emitted

src/nat-manager.js
```diff
@@ -0,0 +1,158 @@
...
+  /**
+   * @class
+   * @param {object} options
+   * @param {PeerId} options.peerId - The peer ID of the current node
+   * @param {TransportManager} options.transportManager - A transport manager
+   * @param {AddressManager} options.addressManager - An address manager
+   * @param {boolean} options.enabled - Whether to enable the NAT manager
+   * @param {string} [options.externalIp] - Pass a value to use instead of auto-detection
+   * @param {string} [options.description] - A string value to use for the port mapping description on the gateway
+   * @param {number} [options.ttl] - How long UPnP port mappings should last for in seconds
```
vasco-santos, line 35:
```suggestion
   * @param {number} [options.ttl] - How long UPnP port mappings should last for in seconds (minimum 1200)
```

vasco-santos: Per Configuration.md this should be a minimum of 20 minutes, right?

src/nat-manager.js
```diff
@@ -0,0 +1,158 @@
...
+  constructor ({ peerId, addressManager, transportManager, ...options }) {
+    this._peerId = peerId
+    this._addressManager = addressManager
+    this._transportManager = transportManager
+
+    this._enabled = options.enabled
+    this._externalIp = options.externalIp
+    this._options = {
+      description: options.description || `${pkg.name}@${pkg.version} ${this._peerId}`,
+      ttl: options.ttl || 7200,
```
vasco-santos, line 50: Should we validate that this is bigger than 1200? What happens if the NatAPI module receives an invalid ttl?

src/nat-manager.js
```diff
@@ -0,0 +1,158 @@
...
+    for (const addr of addrs) {
+      // try to open uPnP ports for each thin waist address
+      const { family, host, port, transport } = addr.toOptions()
+
+      if (!addr.isThinWaistAddress() || transport !== 'tcp') {
+        // only bare tcp addresses
+        continue
+      }
+
+      if (host === '127.0.0.1' || host === '::1') {
```
vasco-santos, line 84: We can just use `libp2p-utils` here: https://github.com/libp2p/js-libp2p-utils/blob/master/API.md#multiaddr-isloopbackma

vasco-santos: I would say filter them out of addrs before the for loop. What do you think?

src/index.js
```diff
@@ -458,8 +473,14 @@ class Libp2p extends EventEmitter {
 
     const announceFilter = this._options.addresses.announceFilter || ((multiaddrs) => multiaddrs)
 
+    // dedupe multiaddrs
```
vasco-santos, line 476: Per https://github.com/libp2p/js-libp2p/pull/810/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json#r534270687 we need to understand if the logic above is what we really want. Currently, (per 0.30) if `announceAddrs` were configured, only them will be advertised in the network. This will mean that the observed addresses will not be included. I think they should probably be included with the announceAddrs otherwise we would not even run the NAT Mgr if we had announceAddrs. Moreover, signed peer records will be created but their multiaddrs will be the same.
What do you think? @achingbrain @jacobheun

test/nat-manager/nat-manager.node.js
```diff
@@ -0,0 +1,227 @@
...
+    return {
+      natManager,
+      addressManager,
+      transportManager
+    }
+  }
+
+  afterEach(() => Promise.all(teardown))
+
+  it('should map TCP connections to external ports', async () => {
```
vasco-santos, line 72: Should we also check if the consumePeerRecord is called? That would mean that the flow of an observed address being added will result in a new signed peer record to be created which will mean this will be send on identify push to the connected peers

vasco-santos (changes requested):

src/nat-manager.js
achingbrain: > Should we validate that this is bigger than 1200?

Can do.

> What happens if the NatAPI module receives an invalid ttl?

It'll ignore anything that's not a number or numeric string, reverting to a default value of thirty minutes.

src/nat-manager.js
achingbrain: It would involve passing over the list of addrs multiple times.  Do you think it'll make it more readable?

src/nat-manager.js
achingbrain: As I see it there are a number of asks here:

- [x] Store NAT punched addresses as observed addresses using the address manager
- [ ] Store observed addresses from identify using the address manager
- [x] Make AddressManager an event emitter that triggers updateSelfPeerRecord
- [ ] Send configured observed addresses plus either configured announceAddrs or the listenAddr during identify
- [ ] Invoke consumePeerRecord when addresses change

vasco-santos: @achingbrain 

>  Send configured observed addresses plus either configured announceAddrs or the listenAddr during identify
>  Invoke consumePeerRecord when addresses change

Your PR already does this. Identify listens for events of the PeerStore when an address change. Considering that you trigger the creation of a new signed peer record when new observed addresses is added, this will call consumePeerRecord which will emit the address change (if in fact it is a new address for the peer). This will trigger identify push to happens.

The "concern" here is when using announceAddresses. This is because we create a signed peer record by fetching `libp2p.multiaddrs` which are the addresses that a peer intends to let other peers know of. So, according to the current logic, if we have `announceAddresses` configured, `libp2p.multiaddrs` will always be the announceAddresses. Which makes, observed addresses never used. When a new signed peer record is created, it will include exactly the same addresses as before (the announce ones).

src/nat-manager.js
vasco-santos: It would simplify it, but an extra iteration. As the list of addresses should be small it does not seem problematic. However, we can probably go with just using `isLoopbackMultiaddr` in this if. At least, it will not be necessary to add the comment "ignore loopback addresses" as the function naming will be enough for readability

src/nat-manager.js
achingbrain: > Your PR already does this

Yeah, I got distracted while double checking the PR for the checkboxes..

test/nat-manager/nat-manager.node.js
achingbrain: Done!

test/addresses/address-manager.spec.js
achingbrain: Done!

src/address-manager/index.js
```diff
@@ -11,19 +13,26 @@ const multiaddr = require('multiaddr')
  * @property {string[]} [listen = []] - list of multiaddrs string representation to listen.
  * @property {string[]} [announce = []] - list of multiaddrs string representation to announce.
  */
-class AddressManager {
+class AddressManager extends EventEmitter {
```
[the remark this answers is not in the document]
achingbrain: Done!

src/nat-manager.js
achingbrain: Done!

src/nat-manager.js
achingbrain: Done!

src/index.js
achingbrain: The logic is changed - previously announceAddrs overrode allAddrs, subject to any announceFilter the user had configured.  Now announceAddrs overrides allAddrs, but is also joined with observedAddrs, and is subject to any announceFilter the user  has configured.

achingbrain:
@vasco-santos what else needs doing on this PR to get it merged?

doc/API.md
```diff
@@ -2055,6 +2055,13 @@ This event will be triggered anytime we are disconnected from another peer, rega
 - `peerId`: instance of [`PeerId`][peer-id]
 - `protocols`: array of known, supported protocols for the peer (string identifiers)
 
+#### Our addresses have changed
```
vasco-santos, line 2058:
```suggestion
### libp2p.addressManager

#### Our addresses have changed
```

src/index.js
```diff
@@ -446,22 +465,32 @@ class Libp2p extends EventEmitter {
   }
 
   /**
-   * Get peer advertising multiaddrs by concating the addresses used
-   * by transports to listen with the announce addresses.
-   * Duplicated addresses and noAnnounce addresses are filtered out.
+   * Get a deduplicated list of peer advertising multiaddrs by concatenating
+   * the listen addresses used by transports with any configured the
```
vasco-santos, line 469:
```suggestion
   * the listen addresses used by transports with any configured
```

vasco-santos (commented):
@achingbrain last suggestions left.

Can you also rebase this PR with master and update https://github.com/libp2p/js-libp2p/blob/master/doc/API.md#multiaddrs to reflect the changes this PR introduces?

vasco-santos (approved):
LGTM! 🚀

Issue #811, opened 2020-11-26
Preferred way to instantiate Libp2p bundle? Inherit or static create?


a1300:
- **Version**: 0.29.3
- **Platform**: Linux
- **Subsystem**:

#### Type: Question

I found in the examples both the `new Libp2p()` and `Libp2p.create()` methods. In the [API](https://github.com/libp2p/js-libp2p/blame/master/doc/API.md) documentation only the `new Libp2p()` way is shown.

Are they the same? Is it safe to inherit from `Libp2p` and add extra methods?

Example with `Libp2p.create()`
https://github.com/libp2p/js-libp2p/blob/d0a9fada32677ee3b810d5c2fb8d5e2c4f5a9fef/examples/discovery-mechanisms/1.js#L24-L27 

Example with `new Libp2p()`
https://github.com/libp2p/js-libp2p/blob/d0a9fada32677ee3b810d5c2fb8d5e2c4f5a9fef/examples/chat/src/dialer.js#L16-L21

vasco-santos:
Hello @a1300 

The preferred way to instantiate Libp2p is using `Libp2p.create`. When you create Libp2p via its constructor a PeerId must be provided, while create is a more out of the box solution that can receive the peerId or not. If you would like to PR the example dialer to use `Libp2p.create`, it would be welcome.

Regarding the docs, we actually have both documented in https://github.com/libp2p/js-libp2p/blob/master/doc/API.md#create . The `new Libp2p` is documented in a toggle that says alternative, as we really don't want people to use that. However, the class being exported enables some use cases where we want to create a Libp2p node but extend it with some behaviour.

a1300:
Hi @vasco-santos 
thanks for the fast answer! Didn't saw the alternative toggle. I think the alternative example (`new Libp2p)`) is missing a comment that specifies that a `peerId` is required. 

Will send a PR for the dialer example

vasco-santos:
> I think the alternative example (new Libp2p)) is missing a comment that specifies that a peerId is required.

If you would like to also PR this change, feel free to do it :)

a1300:
Update docs examples in PR: #814

PR #814, opened 2020-11-26
docs: use Libp2p.create() in examples (#811)

a1300:
I didn't changed the use of `new Libp2p()` in the `examples/delegated-routing` directory because this is already in progress in PR #507

https://github.com/libp2p/js-libp2p/blob/d0a9fada32677ee3b810d5c2fb8d5e2c4f5a9fef/examples/delegated-routing/src/libp2p-bundle.js#L23-L30

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -114,12 +114,25 @@ For Libp2p configurations and modules details read the [Configuration Document](
 
 ```js
 const Libp2p = require('libp2p')
+const TCP = require('libp2p-tcp')
+const MPLEX = require('libp2p-mplex')
+const { NOISE } = require('libp2p-noise')
 
-// specify options
-const options = {}
+async function main () {
+  // specify options
+  const options = {
+    modules: {
+      transport: [TCP],
+      streamMuxer: [MPLEX],
+      connEncryption: [NOISE]
+    }
+  }
 
-// create libp2p
-const libp2p = await Libp2p.create(options)
+  // create libp2p
+  const libp2p = await Libp2p.create(options)
+}
+
+main()
 ```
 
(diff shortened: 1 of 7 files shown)


doc/API.md
```diff
@@ -131,12 +131,20 @@ As an alternative, it is possible to create a Libp2p instance with the construct
...
-const options = {}
+;(async () => {
+  const peerId = await PeerId.create();
 
-// create libp2p
-const libp2p = new Libp2p(options)
+  // specify options
+  // peerId is required when Libp2p is instantiated via the constructor
+  const options = {
+    peerId
```
vasco-santos, line 142: Perhaps we should add to this example and the one above the modules as it is a mandatory property in the options. Can you add it like:

```js
const TCP = require('libp2p-tcp')
const MPLEX = require('libp2p-mplex')
const { NOISE } = require('libp2p-noise')
// ....
modules: {
    transport: [TCP],
    streamMuxer: [MPLEX],
    connEncryption: [NOISE]
}
```

doc/API.md
```diff
@@ -131,12 +131,20 @@ As an alternative, it is possible to create a Libp2p instance with the construct
 
 ```js
 const Libp2p = require('libp2p')
 
-// specify options
-const options = {}
+;(async () => {
```
vasco-santos, line 136: Perhaps we should do 

```js
async function main () {
  // ...
}

main()
```

This is a more generic approach than the async function being immediately called

vasco-santos (changes requested):
Thanks for the PR. Except for my two comments in the API doc, everything looks good

vasco-santos (approved):
LGTM

PR #821, opened 2020-12-03
chore: add typedefs peerstore book template

vasco-santos:
This is a WIP PR for adding templates to the PeerStore Book (only the last commit is relevant)

Unfortunately, it seems that there is a typescript issue on writing the `ts.d` files:

```js
/**
 * @typedef {Object} Address
 * @property {Multiaddr} multiaddr peer multiaddr.
 * @property {boolean} isCertified obtained from a signed peer record.
 *
 * @typedef {Object} CertifiedRecord
 * @property {Uint8Array} raw raw envelope.
 * @property {number} seqNumber seq counter.
 *
 * @typedef {Object} Entry
 * @property {Address[]} addresses peer Addresses.
 * @property {CertifiedRecord} [record] certified peer record.
 */
/**
 * @extends {Book<Entry, Address[], Multiaddr[]>}
 */
class AddressBook extends Book {
  /**
   * The AddressBook is responsible for keeping the known multiaddrs of a peer.
   *
   * @class
   * @param {PeerStore} peerStore
   */
  constructor (peerStore) {
```

generates:

```js
/**
 * @extends {Book<Entry, Address[], Multiaddr[]>}
 */
declare class AddressBook extends AddressBook_base {
```

instead of: 

```js
/**
 * @extends {Book<Entry, Address[], Multiaddr[]>}
 */
declare class AddressBook extends AddressBook_base<Entry, Address[], Multiaddr[]> {
```

Book:

```js
/**
 * @template Data, GetData, EventData
 */
class Book {
```


Gozala:
> Unfortunately, it seems that there is a typescript issue on writing the `ts.d` files:

Can we figure out what is the TS problem here ? I have attempted to reproduce this to make a report but I don't seem to be able to do that https://www.typescriptlang.org/play?experimentalDecorators=false&emitDecoratorMetadata=false&target=5&jsx=0&ts=4.0.5&useJavaScript=true#code/PQKhFgCgAIWgBALgUwLYAcA2BDF0AqUswUAxjgM4XQBCA9gB7QDeR0oEM0sC62ATtlQt8AX2gA3bJgCuyNsTak6AOwqJ+M0ojr8AFFNnIAlCwXREACwCWFAHSG50ALyTpctqLYyVAI0Z6pqxc3PzIiDL8KhY29o7yXF6QSVAcRHDwyAwoKgAm1Mz0DAA8KjKovsj8AHxJipDk2FS0jMi5AHLllfzQWTn5LUzB3GkhGXyCwsw+ANYqdADuKuLWiAoZYRFRBf50mMjYy+skXMgAjjLSFHqrQeabkdGr0NZqiIekyHQAZoPQAGT-F6IOw+fwMQIuZyuKy2UF+ALGTxQURAA even with the oldest version of TS.

There was a similar problem with `@implements`  that got fixed recently https://github.com/microsoft/TypeScript/issues/41258, I wonder if they are somehow related.

I also find `AddressBook_base` surprising here as in where does that `_base` comes from.

Gozala:
I did bit more investigation here and as it turns out this is is the case with typescript "4.0.3" but seems to work as expected with "4.0.5". I'm not sure why do we get here 4.0.3 because aegir depends on 4.0.x but resolving that should address this.

Gozala:
> I did bit more investigation here and as it turns out this is is the case with typescript "4.0.3" but seems to work as expected with "4.0.5". I'm not sure why do we get here 4.0.3 because aegir depends on 4.0.x but resolving that should address this.

I was wrong it was fixed in 4.1.x and not 4.0.5

Gozala:
I have submitted #831 against this branch, which fixes ts issues in the summary by using typescript@4.1.x

vasco-santos:
Closing in favour of #831

PR #823, opened 2020-12-04
chore: update pubsub example by disabled emit self

vasco-santos:
This updates the pubsub examples docs to the new default emitSelf: false.


jacobheun (approved):
pending CI LGTM

Issue #830, opened 2020-12-09
Followup: Types


Gozala:
https://github.com/libp2p/js-libp2p/pull/802 has deliberately opt-ed out couple of modules from TS checking with an intent to address that in followups. This is an issue to track that effort. 

What are we missing / What is next?

- [ ] https://github.com/libp2p/js-libp2p/pull/831
- [ ] Event Emitters
  - @Gozala created https://www.npmjs.com/package/typed-events.ts for the EventEmitter typedefs ❤️ .
After syncing with him, we might want to get the declaration file from there in and use https://www.npmjs.com/package/proper-event-emitter per https://github.com/libp2p/js-libp2p/pull/761 (we also have emittery and some modules and we need to double check ig we should just use it).
- [ ] Improve Circuit Message definitions
  - https://github.com/libp2p/js-libp2p/pull/831#issuecomment-742494373
- [ ] DHT should have typedefs for its options
- [ ] Metrics and Latency monitor needs to be finished (currently @ts-nocheck)
- [ ] Keychain (currently @ts-nocheck)
  - Needs https://github.com/ipfs/interface-datastore/pull/56 + keychain types
- [ ] Configuration Options
  - Probably together with https://github.com/libp2p/js-libp2p/issues/576

vasco-santos:
Thanks @Gozala ! Just added a checklist of things that need to be done in follow up

ryanio:
I found that the [`Libp2pModules`](https://github.com/libp2p/js-libp2p/blob/58d4f9a91542f180907689319d159b326f71096d/src/index.js#L66) type for the config is missing keys for `peerDiscovery` and `dht`.

vasco-santos:
Yes, thanks for adding that! We still need to create interfaces for `peerDiscovery`, `peerRouting` and so on to add the types here. For now they should be accepted as any while we finish that

peterbraden:
Has anyone started working on the types for DHT etc? If not, I plan on creating a PR for these today.

vasco-santos:
Hey @peterbraden 

Thanks for reaching out. It is not released yet but we just got types for the DHT per https://github.com/libp2p/js-libp2p-kad-dht/pull/214 This does not include integration of the DHT config types in libp2p core as this issue is tracking.

You can see the broad state of types in https://github.com/libp2p/js-libp2p/issues/659 . If you would like to help out in any other module let me know

peterbraden:
[...]

PR #831, opened 2020-12-10
Chore/add typedefs peerstore book template 2

Gozala:
This pull request does following:

- Adds aegir from https://github.com/ipfs/aegir/pull/689 so that new typescript@4.1.x is used
  - Which resolves issues with generic class extensions
- It addresses bunch of issues that type checker reported
- I merged #802 into which I think had some fixes that were missing here

@vasco-santos one other thing I want to call out here is that this change turns `StreamHandler` into non generic, that is because it uses `CircuitPB` encode / decode which in turn expects concrete structs. I am however not exactly sure I got all the details right, specifically:

1. According to `CircuitRelay` protobuf all fields are optional.
2. But than typedef in #802 was defining `CircuitRequest` with all fields (but `code`) mandatory.
4. Writes to `StreamHandler` seems to always contain `code & type`
5. Reads values from `StreamHandler` are assumed to have `type` `srcPeer`, `dstPeer`.

Which is what were reflected in the StreamHandler API reads return `CircuitRequest`s with all the mandatory fields and writes take `CircuitMessage`s with all the optional fields. However I am guessing that `CircuitMessage` in fact is more of a union of different message types where each corresponds to `Type` enum and has different subset of fields & probably each must have some subset of the the fields and that is why protobuf just marks them all optional.

If above hypothesis is correct it would make much more sense to represent each `message` type via actual type definition with it's fields and represent `CircuitMessage` as a type union. It would be good to sync up on this to fix those details up.


vasco-santos:
Added a commit merging in the new updates of the base types branch to check this

src/upgrader.js
```diff
@@ -231,6 +231,7 @@ class Upgrader {
             const { stream, protocol } = await mss.handle(Array.from(this.protocols.keys()))
             log('%s: incoming stream opened on %s', direction, protocol)
             if (this.metrics) this.metrics.trackStream({ stream, remotePeer, protocol })
+            // @ts-ignore - metadata seems to be required
```
vasco-santos, line 234: Fixed on https://github.com/libp2p/js-libp2p-interfaces/pull/74

vasco-santos:
This looks good to me! However, a new issue appeared as a result of this update...

```
File './node_modules/@types/node/events.d.ts' is not a module.
import { EventEmitter } from "node/events"
```

This only happens with the new ts. If you look into the dist on `libp2p/dist/src/connection-manager/index.d.ts:153:30` you will see the issue with vscode.

vasco-santos:
> Which is what were reflected in the StreamHandler API reads return CircuitRequests with all the mandatory fields and writes take CircuitMessages with all the optional fields. However I am guessing that CircuitMessage in fact is more of a union of different message types where each corresponds to Type enum and has different subset of fields & probably each must have some subset of the the fields and that is why protobuf just marks them all optional.

> If above hypothesis is correct it would make much more sense to represent each message type via actual type definition with it's fields and represent CircuitMessage as a type union. It would be good to sync up on this to fix those details up.


Yes, this sounds good! I will add it to the follow up list to not block this anymore

lidel:
Discussion continued in #830

PR #840, opened 2020-12-21
chore: add chat example test

vasco-santos:
Add chat example test and CI job.
`libp2p-bundle` file was also renamed into `libp2p`


examples/chat/test.js
```diff
@@ -0,0 +1,77 @@
...
+const execa = require('execa')
+const pDefer = require('p-defer')
+const uint8ArrayToString = require('uint8arrays/to-string')
+
+function startProcess(name) {
+  return execa('node', [path.join(__dirname, name)], {
+    cwd: path.resolve(__dirname),
+    all: true
+  })
+}
```
hugomrdias, line 13: i would add an `onData` as a param, do a the `.all.on('data', onData)` here and avoid the pDefer below

hugomrdias (approved):
LGTM, left a suggestion if you feel it makes sense.

examples/chat/test.js
vasco-santos: That is a good suggestion. However in this case we have to coordinate multiple parts. In the listener side, we need to wait for the listener to be ready to run the dialer process, but we also need to wait for a message to be received in the same process. This would not be possible if we block this waiting for an answer

PR #846, opened 2020-12-28
chore: add browser example test

vasco-santos:
This PR adds a test for the libp2p in the browser example. It also includes the browser setup for testing examples based on https://github.com/ipfs/js-ipfs/pull/2528


hugomrdias:
```js
'use strict'

const execa = require('execa')
const assert = require('assert')
const { chromium } = require('playwright');

async function run () {
  let url = ''
  const proc = execa('parcel', ['./index.html'], {
    preferLocal: true,
    localDir: __dirname,
    cwd: __dirname,
    all: true
  })
  
  proc.all.on('data', async (chunk) => {
    /**@type {string} */
    const out = chunk.toString()

    if(out.includes('Server running at')){
      url = out.replace('Server running at ', '')
    }

    if(out.includes('✨  Built in ')){
      try {
        const browser = await chromium.launch();
        const page = await browser.newPage();
        await page.goto(url);
        await page.waitForFunction(selector => document.querySelector(selector).innerText === 'libp2p started!', '#status')
        await page.waitForFunction(
          selector => { 
            const text = document.querySelector(selector).innerText 
            return text.includes('libp2p id is') &&
            text.includes('Found peer') &&
            text.includes('Connected to')
          }, 
          '#output',
          { timeout: 5000 }
        )
        await browser.close();
        
      } catch (err) {
        console.error(err)
        process.exit(1)
      } finally {
        proc.cancel()
      }
    }
  })

}

module.exports = run

```

With the above snippet in the test.js there no need for changes in the others files. What do you think ?

examples/libp2p-in-the-browser/package.json
```diff
@@ -28,6 +29,7 @@
     "babel-plugin-syntax-async-functions": "^6.13.0",
     "babel-plugin-transform-regenerator": "^6.26.0",
     "babel-polyfill": "^6.26.0",
+    "p-retry": "^4.2.0",
```
hugomrdias, line 32: probably not needed with the new setup

examples/nightwatch.conf.js
```diff
@@ -0,0 +1,35 @@
+'use strict'
```
hugomrdias, line 1: probably not needed with the new setup

examples/package.json
```diff
@@ -8,9 +8,15 @@
   },
   "license": "MIT",
   "dependencies": {
+    "chromedriver": "^87.0.0",
```
hugomrdias, line 11: probably not needed with the new setup

examples/package.json
```diff
@@ -8,9 +8,15 @@
...
     "execa": "^2.1.0",
     "fs-extra": "^8.1.0",
     "p-defer": "^3.0.0",
+    "http-server": "~0.11.1",
+    "nightwatch": "^1.2.4",
```
hugomrdias, line 16: probably not needed with the new setup

examples/test.js
```diff
@@ -7,6 +7,7 @@ const fs = require('fs-extra')
 const path = require('path')
 const execa = require('execa')
 const dir = path.join(__dirname, process.argv[2])
+const { startServer } = require('./utils')
```
hugomrdias, line 10: probably not needed with the new setup

examples/libp2p-in-the-browser/package.json
```diff
@@ -8,6 +8,7 @@
   ],
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "parcel build index.html",
```
hugomrdias, line 11: probably not needed with the new setup

hugomrdias (changes requested):

examples/libp2p-in-the-browser/package.json
vasco-santos: Just tried this out, but we need it otherwise the tests do not run.

vasco-santos:
Thanks for the suggestions @hugomrdias . This should now be good. Just waiting on CI

hugomrdias (approved):

Issue #838, opened 2020-12-19
[Feature Request] Expose custom transport data streams


HexaField:
WebRTC and potentially other transports support media streams (such as video and audio). It would be great to use libp2p directly to expose the relevant hooks/interfaces to enable this functionality, as opposed to opening a second connection and reimplementing a lot of what simple-peer / libp2p-webrtc-peer and libp2p-webrtc-star do. 

The requirements I can identity thus far are only to add a way to (dynamically pass a stream)[https://github.com/ipfs-shipyard/simple-peer#dynamic-videovoice] via the `peer.addStream` function and add an event hook for the `peer.on("stream")` event.

vasco-santos:
Thanks for reaching out @HexaField 

When a transport connection is initially created we create a `MultiaddrConn`. This `MultiaddrConn` has the raw connection, which in the `webrtc-star` transport is the SimplePeer instance. The libp2p upgrader will upgrade it into a libp2p [connection](https://github.com/libp2p/js-libp2p-interfaces/tree/master/src/connection) which does not include the raw connection (SimplePeer instance in `webrtc-star`). With this in mind, adding the raw connection into `libp2p-connection` is quite simple.

We need to take a few things into account though. Libp2p is a modular stack with multiple transports and consequently, transport agnostic. Exposing the raw connection will mean that the underlying API for the raw connection will be different according to the transport being used. This might create issues in the application layer as users will need to take into account these differences in the application layer. Other than that, exposing the raw connection might lead into unexpected behaviours in libp2p, such as if a raw connection is directly closed in the raw connection.

Overall, I think that this feature is valuable for several application use cases. Perhaps we can try to find a balance and expose raw connection functions per configuration.

A potential way of doing this would be to add to the `interface-connection`:

```js
{
  transportFunctions: { }
}
```

In the transport layer, each transport would have a `exposeTransportFunctions`:

```js
  exposeTransportFunctions: (rawConn) => { addStream: rawConn.addStream }
```

With a design like this, the user would be able to expose transport based functions to use in the application layer (if they exist). A simpler approach could be to add a transport string identifier in the `interface-connection` so that users could validate that a connection is using a given transport before trying to do something on the raw connection (directly exposed).

[...]

PR #854, opened 2021-01-01
feat: add config option to expose raw connections

HexaField:
Adds an option to the libp2p configuration `config.exposeRawConn` (default: false)

This option is then used by the upgrader to expose the raw connection and transport tag from a `MultiaddrConnection` to the `Connection`

Requires https://github.com/libp2p/js-libp2p-interfaces/pull/79

Solves https://github.com/libp2p/js-libp2p/issues/838


jacobheun:
Closing this PR as we'd like to just expose the underlying connection per https://github.com/libp2p/js-libp2p/issues/838#issuecomment-766799676

Issue #756, opened 2020-09-16
[docs] Add configuration example in docs for Bootstrap


jacobheun:
We currently don't have any docs in the configuration readme on how to setup libp2p-bootstrap. Additionally the readme in that repo doesn't provide a useful example of how to use it with libp2p.

We should:
1. Update https://github.com/libp2p/js-libp2p/blob/master/doc/CONFIGURATION.md to include a configuration code snippet for Bootstrap
2. Update the https://github.com/libp2p/js-libp2p-bootstrap#usage section in `libp2p-bootstrap` to show how it's actually configured in `libp2p`.

a1300:
Relates to #807

PR #859, opened 2021-01-13
docs: Add bootstrap to custom peer discovery

acolytec3:
Addresses point 1 under #756.  

cc: @jacobheun @vasco-santos 

I wasn't clear from the issue description if the snippet desired is just an addition to the existing custom peer discovery or a whole separate section.  If the latter, let me know and I'm happy to re-write.

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -261,13 +261,14 @@ const TCP = require('libp2p-tcp')
 const MPLEX = require('libp2p-mplex')
 const { NOISE } = require('libp2p-noise')
 const MulticastDNS = require('libp2p-mdns')
+const Bootstrap = require('libp2p-bootstrap')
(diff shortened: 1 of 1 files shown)


vasco-santos (changes requested):
This looks good, as we also had the options for bootstrap below. Thanks for the PR

I would like to ask you a few changes on the configuration though:
- Use `[Bootstrap.tag]` in the config part instead of `Bootstrap`. This is the same, but is less "magical" and will help users understand what this key really is
- Add `enabled: true` to the bootstrap config. It is `true` by default but I think it is helpful to have it there.
- Move the comment `// ...other discovery module options` to the line after, that is, outside the bootstrap configuration option.

vasco-santos (approved):
LGTM! Thanks 👍🏼

PR #861, opened 2021-01-18
chore: fix close for ConnectionManager

samlior:
fix close for ConnectionManager

### DIFF
--- a/src/connection-manager/index.js
+++ b/src/connection-manager/index.js
@@ -160,7 +160,7 @@ class ConnectionManager extends EventEmitter {
       }
     }
 
-    await tasks
+    await Promise.all(tasks)
     this.connections.clear()
   }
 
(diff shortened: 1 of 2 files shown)


vasco-santos (changes requested):
Thanks for fixing this. Can you add a test to avoid future regressions?
The CI failure is not related to this, I will see what is happening there

samlior:
I think this is an obvious bug, no need to add a specific test.

vasco-santos:
> I think this is an obvious bug, no need to add a specific test.

It is a bug, and a test will guarantee that if this code is changed it will not be created again. A simple test should be added [here](https://github.com/libp2p/js-libp2p/blob/master/test/connection-manager/index.node.js#L270) where a peer connects and than immediately after stopping, it should have no connections

samlior:
The PR with master has been rebased, and tests have been added for ConnetionManager.

vasco-santos (approved):
LGTM! Thanks @samlior

PR #856, opened 2021-01-05
feat: rendezvous client

vasco-santos:
Adds rendezvous client

Needs:
- [ ] https://github.com/libp2p/js-libp2p-rendezvous/pull/6

PR #862, opened 2021-01-18
chore: rendezvous example

vasco-santos:
This PR adds the rendezvous usage example.

Needs:

- [ ] https://github.com/libp2p/js-libp2p-rendezvous/pull/6
- [ ] #856 

Other notes:

- In this PR, or as a follow up, we should restructure the `discovery-mechanisms` examples folder. Currently example 1 is bootstrap, 2 mdns and 3 pubsub discovery, while rendezvous was created in a new folder. I think we should create a folder per type of discovery and partially move the Readme content within each context


achingbrain:
Closing due to staleness

PR #863, opened 2021-01-18
chore: use emittery

vasco-santos:
EventEmitter is not isomorphic and using this will be problematic with Webpack5. We should consider changing it in favour of emittery or similar. This will be a breaking change as the API is slightly different.


achingbrain:
Instead of emitter/emittery/etc we're now using [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) which is part of the language.

Issue #680, opened 2020-06-19
PeerId string identifier


wemeetagain:
#### Type:
Question

#### Description:
Currently, js-libp2p uses PeerId objects to identify peers, which are compared using `equals` method and printed with `toB58String`. In many cases, the b58 string is used to index peers (eg: in Maps, Objects, Sets, etc) and there, implicitly used for id equality.

Has any thought been given to using b58 string encoded peer-ids as the canonical peer identifier throught the codebase? In this case, the keystore would be more heavily relied on to retrieve public/private keys since they wouldn't be (optionally) attached as in PeerId objects. In many cases, checking the validity of a string identifier would still be required, but strings are convenient to use for indexing and equality checking. Perhaps this may also result in general speedups and reduced memory usage, as string equality checking is generally faster than Buffer equality checking and public keys can all be stored in one place.

I definitely haven't thought thru the intricacies, just more an idea after seeing lots of `id.toB58String()` in my own code, and seeing that `peer.ID` is also a string in go-libp2p. 

Curious more than anything if this is being considered or on the roadmap, and if this has been discussed before.

vasco-santos:
This seems a good idea! Totally in favour of moving on a direction of using a string identifier and rely on the PeerStore.
We plan to move away from the `id.toB58String()` in favour of `id.toString()`. When we work on this change, I think we should definitely consider this suggestion.

wemeetagain:
I noticed, profiling lodestar, that requesting peers from the peer store (calling `libp2p.peerStore.peers`) is not very performant, taking roughly 500ms per call (with roughly 50 or so peers).
It appears to mainly be a result of creating `PeerId` objects.

![get-peers-flame](https://user-images.githubusercontent.com/1348242/101543286-a7882980-3969-11eb-8bee-bf30596fee06.png)

I'd expect that resolving this issue in the direction of using string peer-ids directly will result in a significant performance improvement.

vasco-santos:
Thanks for the data @wemeetagain 
So, once we use peerIdStr everywhere we will be able to return also the string in the `peerStore.peers`.

However, we should look into what is causing this performance degradation. The bottleneck should probably be in https://github.com/libp2p/js-peer-id/blob/master/src/index.js#L27 creating the b58string which we already have, or creating the cid in https://github.com/libp2p/js-peer-id/blob/master/src/index.js#L248 . Both have sync encodings/decodings.

[...]

PR #865, opened 2021-01-20
fix: store provider multiaddrs during find providers

achingbrain:
Changes the behaviour of `libp2p.contentRouting.findProviders` to store
the multiaddrs reported by the routers before yielding results to
the caller, so when they try to dial the provider, the multiaddrs are
already in the peer store's address book.

Also dedupes providers reported by routers but keeps all of the addresses
reported, even for duplicates.

Also, also fixes a performance bug where the previous implementation would
wait for any router to completely finish finding providers before sending
any results to the caller.  It'll now yield results as they come in which
makes it much, much faster.

Follows on from https://github.com/ipfs/js-ipfs-bitswap/pull/286#discussion_r560799814


src/content-routing.js
```diff
@@ -36,32 +43,76 @@ class ContentRouting {
...
+          }
+
+          seen.add(provider.id.toString())
+
+          return true
+        })
+      },
+      function returnAllAddresses (source) {
+        return map(source, (provider) => {
+          const addresses = peerStore.addressBook.get(provider.id) || []
```
vasco-santos, line 97: This is interesting. We might return addresses other than the ones from the router, in case we already knew the peer. Do we have any advantage with it though? For further dial, it will be good as we will attempt to dial other multiaddrs, but here it might introduce strange behaviors where the results might differ from the ones obtained with `go-ipfs`, when using the DHT.

Perhaps, we should just get the addresses from the provider record, which also means we don't need to hit the peerStore. What do you think?

vasco-santos (commented):
This looks great @achingbrain , thanks for looking into improving the async iteration. Left a suggestion.

src/content-routing.js
achingbrain: I agree, I've actually pulled this out in a further iteration.  The reason being if we've never seen the peer before (likely) and we get a second result for the same peer with different addresses (maybe likely), we'll never yield that second result because we've already yielded the first and we filter based on peer id.

Part of me thinks we shouldn't return multiaddrs at all, just the peer id, since the user now uses a peer id to connect to a remote peer instead of multiaddrs.  One for the future perhaps.

vasco-santos: > Part of me thinks we shouldn't return multiaddrs at all, just the peer id, since the user now uses a peer id to connect to a remote peer instead of multiaddrs. One for the future perhaps.

This seems a good call. We will be moving from returning `PeerId` to return its string identifier in the future for performance reasons. We can look into doing that as part of that breaking change in https://github.com/libp2p/js-libp2p/issues/680

src/content-routing/index.js
```diff
@@ -35,34 +41,28 @@ class ContentRouting {
   }
 
   /**
-   * Iterates over all content routers in series to find providers of the given key.
-   * Once a content router succeeds, iteration will stop.
+   * Iterates over all content routers in parallel to find providers of the given key.
```
vasco-santos, line 44: Can you change the above comment in the constructor that makes the DHT first? As now the order is irrelevant

src/peer-routing.js
```diff
@@ -71,7 +78,7 @@ class PeerRouting {
   }
 
   /**
-   * Iterates over all peer routers in series to find the given peer.
+   * Iterates over all peer routers in parallel to find the given peer.
```
vasco-santos, line 81: Same as content-routing with the constructor comment

src/content-routing/utils.js
```diff
@@ -0,0 +1,89 @@
...
+    return peer
+  })
+}
+
+/**
+ * Filter peers by unique peer id
+ *
+ * @param {AsyncIterable<{ id: PeerId, multiaddrs: Multiaddr[] }>} source
+ */
+function uniqueProviders (source) {
```
vasco-santos, line 33: What do you think on renaming this to `uniquePeers`? It would suit better in the `peerRouting.getClosestPeers`

test/peer-routing/peer-routing.node.js
```diff
@@ -277,70 +291,127 @@ describe('peer-routing', () => {
...
+      sinon.stub(node._dht, 'getClosestPeers').callsFake(function * () { })
+
+      sinon.stub(delegate, 'getClosestPeers').callsFake(function * () {
+        yield result
+      })
+
+      await drain(node.peerRouting.getClosestPeers('a cid'))
+
+      expect(spy.calledWith(result.id, result.multiaddrs)).to.be.true()
+    })
```
vasco-santos, line 414: Probably worth adding the `should dedupe results` test to guarantee the util is called

vasco-santos (changes requested):
This looks good! Great work @achingbrain 

Left a few minor comments

src/content-routing/index.js
achingbrain: Done!

src/peer-routing.js
achingbrain: Done!

src/content-routing/utils.js
achingbrain: Done!

test/peer-routing/peer-routing.node.js
achingbrain: Done!

hugomrdias (approved):

vasco-santos (approved):
LGTM - waiting on CI

vasco-santos:
`0.30.2` is in the interwebs 🚢

PR #868, opened 2021-01-24
Add an example of webrtc-direct

vogdb:
Add an example of custom libp2p in IPFS for a file exchange between a Node and a browser IPFS instances.

Added according to a discussion in https://github.com/libp2p/js-libp2p-webrtc-direct/issues/94#issuecomment-760313699

cc @vasco-santos please review


vasco-santos:
Hey @vogdb 
I apologise if I was not concrete enough with my suggestion for the example. If this lives in the libp2p repo, this should not be via IPFS. The way to go would be to have 2 libp2p nodes interacting with `webrtc-direct`. Then, you could follow up that with some references in the IPFS `libp2p-custom-example` to examples such as the `webrtc-direct` one.
The main reason for this is that IPFS uses libp2p, so it should be IPFS who shows up on to configure libp2p and not the other way around. Anyway, this should be easy to update .
As a fast feedback, it would be worth for the readme to have a better indication of the steps like in https://github.com/libp2p/js-libp2p/tree/master/examples/auto-relay

vogdb:
ok then. I'll do it on the weekend.

vogdb:
Might postpone it to the next weekend, didn't have much spare time.

vasco-santos:
No problem, thanks for the update :)

vogdb:
Please review. Also added you as a collaborator to my fork so you can quickly adjust if needed.

examples/webrtc-direct/README.md
```diff
@@ -0,0 +1,30 @@
...
+## 0. Run a nodejs libp2p client
+When in the root folder of this example, type `node node.js` in terminal. You should see an address that listens for
+incoming connections. Below is just an example of such address. In your case the suffix hash will be different.
+```bash
+$ node node.js 
+Listening on:
+/ip4/127.0.0.1/tcp/9090/http/p2p-webrtc-direct/p2p/QmUKQCzEUhhhobcNSrXU5uzxTqbvF1BjMCGNGZzZU14Kgd
+```
+
+## 1. Prepare a browser libp2p client
```
vasco-santos, line 14:
```suggestion
## 1. Prepare a browser libp2p dialer
```

examples/webrtc-direct/README.md
```diff
@@ -0,0 +1,30 @@
...
+Copy-paste the above address into the field `list` in `public/browser.js`:
```
vasco-santos, line 15:
```suggestion
Copy-paste the above address into the field `list` in `public/dialer.js`:
```

examples/webrtc-direct/README.md
```diff
@@ -0,0 +1,30 @@
...
+    peerDiscovery: {
+      [Bootstrap.tag]: {
+        enabled: true,
+        // paste the address into `list`
+        list: ['/ip4/127.0.0.1/tcp/9090/http/p2p-webrtc-direct/p2p/QmUKQCzEUhhhobcNSrXU5uzxTqbvF1BjMCGNGZzZU14Kgd']
+      }
+    }
+```
+
+## 2. Run a browser libp2p client
```
vasco-santos, line 26:
```suggestion
## 2. Run a browser libp2p dialer
```

examples/webrtc-direct/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "ipfs-file-exchange",
```
vasco-santos, line 2:
```suggestion
  "name": "webrtc-direct",
```

vasco-santos (changes requested):
Thanks for the first iteration @vogdb 

I provided some feedback on the README. We should change the namings do dialer/listener to be consistent with the remaining examples and enable users to assimilate concepts faster.

This will also need some tests. For inspiration you can check:
- https://github.com/libp2p/js-libp2p/blob/master/examples/auto-relay/test.js
- (run a browser node) https://github.com/libp2p/js-libp2p/blob/master/examples/libp2p-in-the-browser/test.js

This test job should then be added to the github actions workflow: https://github.com/libp2p/js-libp2p/blob/master/.github/workflows/main.yml#L117

The CI is currently failing, if it persists that might be a dependency that was updated and the bundle increased. Feel free to update the bundle size expectation to the new value https://github.com/libp2p/js-libp2p/blob/master/.aegir.js#L51 if the CI keeps failing.

Other than this, we should be ready to get this merged. Once again, thanks for creating this!

examples/webrtc-direct/test.js
```diff
@@ -0,0 +1,50 @@
...
+    process.stdout.write('listener.js\n')
+
+    const listenerProc = startProcess('listener.js')
+    listenerProc.all.on('data', async (data) => {
+        process.stdout.write(data)
+
+        listenerOutput += uint8ArrayToString(data)
+        if (listenerOutput.includes('Listening on:') && listenerOutput.includes('/p2p/')) {
+            listenerAddr = listenerOutput.trim().split('Listening on:\n')[1].split('\n')[0]
+            //TODO What is the recommended way to paste `listenerAddr` into the `dialer.js`?
```
vogdb, line 31: What would be a recommended way to paste `listenerAddr` into `dialer.js`

examples/webrtc-direct/README.md
```diff
@@ -0,0 +1,33 @@
+### Webrtc-direct example
+
+An example that uses [js-libp2p-webrtc-direct](https://github.com/libp2p/js-libp2p-webrtc-direct) for connecting a
+nodejs libp2p and a browser libp2p clients. To run the example:
```
vasco-santos, line 4:
```suggestion
nodejs libp2p and browser libp2p clients. To run the example:
```

examples/webrtc-direct/test.js
vasco-santos: yes, there is no good way of doing this...

My suggestion is to allow the listener to receive a peerId via a parameter. Then, we can have the dialer browser with the default bootstrap id to be that one.

You can install `peer-id` module globally and do:

```sh
peer-id --type=ed25519 > id.json
```

This `id.json` file would also be commited in this example. Then, you can run the listener process providing a `peerId` file as a parameter like: https://github.com/libp2p/js-libp2p-relay-server/blob/v0.1.3/src/bin.js#L41

Another idea is really make this part of the example. So, use this `peer-id` file by default in the example and people can provide their own to override it. This would have the benefit that everything would just work and users would not need to change the dialer code to try this out. However, we need to make clear in the README that we are using a pregenerated peer-id by default. What do you think on this?

vogdb:
> Another idea is really make this part of the example. So, use this peer-id file by default in the example and people can provide their own to override it.

I like this approach. It also will remove a manual step of copy-pasting. Again, I will implement it only on the weekend.

vasco-santos:
@vogdb thanks for the back and forth to get this in ❤️

vogdb:
No worries, it is useful for me. Someday I hope to work on this project seriously.

examples/webrtc-direct/test.js
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        if (out.includes('✨  Built in ')) {
+            try {
+                const browser = await chromium.launch();
+                const page = await browser.newPage();
+                await page.goto(dialerUrl);
+                await page.waitForFunction(selector => document.querySelector(selector).innerText === 'libp2p started!', '#status')
+                await page.waitForFunction(
+                  selector => {
```
vogdb, line 63: Don't get inside this function. What could be the possible problem? Locally the above line is executed but this one is not.

vasco-santos: cc @hugomrdias 

playwright using chromium seems to not execute the function in: `page.waitForFunction`. Ideas?

vasco-santos: I added a `return false` inside the function and it never progressed. So it seems to be working, perhaps not does logs within the callback scope if that was what you were trying. 
I commited a process kill to stop the dialer which seemed to be the missing piece to fully run the test and stop

vasco-santos: wait, you man it fails in CI, but not locally?
That seems the case now

vogdb: > perhaps not does logs within the callback scope if that was what you were trying.

Yep. I was trying to do that. Seems that CI passes now after your fix. Please review the PR.

examples/webrtc-direct/README.md
```diff
@@ -0,0 +1,33 @@
...
+incoming connections. Below is just an example of such address. In your case the suffix hash (`peerId`) will be different.
+
+```bash
+$ node listener.js 
+Listening on:
+/ip4/127.0.0.1/tcp/9090/http/p2p-webrtc-direct/p2p/QmUKQCzEUhhhobcNSrXU5uzxTqbvF1BjMCGNGZzZU14Kgd
+```
+
+## 1. Prepare a browser libp2p dialer
+Copy-paste the above address into the field `list` in `public/dialer.js`:
```
vasco-santos, line 18:
```suggestion
Confirm that the above address is the same as the field `list` in `public/dialer.js`:
```

examples/webrtc-direct/README.md
```diff
@@ -0,0 +1,33 @@
...
+        enabled: true,
+        // paste the address into `list`
+        list: ['/ip4/127.0.0.1/tcp/9090/http/p2p-webrtc-direct/p2p/QmUKQCzEUhhhobcNSrXU5uzxTqbvF1BjMCGNGZzZU14Kgd']
+      }
+    }
+```
+
+## 2. Run a browser libp2p dialer
+When in the root folder of this example, type `npm run dev` in terminal. You should see an address where you can browse
+the running client. Open this address in your browser. In my case the address was `http://127.0.0.1:12345`. In console
```
vasco-santos, line 31:
```suggestion
the running client. Open this address in your browser. In console
```

vasco-santos (approved):
LGTM!
Thanks for the PR @vogdb

vogdb:
Thank you for guidance. Happy to contribute.

Issue #873, opened 2021-01-28
In an electron app just importing libp2p in main.ts causes "self is not defined"


raphael10-collab:
In in electron app just importing libp2p in main.ts : 

    const Libp2p = require('libp2p');

causes this error:

    A JavaScript error occurred in the main process
    Uncaught Exception:
    ReferenceError: self is not defined
        at Object../node_modules/multihashing-async/src/sha.browser.js (/home/marco/webMatters/electronMatters/OnlyLibp2p
    /.webpack/main/index.js:34203:16)
        at __webpack_require__ (/home/marco/webMatters/electronMatters/OnlyLibp2p/.webpack/main/index.js:21:30)
        at Object../node_modules/libp2p-crypto/src/keys/rsa-class.js (/home/marco/webMatters/electronMatters/OnlyLibp2p
    /.webpack/main/index.js:19523:13)
        at __webpack_require__ (/home/marco/webMatters/electronMatters/OnlyLibp2p/.webpack/main/index.js:21:30)
        at Object../node_modules/libp2p-crypto/src/keys/index.js (/home/marco/webMatters/electronMatters/OnlyLibp2p/.webpack
    /main/index.js:19074:8)
        at __webpack_require__ (/home/marco/webMatters/electronMatters/OnlyLibp2p/.webpack/main/index.js:21:30)
        at Object../node_modules/peer-id/src/index.js (/home/marco/webMatters/electronMatters/OnlyLibp2p/.webpack
    /main/index.js:58259:20)
        at __webpack_require__ (/home/marco/webMatters/electronMatters/OnlyLibp2p/.webpack/main/index.js:21:30)
        at Object../node_modules/libp2p/src/index.js (/home/marco/webMatters/electronMatters/OnlyLibp2p/.webpack
    /main/index.js:24403:16)
        at __webpack_require__ (/home/marco/webMatters/electronMatters/OnlyLibp2p/.webpack/main/index.js:21:30)

tsconfig.json : 

    {
      "compilerOptions": {
        "jsx": "react",
        "allowJs": true,
        "target": "ES6",
        "module": "ESNext",
        "skipLibCheck": true,
        "esModuleInterop": true,
        "noImplicitAny": false,
        "sourceMap": true,
        "baseUrl": ".",
        "outDir": "dist",
        "moduleResolution": "node",
        "resolveJsonModule": true,
        "paths": {
          "*": ["node_modules/*"],
          "@app/*": ["./src/app/*"],
[...]

PR #875, opened 2021-01-28
fix: create has optional peer id type

vasco-santos:
When using the libp2p create static function, the `peerId` parameter is not required.

I tried to simply make it optional in `CreateOptions`, but sadly it does not override the required in `Libp2pOptions`. So, I needed to create a new typedef.

Context https://github.com/libp2p/js-libp2p/issues/873

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -82,10 +82,12 @@ const { updateSelfPeerRecord } = require('./record/utils')
  * @property {import('./transport-manager').TransportManagerOptions} [transportManager]
  * @property {PeerStoreOptions & import('./peer-store/persistent').PersistentPeerStoreOptions} [peerStore]
  * @property {Libp2pConfig} [config]
+ *
+ * @typedef {Object} constructorOptions
  * @property {PeerId} peerId
  *
  * @typedef {Object} CreateOptions
- * @property {PeerId} peerId
+ * @property {PeerId} [peerId]
  *
  * @extends {EventEmitter}
  * @fires Libp2p#error Emitted when an error occurs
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -101,20 +103,22 @@ class Libp2p extends EventEmitter {
    */
   static async create (options) {
     if (options.peerId) {
+      // @ts-ignore 'Libp2pOptions & CreateOptions' is not assignable to 'Libp2pOptions & constructorOptions'
```
hugomrdias, line 106: maybe these `Libp2pOptions`, `CreateOptions` and `constructorOptions` types need to be refactored to be more composable ?

hugomrdias (approved):

src/index.js
vasco-santos: Do you have any suggestion? I think that the issue is that one says that the peerId is required while the other says it is optional. This causes:

```
Argument of type 'Libp2pOptions & CreateOptions' is not assignable to parameter of type 'Libp2pOptions & constructorOptions'.
  Type 'Libp2pOptions & CreateOptions' is not assignable to type 'constructorOptions'.
    Types of property 'peerId' are incompatible.
      Type 'PeerId | undefined' is not assignable to type 'PeerId'.
        Type 'undefined' is not assignable to type 'PeerId'
```

hugomrdias: i have been dealing with this type of stuff lately, specifically with constructor options that have options props but internally we have defaults for all or some.

So what im doing now is:
```js
class Libp2p {
  
  constructor(options: Libp2pOptions) {
    const opts: Required<Libp2pOptions> = {...options, ...defaults} 
 }
}
```
This avoids having to truthy check everywhere, but also makes all options required so kinda forces you to have defaults for everything which might sounds troublesome but i think its good as a general rule.

vasco-santos: Ok, we might be able to move into that in the future once we revamp libp2p config, but not now. This would mean we need to have libp2p modules as dependencies here.

PR #881, opened 2021-02-01
feat: add confidence in observed addresses

achingbrain:
Require at least four peers to report the same observed address before
we start returning it as part of `node.multiaddrs`.

In order to not let the list of observed addresses grow forever we
enforce a 10 minute timeout, that is for any observed address, at least
four different peers have to report it to us within ten minutes otherwise
we discard the address.

The number of peers and timeout are both configurable.

### DIFF
--- a/src/address-manager/index.js
+++ b/src/address-manager/index.js
@@ -41,10 +41,7 @@ class AddressManager extends EventEmitter {
     this.announce = new Set(announce.map(ma => ma.toString()))
     this.observed = new Map()
     this.config = {
-      observedAddresses: {
-        minConfidence: observedAddresses.minConfidence || 4,
-        maxLifetimeBeforeEviction: observedAddresses.maxLifetimeBeforeEviction || (60 * 10) * 1000
-      }
+      observedAddresses
     }
   }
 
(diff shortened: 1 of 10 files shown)


src/address-manager/index.js
```diff
@@ -31,14 +31,21 @@ class AddressManager extends EventEmitter {
...
     super()
 
     this.peerId = peerId
     this.listen = new Set(listen.map(ma => ma.toString()))
     this.announce = new Set(announce.map(ma => ma.toString()))
-    this.observed = new Set()
+    this.observed = new Map()
+    this.config = {
+      observedAddresses: {
+        minConfidence: observedAddresses.minConfidence || 4,
```
vasco-santos, line 45: The defaults should already exist from the constructor default values

vasco-santos (changes requested):

src/address-manager/index.js
achingbrain: ts kept telling me the values might be undefined, it's not doing it now, will remove.

vasco-santos:
We will merge #882 and think more about how to add confidence to the observed addresses

lidel:
Closing for now, can be picked up as part of  https://github.com/libp2p/js-libp2p/issues/703 when that gets prioritized again.

PR #887, opened 2021-02-12
fix: address book should not emit peer event if no addresses are known

vasco-santos:
The AddressBook.add has a bug where if a libp2p module does `addressBook.add(peerId, [])` the peer event will be emitted. This leads into two consequences:
- (with autoDial enabled) a dial will be attempted to the discovered peer, even though we can't really dial it
  - dial on discovery is wrapped with a `try ... catch` and it will just fail and life goes on  
- when we really discover addresses for the peer and do a new `addressBook.add`, the event will not be emitted and the peer will not be dialled (if autoDial enabled)

Other than this bug and related to #886 , the `libp2p.dial` / `libp2p.dialProtocol` should be wrapped by a `try ... catch` as a dial fail should not make libp2p protocols fail to properly work. This PR also wraps the `autoRelay` dial to avoid a unhandled promise rejection. Moreover, the dials to found providers would be stopped if one dial fails, now it will continue iterating on the providers if dials fail.

### DIFF
--- a/src/circuit/auto-relay.js
+++ b/src/circuit/auto-relay.js
@@ -262,6 +262,9 @@ class AutoRelay {
     }
   }
 
+  /**
+   * @param {PeerId} peerId
+   */
   async _tryToListenOnRelay (peerId) {
     try {
       const connection = await this._libp2p.dial(peerId)
--- a/src/circuit/auto-relay.js
+++ b/src/circuit/auto-relay.js
@@ -31,6 +31,7 @@ const {
  *
  * @typedef {Object} AutoRelayOptions
  * @property {number} [maxListeners = 1] - maximum number of relays to listen.
+ * @property {(error: Error, msg?: string) => {}} [onError]
  */
 
 class AutoRelay {
@@ -40,7 +41,7 @@ class AutoRelay {
    * @class
    * @param {AutoRelayProperties & AutoRelayOptions} props
    */
-  constructor ({ libp2p, maxListeners = 1 }) {
+  constructor ({ libp2p, maxListeners = 1, onError }) {
     this._libp2p = libp2p
     this._peerId = libp2p.peerId
     this._peerStore = libp2p.peerStore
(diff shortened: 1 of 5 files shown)


vasco-santos:
codecov is complaining but the existing tests already cover the lines in question 🤷🏼‍♂️

src/circuit/auto-relay.js
```diff
@@ -231,8 +231,7 @@ class AutoRelay {
 
     // Try to listen on known peers that are not connected
     for (const peerId of knownHopsToDial) {
-      const connection = await this._libp2p.dial(peerId)
```
Gozala, line 234: Can you please add a test to exercise this code path ? Or if it's a problem with codecov, maybe add a codecov specific comment.

Gozala (changes requested):
Changes look good! Codecov seems to suggest new code isn't covered by tests though, I think that should be resolved before landing this.

src/circuit/auto-relay.js
vasco-santos: Yes, this is codecov having a bad time.. This [test](https://github.com/libp2p/js-libp2p/blob/v0.31.0-rc.3/test/relay/auto-relay.node.js#L333) will exercise this code path

vasco-santos: I will add a sinon check in that test

src/circuit/auto-relay.js
```diff
@@ -263,6 +261,18 @@ class AutoRelay {
...
+
+  /**
+   * @param {PeerId} peerId
+   */
+  async _tryToListenOnRelay (peerId) {
+    try {
+      const connection = await this._libp2p.dial(peerId)
+      await this._addListenRelay(connection, peerId.toB58String())
+    } catch (err) {
+      log.error(`could not connect and listen on known hop relay ${peerId.toB58String()}`)
```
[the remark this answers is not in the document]
vasco-santos: added an onError 👍🏼  It would be super cool to have a general error handling approach, as this type of handling is common. In this case, the auto relay lifecycle does not care about the errors as it can try other peers.

Gozala (approved):

PR #891, opened 2021-02-22
fix: update to aegir v31

hugomrdias:
This is just a draft to get some metrics from aegir v31 in this repo, theres some types checking failing, i can only bundle with yarn and not npm, browser cov hangs.  

ref:
https://github.com/ipfs/aegir/pull/737

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const path = require('path')
 const Libp2p = require('./src')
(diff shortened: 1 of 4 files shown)


vasco-santos:
As you mentioned before, one of the things that needs to be changed here is the `types.ts` file.

hugomrdias:
Yes the migration guide for v31 will cover all of that, i just mentioned in the OP issues i found specific to this repo.

PR #893, opened 2021-02-25
fix: transport manager fault tolerance should include tolerance to transport listen fail

vasco-santos:
When we added a fault tolerance option to libp2p's transport manager, this only included a scenario where the provided multiaddrs were not compatible to the configured transport resulting in no listen multiaddrs used to create listeners in the transports.

This fault tolerance should also include tolerance for failure on trying to listen on multiaddrs that are compatible with the configured transport but the listen fails. This is specially important for remote listeners like relays and webrtc signal servers.

Context: https://github.com/ipfs-shipyard/ipfs-share-files/pull/121

### DIFF
--- a/src/transport-manager.js
+++ b/src/transport-manager.js
@@ -196,7 +196,7 @@ class TransportManager {
       // listening on remote addresses as they may be offline. We could then potentially
       // just wait for any (`p-any`) listener to succeed on each transport before returning
       const isListening = results.find(r => r.isFulfilled === true)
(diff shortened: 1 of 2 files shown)


lidel (approved):
Tested this with https://github.com/ipfs-shipyard/ipfs-share-files/pull/121 and works as expected:

> 


@vasco-santos Is it possible to ship this as a patch release?

jacobheun (approved):
LGTM

Issue #888, opened 2021-02-14
Warning ipfs > ipfs-core > libp2p > promisify-es6@1.0.3: Unmaintained


raphael10-collab:
Installing ipfs : 

Warning ipfs > ipfs-core > libp2p > promisify-es6@1.0.3: Unmaintained

- O.S. : Ubuntu 18.04 Desktop
- node: v14.5.0

vasco-santos:
Thanks for reporting. Would you like to PR libp2p and change this?

raphael10-collab:
@vasco-santos 

Yesterday, which was Sunday, I wrote SIX of those upgrading requests. And received for most of them a "send a PR" as response.
If the library's maintainer thinks it's a low priority to upgrade/maintain her/his library's packages, why its users should send a PR?
Shouldn't be the duty and responsability of the library's maintainers to keep library' packages upgraded/maintained?
It's like my children's room: it' true, we live in the same house, but it's their duty and responsability to keep their room tidy.
If everyone does his/her duty, the house keeps going safe and tidy.

Anyway, if you still think I, as final user of YOUR library, has to make the PR for upgrading the packages your library relies on, whenever I will have spare time, I will do it

TJKoury:
Just submitted a pull. Tests work on my machine, pretty straightforward.

BigLep:
Closing since PR was merged

PR #896, opened 2021-03-09
swapped dep to maintained package

TJKoury:
[Solves this issue.](https://github.com/libp2p/js-libp2p/issues/888)

Went with a different library ([es6-promisify](https://www.npmjs.com/package/es6-promisify)) since the package maintainer's solution (use Node's native [util.promisify](https://nodejs.org/api/util.html#util_util_promisify_original)) would require an additional shim for browser use.


vasco-santos:
@TJKoury it seems that a release of `moving-average` module broke the build. I will check it out to unblock this

TJKoury:
@vasco-santos Is there a task for running these typescript checks locally?

TJKoury:
That’s strange, I’ve run the build task a few times but no errors.

TJKoury:
Copy.  Let me get to where you are and I can help tackle the other issues.
I’ll be good to go after a few iterations here.

vasco-santos (approved):
LGTM

PR #903, opened 2021-03-23
chore: delegates config md properties for client

vasco-santos:
missing some commas on the properties to create the http client for the delegates


jacobheun (approved):
lgtm!

PR #904, opened 2021-03-26
chore: config typescript

vasco-santos:
This PR aims to:

- update dependencies
- switch protons to protobufjs
  - a bunch of files were automatically generated by this, which is a significant part of the PR changes
    - "src/circuit/protocol/index.js" and "...d.ts"
    - "src/insecure/proto.js" and "...d.ts"
    - "src/peer-store/persistent/pb/address-book.js" and "...d.ts"
    - "src/peer-store/persistent/pb/proto-book.js" and "...d.ts"
    - "src/record/peer-record/peer-record.js" and "...d.ts"
    - "src/record/envelope/envelope.js" and "...d.ts"
- update aegir to [latest version](https://github.com/ipfs/aegir/blob/master/md/migration-to-v31.md)
  - fix types per no explicitly `any` allowed
  - added `esbuild`
- guarantee all the possible libp2p configurations documented on [GettingStarted.md](https://github.com/libp2p/js-libp2p/blob/master/doc/GETTING_STARTED.md) and [Configuration.md](https://github.com/libp2p/js-libp2p/blob/master/doc/CONFIGURATION.md) completely work with typescript projects
  - added test ts project
- some bugs fixed, comments inline in review
- restructure github actions into a new workflow for examples

There are internal breaking changes in the Dialer and Keychain property names to match what comes from the external config

Needs:
- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/87
- [x] https://github.com/libp2p/js-libp2p-utils/pull/16
- [x] https://github.com/libp2p/js-libp2p-crypto/pull/187
- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/89
- [x] https://github.com/libp2p/js-libp2p-utils/pull/17
- [x] https://github.com/libp2p/js-libp2p-floodsub/pull/114

Unblocks:
- https://github.com/ChainSafe/js-libp2p-gossipsub/pull/153
- https://github.com/libp2p/js-libp2p-kad-dht/pull/217
- IPFS

Closes: #891 
Solves some of the items from #830 (DHT, Config and Keychain)

BREAKING CHANGES:

top level types were updated, multiaddr@9.0.0 is used, dialer and keychain internal property names changed and connectionManager minPeers is not supported anymore


.github/workflows/main.yml
```diff
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - run: yarn
-    - run: yarn lint
+    - run: npm install
```
vasco-santos, line 15: Because yarn does not run prepare in the dependencies, libp2p-interfaces branch does not have types

src/dialer/index.js
```diff
@@ -63,18 +63,18 @@ class Dialer {
     transportManager,
     peerStore,
     addressSorter = publicAddressesFirst,
-    concurrency = MAX_PARALLEL_DIALS,
-    timeout = DIAL_TIMEOUT,
-    perPeerLimit = MAX_PER_PEER_DIALS,
+    maxParallelDials = MAX_PARALLEL_DIALS,
```
vasco-santos, line 66: This is an internal breaking change. No one should be creating a dialer from here though. If we don't want to change this, we can create a separate typedef for core options and keep things as they were. However, keeping different names for the same things does not seem the best call

src/index.js
```diff
@@ -43,41 +43,75 @@ const { updateSelfPeerRecord } = require('./record/utils')
...
  */
 
 /**
+ * @typedef {Object} RandomWalkOptions
+ * @property {boolean} [enabled = false]
+ * @property {number} [queriesPerPeriod = 1]
+ * @property {number} [interval = 300e3]
+ * @property {number} [timeout = 10e3]
+ *
+ * @typedef {Object} DhtOptions
```
vasco-santos, line 62: This should probably go away when we work on the configuration improvements project. We currently also specify the DHT options in https://github.com/libp2p/js-libp2p/blob/master/src/config.js#L52 , but in theory we should accept other DHT implementations and this might mean different configurations.

src/keychain/index.js
```diff
@@ -118,8 +130,8 @@ class Keychain {
     this.opts = mergeOptions(defaultOptions, options)
 
     // Enforce NIST SP 800-132
-    if (this.opts.passPhrase && this.opts.passPhrase.length < 20) {
-      throw new Error('passPhrase must be least 20 characters')
+    if (this.opts.pass && this.opts.pass.length < 20) {
```
vasco-santos, line 133: Same situation as in the Dialer

src/connection-manager/index.js
```diff
@@ -362,7 +365,7 @@ class ConnectionManager extends EventEmitter {
    */
   _maybeDisconnectOne () {
     if (this._options.minConnections < this.connections.size) {
-      const peerValues = Array.from(this._peerValues).sort(byPeerValue)
+      const peerValues = Array.from(new Map([...this._peerValues.entries()].sort((a, b) => a[1] - b[1])))
```
vasco-santos, line 368: This was a bug! `this.peerValues` is a map<string, number>

src/registrar.js
```diff
@@ -103,12 +115,11 @@ class Registrar {
...
    *
    * @param {Connection} connection
-   * @param {Error} [error]
    * @returns {void}
    */
-  _onDisconnect (connection, error) {
+  _onDisconnect (connection) {
     for (const [, topology] of this.topologies) {
-      topology.disconnect(connection.remotePeer, error)
+      topology.disconnect(connection.remotePeer)
```
vasco-santos, line 122: topology.disconnect only has one parameter

examples/libp2p-in-the-browser/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "libp2p-in-browser",
   "version": "1.0.0",
   "description": "A libp2p node running in the browser",
-  "main": "index.js",
+  "main": "dist/index.html",
```
rvagg, line 5: is .html right?

examples/webrtc-direct/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.1",
   "private": true,
   "description": "",
+  "main": "dist/index.html",
```
rvagg, line 6: I'm guessing it's a parcel thing ..?

package.json
```diff
@@ -127,22 +137,24 @@
     "libp2p-delegated-peer-routing": "^0.8.0",
```
rvagg, line 137: lots of `^` in here for <1 dependencies, seems over-optimistic about how this is going to get consumed into the future

src/dialer/index.js
rvagg: _"No one should be"_ famous last words, how much confidence do you have in this?

src/insecure/plaintext.js
```diff
@@ -4,18 +4,23 @@ const debug = require('debug')
...
+// @ts-ignore protons do not export types
 const { Exchange, KeyType } = require('./proto')
 const protocol = '/plaintext/2.0.0'
 
 /**
  * @typedef {import('libp2p-interfaces/src/connection').Connection} Connection
  */
 
+/**
+ * @param {{ id: Uint8Array; pubkey: { Type: any, Data: Uint8Array; }; }} exchange
```
rvagg, line 22: potential candidate for a named type?

src/metrics/stats.js
```diff
@@ -82,7 +82,7 @@ class Stats extends EventEmitter {
   /**
    * Returns a clone of the internal movingAverages
    *
-   * @returns {Object}
+   * @returns {typeof Object.assign}
```
rvagg, line 85: weird, can you explain this one?

rvagg (commented):
what's the versioning strategy going to be with this one? seems like a lot of potential to break for users - aside from the Dialer options the types might introduce breakage?

examples/libp2p-in-the-browser/package.json
vasco-santos: parcel thing per their new updates

examples/webrtc-direct/package.json
vasco-santos: parcel thing per their new updates

src/insecure/plaintext.js
vasco-santos: fixed by relying on protobufjs instead of protons, which offers us all the types for protobuf definitions

package.json
vasco-santos: In the future, I would like that we get confidence in the APIs to ship v1 of all the things.

vasco-santos:
> what's the versioning strategy going to be with this one? seems like a lot of potential to break for users - aside from the Dialer options the types might introduce breakage?

My initial idea was to ship a patch, but with all the types changes, I feel it will be safer to release a breaking change. This will have consequences on the release chain of upcoming releases (pinned issues), but they will probably be affected either way.

package.json
rvagg: My personal preference is to get to 1.0.0 ASAP (I've even set my default version to 1.0.0) because of the semver mess below 1 is just madness for signalling and ranges. The rules are basically "YOLO" which is useless for downstream users.
The problem with <0 is that your only signal for "breaking" changes is the minor version, but we've got `^` in here so there's no way to block breaking changes slipping in when we update one of these components. For <1, it's usually best to use `~` to account for the semver-minor bumps for breaking changes. Anyone using the current version of the library, even if they pin to the specific version, is going to get a newer, broken transitive dependency (like libp2p-delegated-peer-routing) when they do their next `npm install` because we've shipped with `^`.
Alternatively, just don't treat 1.0.0 as anything special and ship it so we can use semver properly from there on.

src/dialer/index.js
vasco-santos: Does not make sense to require the `Dialer` and use it as a specific piece. However, I agree and it will be safe to just ship a breaking change at this point

src/index.js
```diff
@@ -152,7 +192,9 @@ class Libp2p extends EventEmitter {
     this._discovery = new Map() // Discovery service instances/references
 
     // Create the Connection Manager
-    if (this._options.connectionManager.minPeers) { // Remove in 0.29
+    // @ts-ignore deprecated, needs to be removed on breaking change
```
vasco-santos, line 195: Since we are doing a breaking change, I am going to remove this!

package.json
vasco-santos: Yes,I totally agree on this. The team has had a lot of concerns about shipping 1.0.0, but we need to have a plan to get to this point. I think that as part of sudo work we should create a plan on where should we be to land 1.0.0 on everything

src/metrics/stats.js
vasco-santos: No need for this, it can be inferred automatically. Good catch

src/dialer/index.js
achingbrain: If you're adding types it's good to ship as a major anyway as there will almost certainly be some disruption.

package.json
achingbrain: Please ship [v1 of things](https://github.com/ipfs/js-ipfs/issues/2657)!

> Alternatively, just don't treat 1.0.0 as anything special and ship it so we can use semver properly from there on.

Yes!

.github/workflows/main.yml
```diff
@@ -34,111 +34,34 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node }}
-      - run: yarn
+      - run: npm install
       - run: npx nyc --reporter=lcov aegir test -t node -- --bail
```
achingbrain, line 38: nyc isn't necessary any more

```suggestion
      - run: npx aegir test -t node --cov --bail
```

.github/workflows/main.yml
```diff
@@ -34,111 +34,34 @@ jobs:
...
+      - run: npm install
       - run: npx aegir test -t browser -t webworker --bail
   test-firefox:
     needs: check
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - run: yarn
+      - run: npm install
       - run: npx aegir test -t browser -t webworker --bail -- --browsers FirefoxHeadless
```
achingbrain, line 53:
```suggestion
      - run: npx aegir test -t browser -t webworker --bail -- --browser firefox
```

.github/workflows/main.yml
```diff
@@ -12,11 +12,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - run: yarn
-    - run: yarn lint
-    - uses: gozala/typescript-error-reporter-action@v1.0.8
```
achingbrain, line 17: Did this get pulled out intentionally? It is quite useful.

test/core/ping.node.js
```diff
@@ -56,7 +58,7 @@ describe('ping', () => {
               if (firstInvocation) {
                 firstInvocation = false
 
-                for await (const data of stream) {
+                for await (const data of stream) { // eslint-disable-line
```
achingbrain, line 61: Please add the name of the rule that's being disabled here

package.json
vasco-santos: https://github.com/libp2p/js-libp2p/issues/680

we will need to do this refactor as part of the maintenance work and will affect most of our repos. So, this might be a good opportunity to ship v1 of libp2p modules

src/types.d.ts
```diff
@@ -0,0 +1,18 @@
+export interface EventEmitterFactory {
```
achingbrain, line 1: Is defining types for `EventEmitter` still necessary?  I just checked the branch out, removed this file and all the imports referencing it and could build types without errors.

achingbrain (changes requested):
LGTM, just some small nits.

.github/workflows/main.yml
vasco-santos: I changed based on: https://github.com/ipfs/aegir/blob/master/md/github-actions.md

Thought the recommendation was `ts -p check`

src/types.d.ts
vasco-santos: Good catch! Removing it

.github/workflows/main.yml
achingbrain: The error reporter action shows you which lines have errors in a visual form as part of the PR, the aegir command makes you dig through the console output.  Both do the same thing and use the same rules but the action is a little more user friendly IMHO.

achingbrain (approved):

PR #905, opened 2021-03-26
fix: the API of es6-promisify is not the same as promisify-es6

achingbrain:
Fixes a regression introduced by #896

`promisify-es6` was swapped for `es6-promisify` but the APIs of the two modules are no the same so the NAT hole punching code broke and stopped libp2p nodes UDP ports open which prevented processes from exiting.

Also changes port numbers to 0 in the tests as other processes can be listening on those ports which causes spurious test failures.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,6 +109,7 @@
   },
   "devDependencies": {
     "@nodeutils/defaults-deep": "^1.1.0",
+    "@types/es6-promisify": "^6.0.0",
     "abortable-iterator": "^3.0.0",
     "aegir": "^29.2.0",
     "chai-bytes": "^0.1.2",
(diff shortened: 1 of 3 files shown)


vasco-santos (commented):
Can we add a test for NAT hole punching code broke and stopped libp2p nodes UDP ports open which prevented processes from exiting, so that we avoid future refressions

achingbrain:
The UDP ports are opened by the `@motrix/nat-api` library which is functioning correctly so I'm not sure that needs testing - the ports were left open because `.destroy()` was not being called on the client instance on shut down as the client creation was throwing due to the wrong API being used.

I've added a test that would have caught the error.

vasco-santos (approved):
thanks! LGTM

PR #914, opened 2021-04-16
fix: dial protocol should throw if no protocol is provided

vasco-santos:
This makes ts projects using dialProtocol to have to validate wether a stream pair object or a connection is returned. Moreover, returning different types conditionally is not really a good practise.

We should consider in the future to just deprecate `dialProtocol` as users can now easily open streams from a given connection

BREAKING CHANGE: dialProtocol does not return connection when no protocols are provided


achingbrain (approved):
LGTM

src/index.js
```diff
@@ -462,26 +462,24 @@ class Libp2p extends EventEmitter {
...
-   * and the `Connection` will be returned
+   * and the `MuxedStream` will be returned together with the successful negotiated protocol.
    *
    * @async
    * @param {PeerId|Multiaddr|string} peer - The peer to dial
    * @param {string[]|string} protocols
    * @param {object} [options]
    * @param {AbortSignal} [options.signal]
-   * @returns {Promise<Connection|{ stream: MuxedStream; protocol: string; }>}
+   * @returns {Promise<{ stream: MuxedStream; protocol: string; }>}
```
achingbrain, line 474: Could remove the @returns annotation altogether?

vasco-santos, line 474:
```suggestion
```

PR #915, opened 2021-04-16
fix: give stats initial values

achingbrain:
Otherwise the compiler cannot derive the type and thinks `stats.snapshot` returns `{}`

### DIFF
--- a/src/metrics/stats.js
+++ b/src/metrics/stats.js
@@ -19,7 +19,10 @@ class Stats extends EventEmitter {
 
     this._options = options
     this._queue = []
-    this._stats = {}
+    this._stats = {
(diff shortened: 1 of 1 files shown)


vasco-santos (approved):

achingbrain:
The resulting types:

```
    /**
     * Returns a clone of the current stats.
     */
    get snapshot(): {
        dataReceived: Big;
        dataSent: Big;
    };
    /**
     * Returns a clone of the internal movingAverages
     */
    get movingAverages(): {
        dataReceived: MovingAverage[];
        dataSent: MovingAverage[];
    };
```

vasco-santos (approved):

PR #916, opened 2021-04-19
fix: remove inline arg types from function definitions

achingbrain:
Fixes problem with create-react-app's default webpack config that
transpiles the functions to non-async methods.

Depends on:

- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/91

### DIFF
--- a/src/upgrader.js
+++ b/src/upgrader.js
@@ -222,6 +222,7 @@ class Upgrader {
   }) {
     /** @type {import("libp2p-interfaces/src/stream-muxer/types").Muxer} */
     let muxer
+    /** @type {import("libp2p-interfaces/src/connection/connection").CreatedMuxedStream | undefined} */
     let newStream
     /** @type {Connection} */
     let connection // eslint-disable-line prefer-const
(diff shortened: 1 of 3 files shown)


src/upgrader.js
```diff
@@ -309,8 +310,8 @@ class Upgrader {
       },
       newStream: newStream || errConnectionNotMultiplexed,
       getStreams: () => muxer ? muxer.streams : errConnectionNotMultiplexed(),
-      close: async (/** @type {Error | undefined} */ err) => {
-        await maConn.close(err)
+      close: async () => {
+        await maConn.close()
```
achingbrain, line 314: Looking at the [use of this callback](https://github.com/libp2p/js-libp2p-interfaces/blob/master/src/connection/connection.js#L224) no error is passed in, which is reflected in the [callback typedef](https://github.com/libp2p/js-libp2p-interfaces/blob/master/src/connection/connection.js#L33) so this function probably shouldn't accept an error param?

vasco-santos (approved):

Issue #576, opened 2020-03-02
Improve libp2p config


vasco-santos:
- **Version**: N/A
- **Platform**: N/A
- **Subsystem**: N/A

#### Description:

The developer experience of the current libp2p config is one of the pain points for libp2p users, particularly new users. Moreover, there are a few inconsistencies in the config structure and the distance between the modules setup and the configuration for those modules causes some confusion among libp2p users.

Here follows a few ideas already discussed around this topic:

#### User friendly libp2p configs

* Pre baked configs (libp2p-browser, libp2p-node, etc)
* Potentially test against js-libp2p suite

#### Known issues

* module configuration separated from the modules
* DHT not part of content routing / peer routing
* transports with discovery built-in cannot be added to the discovery modules
* transports with encryption & multiplexing built-in
* sane defaults

vasco-santos:
`go-libp2p` uses default transports, multiplexers, encryption, listenAddr, among other things:

https://github.com/libp2p/go-libp2p/blob/v0.5.2/libp2p.go#L33-L55

We might consider a similar approach

vasco-santos:
We should also consider #202  (listen, announce and noAnnounce addresses here)

jacobheun:
Here are my thoughts on moving js-libp2p configuration to a functional configuration model. I think this would allow us to make libp2p more extensible and improve the UX around configuration.

## The Problem

Currently js-libp2p is diffult to configure, esecially for new users. Due to its modularity there are a variety of options to configure and configuring things improperly can result in a node that doesn't do what is needed.

## Requirements

* Users should be able to select a default configuration for their environment.
* Users should be alerted if they are missing any modules, **and** how to resolve that.
* Users should be able to easily extend basic configurations.
* Users should be able to fully customize libp2p

## Considerations

This propsal will require internal updates to libp2p to ensure its API supports making it easily extendable. There are likely some internal order of operations issues that will need to be resolved as part of this work, as we'll need to ensure all internals like the upgrader and registrar are created prior to calling configuration functions.

## The Proposal: functional configuration

This proposal aims to migrate `Libp2p.create()` away from an object based configuration, to a functional configuration. There are many advantages to this approach:

* Modules will be able to more easily extend libp2p (dht adding itself to contentRouting, peerRouting and peerDiscovery)
  * Modules can export their libp2p configuration functions so users can more easily extend their nodes
* js-libp2p can export a variety of configuration options, with runtime module includes
[...]

PR #922, opened 2021-04-21
fix: demand pubsub class instead of pubsub instance

achingbrain:
Changes the `Libp2pModules.pubsub` property to be a class that maybe
extends `PubsubBaseProtocol` instead of an instance of that class.

The `.dht` property could use the same treatment but that can be
done in a follow up as the current `object` type doesn't cause an error.


vasco-santos (approved):
LGTM

vasco-santos:
The DHT case needs to be better thought as we want to not have the DHT as dependency/devDependency in libp2p. We will defer this for #576

PR #924, opened 2021-04-21
fix: keychain optional pw and use interfaces for validators and selectors instead

vasco-santos:
Needs:

- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/94


achingbrain (approved):
LGTM, pending `libp2p-interface` release and removal of gh url.

PR #938, opened 2021-05-05
chore: add GitHub issue templates

vasco-santos:
This PR moves the Release Template to an Issue Template.

Other changes:
- Release template does not track libp2p-examples
- Generic Issue template update: currently we have one (defined In Project settings?), but it mentions discuss.ipfs.io

Needs current issue template to be removed (I do not have permissions to access Settings)


jacobheun (approved):
LGTM, there's no template setup in settings, likely just default. I'll double check after merge.

Issue #827, opened 2020-12-08
peerRouting.findPeer() trying to find self returns AggregateError instead more suitable error


a1300:
- **Version**: v0.29.3
- **Platform**: Linux
- **Subsystem**:

#### Steps to reproduce the error:

```bash
git clone https://gist.github.com/a1300/be916e343940c321aae587fd10cc3478 libp2p-peerRouting-find-self
cd libp2p-peerRouting-find-self
npm install
node index
```

Output:
```
(node:9829) UnhandledPromiseRejectionWarning: AggregateError: 
    Error: No peer found
        at Object.findPeer (/home/a1300/test/gny/libp2p-peerRouting-find-self/node_modules/libp2p-kad-dht/src/peer-routing/index.js:191:2
3)
    at maybeSettle (/home/a1300/test/gny/libp2p-peerRouting-find-self/node_modules/p-some/index.js:31:11)
    at /home/a1300/test/gny/libp2p-peerRouting-find-self/node_modules/p-some/index.js:69:23
(node:9829) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async fu
nction without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:9829) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not han
dled will terminate the Node.js process with a non-zero exit code.
```

jacobheun:
Throwing a dial self error here doesn't really make sense. Ideally the `findPeer(self)` should never happen, and applications should probably try to avoid this. There is the possibility that we may wish to just return ourselves from local, but I don't think this is a great idea.

Is there a particular reason you're trying to find self, or is this just do to generic findPeer(peer) and peer happens to be self in some instances?

a1300:
@jacobheun 

Background:
I need to know all peers of the network in a few seconds. Therefore every node announces a new found peer to the network (via pubsub broadcast event). I have 1 rendezvous node. If another node tells me about myself, I don't want to try to find myself with `peerRouting.findPeer()` myself. I have a check in place now.

But thats how I found this bug. The AggregateError Message is in my opinion not very describing. Maybe a new Error message `ERR_PEERROUTING_FIND_SELF_ERR`?

Is there a more elegant way to find all peers in a network in a few seconds?  I tried the randomWalk of the DHT, but it took too long.

jacobheun:
[...]

PR #941, opened 2021-05-11
fix: peerRouting.findPeer() trying to find self

zeim839:
Resolves an issue where nodes running findPeer() on themselves return abstract aggregateErrors instead of something "more suitable" like ERR_DIALED_SELF. Original issue can be found [here](https://github.com/libp2p/js-libp2p/issues/827).

This is accomplished by comparing the _idB58String of the ID parameter to this._peerId._idB58String at [peer-routing.js](https://github.com/libp2p/js-libp2p/blob/master/src/peer-routing.js#L102). if the two values are equal, an error is thrown alerting the user of the mistake. For example:

```javascript
async findPeer (id, options) { // eslint-disable-line require-await
    ...

    if (id._idB58String == this._peerId._idB58String) throw errCode(new Error('Cannot search for self'), 'ERR_DIALED_SELF')
    
   ...
}
```

### DIFF
--- a/src/peer-routing.js
+++ b/src/peer-routing.js
@@ -104,6 +104,8 @@ class PeerRouting {
       throw errCode(new Error('No peer routers available'), 'NO_ROUTERS_AVAILABLE')
     }
 
+    if (id._idB58String == this._peerId._idB58String) throw errCode(new Error('Cannot search for self'), 'ERR_DIALED_SELF')
+
     const output = await pipe(
       merge(
         ...this._routers.map(router => [router.findPeer(id, options)])
--- a/src/peer-routing.js
+++ b/src/peer-routing.js
@@ -104,7 +104,7 @@ class PeerRouting {
       throw errCode(new Error('No peer routers available'), 'NO_ROUTERS_AVAILABLE')
     }
 
-    if (id.toB58String()== this._peerId.toB58String()) {
+    if (id.toB58String() === this._peerId.toB58String()) {
       throw errCode(new Error('Should not try to find self'), 'ERR_FIND_SELF')
     }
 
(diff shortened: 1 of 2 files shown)


src/peer-routing.js
```diff
@@ -104,6 +104,8 @@ class PeerRouting {
       throw errCode(new Error('No peer routers available'), 'NO_ROUTERS_AVAILABLE')
     }
 
+    if (id._idB58String == this._peerId._idB58String) throw errCode(new Error('Cannot search for self'), 'ERR_DIALED_SELF')
```
vasco-santos, line 107:
```suggestion
    if (id.toB58String()== this._peerId.toB58String()) {
      throw errCode(new Error('Should not try to find self'), 'ERR_FIND_SELF')
    }
```

vasco-santos (changes requested):
Thanks @zeim839 

Can you please add a test in https://github.com/libp2p/js-libp2p/blob/master/test/peer-routing/peer-routing.node.js#L53 and https://github.com/libp2p/js-libp2p/blob/master/test/peer-routing/peer-routing.node.js#L274 with a fail expect such as in https://github.com/libp2p/js-libp2p/blob/master/test/peer-routing/peer-routing.node.js#L37 ?

zeim839:
@vasco-santos Added the request tests to the best of my knowledge. They both pass. Let me know if there's anything else I need to change/improve.

src/peer-routing.js
```diff
@@ -104,6 +104,10 @@ class PeerRouting {
       throw errCode(new Error('No peer routers available'), 'NO_ROUTERS_AVAILABLE')
     }
 
+    if (id.toB58String()== this._peerId.toB58String()) {
```
vasco-santos, line 107:
```suggestion
    if (id.toB58String() === this._peerId.toB58String()) {
```

vasco-santos (approved):
LGTM (pushed commit fixing linting)

Thanks @zeim839

PR #942, opened 2021-05-12
Fix: CONFIGURATION.md keychain typo

zeim839:
Fixed a typo at [setup keychain](https://github.com/libp2p/js-libp2p/blob/master/doc/CONFIGURATION.md#setup-with-keychain) that was referencing "libp2p"  instead of "node." See below:

```javascript
const node = await Libp2p.create({...})

//Typo:
await libp2p.loadKeychain()
//Correction:
await node.loadKeychain()
```

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -515,7 +515,7 @@ const node = await Libp2p.create({
   }
 })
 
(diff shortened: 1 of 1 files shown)


vasco-santos (approved):
Thanks @zeim839

PR #943, opened 2021-05-12
fix: store remote agent and protocol version during identify

achingbrain:
These values were not being stored during identify.

### DIFF
--- a/test/identify/index.spec.js
+++ b/test/identify/index.spec.js
@@ -454,7 +454,7 @@ describe('Identify', () => {
       await pWaitFor(() => connection.streams.length === 0)
       await connection.close()
 
-      const remotePeer = PeerId.createFromB58String('12D3KooWHFKTMzwerBtsVmtz4ZZEQy2heafxzWw6wNn5PPYkBxJ5')
+      const remotePeer = PeerId.createFromB58String(remoteAddr.getPeerId())
 
       const storedAgentVersion = libp2p.peerStore.metadataBook.getValue(remotePeer, 'AgentVersion')
       const storedProtocolVersion = libp2p.peerStore.metadataBook.getValue(remotePeer, 'ProtocolVersion')
--- a/test/identify/index.spec.js
+++ b/test/identify/index.spec.js
@@ -430,6 +430,39 @@ describe('Identify', () => {
       await connection.close()
     })
 
+    it('should store remote agent and protocol versions in metadataBook after connecting', async () => {
+      libp2p = new Libp2p({
+        ...baseOptions,
+        peerId
+      })
+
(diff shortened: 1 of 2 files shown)


achingbrain:
I did have a couple of questions about this - why is `message.agentVersion` stored PascalCase style as `"AgentVersion"`, I would have thought `"agentVersion"` would be more idiomatic.

Why are they stored as `Uint8Array`s when they are deserialized from the incoming message as protobuf strings?

vasco-santos:
> I did have a couple of questions about this - why is message.agentVersion stored PascalCase style as "AgentVersion", I would have thought "agentVersion" would be more idiomatic?

I tracked it down to https://github.com/libp2p/js-libp2p/pull/724 . I am not sure, probably to be similar to Go? We did try to follow go whenever possible for storing stuff in the stores.

> Why are they stored as Uint8Arrays when they are deserialized from the incoming protobuf message as strings?

Because we persist the metadataBook content as Uint8Array, as a Datastore Value. In this case, we considered that anything could be stored as metadata, including a serialized protobuf

test/identify/index.spec.js
```diff
@@ -430,6 +430,39 @@ describe('Identify', () => {
...
+      // Wait for peer store to be updated
+      // Dialer._createDialTarget (add), Identify (consume)
+      await pWaitFor(() => peerStoreSpyConsumeRecord.callCount === 1 && peerStoreSpyAdd.callCount === 1)
+      expect(libp2p.identifyService.identify.callCount).to.equal(1)
+
+      // The connection should have no open streams
+      await pWaitFor(() => connection.streams.length === 0)
+      await connection.close()
+
+      const remotePeer = PeerId.createFromB58String('12D3KooWHFKTMzwerBtsVmtz4ZZEQy2heafxzWw6wNn5PPYkBxJ5')
```
vasco-santos, line 457: Can we do not have the peerId hardcoded?

achingbrain: It's taken from https://github.com/libp2p/js-libp2p/blob/master/test/fixtures/browser.js#L6 where it's also hardcoded?

achingbrain: I thought I would have been able to get it from [remoteLibp2p](https://github.com/libp2p/js-libp2p/blob/master/test/identify/index.spec.js#L392) but that var does't appear to be set anywhere in the tests?

vasco-santos: I was meaning, probably more bullet proof to extract the id from the multiaddr

achingbrain: Aha, have updated it.

vasco-santos (approved):
LGTM

Issue #728, opened 2020-08-05
[keychain] support rotating keychain passphrase


jacobheun:
It should be possible to rotate/update your keychain passphrase. This should require the old and new passphrase. All keys should be imported with the old passphrase and then exported and stored with the new passphrase.

zeim839:
@jacobheun I've been working on a possible implementation that looks somewhat like this:
```js
var newKeychain = new Keychain(this._options.keychain.datastore, { pass: newPassword })
var oldKeychain = await this.keychain.listKeys();
oldKeychain.forEach(async key => {
    if (key.name != "self"){
        var data = await this.keychain.exportKey(key.name, oldPassword)
        await this.keychain.removeKey(key.name)
        await newKeychain.importKey(key.name, data, newPassword)
     }
})
this.keychain = newKeychain;
```
However, it seems that the keys/passwords are somehow bound together in the datastore, meaning I can't import keys into the new keychain with the new password (I'm getting the `Error: Cannot read the key, most likely the password is wrong` error).

There's also the issue of how to operate on the "self" key, as the `importKey()` (and other) functions deem it an "invalid key name". See [here](https://github.com/libp2p/js-libp2p/blob/master/src/keychain/index.js#L408) for example.

Let me know your thoughts/suggestions.

zeim839:
Seems I was confused on how the the keychain and crypto functions work (`keychain.importKey()` vs `crypto.keys.import` and etc.). Anyway, I've come up with a rough solution that looks like this:

```js
async changeKeychainPassword(oldPassword, newPassword){
    if (typeof oldPassword !== 'string' || typeof newPassword !== 'string') {
      throw new Error(`Invalid pass type '${typeof oldPassword}'`);
    }
    if (newPassword.length < 20) {
      throw new Error('pass must be least 20 characters')
    }
    // TODO: Need to decide on how to import/generate opts
    const newDek = newPassword
      ? crypto.pbkdf2(
        newPassword,
        this.opts.dek.salt,
        this.opts.dek.iterationCount,
        this.opts.dek.keyLength,
        this.opts.dek.hash)
      : ''
    const oldDek = privates.get(this).dek
    privates.set(this, { "dek":newDek })
    var keys = await this.listKeys()
    await keys.forEach(async key =>{
      // TODO: Decide how to handle deleting and importing the "self" key. 
          // importKey and removeKey throw an error when handling "self"
      if (key.name != "self"){
[...]

PR #944, opened 2021-05-14
Feat: keychain rotate passphrase

zeim839:
This is a rough implementation supporting [rotating keychain passphrases](https://github.com/libp2p/js-libp2p/issues/728). I need some suggestions on how to proceed. I haven't received a reply so I've copied the original conversation here for reference:

```js
async changeKeychainPassword(oldPassword, newPassword){
    if (typeof oldPassword !== 'string' || typeof newPassword !== 'string') {
      throw new Error(`Invalid pass type '${typeof oldPassword}'`);
    }
    if (newPassword.length < 20) {
      throw new Error('pass must be least 20 characters')
    }
    // TODO: Need to decide on how to import/generate opts
    const newDek = newPassword
      ? crypto.pbkdf2(
        newPassword,
        this.opts.dek.salt,
        this.opts.dek.iterationCount,
        this.opts.dek.keyLength,
        this.opts.dek.hash)
      : ''
    const oldDek = privates.get(this).dek
    privates.set(this, { "dek":newDek })
    var keys = await this.listKeys()
    await keys.forEach(async key =>{
      // TODO: Decide how to handle deleting and importing the "self" key. 
          // importKey and removeKey throw an error when handling "self"
      if (key.name != "self"){
        var keyAsPEM = await this._getPrivateKey(key.name)
        await this.removeKey(key.name)
        this.importKey(key.name, keyAsPEM, oldDek)
      }
    })
  }

```
[src/keychain/index.js](https://github.com/zeim839/js-libp2p/blob/feat/keychain-change-passphrase/src/keychain/index.js#L507)

What's left is to decide how to handle the "self" key (`this.importKey()` and `this.removeKey()` throw an error on "self") and how to handle changing passwords in the context of src/index.js (see [here](https://github.com/zeim839/js-libp2p/blob/feat/keychain-change-passphrase/src/index.js#L209)). I'm trying to be cautious/minimal with the changes I make so I'd appreciate some suggestions on how to proceed.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -207,6 +207,7 @@ class Libp2p extends EventEmitter {
     }
 
     // Create keychain
+      // TODO: Decide how to change these options when the keychain password is being changed
     if (this._options.keychain && this._options.keychain.datastore) {
       log('creating keychain')
 
--- a/src/index.js
+++ b/src/index.js
@@ -207,7 +207,6 @@ class Libp2p extends EventEmitter {
     }
 
     // Create keychain
-      // TODO: Decide how to change these options when the keychain password is being changed
     if (this._options.keychain && this._options.keychain.datastore) {
       log('creating keychain')
 
--- a/src/keychain/index.js
+++ b/src/keychain/index.js
@@ -505,16 +505,16 @@ class Keychain {
   }
 
   async rotateKeychainPass (oldPass, newPass){
-    if (typeof oldPassword !== 'string' || typeof newPassword !== 'string') {
-      throw new Error(`Invalid pass type '${typeof oldPassword}'`);
+    if (typeof oldPass !== 'string' || typeof newPass !== 'string') {
+      throw new Error(`Invalid pass type '${typeof oldPass}'`);
     }
-    if (newPassword.length < 20) {
+    if (newPass.length < 20) {
       throw new Error('pass must be least 20 characters')
     }
     // TODO: Need to decide on how to import/generate opts
-    const newDek = newPassword
+    const newDek = newPass
       ? crypto.pbkdf2(
-        newPassword,
+        newPass,
         this.opts.dek.salt,
         this.opts.dek.iterationCount,
         this.opts.dek.keyLength,
--- a/src/keychain/index.js
+++ b/src/keychain/index.js
@@ -557,7 +557,7 @@ class Keychain {
       batch.put(DsName(key.name), uint8ArrayFromString(keyAsPEM))
       batch.put(DsInfoName(key.name), uint8ArrayFromString(JSON.stringify(keyInfo)))
       await batch.commit()
-    })
+    }
     log('keychain reconstructed')
   }
 }
--- a/src/keychain/index.js
+++ b/src/keychain/index.js
@@ -503,6 +503,36 @@ class Keychain {
       return throwDelayed(errcode(new Error(`Key '${name}' does not exist. ${err.message}`), 'ERR_KEY_NOT_FOUND'))
     }
   }
+
+  async changeKeychainPassword(oldPassword, newPassword){
+    if (typeof oldPassword !== 'string' || typeof newPassword !== 'string') {
+      throw new Error(`Invalid pass type '${typeof oldPassword}'`);
+    }
+    if (newPassword.length < 20) {
+      throw new Error('pass must be least 20 characters')
+    }
+    // TODO: Need to decide on how to import/generate opts
+    const newDek = newPassword
+      ? crypto.pbkdf2(
+        newPassword,
+        this.opts.dek.salt,
+        this.opts.dek.iterationCount,
+        this.opts.dek.keyLength,
+        this.opts.dek.hash)
+      : ''
+    const oldDek = privates.get(this).dek
+    privates.set(this, { "dek":newDek })
+    var keys = await this.listKeys()
+    await keys.forEach(async key =>{
+      // TODO: Decide how to handle deleting and importing the "self" key. 
+          // importKey and removeKey throw an error when handling "self"
+      if (key.name != "self"){
+        var keyAsPEM = await this._getPrivateKey(key.name)
+        await this.removeKey(key.name)
+        this.importKey(key.name, keyAsPEM, oldDek)
+      }
+    })
+  }
 }
 
 module.exports = Keychain
--- a/src/keychain/index.js
+++ b/src/keychain/index.js
@@ -540,7 +540,8 @@ class Keychain {
       const res = await this.store.get(DsName(key.name))
       const pem = uint8ArrayToString(res)
       const privateKey = await crypto.keys.import(pem, oldDek)
-      const keyAsPEM = await privateKey.export(newDek)
+      const password = newDek.toString()
+      const keyAsPEM = await privateKey.export(password)
 
       // Remove key with old pass
       const batch = this.store.batch()
--- a/src/keychain/index.js
+++ b/src/keychain/index.js
@@ -511,7 +511,13 @@ class Keychain {
     if (newPass.length < 20) {
       throw new Error('pass must be least 20 characters')
     }
-    // TODO: Need to decide on how to import/generate opts
+    this.opts = {
+      dek: Keychain.generateOptions().dek,
+      pass: newPass,
+      datastore: this.store
+    }
+
+    const oldDek = privates.get(this).dek
     const newDek = newPass
       ? crypto.pbkdf2(
         newPass,
@@ -520,17 +526,27 @@ class Keychain {
         this.opts.dek.keyLength,
         this.opts.dek.hash)
       : ''
-    const oldDek = privates.get(this).dek
     privates.set(this, { "dek":newDek })
+
     var keys = await this.listKeys()
     await keys.forEach(async key =>{
-      // TODO: Decide how to handle deleting and importing the "self" key. 
-          // importKey and removeKey throw an error when handling "self"
-      if (key.name != "self"){
-        var keyAsPEM = await this._getPrivateKey(key.name)
-        await this.removeKey(key.name)
-        this.importKey(key.name, keyAsPEM, oldDek)
+      var keyAsPEM = await this._getPrivateKey(key.name)
+
+      // Remove key with old pass
+      const batch = this.store.batch()
+      batch.delete(DsName(key.name))
+      batch.delete(DsInfoName(key.name))
+      await batch.commit()    
+
+
+      // Import key with new pass
+      var keyInfo = {
+        name: key.name,
+        id: key.id
       }
+      batch.put(DsName(key.name), uint8ArrayFromString(keyAsPEM))
+      batch.put(DsInfoName(key.name), uint8ArrayFromString(JSON.stringify(keyInfo)))
+      await batch.commit()
     })
   }
 }
--- a/src/keychain/index.js
+++ b/src/keychain/index.js
@@ -504,7 +504,7 @@ class Keychain {
     }
   }
 
-  async changeKeychainPassword(oldPassword, newPassword){
+  async rotateKeychainPass (oldPass, newPass){
     if (typeof oldPassword !== 'string' || typeof newPassword !== 'string') {
       throw new Error(`Invalid pass type '${typeof oldPassword}'`);
     }
--- a/src/keychain/index.js
+++ b/src/keychain/index.js
@@ -526,7 +526,6 @@ class Keychain {
     log('recreating keychain')
     const oldDek = privates.get(this).dek
     this.opts.pass = newPass
-    // eslint-disable-next-line constant condition
     const newDek = newPass
       ? crypto.pbkdf2(
         newPass,
--- a/src/keychain/index.js
+++ b/src/keychain/index.js
@@ -503,7 +503,13 @@ class Keychain {
       return throwDelayed(errcode(new Error(`Key '${name}' does not exist. ${err.message}`), 'ERR_KEY_NOT_FOUND'))
     }
   }
-
+  
+  /**
+   * Rotate keychain password and re-encrypt all assosciated keys
+   *
+   * @param {string} oldPass - The old local keychain password
+   * @param {string} newPass - The new local keychain password
+   */
   async rotateKeychainPass (oldPass, newPass){
     if (typeof oldPass !== 'string' || typeof newPass !== 'string') {
       throw new Error(`Invalid pass type '${typeof oldPass}'`);
@@ -517,7 +523,6 @@ class Keychain {
       datastore: this.store
     }
 
-    const oldDek = privates.get(this).dek
     const newDek = newPass
       ? crypto.pbkdf2(
         newPass,
--- a/src/keychain/index.js
+++ b/src/keychain/index.js
@@ -536,7 +536,7 @@ class Keychain {
       : ''
     privates.set(this, { dek: newDek })
     const keys = await this.listKeys()
-    await keys.forEach(async key => {
+    for (const key of keys) {
       const res = await this.store.get(DsName(key.name))
       const pem = uint8ArrayToString(res)
       const privateKey = await crypto.keys.import(pem, oldDek)
--- a/src/keychain/index.js
+++ b/src/keychain/index.js
@@ -1,6 +1,9 @@
 /* eslint max-nested-callbacks: ["error", 5] */
 'use strict'
-
+const debug = require('debug')
+const log = Object.assign(debug('libp2p:keychain'), {
+  error: debug('libp2p:keychain:err')
+})
 const sanitize = require('sanitize-filename')
 const mergeOptions = require('merge-options')
 const crypto = require('libp2p-crypto')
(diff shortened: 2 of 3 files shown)


src/index.js
```diff
@@ -207,6 +207,7 @@ class Libp2p extends EventEmitter {
     }
 
     // Create keychain
+      // TODO: Decide how to change these options when the keychain password is being changed
```
vasco-santos, line 210: Do you mean the `const keychainOpts = Keychain.generateOptions()`? I think we can just run it again to get the options. What do you think?

src/keychain/index.js
```diff
@@ -503,6 +503,36 @@ class Keychain {
       return throwDelayed(errcode(new Error(`Key '${name}' does not exist. ${err.message}`), 'ERR_KEY_NOT_FOUND'))
     }
   }
+
+  async changeKeychainPassword(oldPassword, newPassword){
```
vasco-santos, line 507:
```suggestion
  async rotateKeychainPass (oldPass, newPass){
```

Perhaps rotate is a better naming, and we use pass everywhere, so I think this should also use `pass`

src/keychain/index.js
```diff
@@ -503,6 +503,36 @@ class Keychain {
       return throwDelayed(errcode(new Error(`Key '${name}' does not exist. ${err.message}`), 'ERR_KEY_NOT_FOUND'))
     }
   }
+
```
vasco-santos, line 506: Please add `jsdoc` here

src/keychain/index.js
vasco-santos: We probably need to specify the key we want to rotate?

src/keychain/index.js
```diff
@@ -503,6 +503,36 @@ class Keychain {
...
+        this.opts.dek.iterationCount,
+        this.opts.dek.keyLength,
+        this.opts.dek.hash)
+      : ''
+    const oldDek = privates.get(this).dek
+    privates.set(this, { "dek":newDek })
+    var keys = await this.listKeys()
+    await keys.forEach(async key =>{
+      // TODO: Decide how to handle deleting and importing the "self" key. 
+          // importKey and removeKey throw an error when handling "self"
```
vasco-santos, line 528: yeah, we should not change this behaviour. I would say we should create internal methods to handle the common logic

vasco-santos (commented):
Thanks for the PR @zeim839 

I think I addressed all the comments

src/index.js
zeim839: Yes that sounds good to me.

zeim839:
@vasco-santos Please take a look at the new changes and let me know what you think. Regarding tests, I've incorporated testing for validation but I'm not exactly sure how to go about the following:
```js
it('can rotate keychain passphrase', async () => {  
    //TODO...
})
```

zeim839:
I think I've come up with a possible solution. If its ok with you then everything should be good to go. I also decided to create a new keychain [here](https://github.com/libp2p/js-libp2p/pull/944/commits/825d70c1718f56700c20b69fa0418b158cf022c3#diff-86874834813c75bef940feea4eabc092bb90aeb6dd04173bbfa55ca8e7d766b1R497) (instead of using ks) to avoid influencing any other/future tests that dont use the new rotated password.

src/keychain/index.js
```diff
@@ -503,6 +503,57 @@ class Keychain {
...
+  
+  /**
+   * Rotate keychain password and re-encrypt all assosciated keys
+   *
+   * @param {string} oldPass - The old local keychain password
+   * @param {string} newPass - The new local keychain password
+   */
+  async rotateKeychainPass (oldPass, newPass){
+    if (typeof oldPass !== 'string' || typeof newPass !== 'string') {
+      throw new Error(`Invalid pass type '${typeof oldPass}'`);
```
vasco-santos, line 515: Probably not essential, but what do you think about using `throwDelayed` like the rest of the keychain functions? And please add an `errCode`

src/keychain/index.js
```diff
@@ -503,6 +503,57 @@ class Keychain {
...
+    }
+    if (newPass.length < 20) {
+      throw new Error('pass must be least 20 characters')
```
vasco-santos, line 518: Same as above

src/keychain/index.js
```diff
@@ -503,6 +503,57 @@ class Keychain {
...
+    }
+    this.opts = {
```
vasco-santos, line 520: Can we add a log before this with `recreating keychain` and after the keys iteration `keychain reconstructed` to be consistent with what we do on creation?

src/keychain/index.js
```diff
@@ -503,6 +503,57 @@ class Keychain {
...
+
+    const newDek = newPass
+      ? crypto.pbkdf2(
+        newPass,
+        this.opts.dek.salt,
+        this.opts.dek.iterationCount,
+        this.opts.dek.keyLength,
+        this.opts.dek.hash)
+      : ''
+    privates.set(this, { "dek":newDek })
```
vasco-santos, line 534:
```suggestion
    privates.set(this, { "dek": newDek })
```

src/keychain/index.js
```diff
@@ -503,6 +503,57 @@ class Keychain {
...
+
+    var keys = await this.listKeys()
+    await keys.forEach(async key =>{
+      var keyAsPEM = await this._getPrivateKey(key.name)
```
vasco-santos, line 538:
```suggestion
      const keyAsPEM = await this._getPrivateKey(key.name)
```

src/keychain/index.js
```diff
@@ -503,6 +503,57 @@ class Keychain {
...
+      ? crypto.pbkdf2(
+        newPass,
+        this.opts.dek.salt,
+        this.opts.dek.iterationCount,
+        this.opts.dek.keyLength,
+        this.opts.dek.hash)
+      : ''
+    privates.set(this, { "dek":newDek })
+
+    var keys = await this.listKeys()
```
vasco-santos, line 536:
```suggestion
    const keys = await this.listKeys()
```

src/keychain/index.js
```diff
@@ -503,6 +503,57 @@ class Keychain {
...
+
+      // Remove key with old pass
+      const batch = this.store.batch()
+      batch.delete(DsName(key.name))
+      batch.delete(DsInfoName(key.name))
+      await batch.commit()    
+
+
+      // Import key with new pass
+      var keyInfo = {
```
vasco-santos, line 548:
```suggestion
      const keyInfo = {
```

src/keychain/index.js
vasco-santos, line 538: this PEM was created with the old dek https://github.com/libp2p/js-libp2p/blob/v0.31.5/src/keychain/index.js#L238. You can look into the export key to see how to get the private key from pem using the previous `dek` and then you need to encrypt it once again as it is done in createKey

vasco-santos (changes requested):
I did an initial iteration on the PR. The new pm does not seem to be created with the new password. I left inline comments on how I think it would look like.

Also, before submitting the PR try to get the CI green please 🙏🏼  I added small inline suggestion to fix the lint, but I probably did not get to all of them. You should run both `npm run lint` and `npm run test`

Thanks @zeim839

zeim839:
> Also, before submitting the PR try to get the CI green please 🙏🏼 I added small inline suggestion to fix the lint, but I probably did not get to all of them. You should run both `npm run lint` and `npm run test`

Sorry about that, I shouldve read the contrib guidelines more carefully. Anyway, I've ran lint and fixed any issues that I could trace back to me but I'm not sure If I can even get the CI to green because linting fails regardless of my edits. I've fixed the mistakes from the previous review and wrote the test as described. 

However, I keep getting `"Error: Cannot read the key, most likely the password is wrong or not a RSA key"` [here](https://github.com/libp2p/js-libp2p/pull/944/files#diff-6bbef5dd5eeccbdf00ec112882c711b59ef3db822873cb85f9244acf9a1e687dR542) when testing on browser, despite it working fine in test:node. Am I missing/overlooking something?

src/keychain/index.js
```diff
@@ -503,6 +506,60 @@ class Keychain {
...
+    if (typeof newPass !== 'string') {
+      return throwDelayed(errcode(new Error(`Invalid new pass type '${typeof newPass}'`), 'ERR_INVALID_NEW_PASS_TYPE'))
+    }
+    if (newPass.length < 20) {
+      return throwDelayed(errcode(new Error(`Invalid pass length ${newPass.length}`), 'ERR_INVALID_PASS_LENGTH'))
+    }
+    log('recreating keychain')
+    const oldDek = privates.get(this).dek
+    this.opts.pass = newPass
+    const newDek = newPass
```
vasco-santos, line 529:
```suggestion
    // eslint-disable-next-line constant condition
    const newDek = newPass
```

test/keychain/keychain.spec.js
```diff
@@ -492,6 +493,84 @@ describe('keychain', () => {
...
+        const oldDek = options.pass
+          ? crypto.pbkdf2(
+            options.pass,
+            options.dek.salt,
+            options.dek.iterationCount,
+            options.dek.keyLength,
+            options.dek.hash)
+          : ''
+
+        const newDek = 'newInsecurePassphrase'
```
vasco-santos, line 558:
```suggestion
        // eslint-disable-next-line no-constant-condition
        const newDek = 'newInsecurePassphrase'
```

src/keychain/index.js
```diff
@@ -503,6 +506,61 @@ class Keychain {
...
+    if (typeof newPass !== 'string') {
+      return throwDelayed(errcode(new Error(`Invalid new pass type '${typeof newPass}'`), 'ERR_INVALID_NEW_PASS_TYPE'))
+    }
+    if (newPass.length < 20) {
+      return throwDelayed(errcode(new Error(`Invalid pass length ${newPass.length}`), 'ERR_INVALID_PASS_LENGTH'))
+    }
+    log('recreating keychain')
+    const oldDek = privates.get(this).dek
+    this.opts.pass = newPass
+    // eslint-disable-next-line constant condition
```
vasco-santos, line 529:
```suggestion
```

vasco-santos:
> Sorry about that, I shouldve read the contrib guidelines more carefully. Anyway, I've ran lint and fixed any issues that I could trace back to me but I'm not sure If I can even get the CI to green because linting fails regardless of my edits. I've fixed the mistakes from the previous review and wrote the test as described.

No problem, thanks for looking into it. I already submitted a suggestion fixing the lint. I will try to replicate the browser issue and get back to you

src/keychain/index.js
```diff
@@ -503,6 +506,60 @@ class Keychain {
...
+      ? crypto.pbkdf2(
+        newPass,
+        this.opts.dek.salt,
+        this.opts.dek.iterationCount,
+        this.opts.dek.keyLength,
+        this.opts.dek.hash)
+      : ''
+    privates.set(this, { dek: newDek })
+    const keys = await this.listKeys()
+    await keys.forEach(async key => {
```
vasco-santos, line 539:
```suggestion
    for (const key of keys) {
```

this `await` does not work like this and the forEach does not return a promise. So this means `rotateKeychainPass` would return before running the cycle. As a side effect, in the browser tests were failing because it was slower + the usage of `setTimeout` in test

Note that the suggestion change below is also needed to close the for cycle correctly

src/keychain/index.js
```diff
@@ -503,6 +506,60 @@ class Keychain {
...
+
+      // Import key with new pass
+      const keyInfo = {
+        name: key.name,
+        id: key.id
+      }
+      batch.put(DsName(key.name), uint8ArrayFromString(keyAsPEM))
+      batch.put(DsInfoName(key.name), uint8ArrayFromString(JSON.stringify(keyInfo)))
+      await batch.commit()
+    })
```
vasco-santos, line 560:
```suggestion
    }
```

test/keychain/keychain.spec.js
```diff
@@ -492,6 +493,85 @@ describe('keychain', () => {
...
+      } catch (err) {
+        expect(err).to.exist()
+      }
+    })
+
+    it('can rotate keychain passphrase', async () => {
+      await kc.createKey('keyCreatedWithOldPassword', 'rsa', 2048)
+      await kc.rotateKeychainPass(oldPass, 'newInsecurePassphrase')
+
+      setTimeout(async () => {
```
vasco-santos, line 543: With the above change, this `setTimeout` can be removed

test/keychain/keychain.spec.js
```diff
@@ -492,6 +493,85 @@ describe('keychain', () => {
...
+          ? crypto.pbkdf2(
+            'newInsecurePassphrase',
+            options.dek.salt,
+            options.dek.iterationCount,
+            options.dek.keyLength,
+            options.dek.hash)
+          : ''
+
+        // Dek with old password should not work:
+        expect(await kc.importKey('keyWhosePassChanged', pem, oldDek)).to.throw()
```
vasco-santos, line 569:
```suggestion
        await expect(kc.importKey('keyWhosePassChanged', pem, oldDek))
          .to.eventually.be.rejected()
```

test/keychain/keychain.spec.js
```diff
@@ -492,6 +493,85 @@ describe('keychain', () => {
...
+        // Dek with new password should work:
+        expect(await kc.importKey('keyWhosePasswordChanged', pem, newDek)).to.have.property('name', 'keyWhosePasswordChanged')
```
vasco-santos, line 571:
```suggestion
        await expect(kc.importKey('keyWhosePasswordChanged', pem, newDek))
          .to.eventually.have.property('name', 'keyWhosePasswordChanged')
```

vasco-santos (changes requested):
Thanks for the new iteration. This is mostly done, I tried the changes I suggested below and I got everything to work.

Moreover, I added a lint skip for the error you had, as the linter was complaining for the ternary condition in the constant in the test, which is fine. When you add the suggestion, please check if the linting is still fine, given that the suggestion spaces could break the spacing.

Also, you will need to pull the changes done in the commit that fixed the linting
when you add the suggestions

src/keychain/index.js
```diff
@@ -503,6 +506,60 @@ class Keychain {
...
+      const privateKey = await crypto.keys.import(pem, oldDek)
+      const password = newDek.toString()
+      const keyAsPEM = await privateKey.export(password)
+
+      // Remove key with old pass
+      const batch = this.store.batch()
+      batch.delete(DsName(key.name))
+      batch.delete(DsInfoName(key.name))
+      await batch.commit()
+
```
vasco-santos, line 551:
```suggestion
```

The last commit should be enough right?

vasco-santos (changes requested):

src/keychain/index.js
zeim839: I thought so too at first but I believe there's some sort of issue where batch cannot both delete and create a key of the same name without separate commits. When I remove the first `batch.commit()` and run `rotateKeychainPass()`, the datastore appears to be completely empty like so: `MemoryDatastore { data: {} }`. If I try `batch.put()` with a different key name, the operation seems to work as expected (i.e the old key gets deleted and the new key is appended). 

Try running the below and let me know If I'm missing something or if this is a legitimate issue:
```js
await kc.createKey('keyCreatedWithOldPassword', 'rsa', 2048)
const dsnameData = new Key('/pkcs8/keyCreatedWithOldPassword')
const dsnameInfo = new Key('/info/keyCreatedWithOldPassword')
const batch = ds.batch()

// Delete key data and info
batch.delete(dsnameData)
batch.delete(dsnameInfo)

// Append key data and info
await batch.put(dsnameData, uint8ArrayFromString('pem'))
await batch.put(dsnameInfo, uint8ArrayFromString('info'))

await batch.commit()
console.log(ds) // Output: MemoryDatastore { data: {} }
```

vasco-santos: Thanks for the details. I see, but rethinking this. A datastore will always be indexed by a key, so the `delete` is redundant as the `put` will write over it

vasco-santos, line 551:
```suggestion
```

src/keychain/index.js
```diff
@@ -503,6 +506,60 @@ class Keychain {
...
+    privates.set(this, { dek: newDek })
+    const keys = await this.listKeys()
+    for (const key of keys) {
+      const res = await this.store.get(DsName(key.name))
+      const pem = uint8ArrayToString(res)
+      const privateKey = await crypto.keys.import(pem, oldDek)
+      const password = newDek.toString()
+      const keyAsPEM = await privateKey.export(password)
+
+      // Remove key with old pass
```
vasco-santos, line 546:
```suggestion
      // Update stored key
```

src/keychain/index.js
vasco-santos: Added two suggestions to get the delete out again.

vasco-santos (changes requested):

zeim839:
@vasco-santos Thanks for the help, I wasn't aware of that. I've removed the redundancy and everything works as expected now.

vasco-santos (approved):
LGTM!
Thanks for landing this @zeim839

vasco-santos:
Released in `0.31.6`

zeim839:
@vasco-santos Thanks for all the help!

PR #945, opened 2021-05-24
fix: chat, TypeError: multiaddr is not a function

mobeiqingkong:


vasco-santos:
Thanks for flagging this @mobeiqingkong 

We should use the Multiaddr constructor exported to be consistent across the entire js-libp2p codebase. So I created #946 to get this fixed

PR #947, opened 2021-05-27
fix: dialer leaking resources after stopping

### LINKED ISSUE #779: libp2p not shutting down properly
- **Version**: 0.29.0
- **Platform**: All
- **Subsystem**: Shutdown, dialer, latency monitor

#### Description:

When libp2p is stopped the dial queue is not being properly emptied. This can cause in progress outbound dials to not be properly terminated which will hold the process open. 

Additionally, the latency monitor is not being stopped by the connection manager which will cause timers to continue.

See https://github.com/ipfs/js-ipfs/issues/3191#issuecomment-706075918

vasco-santos:
This PR fixes the potential leaking of resources in the dialer on stop.

During the `transport.dial` flow, if libp2p is stopped the on going dials were being aborted. However, situations like https://github.com/libp2p/js-libp2p-tcp/issues/142 were possible when a given resolvable multiaddr (`dnsaddr`) is being resolved before getting into the transport level. This would result in new dial attempts after stop and established connections that prevented the process from properly stop.

Unfortunately, node's dns record resolution does not allow the lookup to be abortable. As a consequence, this PR adds a competing Cancellable promise, which will be rejected in case of `libp2p.stop()` => `dialer.destroy()`. With this, the promise that would resolve the `dialTarget` will be rejected and the current on going dial all together.

Moreover, as described in #779 the latency monitor was not being properly closed. So, I also added `start` and `stop` functions to it (instead of the really odd pattern of starting it in the constructor).

Closes #779

### DIFF
--- a/test/dialing/direct.spec.js
+++ b/test/dialing/direct.spec.js
@@ -290,6 +290,34 @@ describe('Dialing (direct, WebSockets)', () => {
     }
   })
 
+  it('should cancel pending dial targets before proceeding', async () => {
+    const dialer = new Dialer({
+      transportManager: localTM,
+      peerStore: {
+        addressBook: {
+          set: () => { }
+        }
+      }
+    })
+
+    sinon.stub(dialer, '_createDialTarget').callsFake(() => {
+      const deferredDial = pDefer()
+      return deferredDial.promise
+    })
+
+    // Perform dial
+    const dialPromise = dialer.connectToPeer(peerId)
+
+    // Let the call stack run
+    await delay(0)
+
+    dialer.destroy()
+
+    await expect(dialPromise)
+      .to.eventually.be.rejected()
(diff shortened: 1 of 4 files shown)


test/dialing/direct.spec.js
```diff
@@ -462,6 +490,41 @@ describe('Dialing (direct, WebSockets)', () => {
...
+
+      // Perform dial
+      const dialPromise = libp2p.dial(remotePeerId)
+
+      // Let the call stack run
+      await delay(0)
+
+      await libp2p.stop()
+      await expect(dialPromise)
+        .to.eventually.be.rejected()
```
achingbrain, line 525: This assertion should be be expanded to ensure the right type of error is being thrown.

achingbrain (changes requested):
LGTM, just a small enhancement to the test needed

PR #960, opened 2021-07-22
Turn interface definition and compliance tests into devDependency

robertkiel:
## Current situation

When installing latest version of `js-libp2p`, it installs `libp2p-interfaces-compliance-tests@1.0.0` as dependency which itself install `aegir@33` which is not needed for `js-libp2p` to run.

## Problem

Installing `aegir` changes the way how some modules are resolved by Node.js, hence tools such as Typedoc fail while loading plugins.

## Solution

Turn dependencies that are only needed for development or releasing NPM packages into devDependencies to make sure they are not installled when installing `js-libp2p`. This also reduces the footprint of `js-libp2p` as a dependency.


vasco-santos:
oh damn, thanks for the PR @robertkiel

package.json
```diff
@@ -151,6 +149,8 @@
     "libp2p-delegated-peer-routing": "^0.10.0",
     "libp2p-floodsub": "^0.27.0",
     "libp2p-gossipsub": "^0.10.0",
+    "libp2p-interfaces": "^1.0.0",
```
vasco-santos, line 152: interfaces should be a dependency, but the compliance totally a devDep

vasco-santos (changes requested):

package.json
robertkiel: done

robertkiel: It could also make sense to turn `aegir` into devDependency within `libp2p-interfaces-compliance-tests` but haven't checked that

vasco-santos (approved):
LGTM

PR #990, opened 2021-09-13
chore: update datastore

achingbrain:
`interface-datastore` now only contains the interface definition, `datastore-core` has the various implementations.


vasco-santos (approved):
LGTM

PR #1008, opened 2021-10-07
feat: support delegated value store in content-routing module

yusefnapora:
This adds support for delegated `put` / `get` operations in the `content-routing` module instead of always using the DHT. It depends on some in-progress PRs, so CI is currently failing.

Dependencies:

- [ ] https://github.com/libp2p/js-libp2p-interfaces/pull/108
- [ ] https://github.com/libp2p/js-libp2p-delegated-content-routing/pull/54

To run the tests locally, you can checkout the PR branch of `js-libp2p-interfaces` and `npm link` it into js-libp2p. For now, I'm depending on a commit hash for the delegated-content-routing PR, but that doesn't work for libp2p-interfaces since it's a multi-module repo.

This adds a new entry to the config `modules` object called `valueStorage` which can contain `ValueStore` implementations. Currently the only non-DHT implementation is in the delegated-content-routing PR above.

I don't think the config change is technically a breaking API change since it's additive - if you don't add the new valueStorage module in your config you just won't get delegated put/get operations, but the DHT will still work if you've enabled it.

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -190,6 +190,23 @@ If you want to know more about libp2p pubsub, you should read the following cont
 - https://docs.libp2p.io/concepts/publish-subscribe
 - https://github.com/libp2p/specs/tree/master/pubsub
 
+### Value Storage
+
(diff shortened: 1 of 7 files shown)


BigLep:
Moving this to draft until we're able to give a concerted effort on IPNS improvements as part of https://github.com/ipfs/js-ipfs/issues/2921

achingbrain:
Closing as this can be implemented by specifying additional content routers.

PR #1009, opened 2021-10-12
feat: update dht

achingbrain:
Changes dht creation to use factory function.

Depends on:

- [x] https://github.com/libp2p/js-libp2p-kad-dht/pull/237
- [x] https://github.com/libp2p/js-libp2p-daemon/pull/49
- [x] https://github.com/libp2p/interop/pull/51


.github/workflows/examples.yml
```diff
@@ -21,82 +24,118 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16
       - run: npm install
-      - run: cd examples && yarn && npm run test -- auto-relay
+      - run: cd examples && npm i && npm run test -- auto-relay
```
achingbrain, line 31: We can put yarn back when we remove `node-fetch@2.x` from ipfs-utils, or when yarn can download tarball dependencies reliably.

This either needs:

1. https://github.com/node-fetch/node-fetch/pull/1172 merging
2. Swap node-fetch for undici
3. Drop CJS support (node-fetch 3 has the above fix but is ESM-only)

achingbrain:
This has been rolled up to js-ipfs here: https://github.com/ipfs/js-ipfs/pull/3947

PR #1011, opened 2021-10-26
feat: allow user to pass in multistream implementation

SahidMiller:
Hi! Just wanted to allow users to easily pass in a multistream implementation. Useful if wanting a custom match function based on RegExp rather than strings.

If not merged, some advice would be helpful the best way to pass in custom modules for multistream / upgrader / protobook / etc.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -105,6 +105,8 @@ const { updateSelfPeerRecord } = require('./record/utils')
  * @property {Object} [dht]
  * @property {{new(...args: any[]): Pubsub}} [pubsub]
  * @property {Protector} [connProtector]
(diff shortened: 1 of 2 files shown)


lidel:
@vasco-santos do you have context to review this ?

BigLep:
2022-01-28: need to decide whether it's worth complicating the API for this specific usecase.

lidel:
@SahidMiller we triaged this and decided that this PR is not the best place to tackle this.
Please open design discussion issue (feeature request) in https://github.com/multiformats/js-multistream-select/ and we will continue there, thanks!

PR #1025, opened 2021-11-18
fixed unhandled promise rejection in case peer lookup fails

andreykiryushkin:
We noticed that in rare cases libp2p leaves unhandled rejected promises in the `findPeer` group of methods. This PR tries to fix them. It's not easily reproduced.

### DIFF
--- a/src/peer-routing.js
+++ b/src/peer-routing.js
@@ -110,7 +110,7 @@ class PeerRouting {
 
     const output = await pipe(
       merge(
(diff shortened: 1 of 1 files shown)


achingbrain:
Thanks for submitting this PR, could you please share a stack trace for the unhandled promise rejection you are seeing?

BigLep:
Going to close this for now.  Please reopen with a stack trace showing the problem if it persists.

PR #1036, opened 2021-11-24
feat: add fetch protocol

stbrody:


src/identify/message.d.ts
```diff
@@ -34,25 +34,40 @@ export class Identify implements IIdentify {
     constructor(p?: IIdentify);
```
stbrody, line 34: all these changes to the random autogenerated protobuf code happened when I ran `npm run build:proto && npm run build:proto-types` on a totally clean checkout before I even started making any of my changes.

stbrody: If you want me to leave those changes out of this PR just let me know and I can unstage those files.

stbrody:
@achingbrain @vasco-santos @aschmahmann  PTAL!

CC @oed @smrz2001

aschmahmann:
@vasco-santos @achingbrain this is needed to implement the updated IPNS over PubSub [spec ](https://github.com/ipfs/specs/blob/master/naming/pubsub.md)(from 2019) and might also be useful for other folks such as Ceramic who just need a basic "give me the data corresponding to `key`" protocol.

src/fetch/README.md
```diff
@@ -0,0 +1,32 @@
...
+```javascript
+var Fetch = require('libp2p/src/fetch')
+
+/**
+ * Given a key (as a string) returns a value (as a Uint8Array), or null if the key isn't found.
+ * @param key - a string
+ * @returns value - a Uint8Array value that corresponds to the given key, or null if the key doesn't 
+ *   have a corresponding value.
+ */
+function lookup(key) {
```
oed, line 23: Maybe make this async for added optionality?

src/fetch/index.js
```diff
@@ -0,0 +1,106 @@
...
+ * responds based on looking up the key in the request via the lookup callback.
+ * @param {MuxedStream} stream
+ * @param {LookupFunction} lookup
+ */
+async function handleRequest(stream, lookup) {
+  const shake = handshake(stream)
+  const request = FetchRequest.decode((await lp.decode.fromReader(shake.reader).next()).value.slice())
+
+  let response
+  const data = await lookup(request.identifier)
```
oed, line 73: Ok i guess it's already async. Maybe update the Readme?

src/fetch/README.md
stbrody: code is right, just missed this in the readme.  Updated

src/fetch/index.js
stbrody: done

stbrody:
> the main technical note I'd give is that you should give room to pull from multiple data sources

Yeah I thought about this a little bit.  The API right now is very simple - when the protocol is registered the user provides a callback to map key to value.  If an app developer wanted to use two different subsystems that both relied on the fetch protocol, the dev could easily wrap the two lookup functions from each subsystem into a single one that dispatches based on the key.

I feel like if we try to build some way to register multiple lookup callbacks to different ranges of the keyspace, it's just going to complicate the API and may wind up being too brittle and prescriptive for every use case.  I think it might genuinely make sense to leave this to users, but I could be swayed if you disagree @aschmahmann

stbrody:
Also it looks like CI found a bunch of compilation errors with typescript support: https://github.com/libp2p/js-libp2p/runs/4316690228?check_suite_focus=true

Is there a way I can test this locally so I can rapidly iterate to fix?  Running `npm run build && npm run test` all worked for me locally.

stbrody:
I seem to be getting failures generating type information, but the errors are all coming from the autogenerated protocol buffer code: https://github.com/libp2p/js-libp2p/runs/4316690228?check_suite_focus=true

I'm wondering if this is related to why a bunch of autogenerated protobuf code got updated when I ran the `build:proto` commands locally even on a clean build.  Maybe my local setup is using a different version of whatever package is generating the protobuf code somehow?  Any advice where to look here to fix this would be appreciated.

src/fetch/index.js
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * @typedef {import('../')} Libp2p
+ * @typedef {import('multiaddr').Multiaddr} Multiaddr
+ * @typedef {import('peer-id')} PeerId
+ * @typedef {import('libp2p-interfaces/src/stream-muxer/types').MuxedStream} MuxedStream
+ * @typedef {(key: string) => Promise<Uint8Array | null>} LookupFunction
+ */
+
+/**
+ * Ping a given peer and wait for its response, getting the operation latency.
```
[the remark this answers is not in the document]
stbrody: Yep, copy-paste fail.  Thanks

stbrody:
Okay, updated to be more modular and allow registering multiple key lookup functions for different fixed prefixes of the keyspace.  Hopefully this addresses your main concern @aschmahmann

@achingbrain @vasco-santos

stbrody:
@aschmahmann, I had an idea that I think would make this a bit simpler to use and understand.  Instead of having a key prefix to dispatch request handlers, what if we added a new string field `namespace` to the fetch request protocol buffer message, alongside the existing `identifier` field, and registered handlers based on namespace?  Then we don't have to worry about collisions on the identifier so long as they are in different namespaces.  The one thing is that this would cause a bit of a divergence from what is currently implemented in go-libp2p, but I think it'd be a cleaner design longer term?  What do you think?

src/fetch/README.md
```diff
@@ -0,0 +1,38 @@
...
+The fetch protocol is a simple protocol for requesting a value corresponding to a key from a peer.
+
+## Usage
+
+```javascript
+var FetchProtocol = require('libp2p/src/fetch')
+
+/**
+ * Given a key (as a string) returns a value (as a Uint8Array), or null if the key isn't found.
+ * All keys must be prefixed my the same prefix, which will be used to find the appropriate key
```
achingbrain, line 19:
```suggestion
 * All keys must be prefixed by the same prefix, which will be used to find the appropriate key
```

test/fetch/fetch.spec.js
```diff
@@ -0,0 +1,115 @@
...
+    fetch.registerLookupFunction(PREFIX_A, generateLookupFunction(PREFIX_A, DATA_A))
+    fetch.mount(receiver)
+
+    try {
+      await Fetch.fetch(sender, receiver.peerId, '/moduleUNKNOWN/foobar')
+      expect.fail("didn't throw")
+    } catch (err) {
+      expect(err).to.be.an('Error')
+      expect(err.message).to.equal('Error in fetch protocol response: No lookup function registered for key: /moduleUNKNOWN/foobar')
+    }
```
achingbrain, line 105: Instead of testing for the error message, please can you test for the `.code` that should be present when using the `err-code` module (see other comments in this PR:

```suggestion
   await expect(Fetch.fetch(sender, receiver.peerId, '/moduleUNKNOWN/foobar')).to.eventually.be.rejected.with.property('code', 'ERR_SOMETHING_SOMETHING')
```

src/fetch/index.js
```diff
@@ -0,0 +1,149 @@
...
+    switch (response.status) {
+      case (FetchResponse.StatusCode.OK): {
+        return response.data
+      }
+      case (FetchResponse.StatusCode.NOT_FOUND): {
+        return null
+      }
+      case (FetchResponse.StatusCode.ERROR): {
+        const errmsg = (new TextDecoder()).decode(response.data)
+        throw new Error('Error in fetch protocol response: ' + errmsg)
```
achingbrain, line 65: Could you please use the `err-code` module to create `Error` objects with consistent `.code` properties as elsewhere in the codebase?

src/fetch/index.js
```diff
@@ -0,0 +1,149 @@
...
+      }
+      default: {
+        throw new Error('Unreachable case')
```
achingbrain, line 68: Please use the `err-code` module as above

src/fetch/index.js
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * Registers a new lookup callback that can map keys to values, for a given set of keys that
+   * share the same prefix.
+   *
+   * @param {string} prefix
+   * @param {LookupFunction} lookup
+   */
+  registerLookupFunction (prefix, lookup) {
+    if (this._lookupFunctions.has(prefix)) {
+      throw new Error("Fetch protocol handler for key prefix '" + prefix + "' already registered")
```
achingbrain, line 125: Please use the `err-code` module as above

src/identify/message.d.ts
achingbrain: Yes please!

src/identify/message.js
```diff
@@ -43,27 +43,27 @@ $root.Identify = (function() {
 
     /**
      * Identify protocolVersion.
-     * @member {string} protocolVersion
```
achingbrain, line 46: Please can you revert the changes to this file.

src/insecure/proto.js
```diff
@@ -36,11 +36,11 @@ $root.Exchange = (function() {
 
     /**
      * Exchange id.
-     * @member {Uint8Array} id
```
achingbrain, line 39: Please can you revert the changes to this file

src/peer-store/persistent/pb/address-book.d.ts
```diff
@@ -89,7 +89,10 @@ export namespace Addresses {
         public multiaddr: Uint8Array;
 
         /** Address isCertified. */
-        public isCertified: boolean;
```
achingbrain, line 92: Please can you revert the changes to this file

src/peer-store/persistent/pb/address-book.js
```diff
@@ -213,11 +213,25 @@ $root.Addresses = (function() {
 
         /**
          * Address isCertified.
-         * @member {boolean} isCertified
```
achingbrain, line 216: Please can you revert the changes to this file

test/fetch/fetch.spec.js
```diff
@@ -0,0 +1,115 @@
...
+  })
+
+  it('Registering multiple handlers for same prefix errors', async () => {
+    const fetch = new Fetch()
+    fetch.registerLookupFunction(PREFIX_A, generateLookupFunction(PREFIX_A, DATA_A))
+    expect(function () {
+      fetch.registerLookupFunction(PREFIX_A, generateLookupFunction(PREFIX_A, DATA_B))
+    }).to.throw('already registered')
```
achingbrain, line 113: Same comment here, it's better to test for a thrown value with `.code` property so we can change the error messages without breaking the tests in future.

achingbrain (changes requested):
This is great, thanks so much for submitting it.  Just a few minor comments for consistency.

The implementation needs to be mounted in https://github.com/libp2p/js-libp2p/blob/master/src/index.js and exposed on the `libp2p` class to let consumers configure lookup functions - see how `.indentifyService` and the ping protocol are configured for an example.

The generated `.js` file also needs to be added to the exclude list in https://github.com/libp2p/js-libp2p/blob/master/tsconfig.json - this plus reverting the changes to the generated files should get linting passing in CI.

Finally please merge master/rebase as it's had some changes made that make the build a lot more stable.

Ideally some tests should be added to https://github.com/libp2p/interop to ensure it works across implementations but I don't know how feasible that is as it uses https://github.com/libp2p/go-libp2p-daemon which is a bit out of date.

src/identify/message.d.ts
stbrody: Yep will do.  I'm still worried though that the code generated for the new fetch protocol protobufs is going to have the same problems as clearly there's something in my local setup causing it to generate protobufs differently than what yall have here.

test/fetch/fetch.spec.js
```diff
@@ -0,0 +1,118 @@
...
+
+    await expect(Fetch.fetch(sender, receiver.peerId, '/moduleUNKNOWN/foobar')).to.eventually.be.rejected.with.property('code', codes.ERR_INVALID_PARAMETERS)
+  })
+
+  it('Registering multiple handlers for same prefix errors', async () => {
+    const fetch = new Fetch()
+    fetch.registerLookupFunction(PREFIX_A, generateLookupFunction(PREFIX_A, DATA_A))
+
+    try {
+      fetch.registerLookupFunction(PREFIX_A, generateLookupFunction(PREFIX_A, DATA_B))
```
stbrody, line 111: I could not figure out how to use a single-line `expect` statement to expect the function to throw an error with a specific code field, so unfortunately I had to use this kind of ugly try/catch

stbrody (commented):
Okay, ready for another round of review.  I reverted the auto-generated protobuf code, merged in master, and added error codes.  PTAL

test/fetch/fetch.spec.js
achingbrain:
```js
expect(() => {
  ... invoke method that throws here
}).to.throw().with.property('code', codes.ERR_KEY_ALREADY_EXISTS)
```

achingbrain (approved):
Thanks so much for adding this.  I've tided some things up to make it more consistent with the existing code, added some extra tests and docs, and exposed the `.fetch` command as a top-level libp2p method.

I've also changed the protocol name to be consistent with go-libp2p, otherwise we won't have interop.

achingbrain:
@stbrody this is now available in libp2p under the `next` tag - could you please try it out and report back any observed issues?

```console
$ npm install libp2p@next
```

stbrody:
All changes look great, thanks @achingbrain!!!  I'm excited to try this out.  Hopefully I'll have time to check it out in the upcoming 2 week sprint.

I'm still curious though what you and @aschmahmann think about the idea of using an explicit `namespace` instead of the 'key prefix' concept that is how it's implemented today.  Doing string parsing always makes me feel gross - I think it'd be much cleaner and easier to understand and reason about with an explicit 'namespace' instead.

PR #1039, opened 2021-11-26
fix: do not auto-dial after shut down

achingbrain:
The `await` in the auto-connect loop means we can start a dial after shutdown but before we test to see if we should break out of the loop.

Instead, use the `this._started` property to break out of the loop, then test again to see if we should use retimer to schedule a further auto-dial attempt.

Fixes: https://github.com/ipfs/js-ipfs/issues/3923


achingbrain:
Superseded by 1041

PR #1041, opened 2021-11-30
fix: stop dht before connection manager

achingbrain:
Stop the dht before the connection manager, otherwise in-flight eviction pings fail and we move on to the next one when we should just abort them all.

Also pulls in the fix from #1039 and splits the auto-dialler out from the connection manager as during shutdown it can get into a weird state where it's simultaneously killing and creating connections so stop auto-dialling things before we cause connections to dip below the low watermark by killing existing connections.

Fixes: https://github.com/ipfs/js-ipfs/issues/3923

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,7 @@ const { codes, messages } = require('./errors')
 
 const AddressManager = require('./address-manager')
 const ConnectionManager = require('./connection-manager')
+const AutoDialler = require('./connection-manager/auto-dialler')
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -380,6 +385,8 @@ class Libp2p extends EventEmitter {
 
       this.relay && this.relay.stop()
       this.peerRouting.stop()
+      this._autodialler.stop()
```
vasco-santos, line 388: why not part of the connection manager start/stop?

PR #1044, opened 2021-12-03
fix: fix uncaught promise rejection when finding peers

robertkiel:
## Quick description

Whenever `dht.findPeer()` is unable to find any peer or fails for other reasons, it produces an unhandled promise rejection.

## Long description

In `peerRouting.js`, the result of many routers is merged into one array by using multiple smaller arrays, using a function that returns an array.

```
router => [router.findPeer(ip, options)]
```

The problem is that this function creates the array without awaiting the result of `router.findPeer()`, hence if `router.findPeer()` throws, no one catches the error because no one awaits the result.

## Fix

Return an async iterable that first awaits the result and catches the error and then return the result.

```
        ...this._routers.map(router => (async function* () {
          try {
            yield [await router.findPeer(id, options)];
          } catch (err) {
            yield;
          }
        })())
```

Makes https://github.com/libp2p/js-libp2p/pull/1025 obsolete


achingbrain:
Thanks for submitting this - CI is failing because of linting, you can fix it with `npm run lint -- --fix`.

Thanks also for adding a test to this PR - it passes without the proposed changes to `peer-routing.js` though (sort of, [this error](https://github.com/libp2p/js-libp2p/blob/master/src/peer-routing.js#L128) should have the code `ERR_NOT_FOUND`, not `NOT_FOUND` that's a bug), so I'm not sure the problem is fixed by this patch.

Looking at the code, it seems you're trying to ignore errors thrown by `router.findPeer(id, options)` - if they are thrown synchronously they should be handled by whatever is awaiting on `peerRouting.findPeer()`.  If they are thrown asynchronously, `it-pipe` should bubble them back up to the same context and again they'll be handled by whatever is awaiting on `peerRouting.findPeer`.

Since you're seeing an `unhandledPromiseRejection` something is not `await`ing when it should be - can you please share a stack trace from the UPR?  Something like this will do:

```js
process.on('unhandledPromiseRejection', (err) => {
  console.error(err)
})
```

robertkiel:
This is the sample code that I used to check the functionality, partially copied from `js-libp2p`

```
const pipe = require('it-pipe')
const merge = require('it-merge')
const filter = require('it-filter')
const first = require('it-first')

async function findPeer () {
  throw Error('foo')
}

async function main () {
  try {
    const output = await pipe(
      merge(
        ...[undefined].map(router =>
          (async function * () {
            try {
              yield [await findPeer()]
            } catch (err) {
              yield undefined
            }
          })()
        )
      ),
      (source) => filter(source, Boolean),
      // @ts-ignore findPeer resolves a Promise
      // (source) => storeAddresses(source, this._peerStore),
      (source) => first(source)
    )

    console.log(output)
  } catch (err) {
    console.log('caught error', err)
  }
}

main()
```

Will produce a warning on Node.js < 15 and an uncaught promise rejection on Node.js 15+ that terminates the process if you don't handle it.

robertkiel:
> `// @ts-ignore findPeer resolves a Promise`

It seems that Typescript has noticed that earlier ;-)

achingbrain (approved):

PR #1047, opened 2021-12-03
fix: do not let closest peers run forever

achingbrain:
The DHT takes a `signal` not a timeout so if a timeout is passed,
create a `TimeoutController` that will abort the query after the
timeout.


achingbrain:
I think longer term we need to sweep the codebase and rationalise all this stuff.  Drop timeouts and switch everything to `AbortSignal`/`AbortController`s.

vasco-santos (approved):
LGTM
Good catch

PR #1048, opened 2021-12-03
Libp2p.pubsub.subscribe returns void, remove await

GGAlanSmithee:
Seems like the correct return type of `Libp2p.pubsub.subscribe` is `void`, so the `await` can be removed: https://github.com/libp2p/js-libp2p/blob/ae21299ade9bcb2cf3fd00150c4389fbe3b24267/src/pubsub-adapter.js#L29


vasco-santos (approved):
LGTM thanks

PR #1051, opened 2021-12-06
fix: remove peer routing search-for-self

achingbrain:
The peer routing module starts a recurring process that searches for
peers close to our peer id.

This makes the DHT module query the network for peers.  Thing is the
DHT module is already doing this because periodically searching for
peers close to us is in the DHT spec so this ends up making redundant
queries.

This PR removes the recurring task configured by the peer routing module.

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -415,14 +415,7 @@ const node = await Libp2p.create({
     contentRouting: [delegatedContentRouting],
     peerRouting: [delegatedPeerRouting],
   },
-  peerId,
-  peerRouting: { // Peer routing configuration
-    refreshManager: { // Refresh known and connected closest peers
-      enabled: true, // Should find the closest peers.
-      interval: 6e5, // Interval for getting the new for closest peers of 10min
-      bootDelay: 10e3 // Delay for the initial query for closest peers
-    }
-  }
+  peerId
 })
 ```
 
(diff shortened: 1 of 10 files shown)


vasco-santos (commented):
What about a libp2p configuration with the delegates only?

achingbrain:
Why would the delegates need to query for themselves?  They'd be doing that in the background to keep their routing tables up to date anyway, right?

vasco-santos:
We use this to find and add peers to our PeerStore (and potentially connect with them), as part of the bootstrap process. These can include relevant peers like relays to use as auto relay nodes, as well as peers to fill in the pubsub mesh.

vasco-santos:
More information on the reasoning in https://github.com/libp2p/js-libp2p/issues/704

achingbrain:
So I guess we either

a - exclude the DHT from the peer routing search-for-self timer
b - disable search for self in the DHT by default
c - make the dht delegate module support the same semantics as the dht spec and query for it's own peer id occasionally

a) breaks the `PeerRoutingModule` abstraction while b) seems like we are managing ambient peer discovery at the wrong level.  c) adds another timer.

achingbrain:
You could even merge `libp2p-delegated-content-routing` and `libp2p-delegated-peer-routing` to create `libp2p-delegated-dht`?

achingbrain:
Option a: https://github.com/libp2p/js-libp2p/pull/1055

PR #1057, opened 2021-12-07
chore: run node tests in ci

achingbrain:
Looks like this project stopped running the `test:node` npm script when it was migrated to gh actions.

Re-enable it and fix all the related test failures.


vasco-santos (approved):
Changes look good, but don't understand the node tests not being running.

I just confirmed they have been running, including this week: https://github.com/libp2p/js-libp2p/runs/4435589812?check_suite_focus=true

achingbrain:
`npx aegir test -t node` does not run the same test files as `npm run test:node`

vasco-santos:
ah, because of the `.spec` ones. Makes sense

PR #1058, opened 2021-12-09
feat: async peerstore backed by datastores

achingbrain:
We have a peerstore that keeps all data for all observed peers in
memory with no eviction.

This is fine when you don't discover many peers but when using the
DHT you encounter a significant number of peers so our peer storage
grows and grows over time.

We have a persistent peer store, but it just periodically writes
peers into the datastore to be read at startup, still keeping them
in memory.

It also means a restart doesn't give you any temporary reprieve from
the memory leak as the previously observed peer data is read into
memory at startup.

This change refactors the peerstore to use a datastore by default,
reading and writing peer info as it arrives.  It can be configured
with a MemoryDatastore if desired.

It was necessary to change the peerstore and *book interfaces to
be asynchronous since the datastore api is asynchronous.

Depends on:

- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/128
- [x] https://github.com/libp2p/js-libp2p-kad-dht/pull/272
- [x] https://github.com/libp2p/js-libp2p-floodsub/pull/124
- [ ] https://github.com/ChainSafe/js-libp2p-gossipsub/pull/182
- [x] https://github.com/libp2p/js-libp2p-daemon/pull/62
- [x] https://github.com/libp2p/js-libp2p-daemon-client/pull/110
- [x] https://github.com/libp2p/interop/pull/55


achingbrain:
Heap usage after running this branch for 24 hours, just idling with the dht self-query running, seems healthier:

image

PR #1059, opened 2021-12-10
fix: clean up pending dial targets

achingbrain:
If the `Promise.race` throws, execution of the function is terminated so the pending dial target is never removed from the map and we leak memory.

This can happen when there are invalid multiaddrs or when a peer reports more dialable addresses than the threshold.

Instead wrap the `Promise.race` in a `try/finally` which will always remove the pending dial target in the event of success or failure.

### DIFF
--- a/src/dialer/index.js
+++ b/src/dialer/index.js
@@ -155,14 +155,16 @@ class Dialer {
       this._pendingDialTargets.set(id, { resolve, reject })
     })
 
-    const dialTarget = await Promise.race([
-      this._createDialTarget(peer),
-      cancellablePromise
-    ])
+    try {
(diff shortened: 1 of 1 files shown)


vasco-santos (approved):

achingbrain:
A graph, running IPFS using libp2p master vs this PR.  Leave it running for an hour, just idling with the DHT enabled running it's query-self every five minutes.

Blue and red lines are heap use in MB.  Blue is master, red is this PR.

image

achingbrain:
Refs https://github.com/ipfs/js-ipfs/issues/3469

vasco-santos:
wow, good job @achingbrain

Issue #1060, opened 2021-12-11
Per-component metrics


achingbrain:
I'd like to get some insight into what's going on inside a libp2p node.  At the moment the metrics are a bit limited, only focussing on bandwidth stats for peers and protocols and few other bits, from what I can see.

It would be really useful to be able to get component-specific metrics to really see what's going on.

How about something like:

```js
libp2p.metrics.updateComponentMetric(component: string, metric: string, value: number)
```

This would set per-component metrics, eg:

```js
libp2p.metrics.updateComponentMetric('kad-dht', 'active-queries', 1)
libp2p.metrics.updateComponentMetric('dialler', 'pending-dials', 10)
// etc
```

Then getting metrics out:

```js
libp2p.metrics.getComponentMetrics() => Map<string, Map<string, number>
```

We'd then report these metrics in the `/debug/metrics/prometheus` endpoint in `ipfs-http-api-server` which would allow graphing these stats over time when `IPFS_MONITORING` is set in the environment similar to how we do with the [number of connected peers](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-http-server/src/api/routes/debug.js#L6).

I don't think we need moving averages, or anything fancy like that - the tool we use to examine the stats can figure all of that out.

vasco-santos:
I think this is super valuable! For further iteration, we should look into getting this closer to what we want to have with https://github.com/libp2p/observer-toolkit/tree/master/packages/proto

On previous hack week, I did a small POC in the daemon https://github.com/libp2p/js-libp2p-daemon/pull/45 and I think this will likely be something we want to use in the daemon. But, there are useful things that we should extract on the libp2p repo that would make easier to consume things in IPFS and the observer toolkit

PR #1061, opened 2021-12-13
feat: allow per-component metrics to be collected

achingbrain:
Implements the idea from #1060 - allows us to get some insight into what's happening in a libp2p node out side of just bandwidth stats.

Configures a few default metrics if metrics are enabled - current connections, the state of the dial queue, etc.

Also makes the `Metrics` class not depend on the `ConnectionManager` class, otherwise we can't collect simple metrics from the connection manager class due to the circular dependency.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -197,10 +197,16 @@ class Libp2p extends EventEmitter {
 
     // Create Metrics
     if (this._options.metrics.enabled) {
-      this.metrics = new Metrics({
-        ...this._options.metrics,
-        connectionManager: this.connectionManager
+      const metrics = new Metrics({
+        ...this._options.metrics
       })
+
+      // listen for peer disconnect events
+      this.connectionManager.on('peer:disconnect', (connection) => {
+        metrics.onPeerDisconnected(connection.remotePeer)
+      })
+
+      this.metrics = metrics
     }
 
     // Create keychain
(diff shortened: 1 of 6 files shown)


achingbrain:
CI is broken due to https://github.com/paulmillr/noble-secp256k1/issues/38

src/index.js
```diff
@@ -197,10 +197,16 @@ class Libp2p extends EventEmitter {
...
-      this.metrics = new Metrics({
-        ...this._options.metrics,
-        connectionManager: this.connectionManager
+      const metrics = new Metrics({
+        ...this._options.metrics
       })
+
+      // listen for peer disconnect events
+      this.connectionManager.on('peer:disconnect', (connection) => {
+        metrics.onPeerDisconnected(connection.remotePeer)
```
vasco-santos, line 206: shouldn't be the connection manager to trigger a metrics change instead of an explicit `onPeerDisconnected` method?

achingbrain: There's quite a bit of logic in the `onPeerDisconnected` method, it'd be better to keep it out of the connection manager, if that's what you mean.

That said I'd rather there was a generic event bus that the metrics collector could listen to, then it could listen for `peer:*` events and count connections without the connection manager knowing anything about metrics at all.

PR #1065, opened 2021-12-17
fix: increase the maxlisteners for timeout controllers

achingbrain:
We use timeout controllers to ensure we're not dialling peers forever but we can end up registering lots of listeners for the `abort` event when peers have a lot of addresses.

In node this means we see an unhelpful `MaxListenersExceededWarning` in the console warning of a potential memory leak.

Increase the max number of listeners on the signal to silence the warning.


vasco-santos (commented):
It looks like ts is not liking it

achingbrain:
@vasco-santos looks healthier now

vasco-santos (approved):

PR #1068, opened 2021-12-20
Fetch all events from findPeer call

robertkiel:
The DHT issues responses using an AsyncIterator that "emits" various events, among the "FINAL_PEER" and "PEER_RESPONSE"

Previously, the function swallowed all results that came with "FINAL_PEER", hence more than necessary calls ended with `not found`. This PR merges incoming events and forwards them to the caller.


src/dht/dht-peer-routing.js
```diff
@@ -26,14 +27,53 @@ class DHTPeerRouting {
...
-      if (event.name === 'PEER_RESPONSE') {
-        const peer = event.closer.find(peerData => peerData.id.equals(peerId))
+      let multiaddrs 
+      switch (event.name) {
+        case 'FINAL_PEER':
+          peerFound = true
+          multiaddrs = event.peer.multiaddrs
+          break
+        case 'PEER_RESPONSE':
+          peerFound = true
```
achingbrain, line 44: `PEER_RESPONSE` just means a network peer responded to a query, it doesn't necessarily mean the peer we're looking for will be in the `closer` field of the response - it could just contain peers that are closer to the peer we are looking for.

src/dht/dht-peer-routing.js
```diff
@@ -26,14 +27,53 @@ class DHTPeerRouting {
    * @param {any} options
    */
   async findPeer (peerId, options = {}) {
```
achingbrain, line 29: I think this function should just wait for the `FINAL_PEER` event, then load the multiaddrs from the address book in the peer store and return a `PeerData` object made from those.

The gotcha is that go-kad-dht emits `FINAL_PEER` events without multiaddrs which is why this function looks for `PEER_RESPONSE` events instead in an effort to follow the same logic.

js-kad-dht emits `peer` events [when it discovers closer or provider peers](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/network.js#L190-L196) - those events are used to populate the address book, so if we listen for `FINAL_PEER`, then use the address book to load the multiaddrs we shouldn't need to juggle both sets of multiaddrs here.

achingbrain (changes requested):

src/dht/dht-peer-routing.js
robertkiel: Oh, in this case, there are some other unit tests wrong - because they don't assume the `FINAL_PEER` event

src/dht/dht-peer-routing.js
robertkiel: >  then load the multiaddrs from the address book in the peer store and return a PeerData object made from those.

I rather suggest to only return those addresses that the DHT request brought us to be able to get the difference.

> The gotcha is that go-kad-dht emits FINAL_PEER events without multiaddrs which is why this function looks for PEER_RESPONSE events instead in an effort to follow the same logic.

Ok, makes sense. `js-kad-dht` definitely seems to emit FINAL_PEER events *with* multiaddrs. And in these cases, the DHT were able to find addresses but `findPeer` still returns with a `NOT_FOUND` error. This is the main reason led me create the PR.

wemeetagain:
bump

PR #1069, opened 2021-12-21
fix: add tracked map

achingbrain:
Small refactor of the component stats - adds a `TrackedMap` which encapsulates updating the metrics and means we don't need to null guard on `this._metrics` everywhere.

If metrics are not enabled a regular `Map` is used.


vasco-santos (approved):
Nice 💯

PR #1072, opened 2021-12-21
fix: abort dial memory leak

achingbrain:
Protocol negotiation or encryption handshaking can hang indefinitely when dialing peers with unreliable network connections.

When this happens the entry in the `_pendingDials` map in `src/dialer/index.js` hangs around forever and the memory is never reclaimed.

Our timeout controller will fire the `abort` event on its signal to cause the dial to abort, the connection to be torn down, and the `_pendingDials` entry to be removed, but something has to be listening for that event so pass the AbortSignal in to mss and noise.

Depends on:

- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/122


achingbrain:
CI is failing intermittently due to:

```
npm ERR! Downloading: ipfs.io/ipfs/QmW6hANCx63LR2Unb8TWwB7UkDC2SNKUV5QbaWzSyjsoPf
npm ERR! Destination: /home/runner/work/js-libp2p/js-libp2p/node_modules/go-libp2p-dep/go-libp2p
npm ERR! Error: aborted
npm ERR! Download failed!
```

achingbrain:
This is the wrong fix - the maconn should be aborted at the source rather than each use of it.  We have a `TimeoutController` that is passed in to the connection that should abort it if the upgrade/crypto handshake isn't completed in time, but at the moment `abortable-iterator` doesn't reject when reads stall indefinitely - fix for that in https://github.com/alanshaw/abortable-iterator/pull/15

PR #1082, opened 2021-12-23
Use WhatWG TextEncoder standard for conversion

TimDaub:
`uint8ArrayToString` and `uint8ArrayFromString` don't exist, so I thought it may be friendlier for the user to suggest WhatWG TextDecoder and TextEncoder that is available on nodejs via standard lib and in many webbrowsers:

- https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder
- https://nodejs.org/api/util.html#class-utiltextencoder

### DIFF
--- a/examples/pubsub/README.md
+++ b/examples/pubsub/README.md
@@ -41,6 +41,8 @@ const node = await Libp2p.create({
 Once that is done, we only need to create a few libp2p nodes, connect them and everything is ready to start using pubsub.
 
 ```JavaScript
+const { TextDecoder, TextEncoder } = require("util");
+
 const topic = 'news'
 
 const node1 = nodes[0]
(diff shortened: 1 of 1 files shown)


examples/pubsub/README.md
```diff
@@ -41,6 +41,8 @@ const node = await Libp2p.create({
 Once that is done, we only need to create a few libp2p nodes, connect them and everything is ready to start using pubsub.
 
 ```JavaScript
+const { TextDecoder, TextEncoder } = require("util");
```
vasco-santos, line 44: Since node 11 it is available in the global scope like in browser, so no need to require it.

However, we use uint8arrays everywhere across the stack and it will likely be important for the users to use them in other scenarios. With this in mind, I think it is friendlier to just rely on them than to do `TextDecoder(TextDecoder.encoding).decode`. Thoughts?

TimDaub: Ah actually I just noticed that you guys are also the maintainer of this library: https://www.npmjs.com/package/uint8arrays

Then I guess it'd make more sense to require the library in the example

Issue #1110, opened 2022-01-09
go-js echo / plaintext fails to send message


gitaaron:
- **Version**: 0.35.8

- **Platform**: Darwin [REDACTED]-Mac-mini.local 20.5.0 Darwin Kernel Version 20.5.0: Sat May  8 05:10:31 PDT 2021; root:xnu-7195.121.3~9/RELEASE_ARM64_T8101 arm64

- **Subsystem**: Insecure

#### Description:
I tried to dial the [JS echo listener example](https://github.com/libp2p/js-libp2p/blob/master/examples/echo/src/listener.js) from the [Go echo dialer example](https://github.com/libp2p/go-libp2p/tree/master/examples/echo) and vice versa over plaintext and in both cases it failed during the handshake.

It seems that the Data property of the pubkey in Exchange is inconsistent where Go simply uses the marshalled value of the encrypted public key and JS includes the entire pubkey object re-encoded as a protobuf.

There also seems to be an issue with the samples themselves where the Go sample uses newline to signify the end of a message and JS does not.

#### Steps to reproduce the error:

1.  Change 'noise' in the js example to 'plaintext'.
2. In the JS sample folder run ```> node listener```
3. In the Go sample folder run ```> ./echo -l 9012 -d /ip4/127.0.0.1/tcp/10333/p2p/QmcrQZ6RJdpYuGvZqD5QEHAv6qX4BrQLJLQPQUrTrzdcgm -insecure```

lidel:
@gitaaron what is the problem / use case  you are trying to solve?

gitaaron:
hi @lidel thanks for looking at this!  The use case is to assist in development/debugging go-js interop issues.

Here is a little more context in case it is helpful.  I was looking at go-js webrtc interop which I believe is still an [issue](https://github.com/libp2p/js-libp2p-webrtc-direct/issues/127) tagged with help wanted.

To assist myself in debugging the problem I figured it might help to look at the traffic over wireshark.  This is when I discovered the go-js interop issue over plaintext (more specifically the discrepancy in the exchange object passed).

So I figured I might create a new issue and submit a PR with the fix in case someone else might also find it useful.

I am new to libp2p so your feedback/advice is appreciated.

wemeetagain:
linked to https://github.com/libp2p/js-libp2p/pull/1111

mpetrunic:
[...]

PR #1111, opened 2022-01-09
fixes go-js echo / plaintext fails to send message #1110

gitaaron:
Ref. https://github.com/libp2p/js-libp2p/issues/1110

fix(go-js-interop): Data property of pubkey in Exchange was encoded as entire pubkey protobuf object; 
instead it should just be the marshaled content of the pubkey

docs(example): include insecure cmd line option so that other clients (ie/ Go) can use plaintext

docs(example): follow newline convention for end of message to interop with other clients (eg/ Go)

docs(example): create PeerId from base58String for remote listener so that it can easily be changed for other clients (eg/ Go)

fix(test): plaintext only handles RSA PeerId encryption for now


lidel (changes requested):
@gitaaron  mind creating a separate plaintext example, and leave existing one as-is? :pray:

gitaaron:
@lidel - I sure can!

Just a caveat that I don't think the current sample will work with the go plaintext sample as is (unless something has changed on the Go side since I looked at it last)

gitaaron:
@lidel - I've created a new sample called 'echo-newline' and mentioned you can optionally pass in an  '-insecure' cmd line param in the README

BTW - I made it a cmd line param to be consistent with the GO echo sample

wemeetagain (approved):
LGTM thanks @gitaaron

BigLep:
2022-03-22 conversation: 
1. We need to add the example test to the list of .yaml file
2. Run CI to make sure it passes
3. Merge

@wemeetagain will take this - thank you!

gitaaron:
> 2022-03-22 conversation:
> 
> 1. We need to add the example test to the list of .yaml file
> 2. Run CI to make sure it passes
> 3. Merge
> 
> @wemeetagain will take this - thank you!

Thanks  @BigLep  and @wemeetagain  !

@BigLep - is 'adding the example test to the list of .yaml' something  I should have done before submitting the PR?

wemeetagain:
This also needs porting to typescript

gitaaron:
> This also needs porting to typescript

Thanks for the feedback - is there a typescript sample or test I can use as reference by any chance?

wemeetagain:
Whoops, I misspoke here, its not typescript thats required, its just using the latest version of libp2p.
The big change is that libp2p and libp2p dependencies are now only using ESM exports.
Here's an example that uses the latest libp2p: https://github.com/libp2p/js-libp2p/blob/master/examples/transports/3.js

gitaaron:
> Whoops, I misspoke here, its not typescript thats required, its just using the latest version of libp2p. The big change is that libp2p and libp2p dependencies are now only using ESM exports. Here's an example that uses the latest libp2p: https://github.com/libp2p/js-libp2p/blob/master/examples/transports/3.js

Sounds good, thanks!  I'll have a look at this.

gitaaron:
Hi Steve,

Sorry for the delay.  I was having trouble testing out the sample I created
with libp2p-noise but I think I narrowed it down to an older version.

Is it alright if I give you an update over the coming week?  Hopefully I
can have it fixed by then or let you know I am stuck.

Thanks,
Aaron

gitaaron:
@BigLep - just looking at this again - it seems that running the samples requires an `npm run build` to build libp2p otherwise I see the following error -
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/asurty/POpen/libp2p/workspace-js-libp2p/js-libp2p/examples/node_modules/libp2p/dist/src/index.js' imported from /Users/asurty/POpen/libp2p/workspace-js-libp2p/js-libp2p/examples/echo/src/libp2p.js
    at new NodeError (node:internal/errors:371:5)
    at finalizeResolution (node:internal/modules/esm/resolve:394:11)
    at moduleResolve (node:internal/modules/esm/resolve:944:10)
    at defaultResolve (node:internal/modules/esm/resolve:1041:11)
    at ESMLoader.resolve (node:internal/modules/esm/loader:530:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:251:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:79:40)
    at link (node:internal/modules/esm/module_job:78:36) {
  code: 'ERR_MODULE_NOT_FOUND'
 ```

If that is the case I can create  a new issue and submit a PR to update the README in the future?

Also  going to close this PR and  open a new one in the future with new commits since this has diverged significantly from master.

PR #1140, opened 2022-01-21
fix: try all peer addresses when dialing a relay

achingbrain:
The order of the addresses can affect our success rate in dialing a relay - if it's a loopback address or similar it won't work.

Instead try dialing every address.

Seems to be the cause of the random CI failures in https://github.com/libp2p/js-libp2p/pull/1136


wemeetagain (approved):
LGTM

Issue #175, opened 2018-03-07
[feature request]: Low-Level Connection Management


kumavis:
### idea:
some additional controls over establishing/denying connections
- hooks for denying connections before they are established (when possible)
- ConnectionManager configuration that can limit the number of connections per transport

### usecase:
MetaMask is building an Ethereum light client for the browser using libp2p.
We intend to use webrtc which is resource intensive and can only safely support 2-5 active connections.
We need to be able to deny incoming webrtc connection attempts after some threshold.

pgte:
@kumavis that's great feedback, thanks!
Opened these issues on [libp2p-connection-manager](https://github.com/libp2p/js-libp2p-connection-manager/issues).

diasdavid:
@kumavis if you have some spare cycles, test out the connection manager already released in libp2p 0.22 and that will also get released in js-ipfs 0.30 https://github.com/ipfs/js-ipfs/issues/1375

jacobheun:
We are aiming to slot connection gating into the 0.30 release.

PR #987, opened 2021-09-08
feat: connection gater

greenSnot:
https://github.com/libp2p/go-libp2p-core/blob/master/connmgr/gater.go

PR #1142, opened 2022-01-24
feat: connection gater

achingbrain:
Port of https://github.com/libp2p/go-libp2p-core/blob/master/connmgr/gater.go

Fixes: https://github.com/libp2p/js-libp2p/issues/175
Refs: https://github.com/libp2p/js-libp2p/issues/744
Refs: https://github.com/libp2p/js-libp2p/issues/769

Follows on from https://github.com/libp2p/js-libp2p/pull/987 - that PR does not allow maintainer commits so I've created a new one.


achingbrain:
I know this is a port of the related go-libp2p functionality and that's where the nomenclature comes from, but I would much rather test against positive action than negative action or ambiguous sentiment.

`interceptPeerDial` to me does not say "return false to allow dialling a peer", it says 'do something with the peer dial'.  Instead I feel something like `allowPeerDial` which returns `true` to allow dialling a peer would be much simpler to reason about.

So, all of these would return `true` to allow the operation to continue:

```
interceptPeerDial -> allowDialPeer
interceptAddrDial -> allowDialMultiaddr
interceptAccept -> acceptConnection
interceptSecured -> acceptSecuredConnection
interceptUpgraded -> acceptUpgradedConnection
```

I also see that we call `interceptAddrDial` before adding a multiaddr for a peer to the addressbook - this is not doing what the method says it is doing.  Suggest adding a `allowStoreMultiaddrForPeer` method or similar.

What do you think @wemeetagain @greenSnot ?

wemeetagain:
@achingbrain 

I agree with your renaming thoughts.

Same with `allowStoreMultiaddrForPeer`. Although, this one feels strange, like the current implementation is incomplete or should be rethought. The flag only currently has effect in a single place in the dialer, and the multiaddr can still be added via `peerStore.addressBook.add(peer, multiaddr)`, and this occurs in several other places.

achingbrain:
Good point, that filtering might be better done in the address book itself rather than just in the dialler.

wemeetagain (approved):
lgtm :+1:

greenSnot:
lgtm, hooray~

PR #1165, opened 2022-02-25
Protocol muxing docs explanation

aomini:
Protocol and stream muxing examples was missing explanation for example `3.js` (Bidirectional connection)


wemeetagain (approved):

mpetrunic (changes requested):
Should be rebased with the latest master. Examples should be written in typescript now.

wemeetagain (approved):

PR #1168, opened 2022-03-11
libp2p:connect: await storing AgentVersion to metadataBook

tuyennhv:
**Motivation**
+ Lodestar would like to query AgentVersion in "libp2p:connect" event handler

**Description**
+ Await "this.identify()" call when handling "libp2p:connect" event in IdentifyService

### DIFF
--- a/src/identify/index.js
+++ b/src/identify/index.js
@@ -78,8 +78,8 @@ class IdentifyService {
     }
 
     // When a new connection happens, trigger identify
-    this.connectionManager.on('peer:connect', (connection) => {
-      this.identify(connection).catch(log.error)
+    this.connectionManager.on('peer:connect', async (connection) => {
+      await this.identify(connection).catch(log.error)
     })
 
     // When self multiaddrs change, trigger identify-push


src/identify/index.js
```diff
@@ -78,8 +78,8 @@ class IdentifyService {
     }
 
     // When a new connection happens, trigger identify
-    this.connectionManager.on('peer:connect', (connection) => {
-      this.identify(connection).catch(log.error)
+    this.connectionManager.on('peer:connect', async (connection) => {
+      await this.identify(connection).catch(log.error)
```
achingbrain, line 82: EventEmitter handlers aren't allow to be async [according to the types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v16/events.d.ts#L330).

achingbrain (changes requested):

achingbrain:
Could you talk about the use case a bit more?

tuyennhv:
> Could you talk about the use case a bit more?
> 
> My feeling is there may be a better way - for example usually an application is interested in peers that talk certain protocols, in which case you'd use a topology because when you are notified of a peer joining the topology, identify has completed and their metadata is available in the peer store.

in lodestar, I want to get AgentVersion right after `peer:connect` event. If we can't make event handler async then I'll try other ways at app layer, for example `sleep(0)` or something like that. Thanks for looking into this @achingbrain

PR #1171, opened 2022-03-18
fix: emit peer:connect after all

tuyennhv:
**Motivation**

In lodestar, when we handle "peer:connect" event, we dial the peer which gives another "peer:connect" event and it causes other issues

**Motivation**

In `onConnect` function, "peer:connect" event should be emitted after we add connection to the `connections` map so that when app dial the peer in "peer:connect" event handler, it uses the same/existing connection

### DIFF
--- a/test/dialing/direct.spec.ts
+++ b/test/dialing/direct.spec.ts
@@ -29,6 +29,7 @@ import { createFromJSON } from '@libp2p/peer-id-factory'
 import Peers from '../fixtures/peers.js'
 import { MULTIADDRS_WEBSOCKETS } from '../fixtures/browser.js'
 import type { PeerId } from '@libp2p/interfaces/peer-id'
+import { pEvent } from 'p-event'
 
 const unsupportedAddr = new Multiaddr('/ip4/127.0.0.1/tcp/9999')
 
@@ -430,19 +431,15 @@ describe('libp2p.dialer (direct, WebSockets)', () => {
 
     const identifySpy = sinon.spy(libp2p.identifyService, 'identify')
     const protobookSetSpy = sinon.spy(libp2p.components.getPeerStore().protoBook, 'set')
-    const connectionPromise = pDefer()
+    const connectionPromise = pEvent(libp2p.connectionManager, 'peer:connect')
 
     await libp2p.start()
-
-    libp2p.connectionManager.addEventListener('peer:connect', () => {
-      connectionPromise.resolve()
-    }, { once: true })
 
     const connection = await libp2p.dial(MULTIADDRS_WEBSOCKETS[0])
     expect(connection).to.exist()
 
     // Wait for connection event to be emitted
-    await connectionPromise.promise
+    await connectionPromise
 
     expect(identifySpy.callCount).to.equal(1)
     await identifySpy.firstCall.returnValue
--- a/test/dialing/direct.spec.ts
+++ b/test/dialing/direct.spec.ts
@@ -434,11 +434,9 @@ describe('libp2p.dialer (direct, WebSockets)', () => {
 
     await libp2p.start()
 
-    libp2p.components.getUpgrader().addEventListener('connection', () => {
+    libp2p.connectionManager.addEventListener('peer:connect', () => {
       connectionPromise.resolve()
-    }, {
-      once: true
-    })
+    }, { once: true })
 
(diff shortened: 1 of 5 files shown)


wemeetagain (approved):
LGTM

achingbrain:
This seems fine to me, and a better way to do it since libp2p should make sure it's internal tracking of connections/addresses/etc is done before informing the outside world a new peer has connected to us.

I'm a little confused though:

> In lodestar, when we handle "peer:connect" event, we dial the peer

`peer:connect` occurs when a peer has connected to us - there's no need to dial them since we already have a connection.

Would you not just open a protocol stream to the peer instead?

That aside, CI appears to be failing due to this change - could you please try running the tests locally and ensure they pass?

tuyennhv:
> peer:connect occurs when a new connection to a peer has been created, e.g. either when we dial them or when they dial us - there's no need to dial them again in another 'peer:connect' handler since we already have a connection.

> Would you not just open a protocol stream to the peer instead?

what we did is to issue a p2p request in lodestar, which does a `dialProtocol()` behind the scene in libp2p. Since the connection is not added, it'd try to open a new one, all of this logic happen in libp2p

> That aside, CI appears to be failing due to this change - could you please try running the tests locally and ensure they pass?

thanks @achingbrain , this is fixed

test/identify/index.spec.ts
```diff
@@ -526,9 +527,15 @@ describe('Identify', () => {
...
-
+      const connectionPromise = pDefer()
+      libp2p.connectionManager.addEventListener('peer:connect', () => {
+        connectionPromise.resolve()
+      }, { once: true })
       const connection = await libp2p.dial(remoteAddr)
+
       expect(connection).to.exist()
+      // Wait for connection event to be emitted
+      await connectionPromise.promise
```
achingbrain, line 538: Please use `pEvent` instead of `pDefer` here, it's more concise and better reflects the intention behind the code.

test/identify/index.spec.ts
```diff
@@ -590,9 +597,15 @@ describe('Identify', () => {
...
```
achingbrain, line 608: Please use `pEvent` instead of `pDefer` here, it's more concise and better reflects the intention behind the code.

test/upgrading/upgrader.spec.ts
```diff
@@ -499,6 +499,13 @@ describe('libp2p.upgrader', () => {
       libp2p.components.getUpgrader().upgradeOutbound(outbound),
       remoteLibp2p.components.getUpgrader().upgradeInbound(inbound)
     ])
+    await new Promise<void>((resolve, reject) => {
```
achingbrain, line 502: Please use `pEvent` instead of the hand-rolled promise here, it's more concise and better reflects the intention behind the code.

test/dialing/direct.spec.ts
```diff
@@ -434,11 +434,9 @@ describe('libp2p.dialer (direct, WebSockets)', () => {
 
     await libp2p.start()
 
-    libp2p.components.getUpgrader().addEventListener('connection', () => {
+    libp2p.connectionManager.addEventListener('peer:connect', () => {
```
achingbrain, line 437: Please use `pEvent` here, it's more concise and better reflects the intention behind the code.

achingbrain (changes requested):
LGTM, just needs some small improvements to the tests

tuyennhv:
thanks @achingbrain, this `pEvent` is super helpful! Could you help with the failed CI check? It's not related to the fix so I have no ideas.

achingbrain:
The error message is in the CI output:

```
[08:36:25] dependency-check (production only) [failed]
Command failed with exit code 1: dependency-check package.json src/**/*.js src/**/*.cjs dist/src/**/*.js utils/**/*.js utils/**/*.cjs --missing --no-dev -i @types/* -i @types/* --extensions cjs:detective-cjs --extensions js:detective-es6
Fail! Dependencies not listed in package.json: @libp2p/tracked-map, it-pair
```

Basically `aegir` has got better at spotting unused/missing deps.  `@libp2p/tracked-map` and `it-pair` are production deps that are erroneously declared as development deps, they just need moving from the `"devDependencies"` block to the the "dependencies" block in the project's `package.json` file.

achingbrain (approved):

achingbrain:
The fix for the final failing build job was merged as part of #1194 - this is good to go.

PR #1172, opened 2022-03-18
feat: convert to typescript

achingbrain:
Converts this module to typescript.  Still a few things to do:

- [ ] Fill out this PR description with list of changes
- [ ] Examples passing
- [ ] Interop passing

BREAKING CHANGE: types are no longer hand crafted, this module is now ESM only


BigLep:
2022-03-22 conversation for someone reviewing this code:
1. Check it out 
2. See if can integrate it with another project.

.github/workflows/main.yml
```diff
@@ -5,40 +5,23 @@ on:
       - master
   pull_request:
     branches:
-      - '**'
+      - master
```
mpetrunic, line 8: isn't ci suppose to run even if you open PR against some old branch. For example if backporting fix to older version?

package.json
```diff
@@ -2,126 +2,228 @@
...
+    "parserOptions": {
+      "sourceType": "module"
+    },
     "ignorePatterns": [
       "!.aegir.js",
       "test/ts-use",
       "*.d.ts"
     ]
   },
+  "release": {
```
mpetrunic, line 75: Seems like libp2p repo doesn't use semantic release so this is useless

package.json
```diff
@@ -2,126 +2,228 @@
...
+    "pretest": "npm run build",
+    "test": "aegir test",
+    "test:node": "npm run test -- -t node -f \"./dist/test/**/*.{node,spec}.js\" --cov",
+    "test:chrome": "npm run test -- -t browser -f \"./dist/test/**/*.spec.js\" --cov",
+    "test:chrome-webworker": "npm run test -- -t webworker -f \"./dist/test/**/*.spec.js\"",
+    "test:firefox": "npm run test -- -t browser -f \"./dist/test/**/*.spec.js\" -- --browser firefox",
+    "test:firefox-webworker": "npm run test -- -t webworker -f \"./dist/test/**/*.spec.js\" -- --browser firefox",
+    "test:examples": "cd examples && npm run test:all",
+    "test:interop": "npm run test -- -t node -f dist/test/interop.js",
+    "release": "semantic-release"
```
mpetrunic, line 178: don't think we have dependency and it seems like its not used in CD

src/circuit/auto-relay.ts
```diff
@@ -0,0 +1,284 @@
...
+    this.components.getConnectionManager().addEventListener('peer:disconnect', this._onPeerDisconnected)
+  }
+
+  /**
+   * Check if a peer supports the relay protocol.
+   * If the protocol is not supported, check if it was supported before and remove it as a listen relay.
+   * If the protocol is supported, check if the peer supports **HOP** and add it as a listener if
+   * inside the threshold.
+   */
+  async _onProtocolChange (evt: CustomEvent<PeerProtocolsChangeData>) {
```
mpetrunic, line 63: We don't use private because of tests or is it too much effort to change them to private?

src/config.ts
```diff
@@ -0,0 +1,101 @@
...
+import * as RelayConstants from './circuit/constants.js'
+import { publicAddressesFirst } from '@libp2p/utils/address-sort'
+import { FAULT_TOLERANCE } from './transport-manager.js'
+import type { Multiaddr } from '@multiformats/multiaddr'
+import type { Libp2pInit } from './index.js'
+import { codes, messages } from './errors.js'
+import errCode from 'err-code'
+import type { RecursivePartial } from '@libp2p/interfaces'
+
+const DefaultConfig: Partial<Libp2pInit> = {
```
mpetrunic, line 14: How come you put Partial here? 

Shouldn't we enforce that whoever adds a new module must put defaults here? Maybe even create some type which would make everything mandatory here 😄

src/errors.ts
```diff
@@ -1,17 +1,19 @@
...
   NOT_STARTED_YET: 'The libp2p node is not started yet',
   DHT_DISABLED: 'DHT is not available',
   CONN_ENCRYPTION_REQUIRED: 'At least one connection encryption module is required',
+  ERR_TRANSPORTS_REQUIRED: 'At least one transport module is required',
+  ERR_PROTECTOR_REQUIRED: 'Private network is enforced, but no protector was provided',
   NOT_FOUND: 'Not found'
 }
 
-exports.codes = {
+export const codes = {
```
mpetrunic, line 10: enum?

achingbrain: Enums end up as numbers at runtime, right?  the use case is:

```js
throw errCode(new Error(messages.FOO), codes.FOO)

// later
if (err.code === codes.FOO) {
  ...
}
```

If someone decided to re-order the enum (alphabetical, add a new code, whatever) this could cause very undesirable side effects in code that did (rightly or wrongly):

```js
if (err.code === 'FOO') {
  ...
}
```

achingbrain: Non-ts code, I mean.  Maybe I'm worrying about nothing.

mpetrunic: Hmm we can use string enums so they don't end up being numbers and have same functionality as we have now: https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

And they shouldn't be replaced by value unless you add const: https://www.typescriptlang.org/docs/handbook/enums.html#const-enums

.github/workflows/main.yml
achingbrain: Yes, I'll revert this change.  Probably worth opening a PR to fix this in the automated CI flows too, [the line is here](https://github.com/protocol/.github/blob/master/templates/.github/workflows/js-test-and-release.yml#L8).

package.json
achingbrain: Good spot - it was added by [check-aegir-project](https://www.npmjs.com/package/check-aegir-project), I've updated it to only do that if there's an npm script called `"release"` with the contents `"semantic-release"`

package.json
achingbrain: The dep is added by aegir but you are quite right it's not used.

src/circuit/auto-relay.ts
achingbrain: The external API for the components are largely defined by interfaces, but it's useful to be able to call internal methods in the tests, but yes, this could change to be private.

I tried out [hash-prefixed private](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) in the [EventEmitter](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/libp2p-interfaces/src/index.ts#L77) class - it works but causes a warning to be logged to the console during the tests, kind of want to experiment with that a bit more before rolling it out everywhere.

In the interim my hope is the interface-as-public api should stop people calling private(ish) methods.

src/config.ts
achingbrain: Without `Partial` is becomes necessary to define things things like `.peerId` and `.datastore` that aren't simple values.

src/errors.ts
achingbrain: Ah, nice - I've updated it.

src/config.ts
mpetrunic:
```suggestion
const DefaultConfig: Required<Omit<Libp2pInit, "peer" | "datastore">> & Libp2pInit = {
```
would probably do the trick but I don't have a strong preference, it might end up being too hard to maintain what to keep optional.

mpetrunic (approved):

src/config.ts
achingbrain: Yeah, it starts getting a bit fiddly like this, maybe let's look at it again later.

doc/CONFIGURATION.md
```diff
@@ -235,189 +235,177 @@ Besides the `modules` and `config`, libp2p allows other internal options and con
...
+import { WebSockets } from '@libp2p/websockets'
+import { Mplex } from '@libp2p/mplex'
+import { Noise } from '@chainsafe/libp2p-noise'
+import { MulticastDNS } from '@libp2p/mdns'
+import { KadDHT } from '@libp2p/kad-dht'
+import { GossipSub } from 'libp2p-gossipsub'
+
+const node = await createLibp2p({
+  transports: [
+    TCP,
```
wemeetagain, line 249: I think this doesn't work anymore, same below where classes are used instead of instances.

doc/CONFIGURATION.md
```diff
@@ -719,19 +693,17 @@ const node = await Libp2p.create({
...
-    transport: [TCP],
-    streamMuxer: [MPLEX],
-    connEncryption: [NOISE]
-  },
+import { createLibp2p } from 'libp2p'
+import { TCP } from '@libp2p/tcp'
+import { Mplex } from '@libp2p/mplex'
+import { Noise } from '@chainsafe/libp2p-noise'
+
+const { FaultTolerance } from 'libp2p/src/transport-manager')
```
wemeetagain, line 701: invalid syntax

doc/STREAMING_ITERABLES.md
```diff
@@ -22,8 +22,8 @@
 Sometimes you may need to wrap an existing duplex stream in order to perform incoming and outgoing [transforms](#transform) on data. This type of wrapping is commonly used in stream encryption/decryption. Using [it-pair][it-pair] and [it-pipe][it-pipe], we can do this rather easily, given an existing [duplex iterable](#duplex).
 
 ```js
-const duplexPair = require('it-pair/duplex')
-const pipe = require('it-pipe')
+const duplexPair from 'it-pair/duplex')
```
wemeetagain, line 25: syntax error

doc/migrations/v0.26-v0.27.md
```diff
@@ -80,7 +82,7 @@ libp2p.dialProtocol(peerInfo, '/echo/1.0.0', (err, conn) => {
 
 **After**
 ```js
-const pipe = require('it-pipe')
+const pipe from 'it-pipe')
```
wemeetagain, line 85: syntax

wemeetagain (commented):
Update LGTM. Some small comments regarding code snippets in docs. We should consider adding something like https://github.com/bbc/typescript-docs-verifier to the CI pipeline.

achingbrain:
@wemeetagain thanks for your review - I've addressed the points in #1182

[bbc/typescript-docs-verifier](https://github.com/bbc/typescript-docs-verifier) - this looks really useful, I didn't know it existed.  Would you like to give adding it to CI a go?

wemeetagain:
Ok, I'll open a PR

PR #1174, opened 2022-03-22
fix: Behavior of LIBP2P_FORCE_PNET

b1018043:
js-libp2p throw error when `LIBP2P_FORCE_PNET` is `0` and don't use Protector. Because, `0` (string) is `true` and js-libp2p enforce to use private network. However, go-libp2p don't throw an error when `LIBP2P_FORCE_PNET` is `0`.

This PR changes js-libp2p not to throw an error when the environment variable `LIBP2P_FORCE_PNET` is `0` (and any value other than `1`) and don't use Protector.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -312,7 +312,7 @@ class Libp2p extends EventEmitter {
     // Attach private network protector
     if (this._modules.connProtector) {
       this.upgrader.protector = this._modules.connProtector
-    } else if (globalThis.process !== undefined && globalThis.process.env && globalThis.process.env.LIBP2P_FORCE_PNET) { // eslint-disable-line no-undef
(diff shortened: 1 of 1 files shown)


mpetrunic:
@b1018043 We would need to revisit all boolean environment variables in this case. Could you open an issue to specify correct inputs and their behavior?

BigLep:
This issue was created: https://github.com/libp2p/js-libp2p/issues/1189

achingbrain:
I don't think we want to encourage interpreting the contents of environmental variables where they are intended to be treated as booleans.

It's important to use idiomatic JavaScript - if you wish the env var to be interpreted as true, set it to a [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) value, if not set it to something [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) (though remember node exposes all `process.env` key/values as strings) or better yet, do not set the variable at all.

PR #1176, opened 2022-03-22
Fix startup sequence

robertkiel:
## TLDR

Change startup sequence to start ConnectionManager before starting modules that might want to connect other peers

## Context

If the ConnectionManager is not started, it unnecessarily blocks (closes) all incoming and outgoing connections because its internal state is not set to started.

## Changes

- start `ConnectionManager` *before* starting modules that might want to establish a connection
- update state variable `_isStarted` *after* work has been done


robertkiel:
> Triage notes: This should be resolved in the new version. New version will also be esm only. Are you okay with migrating to the new version?

thanks for the reply, will check.

BigLep:
@robertkiel: have you confirmed whether updating to the esm only version fixes this?

wemeetagain:
Closing now due to staleness, feel free to reopen if its still an issue in the latest version

PR #1185, opened 2022-04-07
fix: fix unintended aborts in dialer

robertkiel:
Fix a bug where `DialRequest` aborts wrong dial attempts.


achingbrain:
Thanks for opening this - can you please add a test to catch any regressions that may occur in the future?

robertkiel:
> Thanks for opening this - can you please add a test to catch any regressions that may occur in the future?

Currently busy with a different issue, but will create or adapt a unit test tackling this bug soon.

achingbrain:
I put a simple test together, you can add it to `test/dialing/dial-request.spec.ts`:

```js
it('should abort other dials when one succeeds', async () => {
  const connection = mockConnection(mockMultiaddrConnection(mockDuplex(), await createEd25519PeerId()))
  const actions: Record<string, () => Promise<any>> = {
    '/ip4/127.0.0.1/tcp/1231': async () => {
      await delay(10000)
    },
    '/ip4/127.0.0.1/tcp/1232': async () => {
      await delay(10)

      return connection
    },
    '/ip4/127.0.0.1/tcp/1233': async () => {
      await delay(10000)
    }
  }

  const signals: Record<string, AbortSignal | undefined> = {}
  const dialRequest = new DialRequest({
    addrs: Object.keys(actions).map(str => new Multiaddr(str)),
    dialer: new DefaultDialer(new Components(), {
      maxParallelDials: 3
    }),
    dialAction: async (ma, opts) => {
      signals[ma.toString()] = opts.signal
      return await actions[ma.toString()]()
    }
  })

  await expect(dialRequest.run()).to.eventually.equal(connection)

  expect(signals['/ip4/127.0.0.1/tcp/1231']).to.have.property('aborted', true)
  expect(signals['/ip4/127.0.0.1/tcp/1232']).to.have.property('aborted', false)
  expect(signals['/ip4/127.0.0.1/tcp/1233']).to.have.property('aborted', true)
})
```

The thing is it passes against this PR and against master so it might not be exposing the issue you are fixing here.

Could you adapt it so it fails against master and passes against this PR?

BigLep:
@robertkiel: can you add a test that shows what you're fixing?  It's currently not clear what is being fixed.  If there is a problem, we do want to get this merged.

robertkiel:
> @robertkiel: can you add a test that shows what you're fixing? It's currently not clear what is being fixed. If there is a problem, we do want to get this merged.

Thanks a lot for coming back to it. Currently working on the unit test and will give an update once I'm done.

achingbrain:
Thanks!  You may need to merge master again to get the build passing.

robertkiel:
> @robertkiel Could you please fix CI errors?

Omw to fix them. Looks like there are some issues with the `@libp2p/interfaces` package, so I expected them to be fixed independently.

src/connection-manager/dialer/dial-request.ts
```diff
@@ -102,10 +102,23 @@ export class DialRequest {
           }
         }
 
+        // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
```
wemeetagain, line 105: Minor comment, this eslint comment is duplicated here and on the next line.

wemeetagain (changes requested):

src/connection-manager/dialer/dial-request.ts
robertkiel: changed

src/connection-manager/dialer/dial-request.ts
```diff
@@ -86,10 +86,10 @@ export class DialRequest {
         const token = await tokenHolder.shift() // get token
         let conn
         try {
-          const signal = dialAbortControllers[i].signal
+          const signal = (dialAbortControllers[i] as AbortController).signal
```
achingbrain, line 89: Instead of casting, please null-guard on the abort controller instead.

src/connection-manager/dialer/dial-request.ts
achingbrain: If you follow the [suggestions](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md) for the @typescript-eslint/strict-boolean-expressions rule you wouldn't have to disable eslint.

In brief, instead of `if (!conn)` do `if (conn == null)`.

src/connection-manager/dialer/dial-request.ts
```diff
@@ -102,10 +102,23 @@ export class DialRequest {
           }
         }
 
+        // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+        if (!conn) {
+          // Notify Promise.any that attempt was not successful
+          // to prevent from returning undefined despite there
+          // were successful dial attempts
+          throw Error('dialAction led to empty object')
```
achingbrain, line 110: Please wrap the thrown error in `err-code` with an appropriate error code - see examples elsewhere in the codebase for how it's used.

test/dialing/direct.node.ts
```diff
@@ -180,7 +180,9 @@ describe('Dialing (direct, TCP)', () => {
       .and.to.have.property('code', ErrorCodes.ERR_TIMEOUT)
   })
 
-  it('should dial to the max concurrency', async () => {
+  // The implementation of DialRequest is not capable of preventing more dial attempts
+  // than allowed by `maxParallelDials`
+  it.skip('should dial to the max concurrency', async () => {
```
achingbrain, line 185: We should only ever dial to the max concurrency, this test should still pass with these changes - can you please fix it up instead of skipping it?

achingbrain (changes requested):

src/connection-manager/dialer/dial-request.ts
robertkiel: changed

test/dialing/direct.node.ts
robertkiel: This is supposed to be a non-trivial change. In the current version, the test runs through because the bug is covered by this issue.

I would rather fix this in a follow-up PR to keep this PR simple.

robertkiel: Changed - turned out to be easier than expected

src/connection-manager/dialer/dial-request.ts
robertkiel: changed, thx for the suggestion

src/connection-manager/dialer/dial-request.ts
robertkiel: changed, took other occurences as an example

test/dialing/direct.node.ts
achingbrain: Awesome, I had a feeling it might have been 😉 

Please can you re-enable this test and then I think this is good to go.

robertkiel: changeg

Issue #1021, opened 2021-11-12
[Tracker] js-libp2p getting full Type Script support


BigLep:
A few years ago we took the decision on TypeScript to not convert modules to the language, because it's not a standard, it's controlled by a single vendor, it requires lots of extra tooling and there were far more JavaScript developers than TypeScript developers, and all TypeScript developers could write JavaScript but the reverse was not always true.

We were going to use JSDoc comments to generate types so people could use libp2p/ipfs modules in typed environments.  TypeScript developers would get types, and JavaScript developers would get comments, everyone would be happy.

Fast forward a few years and what's changed?  TypeScript has gone from strength to strength, that single-vendor has made some significant plays in open source with acquisitions of npm and GitHub so they're not going anywhere.  Far more developers are writing TypeScript than before.

The JSDoc comment thing didn't work out - it's full of gotchas and introduces so much cruft to the codebase that it makes it unpalatable to both JavaScript developers and TypeScript developers - full TypeScript ends up being much less verbose.  We tried writing complex types in `.d.ts` files but the tooling isn't there to validate them, so we've ended up writing `.ts` anyway.

Meanwhile JavaScript has moved on, as it did from callbacks to promises, now it's away from CJS and towards ESM.

We've been shipping most new code in dual CJS/ESM mode which has meant we've been able to support future-tech like [skypack](https://www.skypack.dev) but also maintain compatibility with older tools that use the [resolve](https://www.npmjs.com/package/resolve) module from [browserify](https://browserify.org/) like [jest](https://jestjs.io/), this is also full of footguns, for example `instanceof` not working between instantiated from CJS then used with `instanceof` in ESM.

Many of our dependencies are now ESM only - since we bundle CJS and ESM we can't update to these as our CJS code would end up pulling in the ESM dependency via a `require` which isn't compatible.

Fast forward to now.  We're standing on the cusp of a CJS to ESM transition, so why not take the opportunity this presents to embrace the peace of mind TypeScript brings at the same time.

We will refactor existing libp2p modules from CJS to TypeScript which will be published as ESM only with export maps and all that good stuff.  This will let us move quickly and safely while supporting new tools and environments.

----

#### Porting

A cheat sheet for porting existing modules.  This will evolve over time as we move some of these steps into Aegir itself in order to reduce the amount of boilerplate required (testing/coding flow, I'm mostly looking at you).

[...]

PR #1205, opened 2022-05-04
fix: strict dependencies to fix build

abetaev:
hi,

at some point i came up with necessity to build js-libp2p, but found that it's not buildable.

[this change was the cause (libp2p/interfaces)](https://github.com/libp2p/js-libp2p-interfaces/commit/a367375accc690d7b4608c9a3313f91df700efd8)

if semver rules were followed this shouldn't have happened.


achingbrain:
Thanks for the PR.  The relevant changes needed to support the updated interfaces have been done as part of #1206 so if you pull from master and reinstall the node deps it should build successfully again.

> if semver rules were followed this shouldn't have happened.

This project is in a bit of a weird space while we [transition to TypeScript](https://github.com/libp2p/js-libp2p/issues/1021) - the interfaces have grown organically over time so are subject to a bit of churn while we apply the learnings of the past few years to them so there will be some breakage.  The good news is this breakage is contained to the `@libp2p/*` modules and `libp2p@next` - the previous `0.36.x` release tree is unaffected because it doesn't use and `@libp2p/*` modules.

Long story short: as soon as #1178 is merged and `0.37.x` is released, semver will be strictly followed again.  In the interim you should use the `0.36.x` release (see [releases](https://github.com/libp2p/js-libp2p/releases) for the latest tag) and not `libp2p@next` or `master`.

abetaev:
thank you for the news and the explanation, it makes things clearer

PR #1208, opened 2022-05-08
fix: transport-manager test, dummy-dht interface compliance

abetaev:
not sure if this might be needed, but it fixes tests and build scripts

should i just wait until 0.37.0 release?

btw, i am just trying to run it in browser, but keep finding dependencies on nodejs APIs (so far: ~~mortice~~, thunky, clean-stack)

### DIFF
--- a/src/dht/dummy-dht.ts
+++ b/src/dht/dummy-dht.ts
@@ -3,8 +3,17 @@ import type { PeerDiscoveryEvents } from '@libp2p/interfaces/peer-discovery'
 import errCode from 'err-code'
 import { messages, codes } from '../errors.js'
 import { EventEmitter } from '@libp2p/interfaces/events'
+import { symbol } from '@libp2p/interfaces/peer-discovery'
 
 export class DummyDHT extends EventEmitter<PeerDiscoveryEvents> implements DualDHT {
+  get [symbol] (): true {
+    return true
+  }
+
+  get [Symbol.toStringTag] () {
+    return '@libp2p/dummy-dht'
+  }
+
   get wan (): SingleDHT {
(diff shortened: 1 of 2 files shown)


achingbrain:
Thanks for submitting this, I actually fixed the issues as part of #1210 before I saw this, sorry!

> should i just wait until 0.37.0 release?

It'll certainly all be much more stable once the release is out.

> btw, i am just trying to run it in browser, but keep finding dependencies on nodejs APIs (so far: ~mortice~, thunky, clean-stack)

```console
% npm ls thunky
libp2p@0.36.2 /Users/alex/Documents/Workspaces/libp2p/js-libp2p
└─┬ @libp2p/mdns@1.0.6
  └─┬ multicast-dns@7.2.4
    └── thunky@1.1.0
```

`thunky` is a dep of `@libp2p/mdns` - this is a node-only module - it's not bundled with libp2p - are you including it as part of your config?

```console
% npm ls clean-stack
libp2p@0.36.2 /Users/alex/Documents/Workspaces/libp2p/js-libp2p
...
└─┬ aggregate-error@4.0.1
  └── clean-stack@4.2.0
```

This is a dep of `aggregate-error` which doesn't actually appear to be used anywhere - I've removed it in #1210 along with some other superfluous deps.

abetaev:
> `thunky` is a dep of `@libp2p/mdns` - this is a node-only module - it's not bundled with libp2p - are you including it as part of your config?
right, after i have written that message i noticed that i had (copypasted from somewhere) mdns in config, it's not needed in a browser, `thunky` is okay

thanks for the update,
i close this one then :)

PR #1239, opened 2022-06-09
feat: constrain stream counts by protocol on a per-connection basis

achingbrain:
Adds a config key `connectionManager.protocolStreamLimits` that lets users define `protocol` -> `maxCount` mappings to limit the number of multiplexed streams that can be opened simultaneously on a per-connection basis.

Applies to both incoming and outgoing streams.

The config key is passed to the upgrader internally because a) we don't expose upgrader config and b) longer term we want to move the upgrader's responsibilities around connection/stream management into the connection manager.


achingbrain:
A nicer API might be to have the registrar accept some additional config when you register a stream handler and would include the number of concurrent streams.

It'd mean the config wouldn't need to know about all protocols at startup time.

wemeetagain (approved):

achingbrain:
Superseded by #1255

PR #1240, opened 2022-06-10
feat: programmatically set agentVersion in identify

### LINKED ISSUE #686: Programmatically set agentVersion in Identify
## Overview

When handling Identify requests we are just sending a static, constant `agentVersion` of `'js-libp2p/0.1.0'`. This is unhelpful and should be dynamic. The agentVersion can be set to any string. What we should be doing is specifying the package name and package version in the agentVersion. This can enable us to better identify applications on the network and see how the public network is upgrading over time.

## Proposed Changes
- Specify a default `agentVersion` of `js-libp2p/<libp2p.version>` ex: (`js-libp2p/0.28.3`)
- When used in another application, such as a js-ipfs daemon, specify an `agentVersion` of `<package.name>/<package.version>` ex: (`ipfs/0.47.0`)
- When running in Node.js, append the node version. ex: (`js-libp2p/<libp2p.version> UserAgent=<process.version>`)
- When running in browser, append the browser UserAgent. ex: (`js-libp2p/<libp2p.version> UserAgent=<navigator.userAgent>`)

go-libp2p already does this for Identify, without the UserAgent information. At a minimum we should include the app name and version name, but the UserAgent would also be useful.

6d7a:
If no `agentVersion` is provided for the Identify protocol, the default `AGENT_VERSION` will now be set to
* `js-libp2p/<libp2p.version> UserAgent=<process.version>` when running in Node.js
* `js-libp2p/<libp2p.version> UserAgent=<navigator.userAgent>` when running in the browser (also when running in a webworker)

Fixes #686


6d7a:
@jacobheun, tagging you as the OP of the issue (:

achingbrain:
Thanks for submitting this, sorry it's taken me so long to get to it.  It looks good but I tried to push some additional commits to this branch but it seems to be out of date so this will be continued in #1296

PR #1271, opened 2022-06-23
fix: do not emit peers on startup

achingbrain:
When starting we emit every peer in the peer store as if they were a new peer.  This makes auto-dial try to dial every peer in the peer store which is very wasteful when there are hundreds if not thousands of peers in the peer store.

Instead any component that thinks it needs a connection certain peers should have tagged those peers with something memorable and should try to connect to them on startup.


wemeetagain (approved):

achingbrain:
Superseded by #1278

PR #1278, opened 2022-06-27
fix: use keep-alive tag to reconnect to peers on startup

achingbrain:
Instead of trying to connect to every peer in the peer store when we start a node, only connect to the peers that have been marked with a `keep-alive` tag.

### DIFF
--- a/src/connection-manager/index.ts
+++ b/src/connection-manager/index.ts
@@ -144,7 +144,7 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
   private started: boolean
   private timer?: ReturnType<retimer>
   private readonly latencyMonitor: LatencyMonitor
-  private startupReconnectTimeout: number
+  private readonly startupReconnectTimeout: number
   private connectOnStartupController?: TimeoutController
 
   constructor (init: ConnectionManagerInit) {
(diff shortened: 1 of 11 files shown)


src/connection-manager/index.ts
```diff
@@ -208,9 +220,43 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
...
     this.components.getUpgrader().addEventListener('connectionEnd', this.onDisconnect)
+
+    // re-connect to any peers with the `keep-alive` tag
+    void Promise.resolve()
+      .then(async () => {
+        const keepAlivePeers: PeerId[] = []
+
+        for (const peer of await this.components.getPeerStore().all()) {
+          const tags = await this.components.getPeerStore().getTags(peer.id)
+          const hasKeepAlive = tags.filter(tag => tag.name === 'keep-alive').length > 0
```
wemeetagain, line 231: Can you define and export the string as a constant? I think downstream users should tag via exported constants if possible to avoid typos and confusion.

achingbrain: I've opened this one: https://github.com/libp2p/js-libp2p-interfaces/pull/263

achingbrain: Mind you, we want to be careful about different modules reusing tag names - otherwise one module could decide a given peer is not important any more and remove the tag, a different module would then be rug-pulled and the bug would be quite hard to track down.

PR #1289, opened 2022-07-04
fix: update muxer behavior

wemeetagain:
- use `direction` in `muxerFactory.createStreamMuxer`
- use `muxer.close` instead of `muxer.streams.forEach(s => s.close())`

PR #1290, opened 2022-07-04
feat: add yamux interop tests

wemeetagain:
Relies on #1289 

Opening this as a draft for discussion. This interop test passes when linking yamux locally.

Changes to the upgrader are needed for yamux, but break mplex :upside_down_face: 

Reason is that upgrader implementation currently either pulls enumerable properties out of the muxed stream or in the case of this PR, attaches new properties to the muxed stream. One way breaks when methods are attached to a class, and the other breaks when methods are functions attached directly to an object. I think ideally, a good solution wouldn't rely on specific implementation details either way...

Perhaps someone has an idea of how to approach the issue.


wemeetagain:
@achingbrain @mpetrunic looking for help on direction to go to solve this

mpetrunic:
@wemeetagain can we update mplex to conform to this changes?

wemeetagain:
@achingbrain I'm not entirely sure why we need to combine the multiselected stream with the muxed stream in the first place.

Perhaps we can just return the multiselected stream?

achingbrain (approved):

PR #1293, opened 2022-07-08
fix: apply identify timeout to opening the stream

achingbrain:
Make sure we can timeout if things like multistream select etc are
slow during identify.

### DIFF
--- a/src/identify/index.ts
+++ b/src/identify/index.ts
@@ -229,16 +229,20 @@ export class IdentifyService implements Startable {
   }
 
   async _identify (connection: Connection, options: AbortOptions = {}): Promise<Identify> {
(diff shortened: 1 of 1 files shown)


src/identify/index.ts
```diff
@@ -229,17 +229,21 @@ export class IdentifyService implements Startable {
...
     if (signal == null) {
       timeoutController = new TimeoutController(this.init.timeout ?? IDENTIFY_TIMEOUT)
       signal = timeoutController.signal
     }
 
+    const stream = await connection.newStream([this.identifyProtocolStr], {
+      ...options,
+      signal
+    })
+    let source: Duplex<Uint8Array> = stream
```
wemeetagain, line 245: `const` ?

PR #1298, opened 2022-07-19
feat: add autonat support

achingbrain:
Implements the [autonat](https://github.com/libp2p/specs/blob/master/autonat/README.md)
spec to give us confidence in our external addresses and to open the
door to things like switching DHT server mode on automatically, hole
punching, v2 circuit relay etc.

Depends on:

- [ ] https://github.com/libp2p/js-libp2p-interfaces/pull/269

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "prepublishOnly": "node scripts/update-version.js",
     "build": "aegir build",
     "generate": "run-s generate:proto:*",
+    "generate:proto:autonat": "protons ./src/autonat/pb/index.proto",
     "generate:proto:circuit": "protons ./src/circuit/pb/index.proto",
     "generate:proto:fetch": "protons ./src/fetch/pb/proto.proto",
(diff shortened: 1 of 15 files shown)


achingbrain:
Testing locally, I get different port numbers from every connection during identify.

E.g. I have uPNP hole punching which opens `/ip4/${PUBLIC_IP}/tcp/4002` - the [observedAddr reported by identify](https://github.com/libp2p/specs/tree/master/identify#observedaddr) is like `/ip4/${PUBLIC_IP}/tcp/61253`, `/ip4/${PUBLIC_IP}/tcp/61254` etc.  We end up with a different observed address for every connection and we can't use autonat to confirm/delete them fast enough so we need to handle this better before merging.

We either:

1. Limit the number of unconfirmed, observed addresses we store
2. Ignore identify observedAddrs
3. Something else?

mpetrunic:
Okay, I managed to bring it up to speed with the current master branch. @achingbrain How do you test that locally and observe?

PR #1334, opened 2022-08-10
fix: type issues

mpetrunic:
I wasn't sure about some type updates.

It seems like multiselect issue surfaced now in unit tests, somehow somewhere number is being sent instead of Uint8Array and it fails to get appended into Uint8ArrayList


achingbrain:
This will be resolved by https://github.com/libp2p/js-libp2p/pull/1335 though it needs https://github.com/ChainSafe/js-libp2p-noise/pull/193 releasing first.

PR #1341, opened 2022-08-11
fix: add successful stream peer to protobook

wemeetagain:
If a protocol stream has been successfully negotiated and is to be
psased to the application, the peerstore should ensure that the peer is
registerd with that protocol.

@achingbrain should the `add` be awaited? or should it be done async?

### DIFF
--- a/src/upgrader.ts
+++ b/src/upgrader.ts
@@ -376,6 +376,10 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
 
               muxedStream.stat.protocol = protocol
 
+              // If a protocol stream has been successfully negotiated and is to be passed to the application,
+              // the peerstore should ensure that the peer is registered with that protocol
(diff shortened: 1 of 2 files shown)


wemeetagain:
fix for https://github.com/ChainSafe/js-libp2p-gossipsub/issues/313

mpetrunic (approved):

wemeetagain:
@achingbrain discussed this with @mpetrunic , decided to stick with not awaiting the `add` here.
Adding something to the protoBook potentially adds an entry to a database, so it is likely unreasonable to await that before returning the stream.

achingbrain:
> Adding something to the protoBook potentially adds an entry to a database, so it is likely unreasonable to await that before returning the stream.

That depends entirely on if the calling code would expect the protobook to be updated by the time the stream is returned.  Not awaiting the promise makes it non-deterministic.

Given that existing code is event based, it might not be a big deal.

achingbrain:
Also, a test for this would be appreciated, given that it'll affect where/when events are emitted.

Issue #1344, opened 2022-08-11
SyntaxError: Named export 'isReactNative' not found. 


wemeetagain:
Using the latest nightly I get this error at runtime
```
file:///home/cayman/Code/lodestar/node_modules/libp2p/dist/src/config.js:10
import { isNode, isBrowser, isWebWorker, isElectronMain, isElectronRenderer, isReactNative } from 'wherearewe';
                                                                             ^^^^^^^^^^^^^
SyntaxError: Named export 'isReactNative' not found. The requested module 'wherearewe' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'wherearewe';
const { isNode, isBrowser, isWebWorker, isElectronMain, isElectronRenderer, isReactNative } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:128:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:194:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:385:24)
    at async file:///home/cayman/Code/lodestar/packages/cli/bin/lodestar.js:3:1
```

wemeetagain:
This `wherearewe` dependency was added in https://github.com/libp2p/js-libp2p/pull/1296

PR #1349, opened 2022-08-12
fix: use default wherearewe import

wemeetagain:
Fixes https://github.com/libp2p/js-libp2p/issues/1344

We can revert this after https://github.com/achingbrain/wherearewe/issues/23


achingbrain:
We can just update to `wherearewe@2.0.0` instead now since it's ESM-only.

achingbrain:
Fixed by #1350

PR #1357, opened 2022-08-17
fix!: load self key into keychain on startup if not present

achingbrain:
To prevent triggering keychain attack prevention on startup, refactor the `KeyChain` class to load the current PeerId as the `'self'` key on startup.

Fixes #1315

BREAKING CHANGE: the `loadKeychain` method has been removed as it is no longer necessary


wemeetagain (approved):
lgtm

PR #1360, opened 2022-08-18
docs: update pubsub message filtering example to listen for correct event

### LINKED ISSUE #1288: bug report: examples/pubsub/message-filtering/1.js
I just tested this example and found a bug, here is the origin code

```
  //subscribe
  node1.pubsub.addEventListener(topic, (evt) => {
    // Will not receive own published messages by default
    console.log(`node1 received: ${uint8ArrayToString(evt.detail.data)}`)
  })
  node1.pubsub.subscribe(topic)

  node2.pubsub.addEventListener(topic, (evt) => {
    console.log(`node2 received: ${uint8ArrayToString(evt.detail.data)}`)
  })

  node3.pubsub.addEventListener(topic, (evt) => {
    console.log(`node3 received: ${uint8ArrayToString(evt.detail.data)}`)
  })
```

addEventListener(**topic**,  

this first param should be fixed as "message", not topic.

==> addEventListener(**"message"**,  

```
  //subscribe
  node1.pubsub.addEventListener("message", (evt) => {
    // Will not receive own published messages by default
    console.log(`node1 received: ${uint8ArrayToString(evt.detail.data)}`)
  })
  node1.pubsub.subscribe(topic)

  node2.pubsub.addEventListener("message", (evt) => {
    console.log(`node2 received: ${uint8ArrayToString(evt.detail.data)}`)
  })

  node3.pubsub.addEventListener("message", (evt) => {
    console.log(`node3 received: ${uint8ArrayToString(evt.detail.data)}`)
  })
```

inverted-capital:
The correct event to listen for is `'message'` and the topic name can be discerned from [the passed event object](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-pubsub/src/index.ts#L34).

Fixes #1288

### DIFF
--- a/examples/pubsub/message-filtering/1.js
+++ b/examples/pubsub/message-filtering/1.js
@@ -40,19 +40,21 @@ const createNode = async () => {
   await node2.dial(node3.peerId)
 
   //subscribe
-  node1.pubsub.addEventListener(topic, (evt) => {
+  node1.pubsub.addEventListener('message', (evt) => {
     // Will not receive own published messages by default
     console.log(`node1 received: ${uint8ArrayToString(evt.detail.data)}`)
   })
   node1.pubsub.subscribe(topic)
 
-  node2.pubsub.addEventListener(topic, (evt) => {
+  node2.pubsub.addEventListener('message', (evt) => {
     console.log(`node2 received: ${uint8ArrayToString(evt.detail.data)}`)
   })
+  node2.pubsub.subscribe(topic)
 
-  node3.pubsub.addEventListener(topic, (evt) => {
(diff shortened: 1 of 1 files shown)


achingbrain:
Thanks for opening this.  I think it needs a bit more of an in-depth fix as the test as written although is run during CI doesn't actually test that the message were received.  If it did then this regression would have been caught.

I have opened #1362 which refactors the test to actually test something, the example to make the test pass and the README to use the new API.

inverted-capital:
Ok thanks - #1362 is a much more thorough fix

PR #1379, opened 2022-09-14
docs: fix examples in documentation

### LINKED ISSUE #1361: Tutorial multiaddrs fix
multiaddrs from this example is undefined. Need to use getMultiaddrs() instead. Also there is no such function as toB58String() but when i used toString() func it seems it was something base58 encoded.

const node = await createNode()

// At this point the node has started
console.log('node has started (true/false):', node.isStarted())
// And we can print the now listening addresses.
// If you are familiar with TCP, you might have noticed
// that we specified the node to listen in 0.0.0.0 and port
// 0, which means "listen in any network interface and pick
// a port for me
console.log('listening on:')
node.multiaddrs.forEach((ma) => console.log(`${ma.toString()}/p2p/${node.peerId.toB58String()}`))

### LINKED ISSUE #1367: Code snippet in Getting Started does not work
[This code snippet on the js-libp2p GitHub Getting Started document](https://github.com/libp2p/js-libp2p/blob/master/doc/GETTING_STARTED.md#running-libp2p) doesn’t work

First reported by here: https://discuss.libp2p.io/t/response-to-a-recent-twitter-debate/1517/6

### Exit Criteria

As a user, I can successfully run the tutorial locally

### Nice to Have

Test the example in CI (not necessary to close this ticket but nice to have)

saul-avikar:
This PR fixes all the errors in the code snippets in the readme files contained in the examples directory and it fixes the doc/GETTING_STARTED.md file.

I have also updated the old package names, e.g. `libp2p-websockets` -> `@libp2p/websockets`.

Resolves #1367 and resolves #1361.

### DIFF
--- a/doc/GETTING_STARTED.md
+++ b/doc/GETTING_STARTED.md
@@ -34,12 +34,12 @@ Now that we have libp2p installed, let's configure the minimum needed to get you
 
 Libp2p uses Transports to establish connections between peers over the network. Transports are the components responsible for performing the actual exchange of data between libp2p nodes. You can configure any number of Transports, but you only need 1 to start with. Supporting more Transports will improve the ability of your node to speak to a larger number of nodes on the network, as matching Transports are required for two nodes to communicate with one another.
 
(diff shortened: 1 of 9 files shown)


mpetrunic (approved):
This is great, thank you so much for doing it!

achingbrain (approved):

achingbrain:
Thanks so much for this, amazing!

PR #1389, opened 2022-09-22
fix: when creating dial targets, encapsulate PeerIds last

achingbrain:
It turns out because `Multiaddr.encapsulate` stringifies the `Multiaddr` it's a [suprisingly expensive operation](https://github.com/multiformats/js-multiaddr/pull/275#issuecomment-1254981709) so here we switch the order of our `Multiaddr` pipeline around so we filter undialable addresses (e.g. unsupported transports etc) before encapsulating the `PeerId` onto a `Multiaddr` we'd then just ignore.


achingbrain:
CPU usage before (left hand side) and after (right hand side) running locally with this change:

image

mpetrunic (approved):

PR #1395, opened 2022-09-27
feat: allow skipping encryption and custom muxer factory in upgrader

ckousik:
Updates the `DefaultUpgrader` to use the new `UpgraderOptions` which allows skipping encryption and/or adding a custom muxer factory for transports which inherently support encryption and muxing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
   },
   "scripts": {
     "clean": "aegir clean",
-    "lint": "aegir lint",
+    "lint": "aegir lint --fix",
     "dep-check": "aegir dep-check",
     "prepublishOnly": "node scripts/update-version.js",
     "build": "aegir build",
@@ -116,7 +116,7 @@
     "@libp2p/interface-pubsub": "^2.0.1",
     "@libp2p/interface-registrar": "^2.0.3",
     "@libp2p/interface-stream-muxer": "^2.0.2",
-    "@libp2p/interface-transport": "^1.0.3",
+    "@libp2p/interface-transport": "file:../js-libp2p-interfaces/packages/interface-transport",
     "@libp2p/interfaces": "^3.0.3",
     "@libp2p/logger": "^2.0.1",
     "@libp2p/multistream-select": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
   },
   "scripts": {
     "clean": "aegir clean",
-    "lint": "aegir lint --fix",
+    "lint": "aegir lint",
     "dep-check": "aegir dep-check",
     "prepublishOnly": "node scripts/update-version.js",
     "build": "aegir build",
@@ -115,7 +115,7 @@
     "@libp2p/interface-peer-store": "^1.2.1",
     "@libp2p/interface-pubsub": "^2.0.1",
     "@libp2p/interface-registrar": "^2.0.3",
-    "@libp2p/interface-stream-muxer": "^2.0.2",
+    "@libp2p/interface-stream-muxer": "file:../js-libp2p-interfaces/packages/interface-stream-muxer",
     "@libp2p/interface-transport": "file:../js-libp2p-interfaces/packages/interface-transport",
     "@libp2p/interfaces": "^3.0.3",
     "@libp2p/logger": "^2.0.1",
--- a/src/upgrader.ts
+++ b/src/upgrader.ts
@@ -265,7 +265,7 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
 
     // Protect
     let protectedConn = maConn
-    if (!skipEncryption) {
+    if ((opts?.muxerFactory) == null) {
       const protector = this.components.getConnectionProtector()
 
       if (protector != null) {
@@ -291,22 +291,21 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
         }
       } else {
         // specify this somehow
-        cryptoProtocol = 'custom'
+        cryptoProtocol = 'native'
+        remotePeer = remotePeerId
       }
 
       upgradedConn = encryptedConn
-      if (opts?.muxerFactory) {
+      if ((opts?.muxerFactory) != null) {
         muxerFactory = opts.muxerFactory
-      } else {
+      } else if (this.muxers.size > 0) {
         // Multiplex the connection
-        if (this.muxers.size > 0) {
-          const multiplexed = await this._multiplexOutbound({
-            ...protectedConn,
-            ...encryptedConn
-          }, this.muxers)
-          muxerFactory = multiplexed.muxerFactory
-          upgradedConn = multiplexed.stream
-        }
+        const multiplexed = await this._multiplexOutbound({
+          ...protectedConn,
+          ...encryptedConn
+        }, this.muxers)
+        muxerFactory = multiplexed.muxerFactory
+        upgradedConn = multiplexed.stream
       }
     } catch (err: any) {
       log.error('Failed to upgrade outbound connection', err)
@@ -314,7 +313,7 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
       throw err
     }
 
-    if (await this.components.getConnectionGater().denyOutboundUpgradedConnection(remotePeerId, {
+    if (await this.components.getConnectionGater().denyOutboundUpgradedConnection(remotePeer, {
       ...protectedConn,
       ...encryptedConn
     })) {
@@ -322,7 +321,7 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
     }
 
     if (metrics != null) {
-      metrics.updatePlaceholder(proxyPeer, remotePeerId)
+      metrics.updatePlaceholder(proxyPeer, remotePeer)
       setPeer(remotePeerId)
     }
 
@@ -427,7 +426,7 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
         }
 
         log('%s: starting new stream on %s', direction, protocols)
-        const muxedStream = muxer.newStream()
+        const muxedStream = await muxer.newStream()
         const metrics = this.components.getMetrics()
         let controller: TimeoutController | undefined
 
--- a/src/upgrader.ts
+++ b/src/upgrader.ts
@@ -12,7 +12,7 @@ import type { MultiaddrConnection, Connection, Stream } from '@libp2p/interface-
 import type { ConnectionEncrypter, SecuredConnection } from '@libp2p/interface-connection-encrypter'
 import type { StreamMuxer, StreamMuxerFactory } from '@libp2p/interface-stream-muxer'
 import type { PeerId } from '@libp2p/interface-peer-id'
-import type { Upgrader, UpgraderEvents } from '@libp2p/interface-transport'
+import type { Upgrader, UpgraderEvents, UpgraderOptions } from '@libp2p/interface-transport'
 import type { Duplex } from 'it-stream-types'
 import { Components, isInitializable } from '@libp2p/components'
 import type { AbortOptions } from '@libp2p/interfaces'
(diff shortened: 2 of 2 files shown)


ckousik:
@mxinden @MarcoPolo

src/upgrader.ts
```diff
@@ -258,56 +260,70 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
...
-
-      if (await this.components.getConnectionGater().denyOutboundEncryptedConnection(remotePeer, {
-        ...protectedConn,
-        ...encryptedConn
-      })) {
-        throw errCode(new Error('The multiaddr connection is blocked by gater.acceptEncryptedConnection'), codes.ERR_CONNECTION_INTERCEPTED)
-      }
+      encryptedConn = protectedConn
+      if (!skipEncryption) {
+        ({
```
MarcoPolo, line 280: Can we make this a `const` declaration so we don't shadow outer vars (I think this is confusing).

src/upgrader.ts
```diff
@@ -240,6 +240,8 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
       throw errCode(new Error('The multiaddr connection is blocked by connectionGater.denyOutboundConnection'), codes.ERR_CONNECTION_INTERCEPTED)
     }
 
+    const skipEncryption = opts?.skipEncryption === true
+
     let encryptedConn
     let remotePeer
```
MarcoPolo, line 246: Can we remove this declaration now that we don't need it here?

src/upgrader.ts
```diff
@@ -258,56 +260,70 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
...
+        // specify this somehow
+        cryptoProtocol = 'custom'
+      }
+
+      upgradedConn = encryptedConn
+      if (opts?.muxerFactory) {
+        muxerFactory = opts.muxerFactory
+      } else {
+        // Multiplex the connection
+        if (this.muxers.size > 0) {
```
MarcoPolo, line 302: else if ?

MarcoPolo (changes requested):
Can we also include the await for async newStream? https://github.com/libp2p/js-libp2p/pull/1396/files#diff-c199200bb38fc06cbe1654f84bb2cfaed1cf0aa9bc9ed71f646932604cbd9252R422

package.json
```diff
@@ -78,7 +78,7 @@
   },
   "scripts": {
     "clean": "aegir clean",
-    "lint": "aegir lint",
+    "lint": "aegir lint --fix",
```
achingbrain, line 81:
```suggestion
    "lint": "aegir lint",
```

src/upgrader.ts
```diff
@@ -258,56 +260,70 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
...
     let protectedConn = maConn
-    const protector = this.components.getConnectionProtector()
+    if (!skipEncryption) {
+      const protector = this.components.getConnectionProtector()
 
-    if (protector != null) {
-      protectedConn = await protector.protect(maConn)
+      if (protector != null) {
+        protectedConn = await protector.protect(maConn)
+      }
```
achingbrain, line 273: We still want to do connection protection here - it's use case is a little different to encrypting the whole channel, instead you're essentially making a private network on top of the encrypted connection because you'll need the PSK to communicate any data.

That is, I can support XYX connection encryption scheme (noise, etc) and connect to a remote node, but I also need to know the PSK to connect to a remote node that uses protected connections.

src/upgrader.ts
```diff
@@ -318,7 +334,7 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
       maConn,
       upgradedConn,
       muxerFactory,
-      remotePeer
+      remotePeer: remotePeerId
```
achingbrain, line 337: This value used to come from the connection encrypter's `encryptOutbound` return value.  Looks like now that's ignored in favour of the one extracted from the multiaddr.

This will be problematic because we want to support discovering the peer id during the noise handshake in future - please can you change it back to use the one from the connection encrypter if it's configured & returned.

achingbrain (changes requested):

src/upgrader.ts
MarcoPolo: This will break if we're using native muxing since there isn't a single underlying "connection". But I agree that we should do this even with native encryption. Maybe skip this if we have a custom stream muxer? fwiw on go-libp2p the QUIC transport doesn't support PSK.

src/upgrader.ts
```diff
@@ -258,56 +260,70 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
...
-        }, this.muxers)
-        muxerFactory = multiplexed.muxerFactory
-        upgradedConn = multiplexed.stream
+        })) {
+          throw errCode(new Error('The multiaddr connection is blocked by gater.acceptEncryptedConnection'), codes.ERR_CONNECTION_INTERCEPTED)
+        }
       } else {
-        upgradedConn = encryptedConn
+        // specify this somehow
+        cryptoProtocol = 'custom'
```
MarcoPolo, line 294: Maybe "native" instead?

src/upgrader.ts
ckousik: @achingbrain Please correct me if I have misunderstood here: For a connection that does not have native encryption, we use the `PreSharedKeyConnectionProtector` to wrap the plaintext stream. Then, the selected encryption scheme runs on top of the already encrypted stream. Is this done for nodes to limit which other nodes can connect to them?

As @MarcoPolo has mentioned, we do not have a single underlying stream in the webrtc spec. Instead, we create an ephemeral data channel for the Noise handshake. Subsequent streams only use the native DTLS encryption.

MarcoPolo: In go-libp2p, we don't support PSK in QUIC or WebTransport.

MarcoPolo: We probably can't support a PSK with native encryption without giving this more thought. I think this change is okay.

MarcoPolo (approved):

MarcoPolo:
Can we move this out of draft @ckousik ?

package.json
```diff
@@ -115,8 +115,8 @@
     "@libp2p/interface-peer-store": "^1.2.1",
     "@libp2p/interface-pubsub": "^2.0.1",
     "@libp2p/interface-registrar": "^2.0.3",
-    "@libp2p/interface-stream-muxer": "^2.0.2",
-    "@libp2p/interface-transport": "^1.0.3",
+    "@libp2p/interface-stream-muxer": "file:../js-libp2p-interfaces/packages/interface-stream-muxer",
+    "@libp2p/interface-transport": "file:../js-libp2p-interfaces/packages/interface-transport",
```
achingbrain, line 119: These need to change to version numbers before merging

src/upgrader.ts
```diff
@@ -258,39 +258,51 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
 
     log('Starting the outbound connection upgrade')
 
+    // If the transport natively supports encryption, skip connection
+    // protector and encryption
+
     // Protect
     let protectedConn = maConn
-    const protector = this.components.getConnectionProtector()
+    if ((opts?.muxerFactory) == null) {
```
achingbrain, line 266: After the [suggested change](https://github.com/libp2p/js-libp2p-interfaces/pull/290#discussion_r988230911) in the interface repo we can make this decision on explicit settings rather than the lack of presence of an overriding muxer factory:

```suggestion
    if (opts.skipProtection !== true) {
```

src/upgrader.ts
```diff
@@ -258,39 +258,51 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
...
 
     try {
       // Encrypt the connection
-      ({
-        conn: encryptedConn,
-        remotePeer,
-        protocol: cryptoProtocol
-      } = await this._encryptOutbound(protectedConn, remotePeerId))
+      encryptedConn = protectedConn
+      if (!opts?.skipEncryption) {
```
achingbrain, line 277: I think this is a bit easier to read.  The linter will also probably complain about not handling boolean values explicitly.

```suggestion
      if (opts?.skipEncryption === false) {
```

src/upgrader.ts
```diff
@@ -258,39 +258,51 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
...
+        }
+      } else {
+        cryptoProtocol = 'native'
+        remotePeer = remotePeerId
       }
 
-      // Multiplex the connection
-      if (this.muxers.size > 0) {
+      upgradedConn = encryptedConn
+      if ((opts?.muxerFactory) != null) {
```
achingbrain, line 296: Doe we need the extra parentheses?

```suggestion
      if (opts?.muxerFactory != null) {
```

src/upgrader.ts
```diff
@@ -318,7 +330,7 @@ export class DefaultUpgrader extends EventEmitter<UpgraderEvents> implements Upg
       maConn,
       upgradedConn,
       muxerFactory,
-      remotePeer
+      remotePeer,
```
achingbrain, line 333: I think this will upset the linter.  You can run `npm run lint` locally to check.

```suggestion
      remotePeer
```

achingbrain (changes requested):
LGTM, some small nits and needs https://github.com/libp2p/js-libp2p-interfaces/pull/290 landing first.

src/upgrader.ts
achingbrain: See https://github.com/libp2p/js-libp2p-interfaces/pull/290#discussion_r988230911 - it's a small change but I think we need to split skipping connection protection out into it's own option - it makes the logic in the changes here easier to follow and also leaves the door open for transports that do their own encryption and muxing but also support PSK.

package.json
```diff
@@ -115,8 +115,8 @@
     "@libp2p/interface-peer-store": "^1.2.1",
     "@libp2p/interface-pubsub": "^2.0.1",
     "@libp2p/interface-registrar": "^2.0.3",
-    "@libp2p/interface-stream-muxer": "^2.0.2",
-    "@libp2p/interface-transport": "^1.0.3",
+    "@libp2p/interface-stream-muxer": "file:../js-libp2p-interfaces/packages/interface-stream-muxer",
```
achingbrain, line 118:
```suggestion
    "@libp2p/interface-stream-muxer": "^3.0.0",
```

package.json
achingbrain, line 119:
```suggestion
    "@libp2p/interface-transport": "^2.0.0",
```

achingbrain:
> I can't push to this branch.

Me neither - @ckousik I noticed the same thing about your [PR to `@libp2p/interfaces`](https://github.com/libp2p/js-libp2p-interfaces/pull/290) - could you please [allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests) on PRs in future?

ckousik:
@achingbrain Should I close this PR and reopen ?

achingbrain:
I think you should be able to grant permission by checking the box on the bottom right of this thread under the PR participants list.  I can't see the box but I see it on PRs I have open to other repos:

image

src/upgrader.ts
MarcoPolo: Wouldn't this return true if `opts` is undefined or `opts.skipEncryption` is undefined?

MarcoPolo: What I mean is that we want the default behavior to be to not skip encryption so we should do `opts?.skipEncryption !== true`. [example](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABFATjCBrAngCjgBygGcB+ALkQG8iMZ9yAjOOAGwFMBDMAXwEoqAUImEiYwRHkKkAdDTqIAvEsTAOLIm36UhI3RARFWbaSzgBzHACIAcnCiI5+fDDBnLvHbu6I26toN09AyMTcytHZ1d3TxFuATiBUEhYBEQGNigoNhRJYnIqR0Zmdi4+AMCxCQI82Vp8RABCZVQQTXLAxH0wQ3ZQixs7BzrItw8O718NdsCunuNTfoiXUZjhOITUdGwcD03MXEo+AQF0zOydk4ysnMPeIA)

MarcoPolo:
I copied this to: https://github.com/libp2p/js-libp2p/pull/1411 with the requested changes.

ckousik:
@achingbrain Since the fork is owned by the little-bear-labs org, I do not get that option at all.

MarcoPolo:
superseded by #1411

PR #1397, opened 2022-09-28
fix!: do not auto-dial peers

achingbrain:
Currently whenever a new peer is discovered we automatically dial them if we are below the connection manager connection limit.

This is incredibly resource intensive and does not necessarily result in high-quality connections (e.g. KAD close, relays or supporting protocols we are interested in).

Now that we can search the DHT for peers, this is no longer necessary so remove the functionality.

BREAKING CHANGE: the old behaviour was to dial any peer we discover, now we just add them to the peer store instead


mpetrunic (approved):

danisharora099:
Was this PR/commit overwritten to reverse the functionality? 

- The latest master currently autodials to peers as can be seen here: https://github.com/libp2p/js-libp2p/blob/bd2bdf7ca0d87ab63b2e9acf7edf7a5752e0559c/src/libp2p.ts#L155
- This commit's version of `libp2p.ts`doesn't exist: https://github.com/libp2p/js-libp2p/blob/d91964f1608e00e18546762f7d01e328ea6483f6/src/libp2p.ts

cc @achingbrain

achingbrain:
The autodialer component ensures the node has active connections to a minimum number of peers (configured by `connectionManager.minConnections`) - if it drops below this the node will dial random peers from the peer store until the number of connections is above `minConnections`.  This is to ensure reachability for things like pubsub messages, DHT queries etc.

The "auto-dial" functionality removed in this PR was that libp2p used to dial every peer it discovered - this is separate to the `minConnections` behaviour.

Hopefully that makes sense.

achingbrain:
Looks like the [docs](https://github.com/libp2p/js-libp2p/blob/master/doc/PEER_DISCOVERY.md) are a little out of date 😩  sorry about that!

achingbrain:
Yes, as per the above, the autodialer component ensures a node keeps `minConnections` open.

danisharora099:
Thanks for clarifying!

PR #1398, opened 2022-09-30
feat: deny incoming connections and add allow/deny lists

achingbrain:
When we reach our connection limit, deny incoming connections before the peer ids are exchanged instead of accepting them and trying to apply the limits after the exchange.

Adds a allow and deny lists of multiaddr prefixes to ensure we can always accept connections from a given network host even when we have reached out connection limit.

Depends on:

- [ ] https://github.com/libp2p/js-libp2p-interfaces/pull/295


src/connection-manager/index.ts
```diff
@@ -598,21 +615,16 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
...
 
   /**
    * If we have more connections than our maximum, select some excess connections
    * to prune based on peer value
    */
-  async _maybePruneConnections (toPrune: number) {
+  async _pruneConnections (toPrune: number) {
     const connections = this.getConnections()
-
-    if (connections.length <= this.opts.minConnections || toPrune < 1) {
```
MarcoPolo, line 612: Is this handled elsewhere?

src/connection-manager/index.ts
```diff
@@ -152,6 +164,8 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
   private readonly startupReconnectTimeout: number
   private connectOnStartupController?: TimeoutController
   private readonly dialTimeout: number
+  private readonly allow: Multiaddr[]
```
MarcoPolo, line 167: Would it make sense to use a `Set` here?

MarcoPolo (approved):

src/connection-manager/index.ts
```diff
@@ -688,34 +702,42 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
...
-      if (allowConnection) {
-        return
-      }
+    if (allowConnection) {
+      return true
+    }
 
-      if (this.getConnections().length < this.opts.maxConnections) {
-        return
+    if (maConn.remoteAddr.isThinWaistAddress()) {
```
MarcoPolo, line 725: Is this just checking ip{4,6} + {tcp,udp}? Where does this name come from?

src/connection-manager/index.ts
```diff
@@ -688,34 +702,42 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
...
-      throw errCode(new Error('maxConnections exceeded'), codes.ERR_TOO_MANY_CONNECTIONS)
-    } catch (err) {
-      await maConn.close()
-      throw err
+    if (this.getConnections().length < this.opts.maxConnections) {
+      return true
     }
+
+    log('maxConnections exceeded')
+    return false
```
MarcoPolo, line 741: The caller no longer knows why this connection wasn't accepted. Probably okay, but maybe better to return an error?

MarcoPolo (approved):

src/connection-manager/index.ts
```diff
@@ -675,4 +701,43 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
...
   }
+
+  async acceptIncomingConnection (maConn: MultiaddrConnection): Promise<boolean> {
+    // check deny list
+    const denyConnection = this.deny.some(ma => {
+      return maConn.remoteAddr.toString().startsWith(ma.toString())
+    })
+
+    if (denyConnection) {
+      log('connection remote address was in deny list')
```
mpetrunic, line 712: Maybe log the remote address that was denied?

src/connection-manager/index.ts
```diff
@@ -132,6 +134,24 @@ export interface ConnectionManagerInit {
    * tagged with KEEP_ALIVE up to this timeout in ms. (default: 60000)
    */
   startupReconnectTimeout?: number
+
+  /**
+   * A list of multiaddrs that will always be allowed to open connections to
```
mpetrunic, line 139:
```suggestion
   * A list of multiaddrs that will always be allowed (except if they are in deny list) to open connections to
```
or some other wording indicating that deny list has higher priority

src/connection-manager/index.ts
```diff
@@ -675,4 +701,43 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
...
+      return true
+    }
+
+    if (maConn.remoteAddr.isThinWaistAddress()) {
+      const host = maConn.remoteAddr.nodeAddress().address
+
+      try {
+        await this.inboundConnectionRateLimiter.consume(host, 1)
+      } catch {
+        log('inboundConnectionThreshold exceeded by host %s', host)
```
mpetrunic, line 731: log remoteAddress as well?

src/connection-manager/index.ts
```diff
@@ -675,4 +701,43 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
...
+        return false
+      }
+    }
+
+    if (this.getConnections().length < this.opts.maxConnections) {
+      return true
+    }
+
+    log('maxConnections exceeded')
```
mpetrunic, line 740: same here

test/connection-manager/index.spec.ts
```diff
@@ -5,12 +5,24 @@ import sinon from 'sinon'
...
 import { KEEP_ALIVE } from '@libp2p/interface-peer-store/tags'
 import pWaitFor from 'p-wait-for'
+import { multiaddr } from '@multiformats/multiaddr'
+import { Components } from '@libp2p/components'
+import { stubInterface } from 'ts-sinon'
+import type { Dialer } from '@libp2p/interface-connection-manager'
+import type { Connection } from '@libp2p/interface-connection'
+
+const defaultOptions = {
+  maxConnections: 10,
```
mpetrunic, line 21: should we relax the default value to 50 (or 100)? I imagine 10 is gonna annoy a lof of users unless they change.

achingbrain: This is in a test?

src/connection-manager/index.ts
achingbrain: > Is this just checking ip{4,6} + {tcp,udp}

Yes.

> Where does this name come from?

I'm actually not sure. I think it refers to just IP + port, no other protocol information - https://www.oilshell.org/cross-ref.html?tag=narrow-waist#narrow-waist

src/connection-manager/index.ts
achingbrain: I don't think so because we call `.some` on it in order to see if the connecting peer's multiaddr is prefixed by anything in the allow list - to do that with a `Set` we'd have to convert it to an array first.

We could have a separate function that iterates over the Set with a `for..of` doing the check that returns early, but `.some` is a one-liner that does the same thing so there seems little benefit?

src/connection-manager/index.ts
```diff
@@ -31,7 +32,8 @@ const defaultOptions: Partial<ConnectionManagerInit> = {
   maxEventLoopDelay: Infinity,
   pollInterval: 2000,
   autoDialInterval: 10000,
-  movingAverageInterval: 60000
+  movingAverageInterval: 60000,
+  inboundConnectionThreshold: 5
```
achingbrain, line 36: This will deny any host not in the allow list that tries to open more than 5x connections a second.  I plucked the number out of thin air, is it reasonable?

src/connection-manager/index.ts
achingbrain: The method prunes the passed number of connections, previously it only did so if we were over the connection limit.  We initiate pruning for a number of reasons - if the event loop is lagging too much or we are sending or receiving too much traffic for example - pruning in these cases would have done nothing if we weren't over the connection limit.

src/connection-manager/index.ts
achingbrain: It logs why the connection was accepted - it's for incoming connections only so it's really just informational for anyone watching the logs.

More idiomatic to throw than return an error but after the conversation [here](https://github.com/libp2p/js-libp2p-interfaces/pull/295#discussion_r986720181) I changed it to return a boolean instead.

mpetrunic (approved):

PR #1418, opened 2022-10-07
fix: update all deps to latest versions

achingbrain:
Pulls in changes from `@libp2p/interface-transport`, `@libp2p/components` etc.


MarcoPolo (approved):
There are also failures that would be fixed by https://github.com/libp2p/js-libp2p/pull/1411. But this looks good regarding the dep updates.

achingbrain:
Superseded by https://github.com/libp2p/js-libp2p/pull/1411

PR #1424, opened 2022-10-12
fix: close stream after sending identify

MarcoPolo:
Close the stream after we finish writing the identify message. This is behavior is specified https://github.com/libp2p/specs/blob/master/identify/README.md#identify

> The peer being identified responds by returning an Identify message and closes the stream.

This shows up when interacting with a go-libp2p node as go-libp2p waits for the stream to close before finishing reading the messages: https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/identify/id.go#L455.

### DIFF
--- a/src/identify/index.ts
+++ b/src/identify/index.ts
@@ -179,14 +179,14 @@ export class IdentifyService implements Startable {
         // make stream abortable
         const source = abortableDuplex(stream, timeoutController.signal)
 
-      await source.sink(pipe(
-        [Identify.encode({
-          listenAddrs,
-          signedPeerRecord,
-          protocols
-        })],
-        lp.encode(),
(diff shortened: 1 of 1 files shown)


MarcoPolo:
bit more context, this is particularly bad when interacting with go-libp2p nodes as go-libp2p will wait for identify to finish before continuing other protocols. So we are blocked until the identify timeout happens before we can do anything else when interacting a js-libp2p node with a go-libp2p node.

For why go does this multi-message merge, see: https://github.com/libp2p/go-libp2p/issues/954#issuecomment-636789416

achingbrain (approved):

achingbrain:
Good catch, thanks!

PR #1427, opened 2022-10-12
fix!: remove @libp2p/components

achingbrain:
`@libp2p/components` is a choke-point for our dependency graph as it depends on every interface, meaning when one interface revs a major `@libp2p/components` major has to change too which means every module depending on it also needs a major.

Switch instead to constructor injection of simple objects that let modules declare their dependencies on interfaces directly instead of indirectly via `@libp2p/components`

Fixes libp2p/js-libp2p-components#6

BREAKING CHANGE: modules no longer implement `Initializable` instead switching to constructor injection


.aegir.js
```diff
@@ -28,14 +28,14 @@ export default {
...
+          webSockets()
         ],
         streamMuxers: [
-          new Mplex()
+          mplex()
         ],
         connectionEncryption: [
-          new Noise(),
-          new Plaintext()
+          () => () => new Noise(),
```
achingbrain, line 37: This is needed until https://github.com/ChainSafe/js-libp2p-noise/pull/229 is released

PR #1440, opened 2022-10-14
fix: enable identify service all the time

achingbrain:
We only enable identify if stream muxers are configured, but new transports can be their own muxers so just enable it all the time.


MarcoPolo (commented):
lgtm

PR #1464, opened 2022-10-22
feat: add options to enable identify and identify push

wemeetagain:
Without rate limiting, identify push may be a DoS vector.

In the mean while it is useful to configure libp2p to completely disable identify and identify push if desired.

FYI was not able to test this test due to type errors installing the latest dependencies :(

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -88,8 +88,8 @@ const DefaultConfig: Partial<Libp2pInit> = {
     maxOutboundStreams: 1,
     maxPushIncomingStreams: 1,
     maxPushOutgoingStreams: 1,
-    disableIdentify: false,
-    disableIdentifyPush: false
+    enable: true,
+    enablePush: true
   },
   ping: {
     protocolPrefix: 'ipfs',
--- a/src/config.ts
+++ b/src/config.ts
@@ -87,7 +87,9 @@ const DefaultConfig: Partial<Libp2pInit> = {
     maxInboundStreams: 1,
     maxOutboundStreams: 1,
     maxPushIncomingStreams: 1,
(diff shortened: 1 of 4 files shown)


src/config.ts
```diff
@@ -87,7 +87,9 @@ const DefaultConfig: Partial<Libp2pInit> = {
     maxInboundStreams: 1,
     maxOutboundStreams: 1,
     maxPushIncomingStreams: 1,
-    maxPushOutgoingStreams: 1
+    maxPushOutgoingStreams: 1,
+    disableIdentify: false,
+    disableIdentifyPush: false
```
mpetrunic, line 92: Do we need this flag? Is there a scenario where we would disable identify push but not identify?

mpetrunic (commented):
Other than that comment, LGTM

src/config.ts
wemeetagain: yes, in lodestar we probably want to disable identify-push but keep identify

achingbrain:
The general pattern is `enabled: true` (defaults to `true`) rather than `disableIdentify: true` (defaults to `false`) (see circuit relay, hop, nat, etc config).

The types seem ok in CI but the tests are failing.

I think adding an `enabled: true` flag is ok as a tactical fix but longer term maybe we want to split the default protocols (identify, identify-push, ping, fetch) out into modules and have them added as part of config, same as transports or peer discovery modules, etc.  What do you think?

wemeetagain:
> and have them added as part of config, same as transports or peer discovery modules, etc. 

Absolutely. the way it works now is a too 'magic'.

p-shahi:
relates to https://github.com/libp2p/js-libp2p/issues/1469

Issue #1455, opened 2022-10-19
Docs: use Aegir + TypeDoc to generate and serve documentation in CI


achingbrain:
## Done Criteria
- [ ] Update js-libp2p to use Aegir + TypeDoc
- [ ] Use GitHub actions to serve new documentation on GitHub pages for each new release of js-libp2p and components

### Original issue
Docs: multaddrs property has been replaced with getMultiaddrs
The docs are out of date, the new getter needs documenting: https://github.com/libp2p/js-libp2p/blob/master/doc/API.md#multiaddrs

p-shahi:
@achingbrain are these docs updated manually?
I was poking around and generated this using [Typedoc](https://github.com/TypeStrong/typedoc)
<img width="1246" alt="image" src="https://user-images.githubusercontent.com/50885601/200445932-fcb15ea8-a4ef-4bc6-a2e7-85edb3f10cd9.png">

Is this something you'd want added to js-libp2p?

achingbrain:
Yes please, the typedoc story has always been a bit hit & miss with JSDoc-style typing but it’s much better with pure typescript.

Aegir has a docs command that generates typedocs and uploads them to gh-pages, that’s be the place to do it, then run that in each repo on release.

maschad:
I created a [Draft PR](https://github.com/libp2p/js-libp2p/pull/1476) but it is currently blocked by [an issue](https://github.com/ipfs/aegir/issues/1105) that I created on AEgir.

PR #1476, opened 2022-11-08
docs: generate Typedocs using Aegir

maschad:
Closes https://github.com/libp2p/js-libp2p/issues/1455


maschad:
Once https://github.com/ipfs/aegir/pull/1107 is merged and a new version of `aegir` is released then this can be reviewed.

README.md
```diff
@@ -52,6 +52,7 @@ If you are looking for the documentation of the latest release, you can view the
   - [Tests](#tests)
     - [Run unit tests](#run-unit-tests)
   - [Packages](#packages)
+- [Used by](#used-by)
```
maschad, line 55: Good catch! thanks

maschad:
Not sure why CI was failing, rebased and pushed again.

maschad:
@achingbrain quite a few of the parameters are not exported as modules and therefore no documentation is generated, **but** they are still referenced by other generated documentation e.g. `ConnectionManager` or `IdentifyService`  - they are also included in the [current API Docs](https://github.com/libp2p/js-libp2p/blob/master/doc/API.md), so do we want to include these?  They may not be modules that we want to export but I don't see the downside of exposing them.

achingbrain:
> quite a few of the parameters are not exported as modules and therefore no documentation is generated, but they are still referenced by other generated documentation 

I haven't tried it yet but this might help? https://www.npmjs.com/package/typedoc-plugin-missing-exports

maschad:
Thanks for that recommendation @achingbrain , once https://github.com/ipfs/aegir/pull/1131 is merged and released then I will re-open this for review.

achingbrain (approved):

PR #1477, opened 2022-11-09
fix: call identify for outgoint connection on dial

simone-sanfratello:
I'm updating `libp2p` in the `e-ipfs` `bitswap-peer` service https://github.com/elastic-ipfs/bitswap-peer/pull/138 from version `0.34` to the latest version `0.40`.

Doing so, there is a difference in the `identify` for outbound connection. While for the inbound ones, the `identify` is been called on the `peer:connect` event 

https://github.com/libp2p/js-libp2p/blob/a74d22a2cddf9ffdca26447fe21a62b5d13e773d/src/identify/index.ts#L105

it doesn't happen anymore for the outbound ones on `dial` method, as it was in previous version.

I suppose it's a regression because of this test

https://github.com/libp2p/js-libp2p/blob/98e80c76b02298af543419f20148e96d4c5f070d/test/identify/service.spec.ts#L43

The issue is reproduced at https://github.com/simone-sanfratello/libp2p-bitswap-issues/tree/master/announce

This PR aims to fix the issue and restore the previous behavior on `dial` method.

It's a draft because I'm not sure adding the `identify` call in `dial` it's the right solution, and also causes other issues (currently marked as skipped tests)

I'm asking for your help to fix

### DIFF
--- a/src/identify/index.ts
+++ b/src/identify/index.ts
@@ -87,6 +87,8 @@ export class IdentifyService implements Startable {
   private readonly init: IdentifyServiceInit
   private started: boolean
 
+  private readonly identifiedPeers: Map<string, Promise<void>>
+
   constructor (components: IdentifyServiceComponents, init: IdentifyServiceInit) {
     this.components = components
     this.started = false
@@ -101,7 +103,9 @@ export class IdentifyService implements Startable {
       ...init.host
     }
 
-    // When a new connection happens, trigger identify
+    this.identifiedPeers = new Map()
+
+    // When a new inbound connection happens, trigger identify
     this.components.connectionManager.addEventListener('peer:connect', (evt) => {
       const connection = evt.detail
       this.identify(connection).catch(log.error)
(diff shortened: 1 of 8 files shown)


src/identify/index.ts
```diff
@@ -291,12 +295,47 @@ export class IdentifyService implements Startable {
     }
   }
 
+  // TODO move to connection outbound close
```
simone-sanfratello, line 298: note: todo in the current PR

src/identify/index.ts
```diff
@@ -291,12 +295,47 @@ export class IdentifyService implements Startable {
...
+
   /**
    * Requests the `Identify` message from peer associated with the given `connection`.
    * If the identified peer does not match the `PeerId` associated with the connection,
    * an error will be thrown.
    */
   async identify (connection: Connection, options: AbortOptions = {}): Promise<void> {
+    // console.log('IdentifyService.identify')
+
+    // TODO test dedupe, error on identifyDeduped
```
simone-sanfratello, line 312: note: todo in the current PR

src/identify/index.ts
```diff
@@ -291,12 +295,47 @@ export class IdentifyService implements Startable {
...
+    const task = this._identifyDeduped(connection, options)
+    this.identifiedPeers.set(id, task)
+    try {
+      // need to return the Promise
+      return task // eslint-disable-line
+    } catch (err) {
+      console.error(err)
+    }
+
+    // TODO clean identifiedPeers on peer disconnect
```
simone-sanfratello, line 334: note: todo in the current PR

src/libp2p.ts
```diff
@@ -377,6 +379,8 @@ export class Libp2pNode extends EventEmitter<Libp2pEvents> implements Libp2p {
     const { id } = getPeer(peer)
 
     await this.components.connectionManager.closeConnections(id)
+    // TODO test connect again, update identify?
```
simone-sanfratello, line 382: note: todo in the current PR

achingbrain:
I'm a little confused here, identify is run on both incoming and outgoing connections.

I've opened https://github.com/libp2p/js-libp2p/pull/1482 which adds a very specific test to this effect.

Backporting that test to 0.34.0 passes as well, as does the existing `'should run identify automatically after connecting'` test in both versions.

What is the issue that this PR resolves?

simone-sanfratello:
This test is failing in the current version

simone-sanfratello:
This test is failing in the current version, while would pass in the old version

https://github.com/libp2p/js-libp2p/pull/1477/files#diff-029c34903d61df5437c532acb62b5a37aa227b00963d009956930acf148b1996R325

You can see the different result in https://github.com/simone-sanfratello/libp2p-bitswap-issues/tree/master/announce/0.34/snip.js and https://github.com/simone-sanfratello/libp2p-bitswap-issues/blob/master/announce/0.40/snip.js

announce

In the old version, the announce addresses are present after the dial, in the latest version they are not

Running `identify` in the `dial` method solves this issues, but it causes other ones, so I'm not sure this is the right path to fix the issue

achingbrain:
To get https://github.com/simone-sanfratello/libp2p-bitswap-issues/blob/master/announce/0.40/snip.js#L33 to pass you need to start the `client`:

```js
// connect to node
const client = await createLibp2p({
  transports: [webSockets()],
  streamMuxers: [mplex()]
})
await client.start() // <--- add this line
```

The handlers for built in protocols are added when the node is started, maybe it needs to throw if you try to dial before it's been started.

simone-sanfratello:
Thanks

PR #1482, opened 2022-11-18
test: add test for inbound and outbound identify

achingbrain:
Ensure identify is run on both inbound and outbound connections.

### DIFF
--- a/test/identify/service.node.ts
+++ b/test/identify/service.node.ts
@@ -1,34 +1,39 @@
 /* eslint-env mocha */
 
-import { expect } from 'aegir/chai'
-import sinon from 'sinon'
-import { createLibp2pNode } from '../../src/libp2p.js'
-import { createNode } from '../utils/creators/peer.js'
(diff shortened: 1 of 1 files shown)


test/identify/service.node.ts
```diff
@@ -0,0 +1,163 @@
...
+      config: {
+        addresses: {
+          listen: ['/ip4/127.0.0.1/tcp/0/ws']
+        }
+      }
+    })
+
+    const connection = await sender.dial(multiaddr(`/ip4/127.0.0.1/tcp/${port}/ws/p2p/${receiver.peerId.toString()}`))
+
+    await delay(1000)
```
mpetrunic, line 124: what are we waiting for here? Can we use some spy instead?

achingbrain: We are waiting for identify.  Since 0.45.x we can wait for the `peer:identify` event instead though.  Have updated.

PR #1488, opened 2022-11-22
docs: add archive notice to webrtc star and direct

p-shahi:
Relates to: https://github.com/libp2p/github-mgmt/pull/80


achingbrain:
Can we please only archive these things when there's an actual alternative for people to use?

Right now [@libp2p/webrtc-star](https://www.npmjs.com/package/@libp2p/webrtc-star) is the only way to do browser to browser communication - once the [@libp2p/webrtc]([@libp2p/webrtc](https://www.npmjs.com/package/@libp2p/webrtc) ) transport has the browser to browser component done, then we can archive WebRTC-Star but given it hasn't even made it on to npm yet this feels a little premature.

p-shahi:
@achingbrain I see your point, the alternatives don't exist or aren't mature enough. However, the primary motivation to archive was to alleviate us of further maintenance work as that's sunken cost (this includes triaging new issues) and let the community know our intention.

If you feel this is a bit too soon, I don't see any problem unarchiving them for now.
However, archiving shouldn't prevent users from still using them and technically they can fork the repos to upstream patches (as onerous as that may be for some users.)

cc: @BigLep as he was also keen to archive the older solutions

p-shahi:
@BigLep poke. we discussed this in last weeks SitRep but can you confirm here as well.

BigLep:
I agree it stinks to pull things away from users, but it's being truthful.  It's not a priority for the js-libp2p maintainers to support it.  We shouldn't spend time triaging or supporting it.  Lets put that effort into the general WebRTC effort.

I think the message we want to get across crystal clear is "use at your own risk.  This is not maintained.  Here is what we're working on to replace it."

I think we're doing this, but if we need to message this better, let's do it.

That said, @p-shahi  it looks like npm hasn't been updated to make clear this is deprecated: https://www.npmjs.com/package/@libp2p/webrtc-star

Can we please do that as well?

maschad:
Given that we released browser to browser in https://github.com/libp2p/js-libp2p-webrtc/releases/tag/v1.1.0 we should be good to merge this once https://www.npmjs.com/package/@libp2p/webrtc-star is deprecated on npm @p-shahi

achingbrain (changes requested):
A couple of extra things to do here:

1. Update webrtc-direct/webrtc-star examples to use new webrtc transport
2. Sweep docs to update references to old webrtc transports to point to new version

examples/libp2p-in-the-browser/README.md
p-shahi, line 1: Should we scrap the libp2p in the browser example and instead point to WebSockets, WebTransport, and WebRTC repos?

doc/CONFIGURATION.md
```diff
@@ -859,72 +857,6 @@ const node = await createLibp2p({
 })
 ```
 
-#### Customizing Transports
```
p-shahi, line 862: Should we replace this section with something other than webrtc-star?

maschad: I agree we could replace it with webrtc and it's [`DataChannelOpts`](https://github.com/libp2p/js-libp2p-webrtc/blob/bc3ae14b9ff2b78fcd35616bcd7bf76c45b7a560/src/stream.ts#L13-L17) or [Websockets and its Multiaddr filter](https://github.com/libp2p/js-libp2p-websockets/blob/3645c2ba72ff1fd0b2a3f385f3cefa5e182f27ff/src/index.ts#L21)

p-shahi:
@maschad Can you take over this PR? i.e. replace the customizing transports with your suggestion

packages/transport-websockets/examples/index.js
```diff
@@ -1,32 +1,20 @@
...
-      webSockets(),
-      webTransport(),
-      webRTC(),
-      webRTCDirect(),
-      circuitRelayTransport({
-        // use content routing to find a circuit relay server we can reserve a
-        // slot on
-        discoverRelays: 1
-      })
+      webSockets()
```
achingbrain, line 14: People copy/paste this config to get started with libp2p in the browser, I think we need to give them the best chance of success with the minimum amount of effort, so this example should have as complete a config as possible.

Please revert the removal of the extra transports.

packages/transport-websockets/examples/package.json
```diff
@@ -0,0 +1,23 @@
+{
```
achingbrain, line 1: Please keep the examples in the examples folder, it's much easier for people visiting the repo to find them that way, since they don't have to dig through the monorepo folders.

doc/CONFIGURATION.md
```diff
@@ -896,66 +863,56 @@ const node = await createLibp2p({
...
-    yamux(),
-    mplex()
+    yamux()
   ],
   connectionEncryption: [
     noise()
   ],
   addresses: {
-    listen: ['/dns4/your-wrtc-star.pub/tcp/443/wss/p2p-webrtc-star'] // your webrtc dns multiaddr
+    listen: ['/dns4/your-wrtc-star.pub/tcp/443/wss/p2p-webrtc'] // your webrtc dns multiaddr
```
achingbrain, line 915: To listen on available relay servers for incoming webrtc connections this should be:

```suggestion
    listen: ['/webrtc']
```

The relay transport should also be configured to search for relays.

doc/CONFIGURATION.md
```diff
@@ -32,7 +30,6 @@
     - [Configuring PeerStore](#configuring-peerstore)
     - [Customizing Transports](#customizing-transports)
     - [Configuring AutoNAT](#configuring-autonat)
-    - [Configuring UPnP NAT Traversal](#configuring-upnp-nat-traversal)
```
achingbrain, line 35: Why has this been removed?

doc/CONFIGURATION.md
```diff
@@ -4,7 +4,6 @@
 - [Modules](#modules)
   - [Transport](#transport)
   - [Stream Multiplexing](#stream-multiplexing)
-    - [Muxer Selection](#muxer-selection)
```
achingbrain, line 7: Why has this been removed?

doc/CONFIGURATION.md
```diff
@@ -73,7 +70,9 @@ Some available transports are:
 - [@libp2p/tcp](https://github.com/libp2p/js-libp2p-tcp) (not available in browsers)
 - [@libp2p/webrtc](https://github.com/libp2p/js-libp2p-webrtc)
 - [@libp2p/websockets](https://github.com/libp2p/js-libp2p-websockets)
-- [@libp2p/webtransport](https://github.com/libp2p/js-libp2p-webtransport) (Work in Progress)
+- [@libp2p/webtransport](https://github.com/libp2p/js-libp2p-webtransport)
+- [@libp2p/webrtc-star](https://github.com/libp2p/js-libp2p-webrtc-star) ([Archived](https://github.com/libp2p/js-libp2p-webrtc-star/blob/master/README.md#%EF%B8%8F%EF%B8%8F%EF%B8%8F%EF%B8%8F%EF%B8%8F%EF%B8%8F))
+- [@libp2p/webrtc-direct](https://github.com/libp2p/js-libp2p-webrtc-direct) ([Archived](https://github.com/libp2p/js-libp2p-webrtc-direct/blob/master/README.md#%EF%B8%8F%EF%B8%8F%EF%B8%8F%EF%B8%8F%EF%B8%8F%EF%B8%8F))
```
achingbrain, line 75: These links should point to monorepo packages where they have been migrated.

doc/CONFIGURATION.md
```diff
@@ -128,11 +127,9 @@ Some available peer discovery modules are:
 - [@libp2p/mdns](https://github.com/libp2p/js-libp2p-mdns)
 - [@libp2p/bootstrap](https://github.com/libp2p/js-libp2p-bootstrap)
 - [@libp2p/kad-dht](https://github.com/libp2p/js-libp2p-kad-dht)
-- [@libp2p/webrtc-star](https://github.com/libp2p/js-libp2p-webrtc-star)
+- [@libp2p/webrtc-star](https://github.com/libp2p/js-libp2p-webrtc-star) ([Archived](https://github.com/libp2p/js-libp2p-webrtc-star/blob/master/README.md#%EF%B8%8F%EF%B8%8F%EF%B8%8F%EF%B8%8F%EF%B8%8F%EF%B8%8F))
```
achingbrain, line 130: These links should point to monorepo packages where they have been migrated.

doc/CONFIGURATION.md
```diff
@@ -128,11 +127,9 @@ Some available peer discovery modules are:
...
 - [@chainsafe/discv5](https://github.com/chainsafe/discv5)
 
-**Note**: `peer-discovery` services within transports (such as `js-libp2p-webrtc-star`) are automatically gathered from the `transport`, via it's `discovery` property. As such, they do not need to be added in the discovery modules. However, these transports can also be configured and disabled as the other ones.
```
achingbrain, line 134: This is still true of modules like `kad-dht` so might be worth mentioning here.

doc/CONFIGURATION.md
```diff
@@ -896,66 +863,56 @@ const node = await createLibp2p({
...
 
 const node = await createLibp2p({
   transports: [
-    webRTC.transport
-  ],
-  peerDiscovery: [
-    webRTC.discovery
+    webRTC({
+      dataChannel: {
+        maxMessageSize: 10
```
achingbrain, line 879: Why is this config key necessary?  People will copy/paste this as good config so it's worth noting why we're overriding the default value here.  If the default value is wrong it should probably be changed instead.

doc/production/README.md
```diff
@@ -8,37 +8,28 @@ This guide aims to guide you from using the public infrastructure into setting u
...
-For nodes that cannot be dialed (including browser), circuit relay nodes should be used.
-
-### `webrtc-star` servers
-
-Regarding `webRTC` connections, a set of star servers are needed to act as a rendezvous point, where peers can learn about other peers (`peer-discovery`), as well as exchange their SDP offers (signaling data).
-
-You can read on how to setup your own star servers in [libp2p/js-libp2p-webrtc-star/DEPLOYMENT.md](https://github.com/libp2p/js-libp2p-webrtc-star/blob/master/DEPLOYMENT.md).
-
-It is worth pointing out that with new discovery protocols on the way, as well as support for distributed signaling, the star servers should be deprecated on the long run.
+In the browser, libp2p supports three transports: `websockets`, `webtransport`, and `webrtc`.
```
achingbrain, line 32: It might be worth expanding this to comment on the differences between them and when you'd expect to use them.

examples/libp2p-in-the-browser/README.md
achingbrain: I disagree - people getting started with libp2p in the browser use this example as a starting point so we should make it as simple, complete and easy to discover as possible.

I think linking to the universal connectivity app as a "next steps" thing would be helpful though, but it does have a whole bunch of extra stuff that can overwhelm new users.

achingbrain (changes requested):

doc/CONFIGURATION.md
achingbrain: Right, looks like the header was removed, https://github.com/libp2p/js-libp2p/pull/1848 restores it

doc/CONFIGURATION.md
maschad: The purpose of this example was to hihglight that you can pass an optional `maxMessageSize` value as mentioned in the above paragraph.

doc/CONFIGURATION.md
maschad: I think the note is a bit confusing overall, since we configure our discovery service via a services module now, intuitively one would configure that separately from the transport. 

So I think it's best to exclude this note.

maschad:
I've move the webRTC, websocket and webtransport examples to their own subdirectories within the `libp2p-in-the-browser` example subdirectory to make the more discoverable as @achingbrain suggested.

achingbrain (approved):

Issue #1342, opened 2022-08-11
Transfer times got twice longer between js-libp2p@0.36.2 and js-libp2p@0.37.3


Alexis-ROYER:
- **Version**: js-libp2p@0.36.2 and js-libp2p@0.37.3
- **Platform**: Windows + node v16.13.2
- **Subsystem**: (to be identified)

#### Severity:

Medium (performance issue)

#### Description:

According to my experimentations, transfer times got twice longer between js-libp2p@0.36.2 and js-libp2p@0.37.3.

If I am true:
- Is the reason known for it? (additional features? refactorings? ...)
- What is the priority for the libp2p team to address this kind of issue?

#### Steps to reproduce the error:

See the following test scripts (one for each version of js-libp2p):
- [libp2p-node.0.36.x.cjs](https://github.com/libp2p/js-libp2p/files/9310751/libp2p-node.0.36.x.cjs.txt)
- [libp2p-node.0.37.x.mjs](https://github.com/libp2p/js-libp2p/files/9310753/libp2p-node.0.37.x.mjs.txt)

Those scripts make it possible to launch 2 nodes:
- a sender, with the following options:
    - `--send`: path to a file which data to send,
    - `--size`: size of the chunks for sending the data.
- a receiver, with the following options:
    - `--connect` option set with the address printed out by the sender.

Timestamps are printed out when the sender starts sending the data, and when the receiver has finished receiving all of it.

With a bit of scripting around it, I made up the following graphs:

![p2p901-2022-08-08a](https://user-images.githubusercontent.com/41943581/184196445-0a20d6de-8ce5-4612-a5f3-6a284554899a.png)
Note: Chunks of 100 bytes are not representative in as much as it causes a lot of overhead around the useful data actually transfered (see [#1343](https://github.com/libp2p/js-libp2p/issues/1343#issuecomment-1212373444)).

![p2p901-2022-08-08b](https://user-images.githubusercontent.com/41943581/184196508-e83b2d04-bc84-4049-87fa-538a9bc8e3c5.png)

![p2p901-2022-08-08c](https://user-images.githubusercontent.com/41943581/184196547-388f3bce-daa0-4c71-a163-40e21557d370.png)

As the graphs show it, js-libp2p@0.37.3 transfer times are about twice those of js-libp2p@0.36.2 from chunk sizes of 1Kb.

Alexis-ROYER:
Possibly related to #1303.

wemeetagain:
[...]

PR #1491, opened 2022-11-23
fix: pipe muxer/connection input/output streams in parallel

achingbrain:
Instead of using `it-pipe` to tie the inputs and outputs of the muxer and underlying connection together, pipe them in parallel.

When sending 105MB in 32b chunks:

## `libp2p@0.36.x`

```
testing 0.36.x
sender 3276810 messages 1638409 invocations  <-- how many mplex messages are sent in how many batches
sender 1638412 bufs 68 b                     <-- how many buffers are passed to the tcp socket and their average size
105 MB in 32 B chunks in 9238ms
```

## `libp2p@0.40.x`

```
testing 0.40.x-mplex
sender 3276811 messages 32 invocations
sender 6553636 bufs 17 b 27476 ms
105 MB in 32 B chunks in 27450ms
```

## With this patch

```
testing 0.40.x-mplex
sender 3276811 messages 17 invocations
sender 6553636 bufs 17 b 23781 ms
105 MB in 32 B chunks in 23753ms
```

## With this patch and https://github.com/libp2p/js-libp2p-mplex/pull/233

```
testing 0.40.x
sender 3276811 messages 1638408 invocations
1638411 bufs 68 b
105 MB in 32 B chunks in 8611ms
```

Refs #1342

### DIFF
--- a/src/upgrader.ts
+++ b/src/upgrader.ts
@@ -1,7 +1,6 @@
 import { logger } from '@libp2p/logger'
 import errCode from 'err-code'
 import * as mss from '@libp2p/multistream-select'
-import { pipe } from 'it-pipe'
 import { codes } from './errors.js'
 import { createConnection } from '@libp2p/connection'
 import { CustomEvent, EventEmitter } from '@libp2p/interfaces/events'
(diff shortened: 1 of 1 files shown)


wemeetagain (approved):

dapplion:
I have seen this pattern of pipe-ing a duplex in many places of libp2p. Can you please review other usages that can cause the same performance issues?

What was the original point of piping the 4 unrelated together with the duplex / pipe pattern?

achingbrain:
> What was the original point of piping the 4 unrelated together with the duplex / pipe pattern?

I think over the years this has just been ported from node streams to pull streams and then to async iterators and they all have some method of directing the output of one stream to the input of another.

Some minimum amount of benchmarking is necessary, particularly where the code is very low level.

PR #1498, opened 2022-11-25
fix: when passed a multiaddr, only dial that multiaddr

### LINKED ISSUE #1497: Dialing a multiaddr should only dial that multiaddr
Currently if we dial a `PeerId`, we check the peer store address book for any known addresses for the `PeerId` and dial them all.

If we dial a `Multiaddr`, we extract the `PeerId` from the `Multiaddr`, add the `Multiaddr` to the peer store then check the peer store for addresses for the `PeerId` and then dial them all.

This is inefficient and also probably not what the user intended.

Instead if passed a `Multiaddr` we should only dial that `Multiaddr`.  If passed a `PeerId` we should try all available addresses as before.

achingbrain:
When a user passes a multiaddr to `libp2p.dial`, only dial that multiaddr instead of extracting the peer id from the multiaddr, using it to look up every known address and dialing them all.

When passing a peer id to `libp2p.dial`, all addresses will be dialed as before.

Fixes #1497

### DIFF
--- a/src/connection-manager/dialer/index.ts
+++ b/src/connection-manager/dialer/index.ts
@@ -1,12 +1,7 @@
 import { logger } from '@libp2p/logger'
-import all from 'it-all'
-import filter from 'it-filter'
-import { pipe } from 'it-pipe'
 import errCode from 'err-code'
-import type { Multiaddr, Resolver } from '@multiformats/multiaddr'
-import { multiaddr, resolvers } from '@multiformats/multiaddr'
+import { isMultiaddr, Multiaddr, Resolver, multiaddr, resolvers } from '@multiformats/multiaddr'
 import { TimeoutController } from 'timeout-abort-controller'
-import { AbortError } from '@libp2p/interfaces/errors'
 import { anySignal } from 'any-signal'
 import { setMaxListeners } from 'events'
 import { DialAction, DialRequest } from './dial-request.js'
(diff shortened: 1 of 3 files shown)


src/connection-manager/dialer/index.ts
```diff
@@ -202,76 +217,83 @@ export class DefaultDialer implements Startable, Dialer {
...
-          yield * await _resolve(a.multiaddr, options)
-        }
-      },
+    let addrs = isMultiaddr(peerIdOrMultiaddr) ? [peerIdOrMultiaddr] : await this._loadAddresses(peerIdOrMultiaddr)
+
+    addrs = (await Promise.all(
+      addrs.map(async (ma) => await _resolve(ma, options))
+    ))
+      // flatten
+      .reduce((acc, curr) => {
```
mpetrunic, line 237: wouldn't just `.flat()` work?

src/connection-manager/dialer/index.ts
```diff
@@ -202,76 +217,83 @@ export class DefaultDialer implements Startable, Dialer {
...
-      }),
-      async (source) => await all(source)
-    )
+      .filter(ma => Boolean(this.components.transportManager.transportForMultiaddr(ma)))
 
+    // deduplicate addresses
     addrs = [...new Set(addrs)]
 
     if (addrs.length > this.maxAddrsToDial) {
-      await this.components.peerStore.delete(peer)
```
mpetrunic, line 265: if user passes ma, we could replace records with given ma?

src/connection-manager/dialer/index.ts
achingbrain: Fixed

src/connection-manager/dialer/index.ts
achingbrain: Not sure, that sounds surprising.  I think it would be better for the user to respond to the error message and update the address book if they need to.

mpetrunic (approved):

PR #1500, opened 2022-11-26
feat!: start libp2p nodes by default

### LINKED ISSUE #1499: Start node by default
Most of the time, after `createLibp2p` is called, the next thing called is `node.start()`.  Sometimes this [catches users out](https://github.com/libp2p/js-libp2p/pull/1477#issuecomment-1321713030).

`createLibp2p` is async so we can just start the node in that function and save the user the trouble.

We can pass `{ start: false }` as an option to disable this behaviour.

achingbrain:
Nodes returned from `createLibp2p` are now started by default. To override this pass `start: false` as an option.

Fixes #1499

BREAKING CHANGE: nodes returned from `createLibp2p` are now started by default

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -53,7 +53,6 @@ export default {
         pipe(stream, stream)
           .catch() // sometimes connections are closed before multistream-select finishes which causes an error
       })
-      await libp2p.start()
 
       return {
         libp2p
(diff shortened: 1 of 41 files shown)


mpetrunic:
@achingbrain Isn't this a breaking change?

mpetrunic (approved):

PR #1504, opened 2022-11-30
fix: registrar is not calling topology on peer reconnect

### LINKED ISSUE #1496: Connection drop causes pubsub to stop working.
- **Version**: 0.40.0

- **Platform**:  MacOS, Node v16.14.0

- **Subsystem**: Pubsub, Relay, ConnectionManager

#### Description:

I have two peers - Alice and Bob, both connected to the same relay server and subscribed to the same topic. 

I need to find all peers subscribed to a given topic, e.g. I call `libp2p.pubsub.getSubscribers("my_topic")` from the Alice side and expect to see Bob's peer id in response.

When we run both peers the first time, everything works as expected - peers sees each other in topic subscribers. But when some of the peers reconnects (doesn't matter Alice or Bob), they does not see each other anymore. But if I restart one of the peers manually, everything works as expected until next reconnect.

I am using `gossibsub` for pubsub, but also tried `floodsub` with the same result.

My issue is very similar to #658.

I created [reproduction repository](https://github.com/AlexMesser/pubsub-issue-reproduction) for better explanation.

mpetrunic:
If the peer reconnects with the same protocols, `change:protocols` is not emitted and topology `onConnect` callbacks are not invoked. This adds listening to connect events in the registrar and checks if we already know protocols for peers before calling the topology `onConnect` callback.

resolves #1496

### DIFF
--- a/src/registrar.ts
+++ b/src/registrar.ts
@@ -240,7 +240,6 @@ export class DefaultRegistrar implements Registrar {
         if (connection == null) {
           continue
         }
-        console.log('topology on connect')
         topology.onConnect(peerId, connection)
       }
     }
--- a/src/registrar.ts
+++ b/src/registrar.ts
@@ -179,7 +179,7 @@ export class DefaultRegistrar implements Registrar {
   }
 
   /**
-   * On peer connected if we already have his protocols. Usually used for reconnects
+   * On peer connected if we already have their protocols. Usually used for reconnects
    * as change:protocols event won't be emitted due to identical protocols.
    */
   _onConnect (evt: CustomEvent<Connection>) {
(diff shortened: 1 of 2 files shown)


mpetrunic:
Seems like regression as it was once fixed in same manner: https://github.com/libp2p/js-libp2p-interfaces/pull/54/files

src/registrar.ts
```diff
@@ -176,12 +178,38 @@ export class DefaultRegistrar implements Registrar {
       })
   }
 
+  /**
+   * On peer connected if we already have his protocols. Usually used for reconnects
```
achingbrain, line 182:
```suggestion
   * On peer connected if we already have their protocols. Usually used for reconnects
```

achingbrain (approved):

AlexMesser:
Thanks for the fix, hope new npm version will be released soon.

achingbrain:
Likely Monday morning - in the interim you can test with the rc with `npm i libp2p@next`.

Issue #1228, opened 2022-05-26
Configuration page needs updates


wemeetagain:
Issues seen while migrating lodestar to latest libp2p:

- configuration page references peer store "persistence", no longer an option, should use `PersistentPeerStore` instead
- `persistent: boolean` listed in an unused interface in `src/index.ts`
- `dialer: {...}` referenced in configuration page, should instead be under `connectionManager`
- ...

achingbrain:
Hand-written docs do get out of sync from time to time.  Would you like to submit a PR that brings everything up to date?

wemeetagain:
I have a branch in progress. waiting on a fix for https://github.com/bbc/typescript-docs-verifier/issues/22

BigLep:
@wemeetagain : we're unblocked here now right?

wemeetagain:
Yes this is unblocked.
It may be useful to break this work up into separate PRs since there's a lot of changes required.
I can start with `doc/CONFIGURATION.md`.

maschad:
@wemeetagain I have a [draft PR here](https://github.com/libp2p/js-libp2p/pull/1505) to introduce [typescript-docs-verifier](https://github.com/bbc/typescript-docs-verifier) to check the [configs](https://github.com/libp2p/js-libp2p/blob/master/doc/CONFIGURATION.md) , I'd be happy to help with updating the config docs f if you could point me to the branches you had started on. 

Given we will have generated docs once https://github.com/libp2p/js-libp2p/issues/1455 is closed, I don't think we would need to document some of the internal options and type descriptions.

@achingbrain do you think we should add [typescript-docs-verifier](https://github.com/bbc/typescript-docs-verifier) as a feature to [aegir](https://github.com/ipfs/aegir) in order to use it here?

Issue #824, opened 2020-12-04
Consider libp2p mono repo


vasco-santos:
Taking into account the overhead that we experience in libp2p development when sometimes for getting a feature in we need to orchestrate several PRs just to land a change, we should consider turning libp2p into a mono repo. This would also help new contributors as they would not need to know where other modules are and that they would need to be updated (like the interop, in case its tests fail in CI).

Some of the repos that should be considered are the daemon, daemon client and libp2p-interfaces.

This issue should be used for this discussion, where we should put together benefits, drawbacks and lessons learned from `js-ipfs` change into a mono repo.

cc @jacobheun @achingbrain @hugomrdias

jacobheun:
The other really nice thing we could do with this not listed is releasing preconfigured libp2p nodes more easily with every libp2p release so users don't have to configure anything if they just want something ready to go.

achingbrain:
Some related modules have been grouped together as monorepos since this was opened:

* https://github.com/libp2p/js-libp2p-daemon
* https://github.com/libp2p/js-libp2p-interfaces
* https://github.com/libp2p/js-libp2p-peer-id
* https://github.com/libp2p/js-libp2p-webrtc-star

The primary piece of learning from `js-ipfs` is that choosing what to add to a monorepo is very important, and that you should not add *everything* unless you like very long build times.

Modules should also be homogeneous in terms of technology, scope and testing - if you have one GUI app in a monorepo and ten tiny utils, releasing changes to the utils takes as long as releasing changes to the GUI app which is a bit nuts.

Also flaky tests become that much painful.

Also, also, contributors tend to get scared off by the scope of a monorepo.

I don't think there are other modules that need grouping together in the libp2p ecosystem so I'm going to close this issue but it's always something worth keeping an eye on.

p-shahi:
candidates to add to monorepo: transports, multiplexers (yamux and mplex are in CS org), interfaces

p-shahi:
The subject of migrating js-libp2p components into a monorepo was brought up by @MarcoPolo and discussed in the 2023-03-07 js-libp2p triage meeting with @achingbrain @wemeetagain @maschad and @p-shahi 

The reasons for moving to a monorepo given were:
* It is painful for maintainers to manage separate repos
* Even though its cleaner conceptually to have separate repos, they are harder to track.
  * In practice there's added overhead even with CI bots to make things smoother.

[...]

PR #1505, opened 2022-12-02
docs: Check TS in docs using typescript-docs-verifier

maschad:
Closes https://github.com/libp2p/js-libp2p/issues/1228


maschad:
Thanks for the feedback.

> I think this should be added to aegir as a `doc-check` command though, similar to `dep-check`.
Good point, I'll add it as a feature

> It can manage the tsconfig.json file for the docs then, in the same way we do it for [linting](https://github.com/ipfs/aegir/blob/master/src/lint.js?rgh-link-date=2022-12-06T12%3A30%3A10Z#L65-L88).

A user may want to use a different  `tsconfig.json` in situations [like this](https://github.com/bbc/typescript-docs-verifier/issues/22) for instance, so it may be best to specify that separately.

> It also doesn't currently check any code blocks in the docs here, because they are fenced with `js` as a type and not `typescript` - `ts` doesn't seem to work either, though maybe there's some config for that?

I think it's because of you were traversing the wrong directory in the script [here](https://github.com/libp2p/js-libp2p/pull/1505/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R83) it works otherwise.

I opened a [ticket on aegir](https://github.com/ipfs/aegir/issues/1117) once this is closed we can update and review this.

maschad:
Once https://github.com/ipfs/aegir/pull/1134 is merged and released then this can be reviewed

doc/LIMITS.md
```diff
@@ -35,32 +35,36 @@ const node = await createLibp2pNode({
     /**
      * The total number of connections allowed to be open at one time
      */
-    maxConnections: number
+    maxConnections: number,
```
maschad, line 38: We will have to find a way to either silence the compiler in regards to this  since `number` only refers to a type, but is being used as a value here.

I know that https://github.com/libp2p/js-libp2p/pull/1456 sort to avoid giving these values but I wonder if we should revert to stating the defaults

doc/LIMITS.md
```diff
@@ -35,32 +35,36 @@ const node = await createLibp2pNode({
...
      * How many connections can be open but not yet upgraded
      */
-    maxIncomingPendingConnections: number
-  }
-})
+    maxIncomingPendingConnections: number,
+  },
+});
+
+export {};
```
maschad, line 53: This `export` could be confusing to end-users but I haven't found another way to avoid the error about redeclaring block-scoped variables unless we change the name of the variable, which would be required for each declaration of `node`

doc/LIMITS.md
achingbrain: It is being used as a value so the compiler is correct in failing.  To use it as a type the example needs to use an actual number (like `50`) or show the interface definition instead.

I think docs are better if the examples can be copy/pasted

doc/LIMITS.md
achingbrain: Usually the `isolatedModules` controls errors in ts code that neither imports nor exports anything - did you try that?

Would rather not have to put `export {}` everywher.

achingbrain (changes requested):
Please can you revert all the formatting changes and stick to the same coding standards as the main codebase.  Single quotes for strings, no semi colons, no extra comma for the final entry in a list, etc.

achingbrain:
The interop suite has a peer dep on aegir@37.x so the dependency tree can't be resolved:

```
npm ERR! Could not resolve dependency:
npm ERR! peer aegir@"^37.2.0" from @libp2p/interop@4.0.2
npm ERR! node_modules/@libp2p/interop
npm ERR!   dev @libp2p/interop@"^4.0.0" from the root project
```

This is blocked until https://github.com/libp2p/interop/pull/82 is merged.

achingbrain:
TBH we may wish to do the aegir upgrade first in it's own PR, then revisit this one once it's merged as adding return types to all the functions will be quite noisy.

doc/METRICS.md
```diff
@@ -1,177 +1,3 @@
 # Libp2p Metrics <!-- omit in toc -->
 
-Metrics allow you to gather run time statistics on your libp2p node.
```
achingbrain, line 3: There is useful information here that isn't in the `@libp2p/prometheus-metrics` docs.

Can you please PR that repo with the information that's been removed here.

You can use the `@packageDocumentation` tag ([example](https://github.com/libp2p/js-libp2p/blob/master/src/pnet/index.ts#L1C1-L23) - [output](https://libp2p.github.io/js-libp2p/modules/pnet.html)) to get general docs to appear on a given page, and the `@example` tag to annotate individual methods, classes or interfaces with code that people can copy/paste to get started.

doc/LIMITS.md
```diff
@@ -223,24 +269,26 @@ const node = await createLibp2p({
      * all connection limits
      */
     allow: [
-      '/ip4/43.123.5.23/tcp/3984',
-      '/ip4/234.243.64.2',
-      '/ip4/52.55',
+      "/ip4/43.123.5.23/tcp/3984",
```
achingbrain, line 272: This should  follow the same linting rules as the main codebase, so use single quotes.

doc/LIMITS.md
```diff
@@ -186,35 +225,42 @@ The [@libp2p/tcp](https://github.com/libp2p/js-libp2p-tcp) transport allows addi
...
+
 ```
 
 ## Allow/deny lists
 
 It is possible to configure some hosts to always accept connections from and some to always reject connections from.
 
-```js
+```ts
```
achingbrain, line 262: The quoting is inconsistent - here it's single quotes, in the `allow` string list below it becomes double quotes.

It should follow the same linting rules as the main codebase, so use single quotes.

doc/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
```
achingbrain, line 1: Can we bundle this config with aegir and load the config from there?

It would be better than copy/pasting this file into every repo.

achingbrain (changes requested):

doc/METRICS.md
maschad: Upon reconsideration I think it makes sense to keep this information here and then also add in the `@packageDocumentation` tags and examples for those who may interact with the metrics lib independently of libp2p core.

doc/tsconfig.json
maschad: Good point, once https://github.com/ipfs/aegir/pull/1255 is merged then we can upgrade `aegir` here to `39.0.6` and then merge.

maschad:
Once https://github.com/ipfs/aegir/pull/1255 is merged then we can update to the newest aegir and then merge this.

p-shahi:
this add deps on prom-metrics in the libp2p repo, will wait for the monorepo work to get in follow up in #824 
have package.json in docs folder to keep it out of libp2p deps

p-shahi:
> this add deps on prom-metrics in the libp2p repo, will wait for the monorepo work to get in follow up in #824 have package.json in docs folder to keep it out of libp2p deps

Unblocked, @maschad adding to this PR itself

doc/LIMITS.md
```diff
@@ -233,14 +279,16 @@ const node = await createLibp2p({
...
+    deny: [
       '/ip4/132.14.52.64/tcp/3984',
       '/ip4/234.243.64.2',
       '/ip4/34.42',
       // etc
-    ]
-  }
-})
+    ],
+  },
```
achingbrain, line 288: These comments should follow the same linting rules as the main codebase; comma-dangle is disallowed.

```suggestion
    ]
  }
```

doc/LIMITS.md
```diff
@@ -233,14 +279,16 @@ const node = await createLibp2p({
...
+});
```
achingbrain, line 289: More weird linting going on here, semi colons are disallowed in the main codebase:

```suggestion
})
```

packages/libp2p/package.json
```diff
@@ -168,15 +169,21 @@
     "@libp2p/daemon-server": "^5.0.2",
     "@libp2p/floodsub": "^7.0.1",
     "@libp2p/interface-compliance-tests": "^3.0.0",
+    "@libp2p/interface-mocks": "^12.0.1",
+    "@libp2p/interface-peer-id": "^2.0.2",
+    "@libp2p/interface-pubsub": "^4.0.1",
+    "@libp2p/interfaces": "^3.3.2",
```
achingbrain, line 175: Please remove the extra deps.

packages/libp2p/package.json
```diff
@@ -168,15 +169,21 @@
...
     "@libp2p/interop": "^8.0.0",
     "@libp2p/kad-dht": "^9.0.0",
     "@libp2p/mdns": "^8.0.0",
     "@libp2p/mplex": "^8.0.0",
+    "@libp2p/pubsub": "^7.0.1",
```
achingbrain, line 180: Please remove the extra deps.

packages/libp2p/package.json
```diff
@@ -95,7 +95,8 @@
   "scripts": {
     "clean": "aegir clean",
     "lint": "aegir lint",
-    "dep-check": "aegir dep-check",
+    "doc-check": "aegir doc-check --inputFiles ../../doc/*.md",
+    "dep-check": "aegir dep-check -i protons @libp2p/prometheus-metrics",
```
achingbrain, line 99: Please revert.

packages/libp2p/package.json
```diff
@@ -168,15 +169,21 @@
...
     "@libp2p/kad-dht": "^9.0.0",
     "@libp2p/mdns": "^8.0.0",
     "@libp2p/mplex": "^8.0.0",
+    "@libp2p/pubsub": "^7.0.1",
     "@libp2p/tcp": "^7.0.0",
     "@libp2p/websockets": "^6.0.0",
     "@types/varint": "^6.0.0",
     "@types/xsalsa20": "^1.1.0",
     "aegir": "^39.0.10",
+    "cborg": "^2.0.1",
```
achingbrain, line 186: Please remove the extra deps.

achingbrain (changes requested):
The `/doc` folder should be added to the `workspaces` list in the root `package.json`.

It should have a `package.json` added that contains:

1. The `doc-check` command
2. The deps needed by the doc examples
3. The `dep-check` command to ensure the deps don't get out of date (assuming we can make it run on markdown, might need to be a follow up)

doc/LIMITS.md
maschad: So the `doc-check` command which utilizes https://github.com/bbc/typescript-docs-verifier only checks that Typescript code snippets compile, it's not a linter. That could be added in a follow up PR but wasn't the original intention of this.

maschad:
I've added the `doc-check` command to the `docs/` workspace. We will need a follow up PR for: 

- doing a `dep-check` than checks the TS snippets in markdown
- Enforcing linting rules on TS snippets in markdown.

Both of these will be features to be added in https://github.com/ipfs/aegir

doc/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "dependencies": {
+    "@chainsafe/libp2p-yamux": "^4.0.2",
+    "@libp2p/interface-metrics": "^4.0.8",
+    "@libp2p/mplex": "^8.0.4",
+    "@libp2p/prometheus-metrics": "^1.1.5",
+    "@libp2p/tcp": "^7.0.3",
+    "aegir": "^39.0.13",
+    "libp2p": "^0.45.9",
+    "prom-client": "^14.2.0"
+  }
```
achingbrain, line 42: We don't want to publish this to npm

```suggestion
  },
  "private": true
```

Issue #1269, opened 2022-06-22
Use/create a more intuitive `errcode`


wemeetagain:
Current usage of errcode is not intuitive.
example:
```ts
throw new errcode(new Error('error in foo'), ERR_FOO, { foo: bar })
```

Prefer an implementation like:
```ts
export class CodeError<T extends Record<string, any>> extends Error {
  code: string
  props: T

  constructor(message: string, code: string, props?: T) {
    super(message)

    this.code = code
    this.props = props ?? {} as T
  }
}
```

with usage then like:
```ts
throw new CodeError('error in foo', ERR_FOO, { foo: bar })
```

achingbrain:
This is a great first issue that would suit someone looking to get started helping build libp2p!

To complete this:

1. Create a new repo
2. Export `CodeError` class (defined in OP)
3. Write tests
4. Open PRs to `@libp2p/*` repos using the new class

tabcat:
Will look at completing this sometime this week now that CodeError is available.

tabcat:
Tracking PRs needed to close this issue.
Have started by adding all relevant `@libp2p/*` packages like requested.
If changes to the list are needed let me know.

- [ ] @libp2p/crypto@1.0.7
- [ ] @libp2p/daemon-client@3.0.6
- [ ] @libp2p/tcp@5.0.2
- [ ] @libp2p/daemon-server@3.0.5
- [ ] @libp2p/tcp@5.0.2
- [ ] @libp2p/interface-mocks@8.0.1
- [ ] @libp2p/kad-dht@6.0.1
- [ ] @libp2p/record@2.0.3
- [ ] @libp2p/mplex@7.0.0
- [ ] @libp2p/multistream-select@3.1.1
- [ ] @libp2p/peer-id@1.1.16
- [ ] @libp2p/peer-record@4.0.4
- [ ] @libp2p/peer-store@5.0.0
- [ ] @libp2p/pubsub@5.0.0
- [ ] @libp2p/tcp@6.0.2
- [ ] @libp2p/topology@3.0.1
- [ ] @libp2p/utils@3.0.2
- [ ] @libp2p/webrtc-star@5.0.3
- [ ] @libp2p/webrtc-peer@2.0.2
- [ ] @libp2p/websockets@5.0.0

*other dependencies (some are indirect)*

- @achingbrain/nat-port-mapper@1.0.7
- @chainsafe/libp2p-yamux@3.0.3
[...]

PR #1532, opened 2023-01-05
chore: use CodeError

tabcat:
Replaces [err-code](https://github.com/IndigoUnited/js-err-code/blob/master/index.js) with [CodeError](https://github.com/libp2p/js-libp2p-interfaces/pull/314)

:laughing:  

Related: #1269

Changes

- needed to upgrade @libp2p/interfaces to use the new CodeError export
- removes err-code from dependencies
- uses CodeError in place of err-code

Before merging:

- [ ] Handle .props breaking change
  - err-code exposes properties inside its props parameter directly from the error object (e.g. error.someProp). CodeError exposes them from the provided props parameter (e.g. error.props.someProp) (see: [js-libp2p/crypto#278](https://github.com/libp2p/js-libp2p-crypto/pull/278#issue-1398406638)). 
  - https://github.com/libp2p/js-libp2p/compare/master...tabcat:js-libp2p:chore/code-error?expand=1#diff-f7c06d53ddde550172379d5951eb28536d95404fa767ee9d986a8fdae34ddd56L127-R130


tabcat:
Two options to handle the .props breaking change wrt err-code -> CodeError

- refactor CodeError to match err-code
  - generic typing for props would not make sense anymore (bad)
- upgrade any internal code relying on this
  - if the keychain's cms decrypt function is public or is used by other libp2p/ipfs packages
    - seems like a breaking change?
    - document change and link to upgrade fix from change log

wemeetagain:
we'll want to go with this option:
> upgrade any internal code relying on this

tabcat:
Ill start looking for those cases then in ipfs/libp2p and other related dependents

tabcat:
From what I could find from the dependents on npm, this doesnt seem to be used internally. It is documented in the code as an API. Will need to update that.

src/keychain/cms.ts
```diff
@@ -124,22 +124,22 @@ export class CMS {
     if (r == null) {
       // @ts-expect-error cms types not defined
       const missingKeys: string[] = recipients.map(r => r.keyId)
-      throw errCode(new Error(`Decryption needs one of the key(s): ${missingKeys.join(', ')}`), codes.ERR_MISSING_KEYS, {
+      throw new CodeError(`Decryption needs one of the key(s): ${missingKeys.join(', ')}`, codes.ERR_MISSING_KEYS, {
         missingKeys
       })
```
tabcat, line 129: This is the only piece of code affected by the `props` change in CodeError across all of the libp2p packages.

src/keychain/cms.ts
```diff
@@ -75,11 +75,11 @@ export class CMS {
    * Reads some protected data.
    *
    * The keychain must contain one of the keys used to encrypt the data.  If none of the keys
-   * exists, an Error is returned with the property 'missingKeys'.  It is array of key ids.
+   * exists, an Error is returned with the property '.props.missingKeys'.  It is array of key ids.
```
tabcat, line 78: Let me know if this needs to be reworded.

p-shahi:
Let's resolve conflicts in this PR after https://github.com/libp2p/js-libp2p/pull/1533 is merged @tabcat

tabcat:
Since the keychain has been moved to it's own repo this is no longer a breaking change.

achingbrain:
Thanks for opening this and having such patience.  We got there!

Issue #1461, opened 2022-10-20
B. 🥊 Decentralized Hole Punching


p-shahi:
**Why:**  P2P networks can have a combination of both public and private nodes. While private nodes can dial nodes on the public Internet, they are unreachable from the outside as they are behind a NAT or a firewall. We need a mechanism to dial them. A [previous DHT crawl found that almost 63%](https://github.com/libp2p/specs/blob/master/ROADMAP.md#-hole-punching-on-tcp-and-quic) of the network was undialable.

**Goal:** Implement Decentralized Hole Punching in js-libp2p and bring it to parity with the Go and Rust implementations.

- [ ] https://github.com/libp2p/js-libp2p/issues/1005
- [ ] https://github.com/libp2p/js-libp2p/issues/1029
- [ ] https://github.com/libp2p/js-libp2p/issues/1459
- [ ] https://github.com/libp2p/js-libp2p/issues/1460

momack2:
can we clarify that this is for js-ipfs in the title please?

p-shahi:
@momack2 Any reason to include js-ipfs specifically? This work is not progressing with it in mind. We will need it for Helia but I would still not include any project in the title since it's better to remain product agnostic

momack2:
sorry - I think I meant "js-libp2p" - the reason is because each of these cards shows up with out the repo title in starmaps - so otherwise you have many "holepunching" milesotnes without knowing they are for different libp2p implementations

SgtPooki:
@momack2 we could include the repo-name on milestone cards in starmap to help mitigate this problem?

p-shahi:
Didn't see the replies above but I support adding the repo-name to the milestone cards

momack2:
Yeah, I think that would help a lot

SgtPooki:
<details>
<summary>JS Colo 2024 discussion</summary>

We don't do hole-punching in TCP, hole-punching is much more successful with UDP (timing is more reliable)

TCP: ~30% success
UDP: ~70% success

</details>


### Things left to do:

1. Incorporate [QUIC transport](https://github.com/libp2p/js-libp2p/issues/1459)
2. Implement missing parts of DCUtR (the half after unilateral connection upgrade)

jost-s:
[...]

Issue #1005, opened 2021-10-01
Autonat support


achingbrain:
An implementation of Autonat compatible with https://github.com/libp2p/specs/tree/master/autonat and https://github.com/libp2p/go-libp2p-autonat is required.

See also: https://docs.libp2p.io/concepts/nat/

tinytb:
backlog grooming: nearly done, but needs some time spent on it.

PR #1533, opened 2023-01-05
feat: implement circuit v2

ckousik:
Reopen the earlier circuit v2 implementation by @mpetrunic with a few fixes to ensure it is up to date with libp2p.

@mpetrunic looks mostly okay, but could you do a rough pass to make sure I haven't missed anything?

Todo:
- [x] Interop testing https://github.com/libp2p/interop/pull/83
- [x] Update relayv2 to proto3: https://github.com/libp2p/specs/issues/509


src/circuit/relay.ts
```diff
@@ -0,0 +1,108 @@
...
+  enabled?: boolean
+  ttl?: number
+}
+
+export interface HopConfig {
+  enabled?: boolean
+  active?: boolean
+}
+
+export interface AutoRelayConfig {
```
mpetrunic, line 29: Do we still use this? We should either remove or rename it

p-shahi:
Is there anything blocking the PR review?

mpetrunic:
> Is there anything blocking the PR review?

I think todo's are:
- revisit configs
- add go interop tests

But I think this shouldn't block reviews^^

achingbrain (changes requested):
First pass of reviewing

- The `RelayConfig`, `HopConfig` `RelayAdvertiseConfig` and `AutoRelayConfig` interface fields all need jsdoc comments to explain what they do
- The `doc/CONFIGURATION.md` sections detailing relay config needs updating

src/circuit/relay.ts
```diff
@@ -0,0 +1,108 @@
...
+import {
+  RELAY_RENDEZVOUS_NS
+} from './constants.js'
+import type { AddressSorter } from '@libp2p/interface-peer-store'
+import type { Startable } from '@libp2p/interfaces/startable'
+import type { Components } from '../components.js'
+
+const log = logger('libp2p:relay')
+
+export interface RelayAdvertiseConfig {
```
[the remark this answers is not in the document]
ckousik: Actually, both those files seem almost identical. Seems like the merge caused some issues. Is it okay to rewrite this?

2color:
I'm curious how this PR relates to AutoNAT (https://github.com/libp2p/js-libp2p/issues/1005).

Isn't AutoNAT generally a prerequisite prior to doing NAT hole punching?

p-shahi:
> I'm curious how this PR relates to AutoNAT (#1005).
> 
> Isn't AutoNAT generally a prerequisite prior to doing NAT hole punching?

@2color Yes AutoNAT is a requirement. You can see our roadmap issue for what's all needed: https://github.com/libp2p/js-libp2p/issues/1461
RelayV2 is required for WebRTC browser-to-browser as well which is why we're doing this right now.

.gitignore
```diff
@@ -4,3 +4,6 @@ dist
 .coverage
 package-lock.json
 yarn.lock
+
+.coverage
```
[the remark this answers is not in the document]
ckousik: fixed

src/circuit/client.ts
```diff
@@ -1,57 +1,88 @@
...
     })
-    this.components.connectionManager.addEventListener('peer:disconnect', this._onPeerDisconnected)
+
+    this.components.connectionManager.addEventListener('peer:disconnect', (evt) => {
+      this._onPeerDisconnected(evt)
+    })
+
+    this.components.connectionManager.addEventListener('peer:connect', (evt) => {
+      log.trace('Connected', evt.detail.remotePeer.toString(), this.components.peerId.toString())
+    })
```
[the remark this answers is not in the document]
ckousik: This seems to have been left in here for logging. I'll remove it.

src/circuit/client.ts
```diff
@@ -60,51 +91,48 @@ export class AutoRelay {
...
+  async _onProtocolChange ({ peerId, protocols }: { peerId: PeerId, protocols: string[] }) {
     const id = peerId.toString()
 
+    if (peerId.equals(this.components.peerId)) {
+      return
+    }
+
     // Check if it has the protocol
-    const hasProtocol = protocols.find(protocol => protocol === RELAY_CODEC)
+    const hasProtocol = protocols.find(protocol => protocol === RELAY_V2_HOP_CODEC)
```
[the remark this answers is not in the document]
ckousik: @achingbrain There are a lot of instances where undefined/null is used for error cases. I will rewrite these to throw explicit errors.

src/circuit/client.ts
```diff
@@ -60,51 +91,48 @@ export class AutoRelay {
...
     }
 
     // If protocol, check if can hop, store info in the metadataBook and listen on it
     try {
       const connections = this.components.connectionManager.getConnections(peerId)
 
-      if (connections.length === 0) {
+      const connection = connections[0]
+
+      if (connection == null) {
```
ckousik, line 124: would the connection manager return a falsy object in the result?

src/circuit/client.ts
```diff
@@ -196,47 +231,52 @@ export class AutoRelay {
...
   async _listenOnAvailableHopRelays (peersToIgnore: string[] = []) {
-    // TODO: The peer redial issue on disconnect should be handled by connection gating
     // Check if already listening on enough relays
-    if (this.listenRelays.size >= this.maxListeners) {
+    if (this.relays.size >= this.maxReservations) {
       return
     }
 
     const knownHopsToDial = []
     const peers = await this.components.peerStore.all()
```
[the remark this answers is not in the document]
ckousik: Yes, and most of the loop can be avoided using a filter.

src/circuit/v1/protocol/index.proto
```diff
@@ -0,0 +1,42 @@
...
+
+  enum Type { // RPC identifier, either HOP, STOP or STATUS
+    HOP = 1;
+    STOP = 2;
+    STATUS = 3;
+    CAN_HOP = 4;
+  }
+
+  message Peer {
+    required bytes id = 1;    // peer id
```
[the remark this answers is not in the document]
ckousik: I seem to have broken this before the merge. `proto3` cannot be used for this spec since enums in proto3 need to start with 0, and the v1 spec does not allow this https://github.com/libp2p/specs/blob/master/relay/circuit-v1.md . I would have to add a dummy value for 0.

src/circuit/relay.ts
ckousik: We use it to automatically initiate a reservation on newly connected peers or if the circuit relay protocol is added to an existing peer.

package.json
```diff
@@ -91,7 +92,8 @@
     "test:firefox": "aegir test -t browser -f \"./dist/test/**/*.spec.js\" -- --browser firefox",
     "test:firefox-webworker": "aegir test -t webworker -f \"./dist/test/**/*.spec.js\" -- --browser firefox",
     "test:examples": "cd examples && npm run test:all",
-    "test:interop": "aegir test -t node -f dist/test/interop.js"
+    "test:interop": "aegir test -t node -f dist/test/interop.js",
+    "test:relay": "aegir test -t node -f \"./dist/test/**/relay.{node,spec}.js\""
```
ckousik, line 96: TODO: Remove this before merging. This is in place to ease testing.

test/relay/relay.node.ts
```diff
@@ -192,8 +198,14 @@ describe('Dialing (via relay, TCP)', () => {
     const dialAddr = relayAddr.encapsulate(`/p2p/${relayLibp2p.peerId.toString()}`)
 
     const connection = await srcLibp2p.dial(dialAddr)
-    const stream = await connection.newStream('/libp2p/circuit/relay/0.1.0')
-
+    /* eslint-disable-next-line no-console */
+    console.log('>>>>>> connection established')
```
ckousik, line 202: NOTE: If the connection is established, `connection.newStream` fails because of a reset.

src/circuit/transport.ts
```diff
@@ -90,102 +128,122 @@ export class Circuit implements Transport, Startable {
...
-    return 'libp2p/circuit-relay-v1'
+    return 'libp2p/circuit-relay-v2'
   }
 
-  async _onProtocol (data: IncomingStreamData) {
+  getPeerConnection (dstPeer: PeerId): Connection | undefined {
+    return this.components.connectionManager.getConnections(dstPeer)[0] ?? undefined
+  }
+
+  async _onProtocolV1 (data: IncomingStreamData) {
```
ckousik, line 138: @mpetrunic Should we disable circuit v1 relaying entirely? There seems to be no reason to register this handler as it will not relay.

src/circuit/index.ts
```diff
@@ -47,88 +14,16 @@ export interface AutoRelayConfig {
   maxListeners: number
 }
 
-export interface RelayInit extends RelayConfig {
-  addressSorter?: AddressSorter
+export interface HopConfig {
+  enabled?: boolean
+  active?: boolean
```
ckousik, line 19: How are active and enabled different? Also, we only use `enabled` in `Relay` to decide whether or not to advertise the node as a relay.

src/circuit/index.ts
```diff
@@ -63,8 +62,6 @@ export interface RelayComponents {
 export class Relay implements Startable {
   private readonly components: RelayComponents
   private readonly init: RelayInit
-  // @ts-expect-error this field isn't used anywhere?
-  private readonly autoRelay?: AutoRelay
```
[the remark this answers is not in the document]
ckousik: This is not required here, but it is useful for enabling the `CircuitService` and the configuration has also been renamed.

src/circuit/relay.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const log = logger('libp2p:circuit:relay')
+
+export interface RelayAdvertiseConfig {
+  bootDelay?: number
+  enabled?: boolean
+  ttl?: number
+}
+
+export interface HopConfig {
```
ckousik, line 24: This is duplicated.

src/circuit/transport.ts
```diff
@@ -62,19 +80,39 @@ export class Circuit implements Transport, Startable {
...
+        log.error(err)
+      })
+    })
+      .catch(err => {
+        log.error(err)
+      })
 
-    await this.components.registrar.handle(RELAY_CODEC, (data) => {
-      void this._onProtocol(data).catch(err => {
+    await this.components.registrar.handle(RELAY_V2_HOP_CODEC, (data) => {
```
ckousik, line 92: Should adding this listener be conditional on `HopConfig.enabled | HopConfig.active === true` ?

src/circuit/v2/reservation-store.ts
```diff
@@ -0,0 +1,67 @@
...
+export class ReservationStore implements IReservationStore, Startable {
+  private readonly reservations = new Map<string, Reservation>()
+  private _started = false;
+  private interval: any
+
+  /**
+   * @param limit - maximum number of reservations to store
+   * @param reservationClearInterval - interval to check for expired reservations in millisecons
+   */
+  constructor (private readonly limit = 15, private readonly reservationClearInterval = 300 * 1000) {
```
ckousik, line 21: Move these values to `constants.ts`.

src/circuit/transport.ts
mpetrunic: Yeah, sure! I was just copying logic that go-libp2p had but I'm pretty sure nowadays, there is no need for that!

src/circuit/transport.ts
mpetrunic: Since circuit v2 kinda changed interaction, maybe it makes sense to make up new configs and names that are more dev friendly

src/circuit/client.ts
```diff
@@ -1,57 +1,100 @@
...
-  maxListeners?: number
+  /**
+   * A callback to invoke when an error occurs in the circuit service.
+   */
   onError?: (error: Error, msg?: string) => void
 }
 
-export class AutoRelay {
-  private readonly components: RelayComponents
+export interface CircuitServiceEvents {
```
ckousik, line 42: @mpetrunic I feel `AutoRelay` is a better name for this class than `CircuitService`. `CircuitService` feels like it should provide the relay service, but in this case it automatically connects to relays. Would `AutoReserve` be a better name?

mpetrunic: RelayReservationManager?

src/circuit/v2/stream-handler.ts
```diff
@@ -0,0 +1,82 @@
...
+   */
+  stream: Stream
+
+  /**
+   * max bytes length of message
+   */
+  maxLength?: number
+}
+
+export class StreamHandlerV2 {
```
achingbrain, line 21: I think you can remove this class by using [it-pb-stream](https://www.npmjs.com/package/it-pb-stream) to read/write pb encoded messages from the stream.

test/circuit/v2/hop.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+      expect(response.status).to.be.equal(Status.OK)
+    })
+
+    it('should fail to connect - invalid request', async function () {
+      await handleHopProtocol({
+        connection: conn,
+        streamHandler,
+        request: {
+          type: HopMessage.Type.CONNECT,
+          // @ts-expect-error
```
achingbrain, line 246: Please add a comment as to what error is expected

test/circuit/v2/hop.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+        request: {
+          type: HopMessage.Type.CONNECT,
+          peer: {
+            id: dstPeer.toBytes(),
+            addrs: []
+          }
+        },
+        reservationStore,
+        circuit,
+        // @ts-expect-error
```
achingbrain, line 274: Please add a comment as to what error is expected

test/dialing/resolver.spec.ts
```diff
@@ -100,6 +117,10 @@ describe('Dialing (resolvable addresses)', () => {
     // Resolver stub
     resolver.onCall(0).returns(Promise.resolve(getDnsRelayedAddrStub(remoteId)))
 
+    // create reservation on relay
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    await pEvent(remoteLibp2p.circuitService!, 'relay:reservation')
```
achingbrain, line 122: Please do not disable linting.

```suggestion
    if (remoteLibp2p.circuitService == null) {
      throw new Error('remote libp2p had no circuit service')
    }

    // create reservation on relay
    await pEvent(remoteLibp2p.circuitService, 'relay:reservation')
```

test/dialing/resolver.spec.ts
```diff
@@ -114,6 +135,17 @@ describe('Dialing (resolvable addresses)', () => {
...
 
+    const { default: Peers } = await import('../fixtures/peers.js')
+
+    // Use the last peer
+    const relayId = await createFromJSON(Peers[Peers.length - 1])
+    // ensure remote libp2p creates reservation on relay
+    await remoteLibp2p.components.peerStore.protoBook.add(relayId, [RELAY_V2_HOP_CODEC])
+
+    // create reservation on relay
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
achingbrain, line 146: Please do not disable linting.

test/dialing/resolver.spec.ts
```diff
@@ -173,6 +205,17 @@ describe('Dialing (resolvable addresses)', () => {
...
```
achingbrain, line 216: Please do not disable linting.

test/relay/relay.node.ts
```diff
@@ -1,19 +1,18 @@
...
 import type { Libp2pNode } from '../../src/libp2p.js'
-import all from 'it-all'
-import { RELAY_CODEC } from '../../src/circuit/multicodec.js'
-import { StreamHandler } from '../../src/circuit/circuit/stream-handler.js'
-import { CircuitRelay } from '../../src/circuit/pb/index.js'
+import { createNode } from '../utils/creators/peer.js'
 import { createNodeOptions, createRelayOptions } from './utils.js'
-import delay from 'delay'
+import all from 'it-all'
+// import delay from 'delay'
```
achingbrain, line 13: Please remove

test/relay/relay.node.ts
```diff
@@ -157,32 +159,30 @@ describe('Dialing (via relay, TCP)', () => {
...
-    const streamHandler = new StreamHandler({ stream })
-    streamHandler.write({
-      type: CircuitRelay.Type.STATUS
-    })
-    const res = await streamHandler.read()
-    expect(res?.code).to.equal(CircuitRelay.Status.MALFORMED_MESSAGE)
-    streamHandler.close()
+    // this should fail as the destination peer has HOP disabled
+    await expect(connections[0].newStream(RELAY_V2_HOP_CODEC))
+      .to.be.rejected()
```
achingbrain, line 164: Please add an assertion to ensure the correct error occurs:


```suggestion
    await expect(connections[0].newStream(RELAY_V2_HOP_CODEC)).to.be.rejected()
      .with.property('code').that.equals('ERR_SOME_ERROR_CODE')
```

test/relay/relay.node.ts
```diff
@@ -157,32 +159,30 @@ describe('Dialing (via relay, TCP)', () => {
...
+    // so we make the message invalid by adding a zeroed byte
 
     // should still be connected
     const dstToRelayConn = dstLibp2p.components.connectionManager.getConnections(relayLibp2p.peerId)
     expect(dstToRelayConn).to.have.lengthOf(1)
     expect(dstToRelayConn).to.have.nested.property('[0].stat.status', 'OPEN')
   })
 
-  it('should time out when establishing a relay connection', async () => {
+  /* eslint-disable-next-line no-warning-comments */
```
achingbrain, line 174: Please do not disable linting

test/relay/relay.node.ts
```diff
@@ -192,13 +192,17 @@ describe('Dialing (via relay, TCP)', () => {
...
-    await stream.sink(async function * () {
-      // delay for longer than the timeout
-      await delay(1000)
-      yield Uint8Array.from([0])
-    }())
+    // this should succeed as the timeout is only effective after
+    // multistream select negotiates the protocol
+    const stream = await connection.newStream([RELAY_V2_HOP_CODEC])
+
+    /* eslint-disable-next-line no-warning-comments */
```
achingbrain, line 199: Please do not disable linting

src/circuit/client.ts
```diff
@@ -0,0 +1,362 @@
...
+  private started: boolean
+
+  constructor (components: Components, init: CircuitServiceInit) {
+    super()
+    this.started = false
+    this.components = components
+    this.addressSorter = init.addressSorter ?? publicAddressesFirst
+    this.maxReservations = init.maxReservations ?? DEFAULT_MAX_RESERVATIONS
+    this.relays = new Set()
+    this.reservationMap = new Map()
```
achingbrain, line 66: You can simplify the code by using the `PeerSet` and `PeerMap` classes from [@libp2p/peer-collections](https://www.npmjs.com/package/@libp2p/peer-collections) - this way you don't need to keep stringifying PeerIds to use them as keys.

src/circuit/client.ts
```diff
@@ -0,0 +1,362 @@
...
+/**
+ * ReservationManager automatically makes a circuit v2 reservation on any connected
+ * peers that support the circuit v2 HOP protocol.
+ */
+export class RelayReservationManager extends EventEmitter<RelayReservationManagerEvents> implements Startable {
+  private readonly components: Components
+  private readonly addressSorter: AddressSorter
+  private readonly maxReservations: number
+  private readonly relays: Set<string>
+  private readonly reservationMap: Map<PeerId, ReturnType<typeof setTimeout>>
```
achingbrain, line 55: Using PeerId objects as a key can be unsafe as they don't pass SameValueZero equality.

E.g.:

```ts
const peerString = 'a-peer-id-string'

const p1 = peerIdFromString(peerString)
const p2 = peerIdFromString(peerString)

console.info(p1 === p2) // false
```

Use `PeerMap` from [@libp2p/peer-collections](https://www.npmjs.com/package/@libp2p/peer-collections) instead.

achingbrain (changes requested):

test/relay/relay.node.ts
ckousik: This is a flaky test that I'm trying to solve. Would it be okay to leave the TODO comments in this case?

test/relay/relay.node.ts
ckousik: This is a flaky test that I'm trying to solve. Would it be okay to leave the TODO comments in this case?

achingbrain: TODOs are fine 👍

src/circuit/v2/stream-handler.ts
```diff
@@ -0,0 +1,82 @@
...
+  private readonly shake: Handshake
+  private readonly decoder: Source<Uint8Array>
+
+  constructor (options: StreamHandlerOptions) {
+    const { stream, maxLength = 4096 } = options
+
+    this.stream = stream
+    this.shake = handshake(this.stream)
+    // @ts-expect-error some type incompatibilities
+    this.decoder = lp.decode.fromReader(this.shake.reader, { maxDataLength: maxLength })
```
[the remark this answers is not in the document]
ckousik: I used it-pb-stream to get rid of the class

src/circuit/client.ts
```diff
@@ -0,0 +1,360 @@
...
+    this.addressSorter = init.addressSorter ?? publicAddressesFirst
+    this.maxReservations = init.maxReservations ?? DEFAULT_MAX_RESERVATIONS
+    this.relays = new PeerSet()
+    this.reservationMap = new PeerMap()
+    this.onError = init.onError ?? noop
+
+    this._onProtocolChange = this._onProtocolChange.bind(this)
+    this._onPeerDisconnected = this._onPeerDisconnected.bind(this)
+
+    this.components.peerStore.addEventListener('change:protocols', (evt) => {
```
mpetrunic, line 73: It recently surfaced that this won't trigger in case node restarts or peer reconnects as protocols won't be changed. I think we need to handle connect event also: https://github.com/libp2p/js-libp2p/blob/5a62f4f22f3d9658ae7356568283d443a76df4ea/src/registrar.ts#L41

src/circuit/multicodec.ts
```diff
@@ -1,2 +1,4 @@
 
-export const RELAY_CODEC = '/libp2p/circuit/relay/0.1.0'
+export const RELAY_V1_CODEC = '/libp2p/circuit/relay/0.1.0'
```
mpetrunic, line 2: didn't we say we are going to remove all v1 stuff?

achingbrain: Yes, all v1 code should be removed.

ckousik: We still allow connecting over v1 which is why this is still in here. Handling relay v1 has been disabled completely but the code has not been removed.

ckousik: All v1 code has been removed

src/circuit/client.ts
ckousik: This case has been handled.

MarcoPolo:
@achingbrain seems like you already took a review pass at this. Let me know if my review would still be helpful

src/circuit/v2/hop.ts
```diff
@@ -0,0 +1,221 @@
...
+      reservation: await makeReservation(relayAddrs, relayPeer, connection.remotePeer, BigInt(result.expire ?? 0)),
+      limit
+    })
+    log('sent confirmation response to %s', connection.remotePeer)
+  } catch (err) {
+    log.error('failed to send confirmation response to %s', connection.remotePeer)
+    await reservationStore.removeReservation(connection.remotePeer)
+  }
+
+  // TODO: how to ensure connection manager doesn't close reserved relay conn
```
achingbrain, line 110: > how to ensure connection manager doesn't close reserved relay conn

This can be done by tagging the relay peer with a value > 0

src/circuit/v2/hop.ts
```diff
@@ -0,0 +1,221 @@
...
+    return
+  }
+
+  hopstr.write({ type: HopMessage.Type.STATUS, status: Status.OK })
+  const sourceStream = pbstr.unwrap()
+
+  log('connection to destination established, short circuiting streams...')
+  // Short circuit the two streams to create the relayed connection
+  return await pipe(
+    sourceStream as Duplex<Uint8ArrayList, Uint8Array | Uint8ArrayList>,
```
achingbrain, line 180: We should probably be counting transferred bytes here to enforce the data transfer limits.

achingbrain, line 180: > sourceStream as Duplex<Uint8ArrayList, Uint8Array | Uint8ArrayList>,

Why is the casting necessary?  Is there a way we can avoid it?

achingbrain: Ah, I think it's because the it-pb-stream unwrap method doesn't return the type of the wrapped stream.

src/circuit/v2/hop.ts
ckousik: Are you refering to https://github.com/libp2p/js-libp2p/blob/master/src/connection-manager/index.ts#L603 ?

achingbrain: I'm referring to [peerStore.tagPeer](https://libp2p.github.io/js-libp2p-interfaces/interfaces/_libp2p_interface_peer_store.PeerStore.html#tagPeer), I think the `TODO:` is referring to that method in the connection manager.

achingbrain:
So far so good, only a couple of things to tidy up I think:

Looking through the spec, something missing here is enforcing limits on data transfer and reservation length.  This will need to be implemented & tested before merging.

Also [tagging relay peers](https://github.com/libp2p/js-libp2p/pull/1533#pullrequestreview-1310315371) so we don't close connections to them under load.

src/circuit/v2/hop.ts
ckousik: Implementing now

ckousik:
> So far so good, only a couple of things to tidy up I think:
> 
> Looking through the spec, something missing here is enforcing limits on data transfer and reservation length. This will need to be implemented & tested before merging.
> 
> Also [tagging relay peers](https://github.com/libp2p/js-libp2p/pull/1533#pullrequestreview-1310315371) so we don't close connections to them under load.

Writing the tests right now.

ckousik:
@achingbrain I added transfer limits and tests for it. For data limits, the Go code discards any data read (silently) after the limit is exceeded, while for the duration limit, it fails with a `ErrDeadlineExceeded` once the duration is completed. In the js case, the current implementation closes the source/sink silently once either limit is exceeded. Is this the required  behaviour?

package.json
```diff
@@ -91,7 +91,8 @@
     "test:firefox": "aegir test -t browser -f \"./dist/test/**/*.spec.js\" -- --browser firefox",
     "test:firefox-webworker": "aegir test -t webworker -f \"./dist/test/**/*.spec.js\" -- --browser firefox",
     "test:examples": "cd examples && npm run test:all",
-    "test:interop": "aegir test -t node -f dist/test/interop.js"
+    "test:interop": "aegir test -t node -f dist/test/interop.js",
+    "test:relay": "aegir test -t node -f \"./dist/test/**/{hop,stop,reservation-store,relay,auto-relay}.{node,spec}.js\" --cov"
```
ckousik, line 95: Will remove before merging

achingbrain: Please can you refrain from adding test code, temporary code, or commented out code etc.  We're all human and these things are too easy to miss.

When ready for review, PRs should be ready to merge and should not need extra pushes after review.

ckousik: Okay, removed

src/circuit/v2/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const RELAYED_TAG = 'relayed'
+
+// circuit v2 connection limits
+// 2 min is the default connection duration
+export const DEFAULT_DURATION_LIMIT = 2 * 60 * 1000
+// 128k is the default data limit
+export const DEFAULT_DATA_LIMIT = 1 << 17
```
ckousik, line 7: constants copied from https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/circuitv2/relay/resources.go#L61-L66

achingbrain: It's worth adding this as a comment in the code for people looking at it later

achingbrain:
> For data limits, the Go code discards any data read (silently) after the limit is exceeded, while for the duration limit, it fails with a ErrDeadlineExceeded once the duration is completed. In the js case, the current implementation closes the source/sink silently once either limit is exceeded. Is this the required behaviour?

Having talked with @Jorropo and @mxinden this isn't the correct behaviour - I've opened https://github.com/libp2p/specs/pull/526 to document what should happen - the stream should be reset.

src/circuit/v2/util.ts
```diff
@@ -0,0 +1,87 @@
...
+const log = logger('libp2p:circuit:v2:util')
+
+type DuplexStream = Duplex<Uint8ArrayList, Uint8ArrayList | Uint8Array>
+export async function createLimitedShortCircuit (source: DuplexStream, destination: DuplexStream, limit?: Limit) {
+  // trivial case
+  if (limit == null) {
+    return await pipe(source, destination, source)
+  }
+  const src = durationLimitDuplex(dataLimitDuplex(source, Number(limit.data ?? 0)), limit.duration ?? 0)
+  const dst = durationLimitDuplex(dataLimitDuplex(destination, Number(limit.data ?? 0)), limit.duration ?? 0)
```
achingbrain, line 18: We shouldn't downcast from `BigInt` to `Number`, otherwise we can't satisfy the limits that may be set.

src/circuit/v2/util.ts
```diff
@@ -0,0 +1,87 @@
+import { abortableDuplex, AbortError } from 'abortable-iterator'
+import { pipe } from 'it-pipe'
```
achingbrain, line 2: `it-pipe` introduces a lot of latency as every operation in the pipeline crosses an async boundary even if the operation is synchronous.  This is likely to be a very hot code path so that latency will be expensive.

src/circuit/index.ts
```diff
@@ -1,135 +1,51 @@
...
-}
-
-export class Relay implements Startable {
-  private readonly components: RelayComponents
-  private readonly init: RelayInit
-  // @ts-expect-error this field isn't used anywhere?
-  private readonly autoRelay?: AutoRelay
-  private timeout?: any
-  private started: boolean
+  enabled?: boolean
```
achingbrain, line 23: Under what circumstances would a user want to disable the reservation manager?

Are there tests for when it's disabled to ensure libp2p can run without it?

ckousik: We would want to disable the reservation manager when we do not want to automatically listen on relays. It replaces the old name `AutoRelay`. https://github.com/libp2p/js-libp2p/pull/1533#discussion_r1086978369.

The relay tests disable this by default https://github.com/libp2p/js-libp2p/pull/1533/files#diff-81971332c1841fd8324647bd37da55f446e332af9e07d4f28632d5768302dac9R27

src/circuit/v2/util.ts
```diff
@@ -0,0 +1,100 @@
...
+import { TimeoutController } from 'timeout-abort-controller'
+import { Uint8ArrayList } from 'uint8arraylist'
+import type { Limit } from './pb/index.js'
+import { logger } from '@libp2p/logger'
+
+const log = logger('libp2p:circuit:v2:util')
+
+type DuplexStream = Duplex<Uint8ArrayList, Uint8ArrayList | Uint8Array>
+const doRelay = (src: DuplexStream, dst: DuplexStream) => {
+  queueMicrotask(() => {
```
ckousik, line 12: These are spawned as microtasks to allow for synchronous sources:

consider `dst.sink(src.source); src.sink(dst.source)`. `dst.sink` will block until `src.source` completes if `src.source` is synchronous.

src/circuit/transport.ts
```diff
@@ -63,126 +78,114 @@ export class Circuit implements Transport, Startable {
...
   get [Symbol.toStringTag] () {
-    return 'libp2p/circuit-relay-v1'
+    return 'libp2p/circuit-relay-v2'
   }
 
-  async _onProtocol (data: IncomingStreamData) {
-    const { connection, stream } = data
+  async onHop ({ connection, stream }: IncomingStreamData) {
+    log('received circuit v2 hop protocol stream from %s', connection.remotePeer)
     const controller = new TimeoutController(this._init.hop.timeout)
```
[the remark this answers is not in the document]
ckousik: This is the time in which the hop protocol must be completed. The hop timeout is much lower than the transport timeout limit, and is only used while 1) a new reservation is being created, or 2) a client is attempting to connect over the relay. Once the operations are complete, the timeout is cleared.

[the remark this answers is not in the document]
ckousik: I think I can avoid the abortableDuplex though.

test/circuit/v2/hop.spec.ts
```diff
@@ -0,0 +1,741 @@
...
+      // source to dest, write 4 bytes
+      const sender = pushable()
+      void pipe(sender, sourceStream)
+      sender.push(uint8arrayFromString('01234'))
+      // source to dest, exceed stream limit
+      sender.push(uint8arrayFromString('extra'))
+      const data = await all(destStream.source)
+      expect(data).to.have.length(1)
+      expect(data[0]).to.have.length(5)
+      expect(srcServerAbort.callCount).to.equal(1)
```
achingbrain, line 507: This sends five bytes, not four.  When four bytes are sent the test fails.

The dialler should send `0123` then `extra`, then the relay should relay five bytes before reset the stream.

Also, the test needs to assert that both streams were reset - from the dialler to the relay and from the relay to the listener.

src/circuit/v2/util.ts
```diff
@@ -16,85 +16,102 @@ const doRelay = (src: DuplexStream, dst: DuplexStream) => {
...
   const dst = durationLimitDuplex(dataLimitDuplex(destination, dataLimit), durationLimit)
-  doRelay(src.value, dst.value)
+
+  doRelay(src, dst)
 }
 
-const dataLimitSource = (source: Source<Uint8ArrayList>, abort: (err: Error) => void, limit: bigint): Source<Uint8ArrayList> => {
-  if (limit === BigInt(0)) {
-    return source
+const dataLimitSource = (stream: Stream, limit: bigint): Stream => {
```
ckousik, line 34: Is this safe? While the implementation of `it-pb-stream` returns the underlying stream, the return type only guarantees returning a `Duplex`. The stream functions could be erased from the returned value in a future release without changing the API, which would cause this to break.

achingbrain: `3.0.1` changes the `unwrap` return type to be the type of the wrapped stream, so it should be safe.

test/circuit/v2/hop.spec.ts
```diff
@@ -0,0 +1,741 @@
...
+      // dest to source, write 4 bytes
+      const sender = pushable()
+      void pipe(sender, destStream)
+      sender.push(uint8arrayFromString('01234'))
+      // dest to source, exceed stream limit
+      sender.push(uint8arrayFromString('extra'))
+      const data = await all(sourceStream.source)
+      expect(data).to.have.length(1)
+      expect(data[0]).to.have.length(5)
+      expect(dstServerStreamAbortStub.callCount).to.equal(1)
```
achingbrain, line 581: Same comment as elsewhere, this sends five bytes not four and the reset needs to be asserted on in both directions.

src/circuit/v2/hop.ts
```diff
@@ -0,0 +1,219 @@
...
+    // result.expire is non-null if `ReservationStore.reserve` returns with status == OK
+    if (result.expire != null) {
+      const ttl = new Date().getTime() - result.expire
+      await peerStore.tagPeer(relayPeer, RELAYED, { value: 1, ttl })
+        .catch((err: CodeError) => {
+          // ignore if peer is already tagged
+          if (err.code !== 'ERR_DUPLICATE_TAG') {
+            throw err
+          }
+        })
```
achingbrain, line 111: If the peer is already tagged, this should update the TTL of the tag if it's going to expire before the reservation rather than swallowing the error.

Also, I can't see any tests for this?

achingbrain (changes requested):
Nearly there, comments inline.

test/circuit/v2/hop.spec.ts
ckousik: Fixed. I was resetting the stream if the sender attempted to send more than the limit.

> Also, the test needs to assert that both streams were reset - from the dialler to the relay and from the relay to the listener.

The relay enforces the limit, so the abort is called from the relay to the dialler and from the relay to the listener. This is checked.

src/circuit/v2/util.ts
```diff
@@ -0,0 +1,140 @@
...
+      let total = 0n
+
+      for await (const buf of source) {
+        const len = BigInt(buf.byteLength)
+        if (total + len > limit) {
+          // this is a safe downcast since len is guarantee to be in the range for a number
+          const remaining = Number(limit - total)
+          try {
+            if (remaining !== 0) {
+              yield buf.slice(0, remaining)
```
achingbrain, line 86:
```suggestion
              yield buf.subarray(0, remaining)
```

`.slice` is a copy operation on `Uint8Array`s (it wasn't on node `Buffer`s)

src/circuit/reservation-store.ts
```diff
@@ -38,8 +46,10 @@ export class ReservationStore implements IReservationStore, Startable {
     this.init = {
       maxReservations: options?.maxReservations ?? 15,
       reservationClearInterval: options?.reservationClearInterval ?? 300 * 1000,
-      applyDefaultLimit: options?.applyDefaultLimit === false,
```
achingbrain: This is fun, it would have turned every js-libp2p node with the relay server enabled into an unlimited relay 😱

achingbrain (commented):
I am going to merge this which will publish an RC to unblock other efforts as it is functional though it still needs some work in a follow-up before being released.

- Split the code into two parts - the transport and the relay server
  - At the moment the transport handles some of the server responsibilities
  - The hop/stop implementations accept a load of context as arguments that might be easier to reason about as a class
- Configure the transport and relay server separately - at the moment the server is partially configured under the `config.relay.hop` key and `config.relay.advertise` - the fact that the docs say "Does not make you a relay" in several places mean it's really not obvious
- Maybe even have a separate CircuitRelay transport that you can configure
- More tests

Issue #1540, opened 2023-01-10
Fix Flakey Tests on CI for PubSub examples


maschad:
#### Severity:

Low

#### Description:

On occasion the [test-example (pubsub)](https://github.com/libp2p/js-libp2p/actions/runs/3861945664/jobs/6583221048#logs) fails with [messages not being received](https://github.com/libp2p/js-libp2p/actions/runs/3861945664/jobs/6583221048#step:5:27) I suspect we could try increasing the delay 

#### Steps to reproduce the error:

Not necessarily easily reproducible as it doesn't happen frequently

tabcat:
I had similar issues and fixed by listening for `subscription-change` events 
https://github.com/opalsnt/welo/commit/3f60808cde11f90f4789a298426e6fdcf9603f98

for some reason even with adding long timeouts some nodes would never peer otherwise

achingbrain:
I think instead of increasing the delay a better fix in the message filtering test might be for `node1` to wait until it sees `node2` and `node3` in it's peer list for the `'fruit'` topic before publishing the messages?

maschad:
I was originally confused by the [logs](https://github.com/libp2p/js-libp2p/actions/runs/3891376024/jobs/6641677031#step:5:25) saying `Error: Not enough banana messages - received 1, expected 2` even though [two _banana_ messages](https://github.com/libp2p/js-libp2p/actions/runs/3891376024/jobs/6641677031#step:5:25) had been received. 

> I think instead of increasing the delay a better fix in the message filtering test might be for `node1` to wait until it sees `node2` and `node3` in it's peer list for the `'fruit'` topic before publishing the messages?

What's interesting is that this approach doesn't seem to work unless `node3` dials `node1` directly,  i.e. `node3` doesn't show up in `node1` 's subscribers  unless `node3` dials `node1` directly (see https://github.com/libp2p/js-libp2p/pull/1543/commits/7a6bff35383c291b27455bf142a1eacf6af7c766#diff-9483cec2def56f6a8975c97c997d70741698e22fb8a971785d83502f7169bbf1R42)

PR #1543, opened 2023-01-11
fix: Update PubSub Example test

maschad:
Closes https://github.com/libp2p/js-libp2p/issues/1540

### DIFF
--- a/examples/pubsub/message-filtering/1.js
+++ b/examples/pubsub/message-filtering/1.js
@@ -28,7 +28,7 @@ const createNode = async () => {
   const [node1, node2, node3] = await Promise.all([
     createNode(),
     createNode(),
-    createNode(),
+    createNode()
   ])
 
   // node1 conect to node2 and node2 conect to node3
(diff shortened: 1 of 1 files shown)


achingbrain:
Ah, whoops - I didn't see this sorry.

It's not necessary for node1 and node3 to be connected as node2 should relay messages to node3.

Instead we need to wait for node1 to know node2 is subscribed to the topic, and that node2 knows node3 is subscribed to the topic.

See the fix here: https://github.com/libp2p/js-libp2p/pull/1549

maschad:
> Ah, whoops - I didn't see this sorry.
> 
> It's not necessary for node1 and node3 to be connected as node2 should relay messages to node3.
> 
> Instead we need to wait for node1 to know node2 is subscribed to the topic, and that node2 knows node3 is subscribed to the topic.
> 
> See the fix here: #1549

Understood. I thought that `node1` needed to also have `node3` in it's subscriber list in order for the messages to be relayed to `node3`

PR #1548, opened 2023-01-14
fix: allow dialing multiaddrs without peer ids

achingbrain:
Most multiaddrs have a peer id, e.g. `/ip4/123.123.123.123/tcp/234/p2p/Qmfoo` but some cannot, e.g. `/unix/tmp/app.sock` - we should still be able to dial these addresses with the caveat that the dialer shoul have some other way of obtaining the peer id to ensure they aren't being man-in-the-middled.

This also allows dialing `dnsaddrs` that are configured as round-robin addresses for load balancing.


mxinden:
Cross-referencing https://github.com/libp2p/specs/issues/458 here for future us.

PR #1556, opened 2023-01-16
docs: Update the peer discovery docs

maschad:
Following https://github.com/libp2p/js-libp2p/pull/1397 being merged the docs need to be updated to reflect the new behaviour of the `AutoDialer` (being removed from the `connectionManager`) along with the new configurations proposed in https://github.com/libp2p/js-libp2p/pull/1511

PR #1557, opened 2023-01-17
docs: update peer discovery docs

achingbrain:
Update docs to more accurately reflect the current state and behaviour of peer discovery in js-libp2p.


maschad:
This is a duplicate of https://github.com/libp2p/js-libp2p/pull/1556 should we close it?

achingbrain:
Fixed by #1556

PR #1511, opened 2022-12-05
feat:  Update `connectionManager` to  allow configuration of incoming and outgoing connection limits

maschad:
Closes https://github.com/libp2p/js-libp2p/issues/1508

PR #1564, opened 2023-01-18
fix: Connections to Allow list peers will not be pruned

### LINKED ISSUE #1515: Connections to Allow list peers should not be pruned
When connection limits are breached, connections can be pruned to bring the node back under it's limits.

Allow lists are supported to allow certain peers to always open connections, connections from these peers should be excluded from connection pruning.

maschad:
Closes #1515

### DIFF
--- a/src/connection-manager/index.ts
+++ b/src/connection-manager/index.ts
@@ -651,7 +651,7 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
       log('too many connections open - closing a connection to %p', connection.remotePeer)
       // check allow list
       const connectionInAllowList = this.allow.some((ma) => {
-        return ma.getPeerId() === connection.remotePeer.toString()
+        return connection.remoteAddr.toString().startsWith(ma.toString())
       })
 
       // Connections in the allow list should be excluded from pruning
--- a/src/connection-manager/index.ts
+++ b/src/connection-manager/index.ts
@@ -649,7 +649,15 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
 
     for (const connection of sortedConnections) {
       log('too many connections open - closing a connection to %p', connection.remotePeer)
-      toClose.push(connection)
+      // check allow list
+      const connectionInAllowList = this.allow.some((ma) => {
+        return ma.getPeerId() === connection.remotePeer.toString()
+      })
+
+      // Connections in the allow list should be excluded from pruning
+      if (!connectionInAllowList) {
+        toClose.push(connection)
+      }
 
       if (toClose.length === toPrune) {
         break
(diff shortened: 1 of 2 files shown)


src/connection-manager/index.ts
```diff
@@ -649,7 +649,15 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
 
     for (const connection of sortedConnections) {
       log('too many connections open - closing a connection to %p', connection.remotePeer)
-      toClose.push(connection)
+      // check allow list
+      const connectionInAllowList = this.allow.some((ma) => {
+        return ma.getPeerId() === connection.remotePeer.toString()
```
achingbrain, line 654: In the test, `allow` is specified as `/ip4/83.13.55.32` - it has no peer id so this check doesn't look right.

I think instead we should be comparing address prefixes here, similar to the `acceptIncomingConnection` method in this class:

```suggestion
        return connection.remoteAddr.toString().startsWith(ma.toString())
```

test/connection-manager/index.spec.ts
```diff
@@ -169,7 +169,72 @@ describe('Connection Manager', () => {
     expect(shortestLivedWithLowestTagSpy).to.have.property('callCount', 1)
   })
 
-  it('should close connection when the maximum has been reached even without tags', async () => {
+  it('should not close connection that is on the allowlist when pruning', async () => {
+    const max = 5
```
achingbrain, line 173: I think this test could be simplified by setting the max to 1, two peers - one with a remote address in the allow list and one not and asserting that the one not in the allow list was disconnected.

test/connection-manager/index.spec.ts
```diff
@@ -169,7 +169,72 @@ describe('Connection Manager', () => {
...
+      const value = i * 10
+      spies.set(value, spy)
+      await libp2p.peerStore.tagPeer(connection.remotePeer, 'test-tag', {
+        value
+      })
+      await connectionManager._onConnect(new CustomEvent('connection', { detail: connection }))
+    }
+
+    // Connect to the peer on the allowed list
+    const connection = mockConnection(mockMultiaddrConnection(mockDuplex(), remoteAddrPeerId))
```
achingbrain, line 208: I may be missing something but I can't see where this connection has it's remote address set to contain the one in the allow list?

achingbrain (changes requested):
Nearly there, just a few changes needed.

test/connection-manager/index.spec.ts
maschad: My apologies, I think when I was cherry-picking this from https://github.com/libp2p/js-libp2p/pull/1511 I may have resolved the conflict incorrectly

test/connection-manager/index.spec.ts
maschad: I think it's still best to give all peers tags and then have the lowest tagged peer in the `allowlist`. Since normally the _lowest tag value_  is pruned first I wanted to highlight that this would not be the case given **that said peer** is in the `allowlist`

achingbrain (approved):

PR #1570, opened 2023-01-28
fix: compare unsanitized number using gte

wemeetagain:
If maxConnections is passed in as a float, this will result in disconnecting all peers due to this strict equality check.

This may happen unintentionally if maxConnections is set dynamically.

This is a patch to avoid this unfortunate failure mode.

Likely more thought should go into a systematic approach to handling these situations.

### DIFF
--- a/src/connection-manager/index.ts
+++ b/src/connection-manager/index.ts
@@ -651,7 +651,7 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
       log('too many connections open - closing a connection to %p', connection.remotePeer)
       toClose.push(connection)
 
-      if (toClose.length === toPrune) {
+      if (toClose.length >= toPrune) {
         break
       }
     }


maschad:
I can't conceptualize a fraction of a connection so I would suggest we convert any value set as a  ‘maxConnection‘ to a positive integer value internally - using something like [Math.trunc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc) - before we perform operations with that value. What do you think?

maschad (changes requested):
I can't conceptualize a fraction of a connection so I would suggest we convert any value set as a `maxConnections` to a positive integer value internally - using something like [Math.trunc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc) - before we perform operations with that value.

maschad:
I opened up https://github.com/libp2p/js-libp2p/pull/1572 in favour of this solution

wemeetagain:
Yea opened this PR mainly for discussion.

I think ideally we can/should validate user options in some more systematic way. Not sure about coersing user values, that could cause some confusion. Perhaps better to just throw when they are invalid.

maschad:
> I think ideally we can/should validate user options in some more systematic way. Not sure about coersing user values, that could cause some confusion. Perhaps better to just throw when they are invalid.

Agreed. That was also [the suggestion made here](https://github.com/libp2p/js-libp2p/pull/1572#pullrequestreview-1274842562)

achingbrain:
Closing as we are going to implement more comprehensive config validation - https://github.com/libp2p/js-libp2p/issues/1573

Issue #1571, opened 2023-01-30
Handle non-integers passed as connection limits


maschad:
If `maxConnections` is passed in as a float, this will result in disconnecting all peers due to this strict equality check.

This may happen unintentionally if `maxConnections` is set dynamically.

Related: https://github.com/libp2p/js-libp2p/pull/1570

maschad:
Closing in favour of https://github.com/libp2p/js-libp2p/issues/1573

PR #1572, opened 2023-01-30
fix: Handle non-integers passed as connection limits 

maschad:
Closes https://github.com/libp2p/js-libp2p/issues/1571


achingbrain (changes requested):
This config is invalid - it should throw instead of trying to convert floats to ints.

PR #1579, opened 2023-02-04
chore: support yamux to wherever mplex is supported 

### LINKED ISSUE #1578: Support `Yamux` where `Mplex` is supported
Add support for the  [Yamux StreamMuxer](https://github.com/ChainSafe/js-libp2p-yamux)

maschad:
Closes #1578


maschad:
It may be wise to merge this after https://github.com/libp2p/js-libp2p/pull/1533 is merged due to potential conflicts

achingbrain (changes requested):
Stream muxers are negotiated in the order in which they are configured - if mplex is first and both peers support mplex, only mplex will be used.

Yamux appears to be configured behind mplex everywhere here so it's not actually going to be used in the tests.

maschad:
> Stream muxers are negotiated in the order in which they are configured - if mplex is first and both peers support mplex, only mplex will be used.

Is there any particular reason for enforcing this? I know that [Yamux should be used over mplex](https://docs.libp2p.io/concepts/multiplex/mplex/) so should we default to **Yamux** and allow **Mplex**  to be configured in the case where a consumer wants to provide backward-compatibility with legacy nodes?

p-shahi:
> > Stream muxers are negotiated in the order in which they are configured - if mplex is first and both peers support mplex, only mplex will be used.
> 
> Is there any particular reason for enforcing this? I know that [Yamux should be used over mplex](https://docs.libp2p.io/concepts/multiplex/mplex/) so should we default to **Yamux** and allow **Mplex** to be configured in the case where a consumer wants to provide backward-compatibility with legacy nodes?

I think @achingbrain is saying that it should be reversed: i.e. `streamMuxers: [mplex(),yamux()],` should be `streamMuxers: [yamux(),mplex()],`

test/upgrading/upgrader.spec.ts
```diff
@@ -619,10 +630,10 @@ describe('libp2p.upgrader', () => {
     const remoteLibp2pUpgraderOnStreamSpy = sinon.spy(remoteLibp2p.components.upgrader as DefaultUpgrader, '_onStream')
 
     const stream = await localConnection.newStream(['/echo/1.0.0'])
-    expect(stream).to.include.keys(['id', 'close', 'reset', 'stat'])
+    expect(stream).to.include.keys(['abortController', 'source', 'id', 'name'])
```
maschad, line 633: These keys are now included in the Yamux stream, I am not sure what this test intended by specifying specific keys.

test/upgrading/upgrader.spec.ts
```diff
@@ -742,7 +757,7 @@ describe('libp2p.upgrader', () => {
     expect(streamCount).to.equal(1)
 
     await expect(localToRemote.newStream(protocol)).to.eventually.be.rejected()
-      .with.property('code', 'ERR_STREAM_RESET')
+      .with.property('code', 'ERR_UNDER_READ')
```
maschad, line 760: Would this now be the expected error code?

achingbrain: Yamux should behave the same as Mplex so the expected error code is the one in the test - `ERR_STREAM_RESET`

maschad: So it seems that according to the test,   _it should limit the number of incoming streams that can be opened using a protocol_ does this mean the error below is correct?

```bash
```

The error is occurring in [it-reader](https://github.com/alanshaw/it-reader/blob/master/src/index.ts#L71) which is a dep of [multistream-select](https://github.com/libp2p/js-libp2p-multistream-select/blob/master/package.json#L155), it happens when trying to [read the protocol response](https://github.com/libp2p/js-libp2p-multistream-select/blob/master/src/select.ts#L81)

maschad: It seems that in js-libp2p-yamux [the `resetController` isn't an abortable source](https://github.com/ChainSafe/js-libp2p-yamux/blob/da397130d09fda8c1b11196128985efaacfe6f6b/src/stream.ts#L129) and so my inclination is that when the [resetController is aborted](https://github.com/libp2p/js-libp2p-mplex/blob/31a32f386f582467b4cda578a823610f4e096a5e/src/stream.ts#L191) the `ERR_STREAM_RESET` is **not** propagated but rather this [lower level error is logged](https://github.com/ChainSafe/js-libp2p-yamux/blob/da397130d09fda8c1b11196128985efaacfe6f6b/src/stream.ts#L146), I'll open a PR

BigLep:
@maschad : what are the next steps here to complete this?  Does https://github.com/ChainSafe/js-libp2p-yamux/pull/26 need to get landed?  I see it as marked as draft.  What do we need to to for it to get reviewed?

This work isn't urgent is important.  It seems like every month in IPFS-land we hit bumps around still needing to fallback to mplex because yamux hasn't been rolled out more broadly or at least isn't on by default in new versions.

maschad:
Yes @BigLep that's correct. I have some remaining work on https://github.com/ChainSafe/js-libp2p-yamux/pull/26 to ensure it's throwing the correct error when the stream resets, before it can be reviewed. I will prioritize that this week.

BigLep:
Thanks @maschad .  I'll defer to @p-shahi on overall priorities as he has better lay of the land and certainly universal connectivity is the key thing to get ahead on.  I just want to make sure this doesn't languish.  Thanks a lot!

maschad:
Once https://github.com/ChainSafe/js-libp2p-yamux/pull/26 is merged we can upgrade to `4.0.1` and then this can be merged

p-shahi:
Should this be included in https://github.com/libp2p/js-libp2p/pull/1707 @achingbrain

achingbrain (approved):

PR #1593, opened 2023-02-19
fix: remove p-settle dependency, use Promise.allSettled instead

jimmywarting:
remove dependency on `p-settle` and use native, spec'ed / builtin `Promise.allSettled` instead.

### DIFF
--- a/test/upgrading/upgrader.spec.ts
+++ b/test/upgrading/upgrader.spec.ts
@@ -406,7 +406,7 @@ describe('Upgrader', () => {
     ])
     expect(results).to.have.length(2)
     results.forEach(result => {
-      expect(result.status).to.equal('rejected')
+      expect(result).to.have.property('status', 'rejected')
       expect(result).to.have.nested.property('reason.code', codes.ERR_UNSUPPORTED_PROTOCOL)
     })
   })
--- a/test/upgrading/upgrader.spec.ts
+++ b/test/upgrading/upgrader.spec.ts
@@ -318,7 +318,7 @@ describe('Upgrader', () => {
     // Ensure both sides fail
     expect(results).to.have.length(2)
     results.forEach(result => {
-      expect(result.status).to.equal('rejected')
+      expect(result).to.have.property('status', 'rejected')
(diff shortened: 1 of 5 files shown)


achingbrain:
`p-settle` is used in the [transport manager](https://github.com/jimmywarting/js-libp2p/blob/rm-pSettle/src/transport-manager.ts#L2) so this will need refactoring too

jimmywarting:
fixed.

test/upgrading/upgrader.spec.ts
```diff
@@ -255,15 +254,15 @@ describe('Upgrader', () => {
...
+    const results = await Promise.allSettled([
       localUpgrader.upgradeOutbound(outbound),
       remoteUpgrader.upgradeInbound(inbound)
     ])
 
     // Ensure both sides fail
     expect(results).to.have.length(2)
     results.forEach(result => {
-      expect(result).to.have.property('isRejected', true)
+      expect(result.status).to.equal('rejected')
```
achingbrain, line 265:
```suggestion
      expect(result).to.have.property('status', 'rejected')
```

test/upgrading/upgrader.spec.ts
```diff
@@ -311,15 +310,15 @@ describe('Upgrader', () => {
...
```
achingbrain, line 321:
```suggestion
      expect(result).to.have.property('status', 'rejected')
```

test/upgrading/upgrader.spec.ts
```diff
@@ -401,13 +400,13 @@ describe('Upgrader', () => {
...
 
-    const results = await pSettle([
+    const results = await Promise.allSettled([
       connections[0].newStream('/unsupported/1.0.0'),
       connections[1].newStream('/unsupported/1.0.0')
     ])
     expect(results).to.have.length(2)
     results.forEach(result => {
-      expect(result).to.have.property('isRejected', true)
+      expect(result.status).to.equal('rejected')
```
achingbrain, line 409:
```suggestion
      expect(result).to.have.property('status', 'rejected')
```

achingbrain (approved):

achingbrain:
Thanks!

PR #1602, opened 2023-02-24
fix: interop tests again

achingbrain:


ckousik:
@achingbrain I'm only seeing these failures on Linux. The tests are passing on macOS. Can you confirm?

ckousik:
@achingbrain It looks like a race condition, but I'm not sure why it only occurs on Linux.
Consider the following line in the js-libp2p-daemon-server: https://github.com/libp2p/js-libp2p-daemon/blob/fedcbdfcf718244dbf043275afbf19535f26897e/packages/libp2p-daemon-server/src/pubsub.ts#L44
When it finishes executing, it is not guaranteed that the server has completed subscribing to the topic. This is because the [subscribe](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/master/src/index.ts#L1703-L1714) method for gossipsub can return once the rpc is scheduled for sending. 

Temporarily, adding a delay (100ms) after [this line](https://github.com/libp2p/interop/blob/master/src/pubsub/gossipsub.ts#L55) fixes the problem.

achingbrain:
If you look a little further on in the test, the publisher [calls](https://github.com/libp2p/interop/blob/master/src/pubsub/gossipsub.ts#L64) `waitForBothSubscribed` before sending a message on the topic.

This function [waits for both peers to see each other in their subscribers list for the topic](https://github.com/libp2p/interop/blob/master/src/pubsub/utils.ts), which you would hope would mean they've both finished subscribing, their subscription lists have been received by the other peer and one could publish a message that the other would see.

They don't always see each other in their subscriber lists though, the question is why?

I do see this timeout running locally on OS X, so it's not operating system related, though it's not very frequent, likely because CI machines are more resource constrained which adds weight to the race condition idea.

Adding an arbitrary delay will bite us in the future, we should detect the thing that occurs during the delay that means the test passes, it's a more robust solution.

ckousik:
@achingbrain I looked into the `waitForBothSubscribed`, and they do see each other bofre the `Publish` rpc is called. That struck me as odd, so I looked into the Go implementation for gossipsub.

Consider the `waitForSubscribe` method:
`waitForSubscribe` causes both peers to join the topic. In the Go impl, this adds an empty list of peers to the mesh, which only gets populated after a heartbeat occurs:

```
go-libp2p:22068:error 2023-02-28T07:59:48.008Z        DEBUG   pubsub  go-libp2p-pubsub/gossipsub.go:1004      direct peers: map[]
  go-libp2p:22068:error 2023-02-28T07:59:48.008Z        DEBUG   pubsub  go-libp2p-pubsub/gossipsub.go:1013      floodsub peers: map[12D3KooWJxZEwNUkRXvizzvQXzzP63JhWAyudCRPJ2G4U97jK12M:{}]
  go-libp2p:22068:error 2023-02-28T07:59:48.008Z        DEBUG   pubsub  go-libp2p-pubsub/gossipsub.go:1022      gmap peers: ok: true, map: map[]
  go-libp2p:22068:error 2023-02-28T07:59:48.008Z        DEBUG   pubsub  go-libp2p-pubsub/gossipsub.go:1049      tosend: map[]
  go-libp2p:22068:error 2023-02-28T07:59:48.008Z        DEBUG   pubsub  go-libp2p-pubsub/gossipsub.go:985       publish returned
  go-libp2p:22068:error  +1ms
  go-libp2p:22068:error 2023-02-28T07:59:48.924Z        DEBUG   pubsub  go-libp2p-pubsub/gossipsub.go:1402      before heartbeat: map[test-topic:map[]]
  go-libp2p:22068:error 2023-02-28T07:59:48.924Z        DEBUG   pubsub  go-libp2p-pubsub/gossipsub.go:1414      HEARTBEAT: Add mesh link to 12D3KooWJxZEwNUkRXvizzvQXzzP63JhWAyudCRPJ2G4U97jK12M in test-topic
  go-libp2p:22068:error  +915ms
  go-libp2p:22068:error 2023-02-28T07:59:48.924Z        DEBUG   pubsub  go-libp2p-pubsub/gossipsub.go:1577      after heartbeat: map[test-topic:map[12D3KooWJxZEwNUkRXvizzvQXzzP63JhWAyudCRPJ2G4U97jK12M:{}]]
``` 

This causes `Publish` to not send to  any peers because [this lookup](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L1013) returns `ok = true` with an empty set of peers.

On removing the `waitForBothSubscribed`, two cases are possible:
1. We try to publish before any peers are added to the topic, which fails.
2. A peers is added to the topic, but not the mesh (yet). In this case the peer [lookup](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L1013) returns `ok = false`. This leads to the router checking if there are any fanout peers to which it can send, failing which it tries [any peers](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L1019-L1023) which are not direct and have a score above the  publishing threshold. 

In case we decide to keep `waitForBothSubscribed`, we should wait until 1 heartbeat is completed for sending to succeed.
In case we remove `waitForBothSubscribed`, we only need to wait until B appears as a subscriber in A.

PR #1604, opened 2023-03-01
feat(@libp2p/protocol-perf): Implement perf protocol

MarcoPolo:
Implements https://github.com/libp2p/specs/pull/478


src/perf/index.ts
```diff
@@ -0,0 +1,133 @@
...
+    let actualRecvdBytes = BigInt(0)
+    for await (const buf of stream.source) {
+      actualRecvdBytes += BigInt(buf.length)
+    }
+
+    if (actualRecvdBytes !== recvBytes) {
+      throw new Error(`Expected to receive ${recvBytes} bytes, but received ${actualRecvdBytes}`)
+    }
+
+    stream.close()
```
mxinden, line 117: Shouldn't the node close the write side of the stream right after having sent the last byte of its payload?

```
4. Close the write side of our stream.
5. Read from the read side of the stream. This  should be the same number of bytes as we told the server in step 2.
```

Or is write-side closing happening implicitly through `stream.sink` above?

mxinden (commented):
:rocket:

BigLep:
What are the next steps on this?  I'd like to make sure js-libp2p isn't getting left behind on the perf dashboarding.

src/perf/index.ts
```diff
@@ -0,0 +1,133 @@
...
+        if (toSend > bytesToSendBack) {
+          toSend = bytesToSendBack
+        }
+        bytesToSendBack = bytesToSendBack - toSend
+        yield uint8Buf.subarray(0, Number(toSend))
+      }
+    }())
+  }
+
+  async startPerfOnStream (peer: PeerId, sendBytes: bigint, recvBytes: bigint, options: AbortOptions = {}): Promise<void> {
```
wemeetagain, line 78: Might call this something simpler?
Identify has an `identify` method and Ping has q `ping` method so you might call this `perf`?

src/perf/index.ts
```diff
@@ -0,0 +1,133 @@
...
+    log('dialing %s to %p', this.protocol, peer)
```
wemeetagain, line 79: Might include other logs:
Log for finishing a perf
Log for handling/finishing an incoming perf

wemeetagain (commented):
The logic and stylistic concerns lgtm

maschad:
> What are the next steps on this? I'd like to make sure js-libp2p isn't getting left behind on the perf dashboarding.

Agreed @BigLep . I am going to refactor the [PerfService](https://github.com/libp2p/js-libp2p/pull/1604/files#diff-dcc7fb1f3a505096c0e8186861a09f833e638979f932873420f28bffe9c01bb1R20) to allow this to be [configured the way configure our other optional services now](https://github.com/libp2p/js-libp2p/pull/1563) following @achingbrain 's [suggestion](https://github.com/libp2p/js-libp2p/pull/1605#issuecomment-1449729020) as well as incorporate @wemeetagain 's [feedback](https://github.com/libp2p/js-libp2p/pull/1604#discussion_r1215154156) and then that should return this PR to a ready state.

mxinden:
Thank you @maschad.

> and then that should return this PR to a ready state.

I will then integrate it into https://github.com/libp2p/test-plans/pull/163/.

src/perf/index.ts
```diff
@@ -0,0 +1,166 @@
...
+    const signal = anySignal([AbortSignal.timeout(this.timeout), options?.signal])
+    const stream = await connection.newStream([this.protocol], {
+      signal
+    })
+
+    // Convert sendBytes to uint64 big endian buffer
+    const view = new DataView(this.databuf)
+    view.setBigInt64(0, recvBytes, false)
+
+    log.trace('sending %i bytes to %p', sendBytes, peer)
```
maschad, line 119: Depends on https://github.com/libp2p/js-libp2p-logger/pull/37

maschad:
[CI check failure](https://github.com/libp2p/js-libp2p/actions/runs/5226097335/jobs/9436454613?pr=1604#step:5:13) is fixed by https://github.com/libp2p/js-libp2p/pull/1801

maschad:
Since this is our first benchmark tests there isn't really a precedent, but do we want to run these types of tests in our regular PR workflow or would it be better suited to be run on the main branch ? cc @achingbrain @wemeetagain @MarcoPolo

mxinden:
What are you referring to with "these types of tests"? For https://github.com/libp2p/test-plans/pull/184 we plan to run it before every release for now.

maschad:
> What are you referring to with "these types of tests"? For [libp2p/test-plans#184](https://github.com/libp2p/test-plans/pull/184) we plan to run it before every release for now.

I was referring to  benchmark tests there (as opposed to our other types of tests such as unit or integration tests). But if the plan is to run it before every release then the answer to my question would be to run these post-merge on the main branch.

MarcoPolo:
In the ideal world, this should run on every PR to highlight any perf regressions or improvements. It should fail CI if perf degrades by more than X% relative to the `main` branch.

maschad:
Currently blocked by a huge refactor in https://github.com/libp2p/js-libp2p/pull/1833

maschad:
> In the ideal world, this should run on every PR to highlight any perf regressions or improvements. It should fail CI if perf degrades by more than X% relative to the `main` branch.

What do you think is an appropriate threshold? I currently have it 20% but [that is subject to failure at times](https://github.com/libp2p/js-libp2p/actions/runs/5351437434/jobs/9705239341?pr=1604#step:14:11). Perhaps we should increase it to 30% ?

mxinden:
> In the ideal world, this should run on every PR to highlight any perf regressions or improvements. It should fail CI if perf degrades by more than X% relative to the `main` branch.

On a high level, the custom benchmarking CI setup introduced in this pull request seems to reinvent what https://github.com/libp2p/test-plans/pull/184 already provides, namely an automated way to run performance benchmarks.

https://github.com/libp2p/test-plans/pull/184 currently has a couple of shortcomings, and can thus not fulfill the use-case needed in this pull request as is:

- It uses somewhat expensive AWS instances (~2 Dollar per run), thus likely not suited to run on  though that can be variable.
- It does not (yet) allow running tests for a single implementation only.

But what it does provide:

- Provisioning of dedicated cloud instances, thus providing reproducible results.
- A runner, building, running and reporting the results back.
- A way to visualize the results https://observablehq.com/@mxinden-workspace/libp2p-performance-dashboard.

I am fine with the direction this pull request is taking, namely to implement a custom performance benchmarking tool for js-libp2p. Though long term, I would suggest to merge with what https://github.com/libp2p/test-plans/pull/184 started, i.e. for js-libp2p to call a workflow within `libp2p/test-plans`.

What are people's thought @maschad @MarcoPolo?

maschad:
> [libp2p/test-plans#184](https://github.com/libp2p/test-plans/pull/184) currently has a couple of shortcomings, and can thus not fulfill the use-case needed in this pull request as is:
> 
> * It uses somewhat expensive AWS instances (~2 Dollar per run), thus likely not suited to run on  though that can be variable.
> * It does not (yet) allow running tests for a single implementation only.

I would add to this @mxinden that https://github.com/libp2p/test-plans  currently does not provide a mechanism for testing across platforms (as opposed to [GIthub CI runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#overview-of-github-hosted-runners) ) and given the current [limitations in testing webRTC on webkit on linux](https://github.com/libp2p/js-libp2p-webrtc/pull/177#issuecomment-1589529908) that constraint could potentially inhibit other benchmarking efforts.

> I would suggest to merge with what https://github.com/libp2p/test-plans/pull/184 started, i.e. for js-libp2p to call a workflow within libp2p/test-plans.

I agree, as I mentioned before that my only concern would for evaluating performance across platforms in the future.

maschad:
I've updated the CI job to output a [comparison matrix](https://github.com/libp2p/js-libp2p/actions/runs/5404907859/attempts/1#summary-14634647501) that passes if there is no performance degradation greater than 20% @mxinden

BigLep:
> I've updated the CI job to output a [comparison matrix](https://github.com/libp2p/js-libp2p/actions/runs/5404907859/attempts/1#summary-14634647501) that passes if there is no performance degradation greater than 20% @mxinden

Driveby - I'm not close here.
Looking at the table, lets make sure we're clear on the units.
20% seems like too much slippage.  Can we have this less like < 5%?
Also, if not's already, do we run each test a few times and then take the average or median to make sure we're not hit by outliers?
image

packages/perf/src/index.ts
```diff
@@ -0,0 +1,166 @@
...
+    return Number((8000n * size) / BigInt(Date.now() - now))
+  }
+
+  // measureUploadBandwidth returns the measured bandwidth in bit per second
+  async measureUploadBandwidth (peer: PeerId, size: bigint): Promise<number> {
+    const now = Date.now()
+    await this.perf(peer, size, 0n)
+    return Number((8000n * size) / BigInt(Date.now() - now))
+  }
+}
```
mxinden, line 162: Based on the `8000` I am assuming that `size`' unit is `kbyte`. For the sake of simplicity and easier understanding I suggest using [SI units](https://en.wikipedia.org/wiki/International_System_of_Units) everywhere. I.e. in this case I suggest using `byte` instead of `kbyte`.

packages/perf/test/index.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    remoteComponents.events.safeDispatchEvent('connection:open', { detail: remoteToLocal })
+
+    // Run Perf
+    const downloadBandwidth = await client.measureDownloadBandwidth(remoteComponents.peerId, 10n << 20n) >> 10
+    // eslint-disable-next-line no-console
+    console.log('Download bandwidth:', downloadBandwidth, 'kiB/s')
+
+    const uploadBandwidth = await client.measureDownloadBandwidth(remoteComponents.peerId, 10n << 20n) >> 10
+    // eslint-disable-next-line no-console
+    console.log('Upload bandwidth:', uploadBandwidth, 'kiB/s')
```
mxinden, line 106: In `measureDownloadBandwidth` you use power of 10, e.g. `8000`, whereas here you use power of 2, e.g. `10n << 20n`. I suggest being consistent, that is use either but not both.

packages/perf/test/index.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    // simulate connection between nodes
+    const [localToRemote, remoteToLocal] = connectionPair(localComponents, remoteComponents)
+    localComponents.events.safeDispatchEvent('connection:open', { detail: localToRemote })
+    remoteComponents.events.safeDispatchEvent('connection:open', { detail: remoteToLocal })
+
+    // Run Perf
+    const downloadBandwidth = await client.measureDownloadBandwidth(remoteComponents.peerId, 10n << 20n) >> 10
+    // eslint-disable-next-line no-console
+    console.log('Download bandwidth:', downloadBandwidth, 'kiB/s')
```
mxinden, line 102: `kiB` is kibibytes, but the base unit here is bit and not byte, right?

mxinden (commented):
Thanks for the work here @maschad.

Definitely in favor of @BigLep suggestions above, especially running multiple times.

How stable are these performance measurements on GitHub runners? Asked differently, how much variance are you seeing across runs?

maschad:
Thanks for the feedback @BigLep and @mxinden , I've incorporated it.

- The threshold is now 5% for performance degradation tolerance.
- Based on @mxinden 's [suggestion](https://github.com/libp2p/js-libp2p/pull/1604#discussion_r1246120682) I have standardized the base unit as `bytes` across the `perf` service, so the units for the bandwidth is thus`B/s`
- The perf tests run 5 times and then an average is taken which are the bandwidths used for comparison.

> How stable are these performance measurements on GitHub runners? Asked differently, how much variance are you seeing across runs?

Since updating to using an _average bandwidth_ that it has failed the 5% threshold on [one occasion](https://github.com/libp2p/js-libp2p/actions/runs/5414774782/jobs/9842312339).

MarcoPolo:
I'm not sure you can run reliably run tight regression tests on GitHub actions CI.

mxinden:
> In the ideal world, this should run on every PR to highlight any perf regressions or improvements. It should fail CI if perf degrades by more than X% relative to the `main` branch.

I think there is a misunderstanding with the above original comment by @MarcoPolo which spawned the automation work in this pull request. I don't think Marco suggested to implement custom benchmarking automation on top of GitHub runners, especially since we now have https://github.com/libp2p/test-plans/issues/63. @MarcoPolo please correct me if I am wrong.

> I'm not sure you can run reliably run tight regression tests on GitHub provided runners.

I share this as the main concern.

@maschad I see [your concerns on cross-platform testing](https://github.com/libp2p/js-libp2p/pull/1604#issuecomment-1610192057) and [the changes required to the output format](https://github.com/libp2p/test-plans/issues/65#issuecomment-1611696876). Still I would suggest prioritizing an integration with https://github.com/libp2p/test-plans/issues/65 over implementing a custom setup.

@maschad do you have time next week to do a quick sync? That might clear up any confusion on both ends.

---

> Based on @mxinden 's [suggestion](https://github.com/libp2p/js-libp2p/pull/1604#discussion_r1246120682) I have standardized the base unit as `bytes` across the `perf` service, so the units for the bandwidth is thus`B/s`

Thank you for the follow-up. Using base units makes the code a lot easier. That said, the base-unit for data rates, i.e. transfer speed, is `bits/s`, never `byte/s`. Also it is fine to use prefixes (e.g. `kilo`, `mega`, ...), though whenever one does, it needs to be explicit, e.g. by adding a unit when printing to stdout. Whenever I see a variable without the prefix e.g. `size` I assume it is the base unit, e.g. `byte` and not `kibyte`.

> The perf tests run 5 times and then an average is taken which are the bandwidths used for comparison.

I suggest using quantiles, e.g. the median, or the 95th percentile. Using the average hides outliers, e.g. among your 5 runs 2 might be crazy fast and 3 crazy slow, the average can still be within the bound.

packages/perf/test/index.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    remoteComponents.events.safeDispatchEvent('connection:open', { detail: remoteToLocal })
+
+    let downloadBandwidth = 0
+    let uploadBandwidth = 0
+
+    for (let i = 1; i < 5; i++) {
+      // Run Perf
+      downloadBandwidth += await client.measureDownloadBandwidth(remoteComponents.peerId, 10485760n)
+      uploadBandwidth += await client.measureUploadBandwidth(remoteComponents.peerId, 10485760n)
+    }
```
mxinden, line 106:
```suggestion
    for (let i = 0; i < 5; i++) {
      // Run Perf
      downloadBandwidth += await client.measureDownloadBandwidth(remoteComponents.peerId, 10485760n)
      uploadBandwidth += await client.measureUploadBandwidth(remoteComponents.peerId, 10485760n)
    }
```

With my limited JS knowledge, this runs only `4` times otherwise.

maschad:
> I think there is a misunderstanding with the above original comment by @MarcoPolo which spawned the automation work in this pull request. I don't think Marco suggested to implement custom benchmarking automation on top of GitHub runners, especially since we now have https://github.com/libp2p/test-plans/issues/63. @MarcoPolo please correct me if I am wrong.

> @maschad I see https://github.com/libp2p/js-libp2p/pull/1604#issuecomment-1610192057 and https://github.com/libp2p/test-plans/issues/65#issuecomment-1611696876. Still I would suggest prioritizing an integration with https://github.com/libp2p/test-plans/issues/65 over implementing a custom setup.

Thanks for clarifying, the rationale for introducing this additional automation work stemmed from [this comment](https://github.com/libp2p/js-libp2p/pull/1604#issuecomment-1606525087) which suggested that I continue in the direction of implementing this custom performance benchmarking tool for js-libp2p, then _post-merge_, [coalesce into a workflow on test plans](https://github.com/libp2p/test-plans/issues/65#issuecomment-1610754398), but there may have been a misunderstanding.

> I'm not sure you can run reliably run tight regression tests on GitHub provided runners.

I would be interested to hear your and @MarcoPolo justification behind this.

> @maschad do you have time next week to do a quick sync? That might clear up any confusion on both ends.

Sure! I will co-ordinate with you on Slack.

> Thank you for the follow-up. Using base units makes the code a lot easier. That said, the base-unit for data rates, i.e. transfer speed, is bits/s, never byte/s. Also it is fine to use prefixes (e.g. kilo, mega, ...), though whenever one does, it needs to be explicit, e.g. by adding a unit when printing to stdout. Whenever I see a variable without the prefix e.g. size I assume it is the base unit, e.g. byte and not kibyte.

> I suggest using quantiles, e.g. the median, or the 95th percentile. Using the average hides outliers, e.g. among your 5 runs 2 might be crazy fast and 3 crazy slow, the average can still be within the bound.

Good points 👍🏾

maschad:
Following a conversation with @mxinden I'm converting this back to draft. The remaining work will seek to integrate this PR with the existing [perf workflow](https://github.com/libp2p/test-plans/tree/master/perf) so that Benchmark results can be visualized through the [perf dashboard](https://observablehq.com/@mxinden-workspace/libp2p-performance-dashboard). 

My next steps are:

- Capture latency and output via stdout into the [prescribed JSON format](https://raw.githubusercontent.com/libp2p/test-plans/master/perf/runner/benchmark-results.json)
- Remove the iterations in the bandwidth test (since iPerf already does that through the [workflow](https://github.com/libp2p/test-plans/blob/master/.github/workflows/perf.yml)
- Refactor these tests to accept  `--server-ip-address` , `--transport`, `--upload-bytes` and `--download-bytes` command line args as input as outlined in the [adding a new implementation](https://github.com/libp2p/test-plans/blob/master/perf/README.md#adding-a-new-implementation)
- Create a JS binary that can be called by the [test-plans](https://github.com/libp2p/test-plans/tree/master) makefile to run the perf tests and open a PR against [test-plans perf](https://github.com/libp2p/test-plans/tree/master/perf)
- Once merged, call that workflow here 

@achingbrain I still think it's worthwhile to have a threshold here ( 5% as Steve suggested maybe?) that we will use to see if PRs introduce performance regressions and so I will modify my [PR workflow](https://github.com/libp2p/js-libp2p/pull/1604/files#diff-b4ef690e62042d857e0be042ebc502f4f2e78760f23db6ea28ce1cc984309005) to call the test-plan's workflow (once it is merged) and then based on the output, [create a performance regression matrix](https://github.com/libp2p/js-libp2p/actions/runs/5415671965/attempts/2#summary-14670985527).

p-shahi:
@maschad one that that wasn't clear to me was whether this perf code will be added in this monorepo or introduce it into test-plans (like go-libp2p https://github.com/libp2p/test-plans/tree/master/perf/impl/go-libp2p)

maschad:
> @maschad one that that wasn't clear to me was whether this perf code will be added in this monorepo or introduce it into test-plans (like go-libp2p [libp2p/test-plans@`master`/perf/impl/go-libp2p](https://github.com/libp2p/test-plans/tree/master/perf/impl/go-libp2p?rgh-link-date=2023-07-05T15%3A02%3A32Z))

My apologies,  so yes my intention is for these tests to live in this monorepo where they will be ran in perf by calling a `Makefile` which downloads the various docker images and builds the JS  based on the _commit hash_ (similar to what happens with [the js interop tests](https://github.com/libp2p/test-plans/blob/master/multidim-interop/impl/js/v0.45/Makefile)) and then runs it using the [perf runner](https://github.com/libp2p/test-plans/blob/master/perf/runner/src/index.ts). This is similar to how we already store the [interop tests](https://github.com/libp2p/js-libp2p/blob/master/interop/test/ping.spec.ts) 

That's what I was alluding to here:

> Create a JS binary that can be called by the [test-plans](https://github.com/libp2p/test-plans/tree/master?rgh-link-date=2023-07-04T20%3A21%3A56Z) makefile to run the perf tests and open a PR against [test-plans perf](https://github.com/libp2p/test-plans/tree/master/perf?rgh-link-date=2023-07-04T20%3A21%3A56Z)

maschad:
I will be OOO from **July 6th - July 17th** so I will resume work on this once I return.

maschad:
Update:

Running these perf tests locally i.e. running `npm run start` with a server and dialler will output a latency measurement to stdout. 

I had spent _a significant amount of time_ trying to run the perf tests using the [terraform provisioned infrastructure](https://github.com/libp2p/test-plans/tree/master/perf#provision-infrastructure)  but to no avail, it ended up being a lot of AWS wrangling which I think is tangential to the essence of this PR, more is outlined [here](https://github.com/libp2p/test-plans/issues/222)

These tests will be ran on [this workflow](https://github.com/libp2p/test-plans/pull/244) but given this PR outputs the measurement of latency to stdout as required by the [test plans implementation setup](https://github.com/libp2p/test-plans/tree/master/perf#adding-a-new-implementation-or-a-new-version) this PR should be ready for review.

.github/workflows/perf-test.yml
```diff
@@ -0,0 +1,158 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ipfs/aegir/actions/cache-node-modules@master
+        with:
+          directories: ./packages/perf/node_modules
+
+      - name: Install playwright dependencies
+        uses: ipfs/aegir/actions/cache-node-modules@master
+      - run: npx playwright install-deps
```
achingbrain, line 22: I don't think this step is necessary any more, playwright-test seems to install it's own deps now.

packages/perf/src/index.ts
```diff
@@ -0,0 +1,172 @@
+import { logger } from '@libp2p/logger'
```
achingbrain, line 1: Please can you add a `@packageDocumentation` block that describes how to use this module.

See [circuit-relay](https://github.com/libp2p/js-libp2p/blob/55e1230ac5dd65ba573c2da4a0e5eae2bac81ed2/packages/libp2p/src/circuit-relay/index.ts#L1-L35) for an example.  It gets rendered [like this](https://libp2p.github.io/js-libp2p/modules/libp2p.circuit_relay.html).

packages/perf/src/main.ts
```diff
@@ -0,0 +1,132 @@
...
...pycy9STKphorpCT83srBVQi59CdFU6Mj+aL/xt0kCPMVigJw8P3/YCEJ9J+rS8BsoWE+xWUEsJvtXoT7vzPHaAtM3ci1HZd302Mz1+GgS8Epdx+7F5p80XAFLDUnELzOzKftvWGZmWfSeDnslwVONkL/1VAzwKy7Ce6hk4SxRE7l2NE2OklSHOzCGU1f78ZzVYKSnS5Ag9YrGjOAmTOXDbKNKN/qIorAQ1bovzGoCwx3iGIatQKFOxyVCyO1PsJYT7JO+kZbhBWRRE+L7l+ppPER9bdLFxs1t5CrKc078h+wuUr05S1P1JjXk68pk3+kQKBgQDeK8AR11373Mzib6uzpjGzgNRMzdYNuExWjxyxAzz53NAR7zrPHvXvfIqjDScLJ4NcRO2TddhXAfZoOPVH5k4PJHKLBPKuXZpWlookCAyENY7+Pd55S8r+a+MusrMagYNljb5WbVTgN8cgdpim9lbbIFlpN6SZaVjLQL3J8TWH6wKBgQDSChzItkqWX11CNstJ9zJyUE20I7LrpyBJNgG1gtvz3ZMUQCn3PxxHtQzN9n1P0mSSYs+jBKPuoSyYLt1wwe10/lpgL4rkKWU3/m1Myt0tveJ9WcqHh6tzcAbb/fXpUFT/o4SWDimWkPkuCb+8j//2yiXk0a/T2f36zKMuZvujqQKBgC6B7BAQDG2H2B/ijofp12ejJU36nL98gAZyqOfpLJ+FeMz4TlBDQ+phIMhnHXA5UkdDapQ+zA3SrFk+6yGk9Vw4Hf46B+82SvOrSbmnMa+PYqKYIvUzR4gg34rL/7AhwnbEyD5hXq4dHwMNsIDq+l2elPjwm/U9V0gdAl2+r50HAoGALtsKqMvhv8HucAMBPrLikhXP/8um8mMKFMrzfqZ+otxfHzlhI0L08Bo3jQrb0Z7ByNY6M8epOmbCKADsbWcVre/AAY0ZkuSZK/CaOXNX/AhMKmKJh8qAOPRY02LIJRBCpfS4czEdnfUhYV/TYiFNnKRj57PPYZdTzUsxa/yVTmECgYBr7slQEjb5Onn5mZnGDh+72BxLNdgwBkhO0OCdpdISqk0F0Pxby22DFOKXZEpiyI9XYP1C8wPiJsShGm2yEwBPWXnrrZNWczaVuCbXHrZkWQogBDG3HGXNdU4MAWCyiYlyinIBpPpoAJZSzpGLmWbMWh28+RJS6AQX6KHrK1o2uw=='
```
achingbrain, line 69: Is hard coding the key part of the protocol?

achingbrain (changes requested):
Looking promising, comments inline.

packages/perf/src/main.ts
maschad: Not explicitly, but I want to have a stable identity i.e. the`peerID`  and given that it's not passed in as an argument, I figured this was the best way to have the client connect to the server without having to do a prior exchange of the server's peerID.

packages/perf/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "any-signal": "^4.1.1",
+    "datastore-core": "9.2.2",
+    "libp2p": "^0.46.1",
+    "p-wait-for": "^5.0.2",
+    "uint8arrays": "^4.0.6",
+    "yargs": "^17.7.2"
+  },
+  "devDependencies": {
+    "aegir": "^40.0.8",
+    "sinon-ts": "1.0.0"
```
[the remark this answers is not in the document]
maschad: Nope just an oversight, thanks!

.github/workflows/perf-test.yml
```diff
@@ -0,0 +1,158 @@
...
+      - name: Run tests and store in test_output
+        working-directory: packages/perf
+        run: |
+          npm run test > test_output.txt
+
+      - name: Extract bandwidth from test output
+        id: extract_bandwidths
+        working-directory: packages/perf
+        shell: bash
+        run: |
```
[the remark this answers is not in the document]
maschad: Given that the bash script already parses the output correctly, is there any benefits to refactoring that now? I wouldn't want to delay the PR unnecessarily.

packages/perf/src/main.ts
mxinden: Unfortunately the libp2p perf runner does not have support for the server to send its peer ID to the client.

In go-libp2p we simply hardcode it:

https://github.com/libp2p/test-plans/blob/a674f734cc625177a91aea84f568a50cfd70ab7d/perf/impl/go-libp2p/v0.29/main.go#L37-L42

rust-libp2p allows connecting to an address without its peer ID and thus does not require hard-coding.

Adding support for the server to send its peer ID to the client is tracked in https://github.com/libp2p/test-plans/issues/63. That said, I think there are higher impact items to tackle beforehand.

packages/perf/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  maxInboundStreams: 1 << 10,
+  maxOutboundStreams: 1 << 10,
+  timeout: 10000,
+  writeBlockSize: BigInt(64 << 10)
+}
+
+export interface PerfService {
+  perf: (connection: Connection, sendBytes: bigint, recvBytes: bigint, options?: AbortOptions) => Promise<void>
+  measureDownloadBandwidth: (connection: Connection, size: bigint) => Promise<number>
+  measureUploadBandwidth: (connection: Connection, size: bigint) => Promise<number>
```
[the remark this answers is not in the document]
mxinden: One might want to test both up- and download bandwidth in a single test in the future. That said, easy to add in the future. Thus `perf` is not a requirement from the general libp2p perspective.

packages/perf/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    await stream.sink(async function * () {
+      while (bytesToSendBack > 0n) {
+        let toSend: bigint = writeBlockSize
+        if (toSend > bytesToSendBack) {
+          toSend = bytesToSendBack
+        }
+        bytesToSendBack = bytesToSendBack - toSend
+        yield uint8Buf.subarray(0, Number(toSend))
```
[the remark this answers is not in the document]
mxinden: Both random bytes and zeros is fine by the specification. See past discussion by you, @achingbrain.

https://github.com/libp2p/specs/pull/478#discussion_r1037393837

All other implementations use zeroes.

packages/perf/src/index.ts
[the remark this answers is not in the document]
maschad: Well we still need the method but perhaps we could utilize these methods in the `main.ts` ,  they were created for unit tests.

packages/perf/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  async handleMessage (data: IncomingStreamData): Promise<void> {
+    const { stream } = data
+
+    const writeBlockSize = this.writeBlockSize
+
+    let bytesToSendBack: bigint | null = null
+
+    for await (const buf of stream.source) {
+      if (bytesToSendBack === null) {
```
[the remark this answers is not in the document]
maschad: I would agree with you but TS compiler complains that it is used before it's assigned otherwise.

packages/perf/src/index.ts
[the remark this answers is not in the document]
maschad: Zero is fine, random bytes can also be fine (as long as it's not a different set of random bytes, because then you may be limited by how fast you can generate random bytes).

.github/workflows/perf-test.yml
```diff
@@ -0,0 +1,158 @@
+name: Performance Test
```
MarcoPolo, line 1: I don't understand why we want this? Shouldn't this be what the test-plans perf runner does? The bash script here is not great (there's scripting inside scripting!). I don't think this is maintainable (I sure don't want to maintain this). We should have a single source of truth for perf – test test-plans perf runner.

I think the desire here is that we'll be able to "catch regressions in CI". That's very hard to do, and probably impossible on shared GitHub Actions runner and in JS. Realistically the path forward to improving performance looks like this:

1. We record a baseline performance in a reproducible and realistic environment. This is @mxinden's work in the test-plans perf runner.
2. We run a profiler to identify bottlenecks. Or intuitively identify bottlenecks.
3. Fix the bottlenecks.
4. Run the test from step 1.

To protect from regressions, for now I suggest we do this on each release rather than per PR.

packages/perf/previousPerf.json
```diff
@@ -0,0 +1,5 @@
+{
```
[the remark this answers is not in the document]
MarcoPolo: see my other comment. I don't think we even want this file.

packages/perf/src/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const MAX_INBOUND_STREAMS = 1 << 10
+export const MAX_OUTBOUND_STREAMS = 1 << 10
+export const PROTOCOL_NAME = '/perf/1.0.0'
+
+export const TIMEOUT = 10000
```
MarcoPolo, line 5:
```suggestion
export const TIMEOUT = 10_000
```

packages/perf/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  protocolName: '/perf/1.0.0',
+  maxInboundStreams: 1 << 10,
+  maxOutboundStreams: 1 << 10,
+  timeout: 10000,
+  writeBlockSize: BigInt(64 << 10)
+}
+
+export interface PerfService {
+  perf: (connection: Connection, sendBytes: bigint, recvBytes: bigint, options?: AbortOptions) => Promise<void>
+  measureDownloadBandwidth: (connection: Connection, size: bigint) => Promise<number>
```
MarcoPolo, line 22: Do we need these methods? afaict there helpers to calling perf with a 0. Would be nicer to have a smaller api.

packages/perf/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  private readonly maxInboundStreams: number
+  private readonly maxOutboundStreams: number
+  private readonly timeout: number
+  private readonly writeBlockSize: bigint
+
+  constructor (components: PerfServiceComponents, init: PerfServiceInit) {
+    this.components = components
+    this.started = false
+    this.protocol = init.protocolName ?? PROTOCOL_NAME
+    this.databuf = new ArrayBuffer(Number(init.writeBlockSize ?? WRITE_BLOCK_SIZE))
```
MarcoPolo, line 53: can we set this.writeBlockSize before this line, then use `this.writeBlockSize` here.

packages/perf/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+    this.maxInboundStreams = init.maxInboundStreams ?? MAX_INBOUND_STREAMS
+    this.maxOutboundStreams = init.maxOutboundStreams ?? MAX_INBOUND_STREAMS
+    this.timeout = init.timeout ?? TIMEOUT
```
MarcoPolo, line 56: Why do you need a timeout? The caller may set one if they care for it. We don't need one in here.

packages/perf/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  async perf (connection: Connection, sendBytes: bigint, recvBytes: bigint, options: AbortOptions = {}): Promise<void> {
+    log('opening stream on protocol %s to %p', this.protocol, connection.remotePeer)
+
+    const uint8Buf = new Uint8Array(this.databuf)
+
+    const writeBlockSize = this.writeBlockSize
+
+    const signal = anySignal([AbortSignal.timeout(this.timeout), options?.signal])
```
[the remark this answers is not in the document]
MarcoPolo: I think you can get rid of this completely. If the caller wants a timeout, they can set it in the signal. We don't need to manage this here.

packages/perf/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+      }
+    }())
+  }
+
+  async perf (connection: Connection, sendBytes: bigint, recvBytes: bigint, options: AbortOptions = {}): Promise<void> {
+    log('opening stream on protocol %s to %p', this.protocol, connection.remotePeer)
+
+    const uint8Buf = new Uint8Array(this.databuf)
+
+    const writeBlockSize = this.writeBlockSize
```
MarcoPolo, line 114: Why?

packages/perf/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+        bytesToSendBack = bytesToSendBack - toSend
+        yield uint8Buf.subarray(0, Number(toSend))
+      }
+    }())
+  }
+
+  async perf (connection: Connection, sendBytes: bigint, recvBytes: bigint, options: AbortOptions = {}): Promise<void> {
+    log('opening stream on protocol %s to %p', this.protocol, connection.remotePeer)
+
+    const uint8Buf = new Uint8Array(this.databuf)
```
MarcoPolo, line 112: should this.databuf be a uint8array? Saves you an object

packages/perf/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    log('performed %s to %p', this.protocol, connection.remotePeer)
+    await stream.close()
+  }
+
+  // measureDownloadBandwidth returns the measured bandwidth in bits per second
+  async measureDownloadBandwidth (connection: Connection, size: bigint): Promise<number> {
+    const now = Date.now()
+    await this.perf(connection, 0n, size)
+    return Number((8000n * size) / BigInt(Date.now() - now))
```
MarcoPolo, line 159: 8 for 8 bits to a byte I assume. But why 1000?

packages/perf/test/index.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    let uploadBandwidth = 0
+
+    for (let i = 0; i < 5; i++) {
+      // Run Perf
+      downloadBandwidth += await client.measureDownloadBandwidth(localToRemote, 10485760n)
+      uploadBandwidth += await client.measureUploadBandwidth(localToRemote, 10485760n)
+    }
+
+    // eslint-disable-next-line no-console
+    console.log('Upload bandwidth:', Math.floor(downloadBandwidth / 5), 'B/s')
```
MarcoPolo, line 101: Why are you dividing by 5?

packages/perf/src/renderResults.ts
```diff
@@ -0,0 +1,51 @@
+interface TestResult {
```
MarcoPolo, line 1: I don't think we need this if we use the test-plans perf runner.

MarcoPolo (commented):
Can only comment. I'd request changes if I could.

I think we can get rid of some of the code here 😄

packages/perf/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+        if (toSend > bytesToSendBack) {
+          toSend = bytesToSendBack
+        }
+        bytesToSendBack = bytesToSendBack - toSend
+        yield uint8Buf.subarray(0, Number(toSend))
+      }
+    }())
+  }
+
+  async perf (connection: Connection, sendBytes: bigint, recvBytes: bigint, options: AbortOptions = {}): Promise<void> {
```
MarcoPolo, line 109: I'm thinking it is reasonable if this function returns a duration instead of having the caller do it.

packages/perf/test/index.spec.ts
maschad: The is an average bandwidth, so run 5 iterations, sum them, then divide by 5.

packages/perf/src/index.ts
MarcoPolo: Milliseconds to seconds :)

.github/workflows/perf-test.yml
maschad: Good point @MarcoPolo and as @achingbrain mentioned there are resource constraints as well.

I will just remove this workflow for now then post-merge of https://github.com/libp2p/test-plans/pull/244 we can introduce a workflow to run on release / main in a follow up PR.

packages/perf/src/index.ts
maschad: To reference the writeBlockSize within the `stream.sink` function scope it would need to be removed from the `this` binding of the block scope, likewise the `this.databuf`

packages/perf/src/index.ts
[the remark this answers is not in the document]
maschad: We still want to retain the abort signal passed in as an option but we can remove the timeout parameter.

packages/perf/test/index.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    // eslint-disable-next-line no-console
+    console.log('Download bandwidth:', Math.floor(uploadBandwidth / 5), 'B/s')
```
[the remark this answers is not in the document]
maschad: I don't think it needs to be removed as it's useful but I can include an assertion as well.

packages/perf/src/index.ts
maschad: See https://github.com/libp2p/js-libp2p/pull/1604/files#r1285108813

maschad:
I've removed the perf workflow as discussed [here](https://github.com/libp2p/js-libp2p/pull/1604#discussion_r1285094158) for now, once https://github.com/libp2p/test-plans/pull/244 is merged we can introduce a workflow to run on release / main in a follow up PRs.

packages/perf/src/index.ts
MarcoPolo: Would still save you an object, but it's fine

packages/perf/test/index.spec.ts
MarcoPolo: If this is just a unit test, can we do a single run and smaller value. Something bigger than a packet size. 8KiB is probably fine.

packages/perf/test/index.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    localComponents.events.safeDispatchEvent('connection:open', { detail: localToRemote })
+    remoteComponents.events.safeDispatchEvent('connection:open', { detail: remoteToLocal })
+
+    const startTime = Date.now()
+
+    // Run Perf
+    await expect(client.measurePerformance(startTime, localToRemote, 1024n, 1024n)).to.eventually.be.fulfilled()
+  })
+
+  it('should output bandwidth', async () => {
```
MarcoPolo, line 62: Is there anything this test does test that the above "should run perf" doesn't test? If not, let's get rid of this test.

packages/perf/src/index.ts
```diff
@@ -0,0 +1,208 @@
...
+import type { IncomingStreamData, Registrar } from '@libp2p/interface-internal/registrar'
+import type { AbortOptions } from '@libp2p/interfaces'
+
+const log = logger('libp2p:perf')
+
+export const defaultInit: PerfServiceInit = {
+  protocolName: '/perf/1.0.0',
+  maxInboundStreams: 1 << 10,
+  maxOutboundStreams: 1 << 10,
+  timeout: 10000,
```
MarcoPolo, line 57: We can get rid of this now.

packages/perf/src/index.ts
```diff
@@ -0,0 +1,208 @@
...
+import type { ConnectionManager } from '@libp2p/interface-internal/connection-manager'
+import type { IncomingStreamData, Registrar } from '@libp2p/interface-internal/registrar'
+import type { AbortOptions } from '@libp2p/interfaces'
+
+const log = logger('libp2p:perf')
+
+export const defaultInit: PerfServiceInit = {
+  protocolName: '/perf/1.0.0',
+  maxInboundStreams: 1 << 10,
+  maxOutboundStreams: 1 << 10,
```
MarcoPolo, line 56: Are these even used? Nothing in the spec mentions "maxInboundStreams". I think we should get rid of them.

packages/perf/src/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  async measurePerformance (startTime: number, connection: Connection, sendBytes: bigint, recvBytes: bigint, options: AbortOptions = {}): Promise<number> {
+    log('opening stream on protocol %s to %p', this.protocol, connection.remotePeer)
+
+    const uint8Buf = new Uint8Array(this.databuf)
+
+    const writeBlockSize = this.writeBlockSize
+
+    const signal = anySignal([options?.signal])
```
MarcoPolo, line 157: Can we get rid of the anySignal usage?

MarcoPolo (commented):
I would approve if I could (my own PR).

packages/perf/src/index.ts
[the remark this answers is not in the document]
achingbrain: Look at me making the same point twice 😆 

> you may be limited by how fast you can generate random bytes

You can generate them ahead of time so this is solvable.  I haven't measured it but in my head at least any network operation will be orders of magnitude slower than a pRNG?

packages/perf/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+ *     perf: perfService()
+ *   ]
+ * })
+ *
+ * const connection = await node.dial(multiaddr(multiaddrAddress))
+ *
+ * await node.services.perf.measurePerformance(connection, BigInt(uploadBytes), BigInt(downloadBytes))
+ *
+ * ```
+ */
```
achingbrain, line 40: The docs here are for circuit relay and not perf, they need updating.

packages/perf/src/index.ts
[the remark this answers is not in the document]
MarcoPolo: Let's measure it! On my machine (M1 Max):

```
$  dd if=/dev/urandom of=/dev/null bs=1500 count=100
100+0 records in
100+0 records out
150000 bytes (150 kB, 146 KiB) copied, 0.000466 s, 322 MB/s
```

That's 2.5 Gigabits/s, while fast is not fast enough to saturate a fairly common 10 gbps line.

(Yes you can fiddle with blocksizes, but I think 1500 is fair as it's a common MTU size.)

But what's the issue with sending zero bytes? Is it that you're afraid we'll silently introduce compression on a transport and thus not correctly measure throughput? If that's the case sending zero bytes is actually ideal because it would be obvious if there is compression happening.

There's also the question of arbitrarily compressing at a transport level. That kind of thing opens you up the [TLS CRIME](https://en.wikipedia.org/wiki/CRIME) style attacks. So I don't think we should support compression without being application aware.

PR #1605, opened 2023-03-01
feat: perf benchmark example

MarcoPolo:
I had to expose some types to get this example to work. I know we discussed this before, but I don't remember where we landed on it. My thought is that if I can't copy the simplest protocol (ping) to implement my protocol either the API design is broken or we need to rewrite ping and see how it feels with the new API design.

This is in draft since it makes other changes to js-libp2p that aren't related to this example.

Summary of run from my localhost:

```
❯ node dist/benchmark/src/index.js
TCP+mplex+noise
Download bandwidth is (mbits/s) 289
Upload bandwidth is (mbits/s) 210
TCP+yamux+noise
Download bandwidth is (mbits/s) 149
Upload bandwidth is (mbits/s) 148
TCP+yamux+plaintext
Download bandwidth is (mbits/s) 1098
Upload bandwidth is (mbits/s) 1964
TCP+mplex+plaintext
Download bandwidth is (mbits/s) 1331
Upload bandwidth is (mbits/s) 1808
```

I've not vetted these numbers yet.

### DIFF
--- a/examples/benchmark/src/index.ts
+++ b/examples/benchmark/src/index.ts
@@ -1,0 +1,133 @@
+/* eslint-disable no-console */
+
+import { tcp } from '@libp2p/tcp'
+import { yamux } from '@chainsafe/libp2p-yamux'
+import { mplex } from '@libp2p/mplex'
+import { noise } from '@chainsafe/libp2p-noise'
+import { createLibp2p, type Libp2pOptions, type Libp2pNode } from 'libp2p'
+import { PerfService } from 'libp2p/perf'
+import { plaintext } from 'libp2p/insecure'
+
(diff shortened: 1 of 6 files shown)


achingbrain:
I think this is exactly the sort of thing that would be configured using the changes in this PR: https://github.com/libp2p/js-libp2p/pull/1563

Extra services that you can use to expand the features of libp2p, but ones that are not going to be shipped by default.

MarcoPolo:
I think that'll work. Basically if I could initialize the ping service without it being a special builtin, then I'm happy.

maschad:
Closing in favour of https://github.com/libp2p/js-libp2p/pull/1604

PR #1612, opened 2023-03-03
fix: Remove stream from registry after it's closed

maschad:


src/connection/index.ts
```diff
@@ -136,7 +136,7 @@ export class ConnectionImpl implements Connection {
    * Remove stream registry after it is closed
    */
   removeStream (id: string) {
-
+    this.streams.splice(this.streams.findIndex(s => s.id === id), 1)
```
achingbrain, line 139: This needs to guard on `findIndex` returning `-1`!

```js
[0, 1, 2].splice(-1)
// [2]
```

A test would help here too.

maschad: Good catch! Thanks

maschad:
Not sure how to fix the [caching issue](https://github.com/libp2p/js-libp2p/actions/runs/4377936537/jobs/7662041643#step:4:71)  which is causing the builds to fail but there is a fix which would be required to upgrade to latest connection interface anyways so I opened that PR https://github.com/libp2p/js-libp2p/pull/1622

achingbrain:
Notes from triage discussion - the muxer should be responsible for managing and storing references to streams, not the connection.

PR #1615, opened 2023-03-07
test: Add multidimensional interop test

MarcoPolo:
Implements: https://github.com/libp2p/test-plans/blob/master/multidim-interop/README.md


MarcoPolo:
Future todos:
- [ ] Figure out why firefox webrtc isn't working
- [ ] Figure out why webkit tests fail to create a libp2p node

MarcoPolo:
@ckousik I'm seeing a failure with firefox -> webrtc -> rust (Chrome works fine).  https://github.com/libp2p/js-libp2p/actions/runs/4369651820/jobs/7643671824#step:5:5535 " localCert.getFingerprints is not a function". Have you seen that before? https://github.com/libp2p/js-libp2p-webrtc/blob/4c8806c6d2a1a8eff48f0e2248203d48bd84c065/src/transport.ts#L215

MarcoPolo:
I don't know why windows is failing. It seems like it's failing in other PRs as well, so probably not related to these changes

ckousik:
> @ckousik I'm seeing a failure with firefox -> webrtc -> rust (Chrome works fine). https://github.com/libp2p/js-libp2p/actions/runs/4369651820/jobs/7643671824#step:5:5535 " localCert.getFingerprints is not a function". Have you seen that before? https://github.com/libp2p/js-libp2p-webrtc/blob/4c8806c6d2a1a8eff48f0e2248203d48bd84c065/src/transport.ts#L215

That's odd, since `getFingerprints` is required in w3c https://w3c.github.io/webrtc-pc/#rtccertificate-interface . Looking into it now.

MarcoPolo:
This ready for review. This could be faster, our build of head takes 10m. I haven't spent any time investigating this, but we could probably speed this up.

p-shahi:
@achingbrain Can we review and get this in this week (meant to follow up on this in triage today)?

achingbrain (approved):
LGTM.

We should add the `interop` folder to the list of files that get linted to keep everything consistent.  At the moment there are some rules violations but we can fix them when they actually start to break something.

Adding 10m to the build time seems a bit excessive - can it be faster?

MarcoPolo:
Per sync discussion:

@achingbrain is hesistant to merge this while it takes >10min. If I bring it down to <10min, we'll merge.

MarcoPolo:
Alright, I spent a good bit of time optimizing this and the tests run in 15 min. I think this is as good as we can get it without reworking the test runner itself. (unfortunately docker run startup time is the limiting factor, with each test taking about 2s to run).

This could improve a bit if we make the rust images smaller, but I wouldn't expect too much.

@achingbrain can I get another approval?

thomaseizinger:
> This could improve a bit if we make the rust images smaller, but I wouldn't expect too much.

Tracking this here: https://github.com/libp2p/rust-libp2p/issues/3881

MarcoPolo:
All tests pass! @achingbrain any objections to merging?

thomaseizinger:
@MarcoPolo Do I understand correctly that with this PR js-libp2p will use the same structure of just pointing to a commit-hash in the test-plans repo?

Much in favor of adopting that!

.github/workflows/interop-test.yml
```diff
@@ -0,0 +1,36 @@
...
+      - uses: actions/checkout@v3
+      - uses: ipfs/aegir/actions/cache-node-modules@master
+        with:
+          directories: |
+            ./interop/node_modules
+      - name: Build interop
+        run: (cd interop && npm i && npm run build)
+      - name: Build images
+        run: (cd interop && make)
+      # - name: Save package-lock.json as artifact
```
maschad, line 22: Why is this commented out? I think that if we want tests to be reproducible for a given commit this would be required otherwise we are vulnerable to dep breakages

interop/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+# FROM node:18
+
+# WORKDIR /app
+# COPY . .
+# COPY --from=builder /app/node_modules /app/node_modules
+# RUN npm run build
+# WORKDIR /app/interop
+# COPY --from=builder /app/interop/node_modules /app/interop/node_modules
+# RUN npm run build
```
maschad, line 22: I am guessing this is irrelevant now, if so we should remove.

MarcoPolo:
> Do I understand correctly that with this PR js-libp2p will use the same structure of just pointing to a commit-hash in the test-plans repo?


Yes

MarcoPolo:
@achingbrain this is blocking more interop tests in test plans and webrtc interop tests. Since this is only test code, I'll assume your previous approval is still valid unless you say otherwise within the next ~24h.

interop/test/ping.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+          break
+        case 'quic':
+          options.connectionEncryption = [noise()]
+          break
+        default:
+          throw new Error(`Unknown secure channel: ${SECURE_CHANNEL ?? ''}`)
+      }
+    } else {
+      // Libp2p requires at least one encryption module. Even if unused.
+      options.connectionEncryption = [noise()]
```
achingbrain, line 105: Fixed in https://github.com/libp2p/js-libp2p/pull/1752

interop/test/ping.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+        console.error('unexpected exception in ping test Errors:', err.errors)
+      } else {
+        console.error('unexpected exception in ping test:', err)
+      }
+      throw err
+    } finally {
+      try {
+        // We don't care if this fails
+        await node.stop()
+      } catch { }
```
achingbrain, line 161: If this isn't part of the test, maybe do this in an `afterEach` block?

interop/test/ping.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+  // eslint-disable-next-line complexity
+  it('should ping', async () => {
+    const TRANSPORT = process.env.transport
+    const SECURE_CHANNEL = process.env.security
+    const MUXER = process.env.muxer
+    const isDialer = process.env.is_dialer === 'true'
+    const IP = process.env.ip ?? '0.0.0.0'
+    const timeoutSecs: string = process.env.test_timeout_secs ?? '180'
+
+    const options: Libp2pOptions<{ ping: PingService }> = {
```
achingbrain, line 34: Can all this config be built in a separate file and invoked in a `beforeEach` block?

It's very verbose and doesn't seem like it's part of the test.

interop/test/ping.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+          handshakePlusOneRTTMillis: handshakePlusOneRTT,
+          pingRTTMilllis: pingRTT
+        }))
+      } else {
+        const multiaddrs = node.getMultiaddrs().map(ma => ma.toString()).filter(maString => !maString.includes('127.0.0.1'))
+        console.error('My multiaddrs are', multiaddrs)
+        // Send the listener addr over the proxy server so this works on both the Browser and Node
+        await redisProxy(['RPUSH', 'listenerAddr', multiaddrs[0]])
+        // Wait
+        await new Promise(resolve => setTimeout(resolve, 1000 * parseInt(timeoutSecs, 10)))
```
achingbrain, line 147: This will always take the maximum amount of time, is there no way to wait for whatever it is we want to happen?  Otherwise these tests may spend a lot of time doing nothing.

interop/test/ping.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+          break
+        default:
+          throw new Error(`Unknown muxer: ${MUXER ?? '???'}`)
+      }
+    }
+
+    const node = await createLibp2p(options)
+
+    try {
+      if (isDialer) {
```
achingbrain, line 126: Conditional logic in tests is always raises an eyebrow as you can end up not testing what you think you are testing.

Can this be split into two separate tests?

For example the dialer test could skip itself if the node isn't the dialer on this run.

interop/.aegir.js
```diff
@@ -0,0 +1,82 @@
...
+      })
+
+      try {
+        // We don't care if this fails
+        await redisClient.disconnect()
+      } catch { }
+    }
+  },
+  build: {
+    bundlesizeMax: '18kB'
```
achingbrain, line 80: Do we care about the bundle size here?  Probably not?  This block can be removed.

interop/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "libp2p": "../",
+    "node-fetch": "^3.3.0",
+    "playwright": "^1.31.2",
+    "redis": "4.5.1"
+  },
+  "browser": {
+    "@libp2p/tcp": false
+  },
+  "devDependencies": {
+    "aegir": "^38.1.0",
```
achingbrain, line 37: Please align aegir version with the main repo.

interop/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "standard": "^17.0.0",
+    "typescript": "^4.9.4"
```
achingbrain, line 39: Aegir comes with standard and typescript, we probably don't need them here?

achingbrain (changes requested):

interop/test/ping.spec.ts
MarcoPolo: This is on purpose. The test runner will shutdown the listener when the dial completes. See https://github.com/libp2p/test-plans/tree/master/multidim-interop#listener

interop/test/ping.spec.ts
MarcoPolo: Why? Every other [implementation](https://github.com/libp2p/go-libp2p/blob/master/test-plans/cmd/ping/main.go#L161) [does](https://github.com/libp2p/rust-libp2p/blob/master/interop-tests/src/bin/ping.rs#L127) [it](https://github.com/libp2p/test-plans/blob/master/multidim-interop/impl/nim/mainv1.nim#L70) [this](https://github.com/MarcoPolo/zig-libp2p/blob/main/interop/main.zig#L118) way.

interop/.aegir.js
MarcoPolo: Vestigal. Will remove.

interop/test/ping.spec.ts
MarcoPolo: I don't see the benefit of having a separate file, but I can move to a beforeEach.

interop/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "overrides": {
+    "playwright-core": "^1.31.2"
+  }
```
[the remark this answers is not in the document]
MarcoPolo: If I recall correctly, we needed a newer browser for some test. This shouldn't be needed after https://github.com/hugomrdias/playwright-test/pull/528/files.

If we package.json supported comments it would have been easier to capture the rational here.

interop/test/ping.spec.ts
achingbrain: Presumably these tests will test more than just ping in the future, at that point we're going to want to set up a libp2p node in more than one place, hence the external file.

We can make that refactor when we need it, but also this is setup code and not test code so it shouldn't be in the `it` function.

interop/test/ping.spec.ts
achingbrain: Looking at the other implementations, none of them are written using a test framework so I don't think they are an example to follow here.

MarcoPolo: Maybe we shouldn’t use the test framework? I only used it because it was convenient. Happy to not use it.

interop/test/ping.spec.ts
MarcoPolo: This is testing whether two libp2p nodes can communicate. A future test may look similar, but I don’t think it makes sense to refactor until we have that future test.

interop/test/ping.spec.ts
MarcoPolo: Do you feel strongly enough against the test framework as to request changes before merging?

I don’t think we should deviate from what the four other implementations do just follow the convention on this test framework, so I would suggest keeping the conditional as is, and , if requested, moving off the test framework. 

I would be against splitting this into two tests.

achingbrain: I think using a test framework to write tests is the right thing to do, as is integrating with the existing tooling of a project because it makes the code instantly familiar to other developers so I think this is the right decision.

I think regardless of framework tests have a certain structure (setup, exercise, verify, teardown), duplication should be minimised and that conditional logic in tests means that later down the line when something goes wrong the author will have to think about which branch is being executed which is a code smell and should be avoided.

I also think that the first test is the most important since people will use it as a starting point for subsequent tests and am somewhat confused as to why you are pushing back on this so hard.

maschad:
> I've split up the test @achingbrain, but what's annoying now is that one or the other test will always be skipped but it will still spin up and shutdown a libp2p node because of beforeEach and afterEach. Any suggestions? We might be able to replace it(... with it.skip(... dynamically and thus skip the test at the higher level, but that seems a bit hacky.

I agree @MarcoPolo it seems to get a bit messy going that route. What do you think about splitting the tests into two separate directories i.e. one for dialler tests and one for listener tests (albeit they are both just one test each at the moment) and we can use just use an `npm` script like 

```
"test": if [ \"$is_dialer\" = \"true\" ]; then aegir test -t node -f \".dist/test/dialer/*.{node,spec}.js\"; else aegir test -t node -f \".dist/test/listener/*.{node,spec}.js\"; fi
```

I know we have been going back and forth on this for a while so if you and @achingbrain agree then I could make these changes so that we can land this.

MarcoPolo:
> > I've split up the test @achingbrain, but what's annoying now is that one or the other test will always be skipped but it will still spin up and shutdown a libp2p node because of beforeEach and afterEach. Any suggestions? We might be able to replace it(... with it.skip(... dynamically and thus skip the test at the higher level, but that seems a bit hacky.
> 
> I agree @MarcoPolo it seems to get a bit messy going that route. What do you think about splitting the tests into two separate directories i.e. one for dialler tests and one for listener tests (albeit they are both just one test each at the moment) and we can use just use an `npm` script like
> 
> ```
> "test": if [ \"$is_dialer\" = \"true\" ]; then aegir test -t node -f \".dist/test/dialer/*.{node,spec}.js\"; else aegir test -t node -f \".dist/test/listener/*.{node,spec}.js\"; fi
> ```

I'm strongly against separate files for this test. I also don't like using bash scripting as the way to check which test to run. Why are we introducing bash into this when we're already working with TypeScript?

maschad:
> I'm strongly against separate files for this test. 

why is that? As it is now [they are two separate tests](https://github.com/libp2p/js-libp2p/pull/1615/files#diff-024eabde3750f3bf282cd9a642bd16ba0357c8c0b3110ffa4c78bfdd60f9ed2fR135) in one file, I think that having them in separate files improves the modularity and reporting (as you wouldn't see the logs of the test being skipped)

>  I also don't like using bash scripting as the way to check which test to run. Why are we introducing bash into this when we're already working with TypeScript?

I can appreciate that but I think they serve two different purposes in this case. I would say it's more common to see bash scripting accessing environment variables than Typescript. I would also say it's more intuitive than [the dynamic way of skipping tests](https://github.com/libp2p/js-libp2p/pull/1615/files#diff-024eabde3750f3bf282cd9a642bd16ba0357c8c0b3110ffa4c78bfdd60f9ed2fR135)

Regardless, I wouldn't let either of those concerns block this from being landed and I see that [tests pass now](https://github.com/libp2p/js-libp2p/actions/runs/5063541633/jobs/9090264246?pr=1615)  🚀. We can always optimize in a follow up PR 😄

MarcoPolo:
> why is that? As it is now [they are two separate tests](https://github.com/libp2p/js-libp2p/pull/1615/files#diff-024eabde3750f3bf282cd9a642bd16ba0357c8c0b3110ffa4c78bfdd60f9ed2fR135) in one file, I think that having them in separate files improves the modularity and reporting (as you wouldn't see the logs of the test being skipped)

Having more files here doesn't improve modularity, it adds indirection. You may as well run the test command and grep for only the tests you want to run.

Is it common to see Bash scripting used in `npm test`? I've done it a couple of times, but only as a last resort.

maschad (approved):

MarcoPolo:
> You may as well run the test command and grep for only the tests you want to run.

Sorry this wasn't clear. I was using this an example of what you wouldn't want to do.

p-shahi:
Poking in here: were the discussions mostly resolved i.e. is this guy almost there to getting merged?

achingbrain:
I meant something like:

```js
it('should be a listener', function () {
  if (is_dialer) {
    return this.skip()
  }

  // listener test code here
})
```

The dynamic skip:

```js
(isDialer ? it.skip : it)('should listen for ping', async () => {
```

is reminiscent of [what we did](https://github.com/ipfs/js-ipfs/blob/master/packages/interface-ipfs-core/src/utils/mocha.js#L98) for the js-ipfs interface tests.  It worked, though typing it all was a little hairy.

I think one file for the ping test is ok, definitely don't want to be debugging bash scripts to figure out why tests aren't running.

I think we should get this in and we can finesse it in follow PRs.

achingbrain (approved):

Issue #1608, opened 2023-03-02
Circuit Relay v2 follow up work items


p-shahi:
I am going to merge this which will publish an RC to unblock other efforts as it is functional though it still needs some work in a follow-up before being released.

_Originally posted by @achingbrain in https://github.com/libp2p/js-libp2p/pull/1533#pullrequestreview-1322333583_

```[tasklist]
### Tasks
- [x] Split the code into two parts - the transport and the relay server
- [x] Configure the transport and relay server separately - at the moment the server is partially configured under the `config.relay.hop` key and `config.relay.advertise` - the fact that the docs say "Does not make you a relay" in several places mean it's really not obvious
- [x] Maybe even have a separate CircuitRelay transport that you can configure
- [x] More tests
```

Issue #1610, opened 2023-03-03
Add hop relay listening functionality to the Connection Manager


maschad:
Whenever we do not have enough relays, we need to try to listen on available hop relay connections. This should be the responsibility of the connection manager as opposed to the [circuit relay client](https://github.com/libp2p/js-libp2p/blob/593deefe1268e0019354964b87b49fac1a3981ed/src/circuit/client.ts#L257)

achingbrain:
The problem with this is that it requires the connection manager to know about relays and their semantics and configuration which makes it hard to make relay functionality opt-in.

achingbrain:
Closing after triage discussion

PR #1619, opened 2023-03-08
fix: circuit relay v2 follow up items

achingbrain:
Code sanitation:

- splits code and config into transport and server
- updates docs
- consolidates relay and circuit tests
- updates relay discovery to use topology
- increases test coverage
- reuses connection gater for acl instead of introducing a new acl mechanism
- adds migration guide

Bug fixes:

- if multiple relays are discovered simultaneously, try to reserve a slot in them in series, otherwise we can end up listening on multiple relays by accident
- relay server tags the reserving peer instead of tagging itself
- incoming remote address was missing `p2p-circuit` tuple so we were allowing relay to relay forwarding
- relay client tags outgoing relay connections
- limit data transfer as total in both directions instead of apply limit to both sides of the relayed stream separately
- allow client to refuse incoming relayed connections

fixes: #1608
refs: #1610


doc/migrations/v0.42-v0.43.md
```diff
@@ -0,0 +1,79 @@
...
+
+A migration guide for refactoring your application code from libp2p v0.42.x to v0.43.0.
+
+## Table of Contents <!-- omit in toc -->
+
+- [Circuit Relay v2](#circuit-relay-v2)
+
+## Circuit Relay v2
+
+libp2p@0.43.x ships with support for [Circuit Relay v2](https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md).  This update to the spec changes the relay from being an open relay by default, and as such quite dangerous to enable, to being a limited relay with a slot reservation mechanism that is much safer.
```
MarcoPolo, line 11: Does the JS impl use ASNs to limit the peers?

src/circuit/server/advert-service.ts
```diff
@@ -0,0 +1,108 @@
...
+const log = logger('libp2p:circuit-relay:advert-service')
+
+export interface AdvertServiceInit {
+  /**
+   * How long to wait after startup to begin advertising the service
+   * - if some configured content routers take a while to warm up (for
+   * example, the DHT needs some peers to be able to publish) this
+   * value should be high enough that they will have warmed up
+   */
+  bootDelay?: number
```
MarcoPolo, line 22: seems like a hack, no? Should the content routers know to hold off on publishing until they are warm enough? This component would have less context on that than the content router itself.

src/circuit/server/index.ts
```diff
@@ -0,0 +1,427 @@
...
+      hopstr.write({ type: HopMessage.Type.STATUS, status: result.status })
+      return
+    }
+
+    try {
+      // tag relay target peer
+      // result.expire is non-null if `ReservationStore.reserve` returns with status == OK
+      if (result.expire != null) {
+        const ttl = new Date().getTime() - result.expire
+        await this.peerStore.tagPeer(connection.remotePeer, RELAY_SOURCE_TAG, { value: 1, ttl })
```
MarcoPolo, line 262: what does `value: 1` mean?

src/circuit/server/index.ts
```diff
@@ -0,0 +1,427 @@
...
+
+      request.peer.addrs.forEach(multiaddr)
+      dstPeer = peerIdFromBytes(request.peer.id)
+    } catch (err) {
+      log.error('invalid hop connect request via peer %p %s', connection.remotePeer, err)
+      hopstr.write({ type: HopMessage.Type.STATUS, status: Status.MALFORMED_MESSAGE })
+      return
+    }
+
+    if (isRelayAddr(connection.remoteAddr)) {
```
MarcoPolo, line 328: why twice? We already did this check above

src/circuit/server/index.ts
```diff
@@ -0,0 +1,427 @@
...
+    const connections = this.connectionManager.getConnections(dstPeer)
+
+    if (connections.length === 0) {
+      log('hop connect denied for destination peer %p not having a connection for %p as there is no destination connection', dstPeer, connection.remotePeer)
+      hopstr.write({ type: HopMessage.Type.STATUS, status: Status.NO_RESERVATION })
+      return
+    }
+
+    const destinationConnection = connections[0]
+    log('hop connect request from %s to %s is valid', connection.remotePeer, dstPeer)
```
MarcoPolo, line 355: Is there a debug log level so that we don't log this all the time under normal operation?

src/circuit/server/index.ts
```diff
@@ -0,0 +1,427 @@
...
+
+    const destinationStream = await this.stopHop({
+      connection: destinationConnection,
+      request: {
+        type: StopMessage.Type.CONNECT,
+        peer: {
+          id: connection.remotePeer.toBytes(),
+          addrs: [multiaddr(`/p2p/${connection.remotePeer.toString()}`).bytes]
```
MarcoPolo, line 363: Do we need the addrs field? It doesn't say so in the spec: https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md#stop-protocol

MarcoPolo (commented):
Quick review. This looks good. I think it would be better if we had some metrics here and can see how this behaves after some time on the real network. Can other nodes use it? Can it use other nodes?

src/circuit/server/index.ts
achingbrain: The connection manager has a max connections setting - when a node breaches this setting the connection manager chooses connections to prune.

It does this by sorting the connections by peer value and peer values are derived from their tags.

Random incoming dials will have a value of 0 so will be pruned first - setting the tag value to 1 here makes the peer ever so slightly more valuable than a random incoming dial but not as valuable as a peer that is kad-close, a bootstrap node during startup or a in a gossipsub topic mesh, for example.

src/circuit/server/index.ts
achingbrain: I don't think so, no.  There's no information there that isn't elsewhere in the message anyway.

src/circuit/server/index.ts
achingbrain: FWIW under normal operation these log lines are all no-ops, stuff is only logged if the `DEBUG` env var is set.

doc/migrations/v0.42-v0.43.md
achingbrain: Not out of the box, but the connection gater interface can be used to do this.

achingbrain:
I'm going to merge this to unblock the other PRs.  We should hold off release until the multiaddr/webrtc naming thing is resolved.

Adding metrics is a good idea, this can be done as a follow up.

Issue #1616, opened 2023-03-08
Batch dialling


achingbrain:
The `libp2p.dial` method should support dialling multiple multiaddrs - it currently supports dialing a peer id (in which case multiaddrs are retrieved from the peer store) or a single multiaddr.

Since only a single connection is returned from a dial, the multiaddrs being dialled should be checked to ensure they either all contain the same peer id or that none of them contain a peer id.

If they all have the same peer id we are guaranteed to return a connection to the expected remote peer, if none of them contain a peer id then the user did not expect a connection to a specific peer so we are ok to return a connection to whatever remote peer we managed to connect to first.

maschad:
This may be a strange case but how would we handle a situation if they have different `peerIDs` for the multiaddrs? return the one we connect to first?

PR #1623, opened 2023-03-11
feat: support batch dialling 

maschad:
Closes https://github.com/libp2p/js-libp2p/issues/1616


src/connection-manager/dialer/index.ts
```diff
@@ -171,10 +175,21 @@ export class DefaultDialer implements Startable, Dialer {
 
     log('creating dial target for %p', peerId)
 
+    if (Array.isArray(multiaddr)) {
+      // Return the first successful connection
+      return (await Promise.all(multiaddr.map(async (addr) => await this.createDialTarget(peerId, addr, options))))[0]
```
achingbrain, line 180: We need to abort the creation of any other connections and/or close them, otherwise this will leak memory and sockets.

achingbrain:
I've pushed a commit here - we already have the capability to dial multiple multiaddrs in one dial target, since when we resolve a dnsaddr, they sometimes resolve to more than one multiaddr.

The change I've made uses this mechanism to dial multiple passed multiaddrs and to handle cancelling other dials once one succeeds.

It also adds more tests around ensuring other dials are actually cancelled, and that we enforce rules around peer ids being consistent (or not specified) in multiple dials.

test/dialing/direct.spec.ts
```diff
@@ -378,6 +378,84 @@ describe('Dialing (direct, WebSockets)', () => {
     expect(dialTarget).to.have.property('addrs').with.lengthOf(1)
     expect(dialTarget.addrs[0].toString()).to.equal(dialMultiaddr.toString())
   })
+
+  it('Should dial multiple multiaddrs and return first successful connection', async () => {
```
achingbrain, line 382:
```suggestion
  it('should dial multiple multiaddrs and return first successful connection', async () => {
```

test/dialing/direct.spec.ts
```diff
@@ -378,6 +378,84 @@ describe('Dialing (direct, WebSockets)', () => {
...
+
+    // first and third dials should have been aborted, second was successful so should not have been
+    expect(tmDialStub.getCall(0)).to.have.nested.property('args[1].signal.aborted', true)
+    expect(tmDialStub.getCall(1)).to.have.nested.property('args[1].signal.aborted', false)
+    expect(tmDialStub.getCall(2)).to.have.nested.property('args[1].signal.aborted', true)
+
+    await dialer.stop()
+  })
+
+  it('Should throw if dialling an empty array is attempted', async () => {
```
achingbrain, line 417:
```suggestion
  it('should throw if dialling an empty array is attempted', async () => {
```

test/dialing/direct.spec.ts
```diff
@@ -378,6 +378,84 @@ describe('Dialing (direct, WebSockets)', () => {
...
+    const dialer = new DefaultDialer(localComponents)
+
+    // Perform dial
+    await expect(dialer.dial([])).to.eventually.rejected
+      .with.property('code', 'ERR_NO_VALID_ADDRESSES')
+
+    await dialer.stop()
+  })
+
+  it('Should throw if dialling multiaddrs with mismatched peer ids', async () => {
```
achingbrain, line 427:
```suggestion
  it('should throw if dialling multiaddrs with mismatched peer ids', async () => {
```

test/dialing/direct.spec.ts
```diff
@@ -378,6 +378,84 @@ describe('Dialing (direct, WebSockets)', () => {
...
+    await expect(dialer.dial([
+      multiaddr(`/ip4/0.0.0.0/tcp/8000/ws/p2p/${(await createPeerId()).toString()}`),
+      multiaddr(`/ip4/0.0.0.0/tcp/8001/ws/p2p/${(await createPeerId()).toString()}`)
+    ])).to.eventually.rejected
+      .with.property('code', 'ERR_INVALID_PARAMETERS')
+
+    await dialer.stop()
+  })
+
+  it('Should throw if dialling multiaddrs with inconsistent peer ids', async () => {
```
achingbrain, line 440:
```suggestion
  it('should throw if dialling multiaddrs with inconsistent peer ids', async () => {
```

PR #1626, opened 2023-03-15
fix!: remove connection manager autodial option

achingbrain:
The connection manager's `autoDial` option originally meant "dial every peer we discover".  Then it changed to "if we have fewer than `minConnections`, dial peers from the peer store until we have more than `minConnections`".

This is confusing and also redundant, since if we don't want to dial peers to ensure we have more than `minConnections` we can just set `minConnections` to `0`.

Adds a section to the upgrade guide covering the removal.

Also fixes a bug where we actually configured two autodialer components.

### DIFF
--- a/doc/PEER_DISCOVERY.md
+++ b/doc/PEER_DISCOVERY.md
@@ -22,6 +22,7 @@
 * To ensure reasonable resource usage, discovered peers are not connected to automatically
 * Applications should lisen for the `peer:connect` event if they wish to take a specific action when new connections are established
 * Libp2p functions best with a good number of network connections to a diverse set of peers. When the number of connected peers a node has falls under the [ConnectionManager](./CONFIGURATION.md#configuring-connection-manager) `minConnections` setting, randomly selected peers from the peer store will be dialed until the node's number of connections rises above this number.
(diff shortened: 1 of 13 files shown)


doc/PEER_DISCOVERY.md
```diff
@@ -22,7 +22,7 @@
 * To ensure reasonable resource usage, discovered peers are not connected to automatically
 * Applications should lisen for the `peer:connect` event if they wish to take a specific action when new connections are established
 * Libp2p functions best with a good number of network connections to a diverse set of peers. When the number of connected peers a node has falls under the [ConnectionManager](./CONFIGURATION.md#configuring-connection-manager) `minConnections` setting, randomly selected peers from the peer store will be dialed until the node's number of connections rises above this number.
-  * Applications can disable this behaviour via the `connectionManager.autoDial` config property, and handle increasing the current number of connections themselves
+  * Applications can disable this behaviour by setting the `connectionManager.minConnections` config property to `0`, and handle increasing the current number of connections themselves
```
maschad, line 25:
```suggestion
* Applications can disable this behaviour by setting the `connectionManager.minConnections` config property to `0`, but they will have to manage increasing the current number of connections themselves.
```

maschad (commented):
LGTM just left a minor linguistic suggestion.

PR #1627, opened 2023-03-16
test: Add webkit test script

MarcoPolo:
This fails on linux. Works on MacOS. I haven't investigated why we fail on linux, but it does block us testing webkit as part of the multidim-inderop tester.


MarcoPolo:
Are we not using https://github.com/protocol/.github/pull/487 in this repo?

achingbrain:
From the CI error it looks like the job just needs to install some extra deps?

```
╔══════════════════════════════════════════════════════╗
║ Host system is missing dependencies to run browsers. ║
║ Please install them with the following command:      ║
║                                                      ║
║     sudo npx playwright install-deps                 ║
║                                                      ║
║ Alternatively, use apt:                              ║
║     sudo apt-get install libvpx7\                    ║
║         libevent-2.1-7\                              ║
║         libopus0\                                    ║
║         libwoff1\                                    ║
║         libharfbuzz-icu0\                            ║
║         libgstreamer-plugins-base1.0-0\              ║
║         libgstreamer-gl1.0-0\                        ║
║         libhyphen0\                                  ║
║         libmanette-0.2-0\                            ║
║         libgles2\                                    ║
║         gstreamer1.0-libav                           ║
║                                                      ║
║ <3 Playwright Team                                   ║
╚══════════════════════════════════════════════════════╝]
```

achingbrain:
Thinking about it a bit more, Webkit is Webkit so do we need to run it on both?

Running it on Linux means it'll pull the cached deps from the `check` job which'll be faster.

MarcoPolo:
The failure is when we do something like this:

```
await crypto.subtle.deriveKey(
    { "name": "PBKDF2", "salt": new Uint8Array(16), "iterations": 32767, "hash": { "name": "SHA-256" } },
    await crypto.subtle.importKey('raw', pass, { "name": "PBKDF2" }, false, ['deriveKey', 'deriveBits']),
    { name: "AES-GCM", length: 128 }, true, ['encrypt'])
```

here: https://github.com/libp2p/js-libp2p-crypto/blob/e7bb8b26655af7d0d8a7387a2923c6c3929952fe/src/ciphers/aes-gcm.browser.ts#L35

MarcoPolo:
Possibly related: https://bugs.webkit.org/show_bug.cgi?id=169468#c1

> I'm guessing these are failing because there is no PBKDF2 implementation on GTK, and that once someone hooks it up things will start working.

MarcoPolo:
This is blocked on https://github.com/libp2p/js-libp2p-crypto/pull/319

src/autonat/index.ts
```diff
@@ -508,7 +508,7 @@ class DefaultAutoNATService implements Startable {
           // they either told us which address worked/didn't work, or we only sent them one address
           const addr = dialResponse.addr == null ? multiaddrs[0] : multiaddr(dialResponse.addr)
 
-          log('Autonat response for %s is %s', addr, dialResponse.status)
+          log('Autonat response for %s is %s', addr.toString(), dialResponse.status)
```
achingbrain, line 511: As a rule we shouldn't cause work to be done while logging things.

If we need to call methods on an object to log it properly a [custom formatter](https://github.com/libp2p/js-libp2p-logger/blob/master/src/index.ts#L9-L37) should be created so the functions are only called if logging is enabled.

src/autonat/index.ts
```diff
@@ -551,7 +551,7 @@ class DefaultAutoNATService implements Startable {
 
           if (results[addrStr].failure === REQUIRED_SUCCESSFUL_DIALS) {
             // we are now unconvinced
-            log('%s is not externally dialable', addr)
+            log('%s is not externally dialable', addr.toString())
```
achingbrain, line 554: Same comment about custom formatters.

test/autonat/index.spec.ts
```diff
@@ -92,19 +96,28 @@ describe('autonat', () => {
...
         type: Message.MessageType.DIAL_RESPONSE,
         dialResponse
       })
-      stream.source = (async function * () {
-        yield lp.encode.single(response)
-      }())
-      stream.sink.returns(Promise.resolve())
+
+      // stub autonat protocol stream
+      const stream = stubObject<Stream>({
```
achingbrain, line 105: Trying to run this branch locally I noticed something weird with the stubbing but only in WebKit, though I didn't get to the bottom of it.

We stub a bunch of peer responses in each test to simulate different scenarios - once a certain number of tests were being run, the stubbed return values started being from the wrong tests.

Needs more investigation really.

achingbrain (changes requested):

test/autonat/index.spec.ts
MarcoPolo: Yes. Hence this change. I'm not sure why exactly this fails on WebKit but seems to work on other browsers. I'm not sure overriding values is supported in Sinon or if it's a hack we do. The change here is to not override and instead be explicit in what gets stubbed vs set.

My guess is the other failures we're seeing is related to something like this.

achingbrain: Have you checked out the Sinon docs? I believe we’re using it as intended, no “hack” involved.

MarcoPolo: I think sinon is not involved here. This should just be setting values on the object here: https://github.com/ttarnowski/ts-sinon/blob/master/src/index.ts#L67. Maybe `Proxy` in WebKit has slightly different behavior? Maybe a bug since this only shows up in certain cases (e.g. run only this test is fine, but run two versions of this test fails).

achingbrain: Yeah, it’s really weird because we don’t reuse the stubbed instances between the tests that I can see 🤔

src/autonat/index.ts
maschad: I've opened https://github.com/libp2p/js-libp2p-logger/pull/34 to address this

MarcoPolo:
@maschad CI is still failing in webkit fyi

maschad:
> @maschad CI is still failing in webkit fyi

Thanks @MarcoPolo I actually created [an issue](https://github.com/libp2p/js-libp2p/issues/1781) as I don't believe this failure [is specific](https://github.com/libp2p/js-libp2p/actions/runs/5081049903/jobs/9128906249) our webkit tests

MarcoPolo:
Is this test more likely to fail on WebKit?

Can we handle the flaky test separately (maybe skip for now) and merge WebKit tests now? It would be good to make sure we don't break WebKit from now on, and it would be useful for WebRTC testing.

maschad:
> Is this test more likely to fail on WebKit?

It seems to fail consistently on Webkit based on the last few job runs. 

> Can we handle the flaky test separately (maybe skip for now) and merge WebKit tests now? It would be good to make sure we don't break WebKit from now on, and it would be useful for WebRTC testing.

Good suggestion, I push a skip for now.

maschad:
> If it fails consistently on WebKit, it doesn't sound like a flaky test to me. Does it fail locally when you run this on a linux box?

Good suggestion for the linux vm, it wasn't failing on Mac. It does fail consistently on linux with a stream reset error, I am investigating now.

maschad:
Turns out the nodes were not shutting down properly in some [identify service tests](https://github.com/libp2p/js-libp2p/pull/1627/commits/a78db93c14ede84379596a2fc687a7ee4620ccf7#diff-bdd56911c7c55ece439a377d4754987ac0090a295fed80e0d063470d1acafe2dR73) 

@achingbrain I do notice that the said tests addressed in a78db93 _intentionally_ close before the connection is established, but that results in this error in the logs.

```bash
WebSocket connection to 'ws://127.0.0.1:15001/p2p/12D3KooWHFKTMzwerBtsVmtz4ZZEQy2heafxzWw6wNn5PPYkBxJ5' failed: WebSocket is closed before the connection is established.
```

it's not too noisy but perhaps we should surpess this?

maschad:
> That's a bit weird - the libp2p node is [shut down in the afterEach](https://github.com/libp2p/js-libp2p/pull/1627/files#diff-bdd56911c7c55ece439a377d4754987ac0090a295fed80e0d063470d1acafe2dR32-R38) for those tests.

That's a good point @achingbrain could it be that the `afterEach` is called after the [failing test](https://github.com/libp2p/js-libp2p/pull/1627/commits/3d6a67ed8fd7949e585bc52384515e6a64110bed#diff-79d61f88bb686a9134031ca95cc22ffa9666d3076dcfe876430d35b007b7f508R672) is run in some sort of race condition ?  Either way I am not sure if this related directly to this PR but seems to be a testing environment issue.

test/autonat/index.spec.ts
MarcoPolo: @maschad did you figure out what the issue was here around Proxy?

maschad: @MarcoPolo I don't think the issue was related to `Proxy` but rather the peer id fixtures, the relay server behaved as if it was creating reservations for peers that _seemed_ to be reconnecting due to the tests reusing peer ids, we resolved this [in this PR](https://github.com/libp2p/js-libp2p/pull/1795)

PR #1636, opened 2023-03-21
fix: interpret circuit relay expiry as seconds

### LINKED ISSUE #1635: circuitRelayTransport repeatedly reconnecting to relay
- **Version**:

- **Platform**:

Linux archlinux 5.15.59-2-lts #1 SMP Sat, 06 Aug 2022 21:34:17 +0000 x86_64 GNU/Linux

- **Subsystem**:

Relay

#### Description:

- Connected a libp2p node to a Go libp2p node serving circuit v2 relay.
- A relay slot was repeatedly being reserved on the Go relay (looked like an infinite loop).
- It should only reserve the slot once and wait for a timeout to trigger.

Instantiate the go relay with `GOLOG_LOG_LEVEL=debug go run main.go`. It should print out a websocket multiaddr.
```go
package main

import (
	"crypto/rand"
	"fmt"

	"github.com/libp2p/go-libp2p"
	"github.com/libp2p/go-libp2p/core/crypto"
	"github.com/libp2p/go-libp2p/core/host"

	// "github.com/libp2p/go-libp2p/core/peer"
	"github.com/libp2p/go-libp2p/p2p/muxer/mplex"
	relay "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/relay"
	ma "github.com/multiformats/go-multiaddr"
)

func main() {
	makeRelayV2()
	select {}
}

func makeRelayV2() host.Host {
	r := rand.Reader
	// Generate a key pair for this host. We will use it at least
	// to obtain a valid host ID.
	priv, _, err := crypto.GenerateKeyPairWithReader(crypto.RSA, 2048, r)
	if err != nil {
		panic(err)
	}

	opts := []libp2p.Option{
		libp2p.DefaultTransports,
		libp2p.ListenAddrStrings(
			"/ip4/0.0.0.0/tcp/4003/ws",
		),
		libp2p.Muxer("/mplex/6.7.0", mplex.DefaultTransport),
		libp2p.Identity(priv),
		libp2p.EnableRelay(),
	}

	host, err := libp2p.New(opts...)
	if err != nil {
		panic(err)
	}

	_, err = relay.New(host)
	if err != nil {
		panic(err)
	}

	// fmt.Println(host.Mux().Protocols())

	for _, addr := range host.Addrs() {
		a, err := ma.NewMultiaddr(fmt.Sprintf("/p2p/%s", host.ID().Pretty()))
		if err != nil {
			panic(err)
		}
		fmt.Println("p2p addr: ", addr.Encapsulate(a))
	}
	return host
}
```

Instantiate the node process with `DEBUG=libp2p:circuit-relay:* node index.js`
```js
import { noise } from "@chainsafe/libp2p-noise";
import { mplex } from "@libp2p/mplex";
import { webSockets } from "@libp2p/websockets";
import * as filters from  "@libp2p/websockets/filters";
import { createLibp2p } from "libp2p";
import { circuitRelayTransport } from "libp2p/circuit-relay";
import { multiaddr } from "@multiformats/multiaddr";

    transports: [
        webSockets({
            filters: filters.all,
        }),
        circuitRelayTransport({
            discoverRelays: 1,
        }),
    ],
    streamMuxers: [mplex()],
});


await node.dial(multiaddr(process.argv[2]))

```
The node and go processes should keep logging a new circuit relay reservation.

ckousik:
This fixes #1635 which causes the circuit relay to repeatedly connect to a discovered relay. If the relay returns connection expiration in seconds, we get a negative ttl when calculating
`expiration - new Date().getTime()`.
This caused the `addRelay` function to set the timeout at 0ms. This timeout instantly triggers and calls `addRelay` recursively.

### DIFF
--- a/src/circuit/transport/reservation-store.ts
+++ b/src/circuit/transport/reservation-store.ts
@@ -8,7 +8,7 @@ import type { ConnectionManager } from '@libp2p/interface-connection-manager'
 import type { Connection } from '@libp2p/interface-connection'
 import type { Reservation } from '../pb/index.js'
 import { HopMessage, Status } from '../pb/index.js'
-import { getExpirationSeconds } from '../utils.js'
+import { getExpirationMilliseconds } from '../utils.js'
 import type { TransportManager } from '@libp2p/interface-transport'
 import type { Startable } from '@libp2p/interfaces/dist/src/startable.js'
 import { CustomEvent, EventEmitter } from '@libp2p/interfaces/events'
@@ -131,7 +131,7 @@ export class ReservationStore extends EventEmitter<ReservationStoreEvents> imple
         const existingReservation = this.reservations.get(peerId)
 
         if (existingReservation != null) {
-          if (getExpirationSeconds(existingReservation.reservation.expire) > REFRESH_WINDOW) {
+          if (getExpirationMilliseconds(existingReservation.reservation.expire) > REFRESH_WINDOW) {
             log('already have reservation on relay peer %p and it expires in more than 10 minutes', peerId)
             return
           }
@@ -162,7 +162,7 @@ export class ReservationStore extends EventEmitter<ReservationStoreEvents> imple
 
         log('created reservation on relay peer %p', peerId)
 
-        const expiration = getExpirationSeconds(reservation.expire)
+        const expiration = getExpirationMilliseconds(reservation.expire)
 
         const timeout = setTimeout(() => {
           this.addRelay(peerId, type).catch(err => {
@@ -171,7 +171,7 @@ export class ReservationStore extends EventEmitter<ReservationStoreEvents> imple
         },
         // TODO: If we get an expiration of 0, this will keep calling
         // addRelay
-        Math.max(expiration * 1000 - REFRESH_TIMEOUT, 0))
+        Math.max(expiration - REFRESH_TIMEOUT, 0))
 
         this.reservations.set(peerId, {
           timeout,
--- a/src/circuit/transport/reservation-store.ts
+++ b/src/circuit/transport/reservation-store.ts
@@ -23,7 +23,7 @@ const REFRESH_WINDOW = (60 * 1000) * 10
 // try to refresh relay reservations 5 minutes before expiry
 const REFRESH_TIMEOUT = (60 * 1000) * 5
 
-// maximum duration before which a reservation should be refereshed (2 hrs)
+// maximum duration before which a reservation should be refereshed
 const REFRESH_TIMEOUT_MAX = (60 * 1000) * 15
 // minimum duration before which a reservation must not be refreshed
 const REFRESH_TIMEOUT_MIN = 30 * 1000
--- a/src/circuit/transport/reservation-store.ts
+++ b/src/circuit/transport/reservation-store.ts
@@ -24,7 +24,9 @@ const REFRESH_WINDOW = (60 * 1000) * 10
 const REFRESH_TIMEOUT = (60 * 1000) * 5
 
 // maximum duration before which a reservation should be refereshed (2 hrs)
-const REFRESH_TIMEOUT_MAX = (60 * 1000) * 120
+const REFRESH_TIMEOUT_MAX = (60 * 1000) * 15
+// minimum duration before which a reservation must not be refreshed
+const REFRESH_TIMEOUT_MIN = 30 * 1000
 
 export interface RelayStoreComponents {
   peerId: PeerId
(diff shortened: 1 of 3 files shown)


src/circuit/transport/reservation-store.ts
```diff
@@ -162,13 +162,16 @@ export class ReservationStore extends EventEmitter<ReservationStoreEvents> imple
...
 
         const timeout = setTimeout(() => {
           this.addRelay(peerId, type).catch(err => {
             log.error('could not refresh reservation to relay %p', peerId, err)
           })
-        }, Math.max(expiration - REFRESH_TIMEOUT, 0))
+        },
+        // TODO: If we get an expiration of 0, this will keep calling
+        // addRelay
+        Math.max(expiration * 1000 - REFRESH_TIMEOUT, 0))
```
ckousik, line 174: could this cause a 32 bit overflow?

src/circuit/transport/reservation-store.ts
```diff
@@ -131,7 +131,7 @@ export class ReservationStore extends EventEmitter<ReservationStoreEvents> imple
         const existingReservation = this.reservations.get(peerId)
 
         if (existingReservation != null) {
-          if (getExpiration(existingReservation.reservation.expire) > REFRESH_WINDOW) {
+          if (getExpirationSeconds(existingReservation.reservation.expire) > REFRESH_WINDOW) {
```
achingbrain, line 134: It looks like you're comparing seconds to milliseconds here?

src/circuit/utils.ts
```diff
@@ -115,6 +115,7 @@ export async function namespaceToCid (namespace: string): Promise<CID> {
 /**
  * returns number of ms between now and expiration time
  */
-export function getExpiration (expireTime: bigint): number {
-  return Number(expireTime) - new Date().getTime()
+export function getExpirationSeconds (expireTime: bigint): number {
```
achingbrain, line 118: The comment still says it returns the number of ms but the function has been renamed to say it returns seconds.

src/circuit/utils.ts
```diff
@@ -115,6 +115,7 @@ export async function namespaceToCid (namespace: string): Promise<CID> {
...
+  const currentTimeSeconds = ((new Date().getTime()) / 1000) | 0
+  return Number(expireTime) - currentTimeSeconds
```
achingbrain, line 120: Why is the downcast to number necessary?

src/circuit/utils.ts
```diff
@@ -115,6 +115,10 @@ export async function namespaceToCid (namespace: string): Promise<CID> {
...
  * returns number of ms between now and expiration time
  */
-export function getExpiration (expireTime: bigint): number {
-  return Number(expireTime) - new Date().getTime()
+export function getExpirationMilliseconds (expireTimeSeconds: bigint): number {
+  const expireTimeMillis = expireTimeSeconds * BigInt(1000)
+  const currentTime = new Date().getTime()
+  // The difference between expireTime and currentTime is much more
+  // likely to fit in a Number type.
+  return Number(expireTimeMillis - BigInt(currentTime))
```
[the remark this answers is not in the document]
ckousik: We would have to downcast anyway to pass a duration to the setTimeout function.

ckousik:
@achingbrain I've bounded the timeout duration between 30 seconds and 15 minutes. The values are arbitrary for now.

src/circuit/transport/reservation-store.ts
```diff
@@ -23,6 +23,11 @@ const REFRESH_WINDOW = (60 * 1000) * 10
 // try to refresh relay reservations 5 minutes before expiry
 const REFRESH_TIMEOUT = (60 * 1000) * 5
 
+// maximum duration before which a reservation should be refereshed
```
achingbrain, line 26:
```suggestion
// maximum duration before which a reservation should be refreshed
```

src/circuit/transport/reservation-store.ts
```diff
@@ -162,13 +167,18 @@ export class ReservationStore extends EventEmitter<ReservationStoreEvents> imple
 
         log('created reservation on relay peer %p', peerId)
 
-        const expiration = getExpiration(reservation.expire)
+        const expiration = getExpirationMilliseconds(reservation.expire)
+
+        // sets a lower bound on the timeout
+        const timeoutDuration = Math.max(expiration - REFRESH_TIMEOUT, REFRESH_TIMEOUT_MIN)
+        // sets an upper bound on the timeout
+        const boundedTimeoutDuration = Math.min(timeoutDuration, REFRESH_TIMEOUT_MAX)
```
achingbrain, line 175: If I'm reading this correctly `timeoutDuration` will either be `expiration - REFRESH_TIMEOUT` or `REFRESH_TIMEOUT_MIN`, (e.g. 30 seconds) whichever is greater.

Why do we need to then take the smaller value of `timeoutDuration` or `REFRESH_TIMEOUT_MAX ` (e.g. 15m)?

I think this means in the case where we get a reservation for an hour, we'll ignore the expiry time and refresh it after 15 minutes?

Can we not just use `timeoutDuration` having set a sensible lower bound?

achingbrain: Having said that, the [go implementation](https://github.com/libp2p/go-libp2p/blob/9e208249b588727cef726a7c2560a9db722404a1/p2p/protocol/circuitv2/client/reservation.go#L89-L91) just errors out if we receive a reservation expiration in the past, maybe that's better.

achingbrain (approved):

PR #1634, opened 2023-03-21
docs(getting_started): add debugging info

SgtPooki:
provide instructions for users using node/browser debugging

PR #1638, opened 2023-03-21
docs: update getting started table of contents

achingbrain:
This was missed in https://github.com/libp2p/js-libp2p/pull/1634


p-shahi (approved):
👍

PR #1647, opened 2023-03-23
fix(content-routing:utils): improve requirePeers error msg

SgtPooki:
This error message of "not found" is completely unhelpful when developing. I imagine providing more context would save devs some time

### DIFF
--- a/src/content-routing/utils.ts
+++ b/src/content-routing/utils.ts
@@ -49,6 +49,6 @@ export async function * requirePeers (source: Source<PeerInfo>, min: number = 1)
   }
 
   if (seen < min) {
- throw new CodeError(`more peers required, seen: ${seen}  min: ${min}`, 'NOT_FOUND')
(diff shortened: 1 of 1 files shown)


src/content-routing/utils.ts
```diff
@@ -49,6 +49,6 @@ export async function * requirePeers (source: Source<PeerInfo>, min: number = 1)
   }
 
   if (seen < min) {
-    throw new CodeError('not found', 'NOT_FOUND')
+    throw new CodeError(`requirePeers failed with ${seen} of ${min} required peers.`, 'NOT_FOUND')
```
maschad, line 52:
```suggestion
 throw new CodeError(`more peers required, seen:${seen}  min: ${min} .`, 'NOT_FOUND')
 ````

maschad (changes requested):
Thanks for this @SgtPooki ! I do wonder if `NOT_FOUND` is more appropriate than `ERROR` as peers have been found, but they are below the threshold.

achingbrain (approved):

maschad (approved):

PR #1651, opened 2023-03-27
fix: add logging around filtered addresses

achingbrain:
To enable debugging errors like 'the dial request had no valid addresses', log the before/after state of addresses filtered by supported transport.

### DIFF
--- a/src/connection-manager/dialer/index.ts
+++ b/src/connection-manager/dialer/index.ts
@@ -260,21 +260,28 @@ export class DefaultDialer implements Startable, Dialer {
       addrs.map(async (ma) => await this._resolve(ma, options))
     ))
       .flat()
-      // Multiaddrs not supported by the available transports will be filtered out.
-      .filter(ma => Boolean(this.components.transportManager.transportForMultiaddr(ma)))
+
+    // Multiaddrs not supported by the available transports will be filtered out.
+    addrs = addrs.filter(ma => Boolean(this.components.transportManager.transportForMultiaddr(ma)))
 
     // deduplicate addresses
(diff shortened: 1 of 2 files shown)


src/connection-manager/dialer/index.ts
```diff
@@ -260,21 +260,28 @@ export class DefaultDialer implements Startable, Dialer {
...
       .flat()
-      // Multiaddrs not supported by the available transports will be filtered out.
-      .filter(ma => Boolean(this.components.transportManager.transportForMultiaddr(ma)))
+
+    // Multiaddrs not supported by the available transports will be filtered out.
+    addrs = addrs.filter(ma => Boolean(this.components.transportManager.transportForMultiaddr(ma)))
 
     // deduplicate addresses
-    addrs = [...new Set(addrs.map(ma => ma.toString()))].map(ma => multiaddr(ma))
+    let filteredAddrs = [...new Set(addrs.map(ma => ma.toString()))].map(ma => multiaddr(ma))
```
maschad, line 268: This is a little bit difficult to read, could we have something like:

```typescript
const filteredAddrs = addrs.filter(ma => Boolean(this.components.transportManager.transportForMultiaddr(ma)))

const dedupedAddrs = [...new Set(filteredAddrs.map(ma => ma.toString()))]

const dedupedMultiaddrs = dedupedAddrs.map(ma => multiaddr(ma))
```

maschad (commented):
just some recommended code clean up to make this a bit easier to understand.

src/connection-manager/dialer/index.ts
achingbrain: Done!

Issue #1574, opened 2023-02-01
libp2p events


achingbrain:
The `libp2p` object returned from [createLibp2p](https://github.com/libp2p/js-libp2p/blob/master/src/index.ts#L200) is an [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) that emits [a few different events](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-libp2p/src/index.ts#L36-L78).

These events are really useful for observing the inner workings of libp2p without having to [tie your code to the internal structure of libp2p](https://github.com/libp2p/js-libp2p/issues/1567).

They also open the door to real-time monitoring of libp2p with better observability and debugging possibilities.

The event list should be expanded to encompass more events.  Some (non-exhausive) suggestions:

* `peer:dial` - when we dial a peer, the event should hold the peer id and which multiaddrs are being dialled
* `peer:dial:success` - the event should hold the peer id and the multiaddr that succeeded
* `peer:dial:error` - the event should hold the peer id and the multiaddr that failed and an `Error` object
* `identify:success` - the identify protocol competed for a peer - the event should hold the peer id
* `identify:error` - the identify protocol failed for a peer - the event should hold the peer id and an `Error` object
* `identify:push:success` - the identify push protocol competed for a peer - the event should hold the peer id
* `identify:push:error` - the identify push protocol failed for a peer - the event should hold the peer id and an `Error` object
* `connection:prune` - we pruned a connection, the event should have the peer id of the connection that was closed
* `connection:upgrade` - a connection was upgraded, the event should have the connection that was upgraded
* ...other things?

It would be good to establish a naming convention for events, the above hasn't had a whole lot of thought applied to it.

maschad:
Great idea @achingbrain this would be very helpful, I would also add to this:

- `peer:dial:pending` when a dial is pending, the event the event should hold the peer id and which multiaddrs are being dialled
- `connection:${status}` when a peer's [status](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-connection/src/status.ts) has changed, the event should  have the peer id and the multiaddr and the status.

@wemeetagain can you think of other events that might be useful?

achingbrain:
We may not need a `peer:dial:pending` event - it is pending after the `peer:dial` event and before the `peer:dial:success`/`peer:dial:error` events.

wemeetagain:
[...]

PR #1652, opened 2023-03-27
fix: invoke onProgress callback with DHT queries during routing

achingbrain:
Allow passing an `onProgress` callback to the peer/content routers that can receive DHT query events.

Refs #1574


maschad (approved):
ProgessEvents! Woot!

test/content-routing/content-routing.node.ts
```diff
@@ -115,6 +115,39 @@ describe('content-routing', () => {
...
+            multiaddrs: [],
+            protocols: []
+          }]
+        }
+        deferred.resolve()
+      })
+
+      const onProgress = sinon.stub()
+
+      await drain(nodes[0].contentRouting.findProviders(CID.parse('QmU621oD8AhHw6t25vVyfYKmL9VV3PTgc52FngEhTGACFB'), {
```
maschad, line 142: Where did this _CID_ come from?

achingbrain: It's used elsewhere in the test suite 🤷

achingbrain:
Converted to draft because we need to be a bit cleverer about how we expose the `ProgressEvent` types.

If we don't expose them all we get errors:

```js
await name.resolve(key, {
  onProgress(evt) {
    if (evt.type === 'content-routing:get:dht:event') {

    }
  }
})
```

```
This comparison appears to be unintentional because the types '"ipns:resolve:start" |
"ipns:resolve:success" | ..(more events here).. and '"content-routing:get:dht:event"'
have no overlap.
```

The problem is we want to avoid exposing all the ProgressEvent types for all the content routing/peer routing implementations at the content routing/peer routing interface level.  Better to somehow aggregate the types based on the configuration passed to the libp2p factory function in a similar way to #1563

achingbrain:
Closing in favour of #1975

Issue #1665, opened 2023-03-31
bug(connection-manager): checkMaxLimit - unset/invalid limit not caught properly


SgtPooki:
https://github.com/libp2p/js-libp2p/blob/290e6103f2497b2a74a8964b7479636c391df2ba/src/connection-manager/index.ts#L582

results in

```
libp2p:connection-manager:trace checking limit of maxConnections. current value: 1 of NaN +0ms
common.js:113 libp2p:connection-manager:trace checking limit of maxConnections. current value: 1 of NaN +946ms
common.js:113 libp2p:connection-manager:trace checking limit of maxConnections. current value: 2 of NaN +818ms
common.js:113 libp2p:connection-manager:trace checking limit of maxConnections. current value: 2 of NaN +637ms
common.js:113 libp2p:connection-manager:trace checking limit of maxConnections. current value: 1 of NaN +1m
common.js:113 libp2p:connection-manager:trace checking limit of maxConnections. current value: 1 of NaN +1s
```

Note that in javascript... `NaN == null // returns false`

PR #1666, opened 2023-03-31
fix:  ensure to check for Infinity in checkMaxLimit function

SgtPooki:
fixes https://github.com/libp2p/js-libp2p/issues/1665

### DIFF
--- a/src/connection-manager/index.ts
+++ b/src/connection-manager/index.ts
@@ -584,8 +584,8 @@ export class DefaultConnectionManager extends EventEmitter<ConnectionManagerEven
       return
     }
     if (limit === Infinity) {
-        log.trace('limit %s is Infinity so it cannot be applied', name);
-        return;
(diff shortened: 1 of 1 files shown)


maschad (approved):
Thanks!

SgtPooki:
note @maschad that we could also change this to `!Number.isSafeNumber(value)` which would catch both NaN and Infinity, but tbh, we should be catching NaN elsewhere. I think the NaN displaying in #1665 is something with `@libp2p/logger` or `debug`

maschad:
Agreed @SgtPooki I think this more explicit handling is more appropriate to see the flow of execution.

achingbrain:
What's the problem this solves?

If the limit is `Infinity` and the check it `if (value > limit) {`, `value` will never be over `limit`.

SgtPooki:
Mostly to tidy up the log.

I saw these errors as a red herring of something else, found out limit was set to infinity but the log confused me.

PR #1668, opened 2023-04-01
feat(auto-dialler): dial in parallel

SgtPooki:
Fixes issues brought up by @achingbrain:

> the libp2p auto dialler (component that ensures we always have some good
> connections) dials remote peers one at a time - one peer that takes ages to
> respond can make this stall, so this needs to become parallelised

### Features:

* Makes auto-dialer dial in parallel
* Prevents dialing to peer.id if peer.id is already being dialed to
* Aborts all dial attempts if autoDialer is stopped 

--- 

Note: With regard to the third item above: I added a parent AbortAbort to the autodialer that is aborted when ’autodialer.stop` is called, which would result in any connection attempts created by the auto-dialer to also be aborted. 

A few changes we could make:

1. Don't cancel pending dials
2. Cancel only pending dials that have been pending for more than X time (add timeout to AbortAbort for the dial attempt)

Other things of note:

* we're not currently passing through any signals (via options) to the auto-dialer. Do we want to?
* If the signal passed to openConnection is aborted, but we have a successful connection, does the connection get aborted or just the dial attempt?


test/connection-manager/auto-dialler.spec.ts
```diff
@@ -107,4 +108,79 @@ describe('Auto-dialler', () => {
...
+    // }
+    // const peer2: Peer = {
+    //   id: await createEd25519PeerId(),
+    //   protocols: [],
+    //   addresses: [{
+    //     multiaddr: multiaddr('/ip4/127.0.0.1/tcp/4001'),
+    //     isCertified: true
+    //   }],
+    //   metadata: new Map()
+    // }
```
SgtPooki, line 131: will cleanup

SgtPooki (commented):
self review. update coming

SgtPooki:
test failure restarted: https://github.com/libp2p/js-libp2p/actions/runs/4585538848/jobs/8097751623?pr=1668

src/connection-manager/auto-dialler.ts
```diff
@@ -48,6 +52,8 @@ export class AutoDialler implements Startable {
     this.options = mergeOptions.call({ ignoreUndefined: true }, defaultOptions, init)
     this.running = false
     this._autoDial = this._autoDial.bind(this)
+    this.abortController = new AbortAbort()
```
achingbrain, line 55: Why not just use a regular `AbortController` here?

Create it in `.start`, pass it's `.signal` to every dial attempt in `._autoDial` and `.abort()` it in `.stop`.

src/connection-manager/auto-dialler.ts
```diff
@@ -37,6 +39,8 @@ export class AutoDialler implements Startable {
   private readonly options: Required<AutoDiallerInit>
   private running: boolean
   private autoDialTimeout?: ReturnType<retimer>
+  private readonly abortController?: AbortAbort
+  connectionMap: Map<string, Promise<Connection>>
```
achingbrain, line 43: You can use [PeerMap](https://libp2p.github.io/js-libp2p-peer-collections/classes/PeerMap.html) from `@libp2p/peer-collections` - no need to stringify the peer ids then.

src/connection-manager/auto-dialler.ts
```diff
@@ -137,14 +150,29 @@ export class AutoDialler implements Startable {
...
+        this.abortController?.addDependant(connectionAbortController)
+
+        /**
+         * we do not need to await the connection attempts because we are
+         * blocking new calls to `openConnection` by checking whether the
+         * peer already has an attempt in the `connectionMap` and we are
+         * removing the peer from the `connectionMap` when the attempt
+         * finishes (successful or not)
+         */
+        const connectionAttempt = this.components.connectionManager.openConnection(peer.id, { signal: connectionAbortController.signal })
```
achingbrain, line 167: `peers` contains every peer in the peer store - I think as written this will dial every peer ever encountered when the number of connections drops below the threshold, potentially swamping the dial queue.

This needs to have a configurable bound by putting the `connectionManager .openConnection` calls in `p-queue` items or something and exposing the `concurrency` option as config for the connection manager.

src/connection-manager/auto-dialler.ts
```diff
@@ -113,7 +121,12 @@ export class AutoDialler implements Startable {
       return true
     })
 
-    // shuffle the peers
+    /**
+     * shuffle the peers
+     *
+     * todo: Be smarter about shuffling peers. We should add weights to peers
+     * that are more likely to succeed
```
achingbrain, line 128: Not one for this PR but we may wish to sort peers by tag value so we try to connect to 'high value' peers first.  

Those that are KAD-close, those in GossipSub meshes, bootstrappers, etc.

achingbrain (changes requested):

SgtPooki:
@achingbrain something I wanted to call out about testing this over the weekend:

works fine with small peer counts.. with large peer counts, it falls over due to lack of exponential backoff and inability to reach a large number of peers. There is a lot of wasted effort due to 

* protocols that are obviously not supported. If we fail to utilize dns6/ip6/etc for any/all peers, but connect successfully on others, we need to filter those out - e.g. ip6 on my machine. 
* peers that are not dialable - should be bumped down in priority for connection attempts.
* peers that are dialable are not prioritized - past successful connections to peers means we should probably try them first.
* multiaddrs for remote peers advertising private IP addresses - not really sure how we would handle this up front, but if we cannot connect to a peer listing a private/local IP, we should probably ignore that multiaddr at some point. 
* large number of webtransport requests don't seem to be limited "inherently" the same as other browser requests - I've seen the dialer kick off 150 requests at the same time, and the browser seems to ignore them all. - we should probably have a config that accepts per protocol/transport/whatever limit instead of only global limits like we have now. 
* too many requests in parallel flooding the browser - we should probably be optimizing requests via batching. i thought that simply setting a config in connection-manager to limit the number of concurrent calls would limit the number of auto-dialer calls, but we can see in the code that auto-dialer is not reading and applying that config

src/connection-manager/auto-dialler.ts
SgtPooki: and those that we've successfully connected to in the past.

src/connection-manager/auto-dialler.ts
SgtPooki: > peers contains every peer in the peer store - I think as written this will dial every peer ever encountered when the number of connections drops below the threshold, potentially swamping the dial queue.

That makes sense, thanks for the context. see https://github.com/libp2p/js-libp2p/pull/1668#issuecomment-1494741602

SgtPooki: > This needs to have a configurable bound by putting the connectionManager.openConnection calls in p-queue items or something and exposing the concurrency option as config for the connection manager.

I imagine we would want to do that in this PR. I will look into it

achingbrain:
Take a look at https://github.com/libp2p/js-libp2p/pull/1678 - it addresses most of the above points.

SgtPooki:
Closing because https://github.com/libp2p/js-libp2p/pull/1678 did a much better job of improving things =)

Issue #1587, opened 2023-02-15
Use of Promise.any() can cause unhandled promise rejection 


inverted-capital:
- **Version**: 0.42.2

- **Platform**: Linux plain 6.1.10-100.fc36.x86_64 

- **Subsystem**:  Dialer

#### Description:
On this line: https://github.com/libp2p/js-libp2p/blob/de73b6a22c7580d2bba5ca97f86465c8f9ed0e06/src/connection-manager/dialer/dial-request.ts#L85 there is a call to `Promise.any()` however inside the promises that are supplied to that call, there appears to be insufficient error handling.  Once the call to any has resolved, any of the other promises may still reject and will cause a node system error.  This issue has been raised before: https://github.com/nodejs/node/issues/43639

This is the error I get back when I run [this test](https://github.com/dreamcatcher-tech/dreamcatcher-stack/blob/ea49390c62ceb2b5083f97c47a7202fc75c2bc49/pkg/interblock/src/w310-interpulse/test/mount.test.js#L75):
```
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[AggregateError: All promises were rejected] {
  code: 'ERR_TIMEOUT',
  [errors]: [
    AbortError: The operation was aborted
        at EventTarget.onAbort (/home/evil_liskov/repos/dreamcatcher-stack/node_modules/@libp2p/websockets/src/index.ts:87:16)
        at EventTarget.[nodejs.internal.kHybridDispatch] (node:internal/event_target:735:20)
        at EventTarget.dispatchEvent (node:internal/event_target:677:26)
        at abortSignal (node:internal/abort_controller:308:10)
        at AbortController.abort (node:internal/abort_controller:338:5)
        at EventTarget.onAbort (/home/evil_liskov/repos/dreamcatcher-stack/node_modules/any-signal/index.js:11:16)
        at EventTarget.[nodejs.internal.kHybridDispatch] (node:internal/event_target:735:20)
        at EventTarget.dispatchEvent (node:internal/event_target:677:26)
        at abortSignal (node:internal/abort_controller:308:10)
        at AbortController.abort (node:internal/abort_controller:338:5)
        at EventTarget.onAbort (/home/evil_liskov/repos/dreamcatcher-stack/node_modules/any-signal/index.js:11:16)
        at EventTarget.[nodejs.internal.kHybridDispatch] (node:internal/event_target:735:20)
[...]

Issue #1630, opened 2023-03-16
bug: heliaInstance.libp2p.addEventListener('peer:discovery', cb) calls cb twice for same peer


SgtPooki:
Using libp2p 0.42.2 via https://cdn.jsdelivr.net/npm/libp2p@0.42.2/dist/index.min.js

Peer discovery should only emit once, unless we are in fact discovering the peer twice via different methods somehow.

Either way, once a peer is discovered, we should cancel any in-progress operations and emit that we discovered the peer only once.

While working on a new helia-script-tag example (via #43), I was able to get the following event log:

```
1263ms - Created LibP2P instance
1263ms - Created Helia instance
2248ms - Discovered peer QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN
2252ms - Discovered peer QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN
2252ms - Discovered peer QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa
2252ms - Discovered peer QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa
2253ms - Discovered peer QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb
2254ms - Discovered peer QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb
2254ms - Discovered peer QmcZf59bWwK5XFi76CZX8cbJ4BhTzzA3gU1ZjYZcYW3dwt
2254ms - Discovered peer QmcZf59bWwK5XFi76CZX8cbJ4BhTzzA3gU1ZjYZcYW3dwt
74326ms - Connected to QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb
77805ms - Connected to QmcZf59bWwK5XFi76CZX8cbJ4BhTzzA3gU1ZjYZcYW3dwt
135298ms - Disconnected from QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb
138782ms - Disconnected from QmcZf59bWwK5XFi76CZX8cbJ4BhTzzA3gU1ZjYZcYW3dwt
```

SgtPooki:
[...]

Issue #1510, opened 2022-12-04
Use standard IP cidr format for allow/deny lists


achingbrain:
Currently the connection manager treats multiaddrs in the allow/deny lists as string prefixes (e.g. `"/ip4/52.55"`), it should use the standard IP cidr format instead (e.g. `"/ip4/52.55.0.0/ipcidr/16"`) which will allow us to validate these as multiaddrs.

This will likely involve changes to the `@multiformats/multiaddr` module to allow using address ranges.

mpetrunic:
> This will likely involve changes to the @multiformats/multiaddr module to allow using address ranges.

@achingbrain Doesn't multiaddr already have `ipcidr` for this purpose?

Seems like we are missing https://github.com/multiformats/go-multiaddr/blob/master/protocols.go#L112 in js?

Should we add cidr based filtering implementation in js-multiaddr (like go) or in js-libp2p-utils?

SgtPooki:
do we still need this @achingbrain ?

achingbrain:
Yes

SgtPooki:
<details>
<summary>JS Colo 2024 rough discussion</summary>
It looks like we already have the code in js-multiaddr, https://github.com/multiformats/js-multiaddr/blob/d4164fbce730d3b43a10682b651bcc96db845e7c/src/convert.ts#L120C17-L135, so we don't need to change it

Should only need changes in js-libp2p 

`convertToIpNet` defined at https://github.com/multiformats/js-multiaddr/blob/d4164fbce730d3b43a10682b651bcc96db845e7c/src/convert.ts#L120C17-L135
</details>

### Action items:

1. Convert `.allow` and `.deny` properties of connection-manager to be `IpNet[]`.
    * `convertToIpNet` at https://github.com/libp2p/js-libp2p/blob/4fd7eb2e14c2ac30150060adb2c8aca4c5e295f3/packages/libp2p/src/connection-manager/index.ts#L200-L201 
1. replace `startsWith` in https://github.com/libp2p/js-libp2p/blob/6ccbb06f0e76dcd1ba20d31e2518f3a1acb0efbc/packages/libp2p/src/connection-manager/connection-pruner.ts#L101-L104 with `Instance<IpNet>.contains(ip)`
[...]

PR #1678, opened 2023-04-06
fix!: refactor connection manager to use a prioritised queue

### LINKED ISSUE #1565: Should not emit peer:connect with a closed connection
- **Version**:
0.41.0

- **Platform**:

`Linux 5.4.0-91-generic #102-Ubuntu SMP Fri Nov 5 16:31:28 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

- **Subsystem**:
connection-manager

#### Severity:
Low

#### Description:
`js-libp2p` may emit `peer:connect` with a closed connection if the new connection is selected to prune https://github.com/libp2p/js-libp2p/blob/40841637ab36c3163e36afce4050a07ea2f614af/src/connection-manager/index.ts#L433

In my opinion, it's not expected to emit a `peer:connect` with a closed connection but if it is, we need to document that so that downstream applications can handle this accordingly

Run lodestar v1.4.0-rc.0

achingbrain:
Refactors the connection manager.

1. Internally it uses a queue to control concurrency instead of dial tokens (e.g. `maxParallelDials`)
2. A second queue is used for each peer to prevent inividual peers from swamping the dial queue (e.g. `maxParallelDialsPerPeer`)
3. The auto dialler now also checks the minimum connection limit on peer discovery and peer disconnect events for fast reconnect
4. Auto dialled peers are sorted based on tag value so valuable peers should be re-dialled first
5. Auto dialled peers are dialled in parallel to prevent a slow peer locking up the auto-dial queue
6. A `getDialQueue` method has been added to libp2p to allow inspection of the dial queue though it needs exposing in the interface
7. The connection gater `denyDialMultiaddr` method now only takes a multiaddr
8. Fixes duplicate connections to fast peers
9. Fixes unhanded promise rejections #1587
10. Connection pruning now happens after `peer:connect` events not before - fixes #1565

BREAKING CHANGE: some connection manager options have changed - please see the upgrade guide for full details


doc/migrations/v0.43-v0.44.md
```diff
@@ -0,0 +1,117 @@
...
+
+    // these should now be ranges expressed as MultiaddrFilters
+    allow: [
+      '/ip4/0.0.0.0/tcp/123'
+    ],
+
+    // these should now be ranges expressed as MultiaddrFilters
+    deny: [
+      '/ip4/0.0.0.0/tcp/123'
+    ]
```
SgtPooki, line 40: Can we get a "allow peer" that allows any appropriate transports given a specific peerId?

I was trying to do that the other day in kubo and realized i needed a full multiaddr. I was trying to grant full-access to/from my helia node and couldn't figure out how to do it within an hour or so and i gave up.

It would be great to be able to do that in js-libp2p

doc/migrations/v0.43-v0.44.md
```diff
@@ -0,0 +1,117 @@
...
+  }
+})
+```
+
+**Before**
```
SgtPooki, line 45:
```suggestion
**After**
```

doc/migrations/v0.43-v0.44.md
```diff
@@ -0,0 +1,117 @@
...
+
+const node = await createLibp2p({
+  connectionManager: {
+    // how many peers to dial at once while trying to ensure the node
+    // is above minConnections
+    autoDialConcurrency: 25,
+
+    // a low value allows user-initiated dials to take priority over
+    // auto dials
+    autoDialPriority: 0,
```
SgtPooki, line 59: Nice

doc/migrations/v0.43-v0.44.md
```diff
@@ -0,0 +1,117 @@
...
+
+    // this was previously named `maxAddrsToDial`
+    maxPeerAddrsToDial: 20,
```
SgtPooki, line 62: PerConnection attempt? Or total(then we banlist the peer as bad??)

Might be good to call out which one and how in jsdoc, or in the name if possible.

doc/migrations/v0.43-v0.44.md
```diff
@@ -0,0 +1,117 @@
...
+
+    // this was previously named `maxDialsPerPeer`
+    maxParallelDialsPerPeer: 20,
```
SgtPooki, line 65: Just so i understand fully, this setting changes whether we can attempt to dial 2 or 20 multiaddrs at once when attempting to connect to a specific peer, correct?

Would 'maxParallelMultiAddrDialsPerPeer' be a more explicit name? Is there a way we can get that explicitness without the verbosity?

doc/migrations/v0.43-v0.44.md
```diff
@@ -0,0 +1,117 @@
...
+
+    // these should now be ranges expressed as MultiaddrFilters
+    allow: [
+      new MultiaddrFilter('/ip4/192.168.1.1/ipcidr/32')
+    ],
+
```
SgtPooki, line 71: As above, can i create a multiaddrFilter consisting only of '/p2p/peerId' ?

doc/migrations/v0.43-v0.44.md
```diff
@@ -0,0 +1,117 @@
...
+```js
+import { createLibp2p } from 'libp2p'
+
+const node = await createLibp2p({
+  connectionGater: {
+    denyDialMultiaddr: (multiaddr) => {
+      if (multiaddr.getPeerId() != null) {
+        // there is a peer id present in the multiaddr
+      }
```
SgtPooki, line 111: 👍

src/connection-manager/auto-dial.ts
```diff
@@ -0,0 +1,148 @@
...
+import { AUTO_DIAL_CONCURRENCY, AUTO_DIAL_PRIORITY, MIN_CONNECTIONS } from './constants.js'
+
+const log = logger('libp2p:connection-manager:auto-dial')
+
+export interface AutoDialInit {
+  /**
+   * The minimum number of connections below which libp2p will start to dial
+   * peers from the peer book. (default: 0)
+   */
+  minConnections?: number
```
SgtPooki, line 15: MinConnections: 0 is the same as saying, I'm allowed to have zero connections, don't autoDial, correct?

Might be useful to call out that setting this to zero disables it

src/connection-manager/auto-dial.ts
```diff
@@ -0,0 +1,148 @@
...
+  private readonly connectionManager: ConnectionManager
+  private readonly peerStore: PeerStore
+  private readonly queue: PQueue
+  private readonly minConnections: number
+  private readonly autoDialPriority: number
+
+  /**
+   * Proactively tries to connect to known peers stored in the PeerStore.
+   * It will keep the number of connections below the upper limit and sort
+   * the peers to connect based on wether we know their keys and protocols.
```
SgtPooki, line 52:
```suggestion
   * the peers to connect based on whether we know their keys and protocols.
```

SgtPooki (commented):
Partial review

doc/migrations/v0.43-v0.44.md
achingbrain: No, the allow/deny lists are for allowing/denying classes of network to make connections - this happens before the peer id exchange so we don't know who the remote peer is at that point.

It's a DoS mitigation strategy to always allow connections from networks you know, control or trust, or disallow connections from networks that are attacking you.

To deny specific peers you can implement a [connection gater](https://github.com/libp2p/js-libp2p/blob/master/doc/CONFIGURATION.md#configuring-connection-gater).

doc/migrations/v0.43-v0.44.md
achingbrain: I'm a little confused, access is allowed by default, it's only disallowed if a configured connection gater denys it or you're at `maxConnections` already.

The `allow` list above enables specific networks to breach connection limits - a DoS mitigation for when your node is being eclipsed.

Maybe you can open an issue on the helia repo with a repro case?

doc/migrations/v0.43-v0.44.md
achingbrain: We only dial multiaddrs so maybe having `MultiAddr` in the name is redundant?

doc/migrations/v0.43-v0.44.md
achingbrain: Did you check [the jsdoc for this config option](https://github.com/libp2p/js-libp2p/blob/fix/refactor-connection-manager/src/connection-manager/index.ts#L79-L83)?  If it's not clear could you suggest an edit?

src/connection-manager/auto-dial.ts
achingbrain: Turns out that default was wrong 🤭  - would much rather generate that info but not sure how.

I've updated [the jsdoc for this setting](https://github.com/libp2p/js-libp2p/blob/fix/refactor-connection-manager/src/connection-manager/index.ts#L39-L44) - if it's still not clear could you suggest an edit?

doc/migrations/v0.43-v0.44.md
achingbrain: But yes, a setting of `2` with a peer that has `10` addresses means we'll dial all 10 addresses but only ever 2 at once.  This stops a peer with a lot of spam addresses blocking the dial queue.

src/connection-manager/dial-queue.ts
```diff
@@ -0,0 +1,498 @@
...
+        }))
+      })
+    ))
+      .flat()
+
+    // filter out any multiaddrs that we do not have transports for
+    const filteredAddrs = resolvedAddresses.filter(addr => Boolean(this.transportManager.transportForMultiaddr(addr.multiaddr)))
+
+    // deduplicate addresses
+    const dedupedAddrs: Map<string, Address> = new Map()
```
maschad, line 305: Couldn't we do something like 

```suggestion
let dedupedMultiaddrs = [...new Set(filteredAddrs.map(addr => { 
	addr.isCertified = addr.isCertified || addr.isCertified || false
})).values()]
```

to avoid the null checks per item

src/connection-manager/dial-queue.ts
```diff
@@ -0,0 +1,498 @@
...
+    const dialAbortControllers: Array<(AbortController | undefined)> = pendingDial.multiaddrs.map(() => new AbortController())
+
+    try {
+      // internal peer dial queue to ensure we only dial the configured number of addresses
+      // per peer at the same time to prevent one peer with a lot of addresses swamping
+      // the dial queue
+      const peerDialQueue = new PQueue({
+        concurrency: this.maxParallelDialsPerPeer
+      })
+      peerDialQueue.on('error', (err) => {
```
maschad, line 383: Should we abort dials here?

src/connection-manager/index.ts
```diff
@@ -73,137 +90,175 @@ export interface ConnectionManagerConfig {
...
 
-    log('options: %o', this.opts)
-
-    this.components = components
-
     /**
      * Map of connections per peer
      */
-    this.connections = new Map()
+    this.connections = new PeerMap()
```
maschad, line 183: Didn't know this existed! Nice!

maschad (changes requested):
Great work on this @achingbrain will definitely improve with connection management overall. I made some comments and suggestions. 

I think these are a lot of changes though and perhaps we should split this into some smaller PRs to make reviewing a bit easier and the changes more manageable. I would suggest separate PRs for:
- https://github.com/libp2p/js-libp2p/issues/1510
- https://github.com/libp2p/js-libp2p/issues/1630
- `denyDialMultiaddr` method now only takes a `multiaddr` and also the other interfaces updates required with it.

src/connection-manager/auto-dial.ts
```diff
@@ -0,0 +1,140 @@
...
+      // Remove peers without addresses
+      if (peer.addresses.length === 0) {
+        return false
+      }
+
+      return true
+    })
+
+    // shuffle the peers
+    peers = peers.sort(() => Math.random() > 0.5 ? 1 : -1)
```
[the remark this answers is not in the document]
achingbrain: So we don't dial the same peers every time.

Sorting is stable so if we start with a random order, then sort by a given field, where a subset of the peers have the same value for the given field they'll still be randomly sorted within that subset.

src/connection-manager/dial-queue.ts
achingbrain: `Address`es are not `Multiaddr`s - they have an `isCertified` field so we can't just stringify them into a set to dedupe, instead we use the keys of a map to dedupe but using the `Address` as map values lets us preserve the `isCertified` field.

src/connection-manager/dial-queue.ts
achingbrain: No, because other dials in the queue may succeed - the error event may be emitted if an individual dial fails.

doc/migrations/v0.43-v0.44.md
```diff
@@ -0,0 +1,117 @@
...
+  connectionManager: {
+    // this option has been removed as auto dial now runs in response to peers
+    // disconnecting and is no longer on a timer
+    autoDialInterval: 0,
+
+    // this has been renamed to `maxPeerAddrsToDial`
+    maxAddrsToDial: 20,
+
+    // this has been renamed to `maxParallelDialsPerPeer`
+    maxDialsPerPeer: 20,
```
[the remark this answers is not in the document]
achingbrain: Done!

achingbrain:
I've removed the attempt at resolving the double-peer discovery bug - I think it needs a bit more thought, basically the logic to determine whether to emit the event is complicated and will be vastly simplified by having an [atomic peer store](https://github.com/libp2p/js-libp2p-interfaces/pull/368).

The auto dial interval has also returned - this is because I found an edge case where we could still end up with no peers if all auto-dials failed - if network connection is lost, for example.

I've reverted the `MultiaddrFilter` change for allow/deny lists too - with a bit more thought it would be better to have the connection gater control allow/deny behaviour, then we aren't splitting responsibility for that between the connection manager and the connection gater.

doc/CONFIGURATION.md
```diff
@@ -600,14 +600,7 @@ const node = await createLibp2p({
   connectionEncryption: [noise()],
   connectionManager: {
     maxConnections: Infinity,
-    minConnections: 0,
-    pollInterval: 2000,
```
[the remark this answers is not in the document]
achingbrain: Done

doc/CONFIGURATION.md
```diff
@@ -600,14 +600,7 @@ const node = await createLibp2p({
...
-    // The below values will only be taken into account when Metrics are enabled
-    maxData: Infinity,
```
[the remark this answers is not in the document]
achingbrain: Done

maschad (approved):
LGTM

achingbrain:
Please hold off on merging until https://github.com/jacobheun/any-signal/pull/25 is merged

doc/migrations/v0.43-v0.44.md
SgtPooki: sure. I will re-open if I run into it.. essentially what I was trying to do was prefer my local node to others, and to allow connections to that one no matter what. it sounds like that may not be necessary

src/connection-manager/auto-dial.ts
SgtPooki: https://github.com/libp2p/js-libp2p/blob/4078082dac944c6feb64e9e45e5275d7f0bf6ae4/src/connection-manager/index.ts#L39-L44 looks much better to me, thanks :)

PR #1680, opened 2023-04-12
fix: encapsulate /p2p-circuit multiaddrs when dialing a known peerid

joeltg:
The logic for creating a dial target tries to ensure multiaddrs are encapsulated with `/p2p/${peerId}` IF the dial target's peer id is known. It uses `addr.getPeerId() != null` to check if this is already the case.

```
> multiaddr("/ip4/127.0.0.1/tcp/4001/p2p/12Dfoo...").getPeerId()
'12Dfoo...'
```

However, circuit relay addresses return the **relay server's** peer id if they don't end in `/p2p/${destinationPeerId}`.

```
> multiaddr("/ip4/127.0.0.1/tcp/4001/p2p/12D3bar.../p2p-circuit").getPeerId()
'12D3bar...'
```

Since the dialer then thinks the multiaddr already has a peer id, it doesn't encapsulate with `/p2p/${destinationPeerId}`, which ultimately results in failed dials to circuit relay destinations (`Circuit relay dial failed as addresses did not have peer id`). If I understand things right, the desired behavior is for circuit relay addresses to get encapsulated with the destination peer id.


joeltg:
Just want to bump this since I'm now confident this is **breaking all dials to circuit-relay addresses**, at least on our network.

achingbrain (approved):

achingbrain:
Thanks!  This fix is be available via the `next` tag in a few minutes - could you try it out and let me know if it fixes the problem?  I'll do a release if so.

There's a short [upgrade guide](https://github.com/libp2p/js-libp2p/blob/master/doc/migrations/v0.43-v0.44.md) to follow but it should be relatively painless unless you are doing a lot of custom stuff with the connection manager.

```console
$ npm i libp2p@next
```

joeltg:
That works! Thank you so much!

achingbrain:
This has been released in `libp2p@0.44.0`

PR #1691, opened 2023-04-16
fix: do not autodial connected peers or those in the dial queue

achingbrain:
If we already have a connection to a peer, or the peer is already in the dial queue, do not select it to autodial.


achingbrain:
The connection manager already applies this optimisation.

PR #1729, opened 2023-05-04
fix: update autonat and upnp-nat services to match fetch etc

achingbrain:
Refactors exports for autonat and upnp-nat to be consistent with other services - fetch, ping, identify, etc.


test/autonat/index.spec.ts
```diff
@@ -36,7 +36,7 @@ const defaultInit: AutonatServiceInit = {
 }
 
 describe('autonat', () => {
-  let service: AutonatService
+  let service: any
```
wemeetagain, line 39: :(

maschad (approved):
LGTM

PR #1733, opened 2023-05-04
fix: remove closed listeners from transport manager

achingbrain:
When a listener closes, remove it from the stored list of listeners.

This wasn't much of a problem before as listeners tended to be very long lived but circuit relay listeners get closed all the time so this fixes a now much more obvious memory leak.

### DIFF
--- a/src/transport-manager.ts
+++ b/src/transport-manager.ts
@@ -158,6 +158,13 @@ export class DefaultTransportManager implements TransportManager, Startable {
   }
 
   /**
+   * Returns all the listener instances
+   */
+  getListeners (): Listener[] {
+    return Array.of(...this.listeners.values()).flat()
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


src/transport-manager.ts
```diff
@@ -216,6 +223,11 @@ export class DefaultTransportManager implements TransportManager, Startable {
             })
         })
         listener.addEventListener('close', () => {
```
maschad, line 225: should we be removing these eventListeners once they are closed as well?

maschad (commented):
Left a comment, LGTM otherwise

src/transport-manager.ts
achingbrain: I don't think so because the listener is created and returned by the transport but not retained.

maschad (approved):
LGTM

PR #1735, opened 2023-05-04
fix!: do not dial private addresses in browsers

achingbrain:
Dialling is very expensive in browsers. If we're dialling addresses that include private networks, chances are they're not going to succeed so filter them out for browsers only.

This can be re-enabled by configuring a permissive connection gater, see the upgrade guide for more information.

BREAKING CHANGE: browsers will no longer try to dial private addresses by default


src/config/connection-gater.browser.ts
```diff
@@ -0,0 +1,31 @@
...
+    denyDialMultiaddr: async (multiaddr: Multiaddr) => {
+      const tuples = multiaddr.stringTuples()
+
+      if (tuples[0][0] === 4 || tuples[0][0] === 41) {
+        return Boolean(isPrivate(`${tuples[0][1]}`))
+      }
+
+      return false
+    },
+    denyInboundConnection: async () => await Promise.resolve(false),
```
wemeetagain, line 22: I don't htink these Promise.resolves are helpful here

achingbrain: Fair, this block was copied from `libp2p.ts` but I think it's been mangled by the linter auto-fix over time.

achingbrain: I've removed them

wemeetagain (approved):

maschad:
Looks like there are quite a few broken tests @achingbrain

maschad (approved):

PR #1740, opened 2023-05-05
fix: do not auto-dial peers in the dial queue

achingbrain:
Similar to #1730, filter auto-dial peers by peers already in the dial queue.

The dial queue will join any pre-existing dial attempts for the selected peers, but it just creates unnecessary work and adds noise to the logs and filtering them out beforehand is cheap so just do that.

### DIFF
--- a/src/connection-manager/auto-dial.ts
+++ b/src/connection-manager/auto-dial.ts
@@ -1,7 +1,7 @@
 import { logger } from '@libp2p/logger'
 import type { PeerStore } from '@libp2p/interface-peer-store'
 import type { ConnectionManager } from '@libp2p/interface-connection-manager'
-import { PeerMap } from '@libp2p/peer-collections'
+import { PeerMap, PeerSet } from '@libp2p/peer-collections'
 import PQueue from 'p-queue'
 import { AUTO_DIAL_CONCURRENCY, AUTO_DIAL_INTERVAL, AUTO_DIAL_PRIORITY, MIN_CONNECTIONS } from './constants.js'
 import type { Startable } from '@libp2p/interfaces/startable'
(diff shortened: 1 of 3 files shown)


src/connection-manager/auto-dial.ts
```diff
@@ -103,6 +103,12 @@ export class AutoDial implements Startable {
 
     const connections = this.connectionManager.getConnectionsMap()
     const numConnections = connections.size
+    const dialQueue = new PeerSet(
+      // @ts-expect-error boolean filter removes falsy peer IDs
+      this.connectionManager.getDialQueue()
```
maschad, line 108: type casting should work here and remove the need for ugly `//@ts-expect-error` s
```suggestion
 const dialQueue = new PeerSet(
      this.connectionManager.getDialQueue()
        .map(queue => queue.peerId)
        .filter(Boolean) as PeerId[]
    )
 ```

File: test/connection-manager/auto-dial.spec.ts [hunk 2]
```diff
@@ -51,6 +52,7 @@ describe('auto-dial', () => {
 
     const connectionManager = stubInterface<ConnectionManager>()
     connectionManager.getConnectionsMap.returns(new PeerMap())
+    connectionManager.getDialQueue.returns([])
```
maschad, line 55: the [getDialQueue mock](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-mocks/src/connection-manager.ts#LL192C1-L193C1) should be sufficient here.

maschad (commented):
Nice! Looks good overall just some recommendations

src/connection-manager/auto-dial.ts
achingbrain: The thing about the ugly `@ts-expect-error` is that should typescript ever be smart enough to work out what `.filter(Boolean)` does, it'll then become an error itself letting you know that the types are correctly inferred now whereas the `as` cast will smother the error forever.

maschad: I see your point but wouldn't a `PeerId[]` type be expected here? I think the type error being returned by `.filter(Boolean)` is legitimate because all values _could_ be falsey, which in that case it would be more appropriate to have an empty array of `PeerId` passed to `new PeerSet` as opposed to `undefined`.

achingbrain: This is a bug in typescript.  The issue reporting the bug is here: https://github.com/microsoft/TypeScript/issues/16655 - it was fixed by https://github.com/microsoft/TypeScript/pull/29955 and then reverted by https://github.com/microsoft/TypeScript/issues/16655#issuecomment-494907057 so remains unfixed.

PR #1741, opened 2023-05-05
feat: emit peer:identify event after identify completes

achingbrain:
To monitor when peers complete the identify protocol, emit a `peer:identify` event with the protocol reply.

This is more reliable than doing an on-demand identify since we can easily breach per-protocol stream limits for identify.


maschad (approved):
Good idea, LGTM

PR #1743, opened 2023-05-05
fix: make address manager responsible for addresses

achingbrain:
Refactors the transport manager to just start and stop transports.

Now the address manager is responsible for writing addresses into the peer store.

Debounces writes to the peer store so we don't get lots of `self:peer:update` events on startup as each transport starts to listen on configured addresses.

### DIFF
--- a/src/address-manager/utils.ts
+++ b/src/address-manager/utils.ts
@@ -1,0 +1,14 @@
+
+export function debounce (func: () => void, wait: number): () => void {
+  let timeout: ReturnType<typeof setTimeout> | undefined
+
+  return function () {
+    const later = function (): void {
+      timeout = undefined
+      func()
+    }
+
+    clearTimeout(timeout)
+    timeout = setTimeout(later, wait)
+  }
+}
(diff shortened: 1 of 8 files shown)


src/address-manager/utils.ts
```diff
@@ -0,0 +1,14 @@
+
+export function debounce (func: () => void, wait: number): () => void {
```
maschad, line 2: This is quite a useful function, I would if we should make it generic for other use cases, such as in DHT queries for example. Something like

```suggestion
function debounce<T extends (...args: any[]) => any>(func: T, delay: number): (...args: Parameters<T>) => void {
  let timeoutId: NodeJS.Timeout;
  return function (...args: Parameters<T>): void {
    clearTimeout(timeoutId);
    timeoutId = setTimeout(() => {
      func(...args);
    }, delay);
  };
}
```

maschad (approved):
LGTM, just made a suggestion

src/address-manager/utils.ts
achingbrain: I think this might cause hard to debug behaviour because only the args passed to the final call will be used, so lets only make this change when we need it.

For example:

```js
const fn = (arg: string) => {
  console.info('got arg:', arg)
}

const debounced = debounce(fn, 100)

debounced('one')
debounced('two')
debounced('three')
// output:
// got arg: three
```

In this case the caller may not know that the function has been debounced so may expect all three invocations to occur and be left scratching their head when they don't.

Issue #1695, opened 2023-04-17
Stream limits and relayed connections


achingbrain:
Need investigation but stream limits may be being applied per-connection over relayed connections - this should be treated as a special case and applied per-relayed peer.

That is, if we have connections to 10 peers on the same relay, we’ll only have on connection so in this case stream limits should be per-peer, not per-connection.

This may be why the stream limits had to be set to high for the universal connectivity demo.

achingbrain:
On further investigation I don't think this is what's happening.

The upgrader is the component that enforces stream limits, and relayed streams are turned into [MultiaddrConnections by the relay transport](https://github.com/libp2p/js-libp2p/blob/master/src/circuit-relay/transport/index.ts#L231-L238) which the upgrader turns into a Connection which has a separate stream count.

joeltg:
@achingbrain any idea what else might be going on here? I'm also getting lots of unexpected `Too many outbound protocol streams for protocol "/libp2p/circuit/relay/0.2.0/hop"` errors in the logs and don't know where to look.

p-shahi:
@joeltg dropping a note here as I saw the same stream outbound/inbound limit error wrt relay hop in an app I was helping build.
The only configurable stream limits related to circuit relay I found were:
```
    relay: circuitRelayServer({
        maxInboundHopStreams: 1000,
        maxOutboundHopStreams: 1000
    }),
```
But those are fore the relay server.
I set here: https://github.com/libp2p/universal-connectivity/pull/53/commits/3fe2dd467539443ebaeb3759569901a849cfa736 but eventually reverted

The underlying issue I saw was this:
```
libp2p:circuit-relay:transport:error Circuit relay dial failed +0ms CodeError: failed to connect via relay with status RESOURCE_LIMIT_EXCEEDED
```
This occurred b/c the rust v2 relay the app used was rejecting connections. Fixed by this: https://github.com/libp2p/universal-connectivity/pull/54

PR #1758, opened 2023-05-16
fix: close circuit-relay streams on connection failure

joeltg:
If a dial to a relayed address fails, e.g. with status `NO_RESERVATION`, the newly created `/libp2p/circuit/relay/0.2.0/hop` stream is **never closed**. This can be particularly bad if the connection manager keeps auto-dialing the address, since the streams just accumulate until they hit the max per-protocol stream limit (64). I believe this is the ultimate issue behind [#1695](https://github.com/libp2p/js-libp2p/issues/1695).

I don't know whether closing or aborting the stream is more correct, or if it's important to catch potential exceptions in `await relayConnection.newStream` call, so this may need a little revision from someone with more context.

Also, I know that you typically want PRs to include tests covering the fix - I gave it a couple tries but couldn't figure out how to set one up for this scenario. Sorry!

### DIFF
--- a/src/circuit-relay/transport/index.ts
+++ b/src/circuit-relay/transport/index.ts
@@ -183,8 +183,11 @@ class CircuitRelayTransport implements Transport {
       disconnectOnFailure = true
     }
 
-    const stream = await relayConnection.newStream([RELAY_V2_HOP_CODEC])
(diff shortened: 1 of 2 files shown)


achingbrain:
Good catch!

Issue #1573, opened 2023-02-01
Configuration validation


achingbrain:
We don't perform much in the way of validation on loaded libp2p config but we really should.

Ideally each module parsing it's config would use something like [superstruct](https://www.npmjs.com/package/superstruct) to validate at creation time and fail fast with good feedback to the user as to how they can fix their broken config.

This should negate the need for PRs like https://github.com/libp2p/js-libp2p/pull/1572 as the config would have been validated in [src/config.ts](https://github.com/libp2p/js-libp2p/blob/master/src/config.ts#L95-L120) already.

## Update 

@achingbrain [explored a few different options](https://github.com/libp2p/js-libp2p/pull/2133#issuecomment-2164986991) and we determined that [zod](https://github.com/colinhacks/zod) is perhaps the best choice. Currently Zod is 8kb minified + zipped but based on [this discussion](https://github.com/colinhacks/zod/issues/2596) there are a few different approaches to making Zod tree-shakeable.

maschad:
Re-opening this as [original PR](https://github.com/libp2p/js-libp2p/pull/1778) which addressed this was reverted and subsequent work has taken place on https://github.com/libp2p/js-libp2p/pull/2133

I've also updated the description

sumanjeet0012:
@maschad I want to work on this issue.
I want to use the latest version of Zod for validation.

PR #1762, opened 2023-05-17
fix: allow services to indicate their dependencies

achingbrain:
Some services depend on others, for example circuit relay v2 and kad-dht don't really work without identify so we need a way to prompt the user to configure these dependencies.

The change is to allow extending the components a service requires with a service map of the dependencies.

The components type in the `Libp2pInit` type is extended by the `ServiceMap` generic type so when the user configures a node without the required dependencies, TypeScript will fail to compile as the components being passed to the factory function don't have type overlap with the required components.

### DIFF
--- a/test/circuit-relay/discovery.node.ts
+++ b/test/circuit-relay/discovery.node.ts
@@ -5,6 +5,7 @@ import { tcp } from '@libp2p/tcp'
 import { expect } from 'aegir/chai'
 import { pEvent } from 'p-event'
 import { circuitRelayServer, type CircuitRelayService, circuitRelayTransport } from '../../src/circuit-relay/index.js'
+import { identifyService } from '../../src/identify/index.js'
 import { createLibp2p } from '../../src/index.js'
 import { plaintext } from '../../src/insecure/index.js'
 import { getRelayAddress, hasRelay, MockContentRouting, mockContentRouting } from './utils.js'
@@ -17,6 +18,7 @@ describe('circuit-relay discovery', () => {
 
   beforeEach(async () => {
     // create relay first so it has time to advertise itself via content routing
+    // @ts-expect-error TODO: typescript cannot infer the service map type
     relay = await createLibp2p({
       addresses: {
         listen: ['/ip4/127.0.0.1/tcp/0']
@@ -34,10 +36,10 @@ describe('circuit-relay discovery', () => {
         mockContentRouting()
       ],
       services: {
+        identify: identifyService(),
+        // @ts-expect-error TODO: typescript cannot infer the service map type
         relay: circuitRelayServer({
-          advertise: {
-            bootDelay: 10
-          }
+          advertise: true
         })
       }
     })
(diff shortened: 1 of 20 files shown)


test/circuit-relay/discovery.node.ts
```diff
@@ -17,6 +18,7 @@ describe('circuit-relay discovery', () => {
 
   beforeEach(async () => {
     // create relay first so it has time to advertise itself via content routing
+    // @ts-expect-error TODO: typescript cannot infer the service map type
```
achingbrain, line 21: I am not sure why TypeScript can't infer this, it seems to work elsewhere.

joeltg:
Super cool!

This does feel like it gets close to the line where TypeScript stops being fun anymore... once you start playing the generics game it's hard to stop. Personally I love going crazy with types but it can start to slow people down; it's up to you to weigh those tradeoffs.

Just a thought: if the idea is that "we have some components that get passed around, which always include certain built-in components, ", then would changing everything to take a `Components extends BuiltInComponents` parameter make more sense than passing around the "service map" just to pass it back into `Components<ServiceMap> = BuiltInComponents & ServiceMap`? Might be a lot of refactoring but it could eliminate the extra "service map" type concept.

maschad:
>  then would changing everything to take a Components extends BuiltInComponents parameter make more sense than passing around the "service map" just to pass it back into Components<ServiceMap> = BuiltInComponents & ServiceMap? Might be a lot of refactoring but it could eliminate the extra "service map" type concept.

That's a good point @joeltg , the original impetus to even have these _optional_  services was to reduce Libp2p surface area in consumers and also to [improve extendability through custom modules](https://github.com/libp2p/js-libp2p/pull/1563). I think at the time we went with the `ServiceMap`  because we were unsure how could have an interface that could facilitate them being accessed from a `services` key.  

I think having the [`DefaultComponents` extend the `BuiltInComponents`](https://github.com/libp2p/js-libp2p/pull/1762/files#diff-af796a020e973df5f3ef424841c9f55fdb45d09ffac0ca4715fc0692f257b2e3R35-R125) and also having that [configuration validated](https://github.com/libp2p/js-libp2p/issues/1573) would solve this issue without a large refactor.

> Edit: I guess the other question hanging over this is that the services refactor kinda suggests that people can name their services anything they want now, but in reality dependent services still rely on specific names. Maybe make sure this is communicated in the docs somewhere?

Whilst we do now support custom modules I agree the docs should reflect those limitation, perhaps we could have a section in the configuration doc where we indicate that some services require others [e.g. setup with relay requires identify](https://github.com/libp2p/js-libp2p/blob/master/doc/CONFIGURATION.md#setup-with-relay)

achingbrain:
Discussion: https://github.com/libp2p/js-libp2p/discussions/2135

achingbrain:
I'm going to close this in favour of the solution proposed in https://github.com/libp2p/js-libp2p/discussions/2135

PR #1768, opened 2023-05-19
fix: import path for interfaces module

spalladino:
Fix import path for type from `@libp2p/interfaces` in `address-manager`. 

Without this fix, any clients building with typescript module resolution `node16` get the following error:

```
node_modules/libp2p/dist/src/address-manager/index.d.ts:5:35 - error TS2307: Cannot find module '@libp2p/interfaces/dist/src/events' or its corresponding type declarations.

5 import type { EventEmitter } from '@libp2p/interfaces/dist/src/events';
```

It can also be reproduced by changing the module resolution of this package to `node16` and rebuilding.


achingbrain (approved):

achingbrain:
Bah, this is vscode messing up the auto-imports. Thanks for the PR!

PR #1771, opened 2023-05-24
docs: update customizing section to reflect service dependencies

### LINKED ISSUE #1763: Docs: Update Customizing section of Libp2p Config to reflect Service constraints
#### Description:

The [customization section of the docs](https://github.com/libp2p/js-libp2p/blob/master/doc/CONFIGURATION.md#customizing-libp2p) should reflect that we do facilitate the [custom modules being passed in as services in the Libp2p config](https://github.com/libp2p/js-libp2p/pull/1563) but also ensure that consumers understand that some services depend on others e.g. such as circuit relay v2 and kad-dht don't really work without identify

maschad:
Closes #1763

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -4,6 +4,7 @@
 - [Modules](#modules)
   - [Transport](#transport)
   - [Stream Multiplexing](#stream-multiplexing)
+    - [Muxer Selection](#muxer-selection)
   - [Connection Encryption](#connection-encryption)
   - [Peer Discovery](#peer-discovery)
   - [Content Routing](#content-routing)
@@ -89,6 +90,10 @@ Some available stream multiplexers are:
 
 - [@libp2p/mplex](https://github.com/libp2p/js-libp2p-mplex)
 - [@chainsafe/libp2p-yamux](https://github.com/chainsafe/js-libp2p-yamux)
+
+Some transports such as WebRTC and WebTransport come with their own built-in stream multiplexing capabilities.
+
+If you configure multiple muxers for use in your application, js-libp2p will choose the first muxer in the list. Therefore, ordering matters.
 
 If none of the available stream multiplexers fulfills your needs, you can create a libp2p compatible stream multiplexer. A libp2p multiplexer just needs to be compliant with the [Stream Muxer Interface](https://github.com/libp2p/js-interfaces/tree/master/src/stream-muxer).
 


vmx:
I was using my on configuration and was just using the `pubsub` service. That isn't enough, it won't find the other peers. You would also need the `identify` service to make it work. That should be added to the docs (thanks @achingbrain for pointing that out).

maschad:
Thanks for pointing that out @vmx I've updated it.

maschad:
> In the other sections the identify service is explicitly mentioned in the description and not only the code. Would it make sense to do the same for pubusb?
 
 That's a good point. When I listed the [Potential methods for discovering peers](https://github.com/libp2p/js-libp2p/pull/1771/files#diff-a97f75020345697a0dc0ab9adb4d6cc5de41f8b057d3601eafe07a4a812123d9R331) I thought that may suggest that identify protocol is implicit, but I think it would be clearer to add it explicitly.

doc/CONFIGURATION.md
```diff
@@ -30,6 +30,7 @@
     - [Configuring Metrics](#configuring-metrics)
     - [Configuring PeerStore](#configuring-peerstore)
     - [Customizing Transports](#customizing-transports)
+  - [Configuring AutoNAT](#configuring-autonat)
```
achingbrain, line 33:
```suggestion
    - [Configuring AutoNAT](#configuring-autonat)
```

doc/CONFIGURATION.md
```diff
@@ -920,6 +943,32 @@ const node = await createLibp2p({
 })
 ```
 
+### Configuring AutoNAT
+While the identify protocol allows peers to inform each other about their observed network addresses, sometimes these addresses are inaccessible as the peer may be located in a private network (i.e., behind a NAT or a firewall).
```
achingbrain, line 947:
```suggestion
#### Configuring AutoNAT

While the identify protocol allows peers to inform each other about their observed network addresses, sometimes these addresses are inaccessible as the peer may be located in a private network (i.e., behind a NAT or a firewall).
```

doc/CONFIGURATION.md
```diff
@@ -920,6 +943,33 @@ const node = await createLibp2p({
...
+  services: {
+    nat: autoNATService({
+      protocolPrefix: 'my-node', // allows overriding the default protocol prefix,The AutoNAT protocol uses the protocol ID /libp2p/autonat/1.0.0 and involves the exchange of Dial and DialResponse messages.
+      timeout: 30000,
+      maxInboundStreams: 1000,
+      maxOutboundStreams: 1000
+    })
+  }
+})
+```
```
achingbrain, line 970:
```suggestion
In order for a node to have confidence that it is publicly dialable, the AutoNAT protocol can be used to instruct remote peers to dial the node on the addresses that it believes to be public.

If enough peers report that this address is dialable, the node is free to change it's relationship to the rest of the network; for example, it could become a DHT server or fulfil some other public role.

For more information see https://docs.libp2p.io/concepts/nat/autonat/#what-is-autonat

```ts
import { createLibp2p } from 'libp2p'
import { autoNATService } from 'libp2p/autonat'

  services: {
    nat: autoNATService({
      protocolPrefix: 'my-node', // this should be left as the default value to ensure maximum compatibility
      timeout: 30000, // the remote must complete the AutoNAT protocol within this timeout
      maxInboundStreams: 1, // how many concurrent inbound AutoNAT protocols streams to allow on each connection
      maxOutboundStreams: 1 // how many concurrent outbound AutoNAT protocols streams to allow on each connection
    })
  }
})
```
```

File: doc/CONFIGURATION.md [hunk 4]
```diff
@@ -344,23 +358,28 @@ const node = await createLibp2p({
       noise()
     ],
     services: {
+      identify: identifyService(),
       pubsub: gossipsub({
         emitSelf: false,                                  // whether the node should emit to self on publish
         globalSignaturePolicy: SignaturePolicy.StrictSign // message signing policy
-      })
+      }),
```
achingbrain, line 365:
```suggestion
      })
```

doc/CONFIGURATION.md
```diff
@@ -323,6 +326,16 @@ const node = await createLibp2p({
...
+Before a peer can subscribe to a topic it must find other peers and establish network connections with them. The pub/sub system doesn’t have any way to discover peers by itself. Instead, it relies upon the application to find new peers on its behalf, a process called ambient peer discovery. This means that pubsub relies on the identify service to exchange peer information with other peers.
+
+Potential methods for discovering peers include:
+
+- [Distributed hash tables](#dht)
+- [Local network broadcasts](https://docs.libp2p.io/concepts/discovery-routing/mdns/)
+- [Centralized trackers or rendezvous points](https://docs.libp2p.io/concepts/discovery-routing/rendezvous/)
+- [Lists of bootstrap peers](https://github.com/libp2p/js-libp2p-bootstrap)
+
+
```
achingbrain, line 338:
```suggestion

```

doc/CONFIGURATION.md
```diff
@@ -344,23 +358,28 @@ const node = await createLibp2p({
...
-  }
-})
+  })
 ```
 
 #### Customizing DHT
 
+As explained in [previous sections](#dht) the kad-dht is a Distributed Hash Table based on the Kademlia routing algorithm, with some modifications.
+
+libp2p uses the DHT as the foundation of its peer routing and content routing functionality. The kadDHT service relies on the Identify service for peer identification, address discovery and protocol support discovery. Upon creating or learning of a new address, the peer can push the new address to all peers it’s currently aware of. This keeps everyone’s routing tables up to date and makes it more likely that other peers will discover the new address.
```
achingbrain, line 374:
```suggestion
libp2p uses the DHT as an implementation of its peer routing and content routing functionality.

The kadDHT service requires the Identify service to discover other peers that support the protocol and which allows it to use them to make network queries.
```

doc/CONFIGURATION.md
```diff
@@ -323,6 +326,16 @@ const node = await createLibp2p({
 
 #### Customizing Pubsub
 
+Before a peer can subscribe to a topic it must find other peers and establish network connections with them. The pub/sub system doesn’t have any way to discover peers by itself. Instead, it relies upon the application to find new peers on its behalf, a process called ambient peer discovery. This means that pubsub relies on the identify service to exchange peer information with other peers.
```
achingbrain, line 329:
```suggestion
Before a peer can subscribe to a topic it must find other peers and establish network connections with them. The pub/sub system doesn’t have any way to discover peers by itself. Instead, it relies upon the application to find new peers on its behalf, a process called ambient peer discovery.

This means that pubsub requires the identify service to be configured in order to exchange peer information with other peers, including lists of supported protocols.
```

achingbrain (approved):

Issue #1755, opened 2023-05-13
Topology on onDisconnect difference from onConnect makes it hard to distinguish between connections


marcus-pousette:
- **Version**:
0.45.0

- **Subsystem**:
Registrar, ConnectionManager

#### Severity:
Medium

#### Description:
I am running into issues when I am testing my protocol handler where I let nodes dial multiple addresses at once that could succeed.

When a peer is dialing multiple addresses of another peer, only one connection [succeed and the rest will aborted](https://github.com/libp2p/js-libp2p/blob/710da918a89b6f6ee5d03c0d14a311efd40a1b8f/src/connection-manager/dial-queue.ts#L452) 

In a protocol handler on the other other hand we can do something like this

```typescript
this.topology = createTopology({
	onConnect: this.onPeerConnected.bind(this),  // ( peerId, connection ) => void
	onDisconnect: this.onPeerDisconnected.bind(this), // ( peerId ) => void
});
```
Which means multiple connection can happen, before some of them gets aborted. When they get aborted, and onDisconnect is called, it is hard to know which connection was actually closed, since the connection argument is not passed as with ```onConnect```

Questions/Thoughts 
- If ```onDisconnect``` also emits the 'connection' object  just as with ```onConnect``` we can actually compare connection ids to keep track events, and determine if there is actually a connection open even if multiple onConnect and onDisconnect happened. [And it seems fairly easy to do?](https://github.com/libp2p/js-libp2p/blob/710da918a89b6f6ee5d03c0d14a311efd40a1b8f/src/registrar.ts#L175)
- Why don't we allow multiple connections (when dialing)? What if, in the future we want to have multiple kinds of connections because they have different qualities? Like Bluetooth + WebRTC ? 
- If we only allow one connection to succeed, why is the fastest one to succeed the right one? What if Webtransport and Websocket are racing, but Websocket wins but Webtransport is preferred?  If I can't have two connections concurrently I have to first try Webtransport in this case, then aftewards, Websocket. If I would be able to have multiple connections concurrently, I could just dial both addresses, and at the protocol handler being able to go from one to the other whenever it is possible without uptime loss.

PR #1773, opened 2023-05-25
fix: use peer events in registrar

achingbrain:
Refactors the registrar to use `peer:connect` and `peer:disconnect` events instead of `connection:open` and `connection:close`.

The connection events are too low-level for the purpose and can cause a race condition when multiple multiaddrs are dialed at the same time.

When this happens the first connection survives and others are closed but their corresponding `connection:open` event has already fired.

Interested parties registering topologies have their onConnect event invoked with the soon-to-be-closed connection.

The `peer:connect` event happens later in the lifecycle by which point the extra connection has been closed.

### DIFF
--- a/src/registrar.ts
+++ b/src/registrar.ts
@@ -197,7 +197,7 @@ export class DefaultRegistrar implements Registrar {
         const connection = this.components.connectionManager.getConnections(peer.id)[0]
 
         if (connection == null) {
-          log('peer %p connected but the connection manager did not have a connection')
+          log('peer %p connected but the connection manager did not have a connection', peer)
           // peer disconnected while we were loading their details from the peer store
           return
         }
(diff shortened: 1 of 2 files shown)


maschad (changes requested):
Looks good, just one comment

maschad (approved):

marcus-pousette:
So this will essentially solve/fix https://github.com/libp2p/js-libp2p/issues/1755 ? I wrote this issue because I had problems when doing multi-dialing.

Is this a breaking change also? Will protocol handlers work differently as only one connection event will be emitted in the toplogies?

achingbrain:
Yes, I think it should solve #1755.

It's not a breaking change because the topology is intended to be notified when a peer that supports the given protocol connects/disconnects, not when they open a connection.

The `connection:open` event is the wrong time to do it anyway since identify hasn't run yet (which is how we discover which protocols a peer supports) - the topology is only notified in this case when a peer has reconnected.

Notifying topologies mostly happens after the `peer:update` event which happens when `libp2p.peerStore.merge` or `libp2p.peerStore.update` is called in a way that updates the peer's protocol list, for example after the first identify.

marcus-pousette:
Aha ok! Thanks for the info

PR #1778, opened 2023-05-29
feat: configuration validation

### LINKED ISSUE #1757: Undefined services causes exception
- **Version**:
0.45.0

#### Severity:
Low

#### Description:
Creating a libp2p instance with services that are undefined creates a null pointer exception even if it is allowed by the types

```typescript 
createLibp2p({
...,
    services: {
         abc: undefined
    }
})
```

To mitigate this I think there needs to be a null check before this
https://github.com/libp2p/js-libp2p/blob/710da918a89b6f6ee5d03c0d14a311efd40a1b8f/src/libp2p.ts#L167
to check if ```createService``` is a function. 

I ended up with this error because I am generating different configs based on conditions, and this leads me to assign "undefined" to services ```relay``` occasionally

### LINKED ISSUE #1903: Swallowed Error in CircuitRelayV2 if reservations.defaultDataLimit is set to number
- **Version**:
0.45.9

- **Platform**:
`Linux master 6.3.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Tue, 30 May 2023 13:44:01 +0000 x86_64 GNU/Linux`
`Brave Browser 114.1.52.117`

- **Subsystem**:
CircuitRelayV2

#### Severity:
Low

#### Description:
When configuring 'reservations' on the CircuitRelayServer, there is a very easy landmine to hit. Setting the 'defaultDataLimit' to any number will cause all relay connections to start failing. This is caused because the configuration value is passed directly through to 'countStreamBytes' with no validation.
https://github.com/libp2p/js-libp2p/blob/master/packages/libp2p/src/circuit-relay/utils.ts#L14

While the operation will throw a TypeError due to the mixed types, it will not be seen because of the way countStreamBytes used in a pipeline submitted to queueMicroTask.

- create a CircuitRelayServer service
- configure 'reservations.defaultDataLimit' to any number (eg. 131072)
- observe CircuitRelay connections will be setup but fail to upgrade

maschad:
Closes https://github.com/libp2p/js-libp2p/issues/1573
Closes #1757 
Closes #1903 

I decided to go with [yup](https://github.com/jquense/yup) for a few reasons:

- Integration was more seamless
- Is used by substantially more projects ( 545k vs 105k)
- Preferred the schema building via chaining as opposed to defining complex functions (especially for defining integers, mins and defaults)
- Was more actively maintained ( most recent [commit 3 weeks ago](https://github.com/jquense/yup/commit/e593f8f72e7195cf0ac48fa8e1cd82d95c1e6bb5) as opposed to [last year November](https://github.com/ianstormtaylor/superstruct/commit/03d65bd37d505d934db7cb48eed691fc41ecba7c)
- More downloads ( 4,201,246 vs 889,754 weekly downloads)


maschad:
Currently blocked by https://github.com/libp2p/js-libp2p/issues/1867

achingbrain:
Looking good.

What's the thinking behind this pattern?:

```js
class ServiceClass implements MyService {
  constructor (components, init) {
     // use init
  }
}

export function myService (init: MyServiceInit = {}): (components: MyComponents) => MyService {
  const validatedConfig = object({
    // config def
  }).validateSync(init)

  return (components) => {
    return new ServiceClass(components, validatedConfig)
  }
}
```

The thing that looks wonky to me is that the class doesn't validate it's args, it's done in the factory instead, so the class has to trust whatever it's given.

The config schema is also created for each object instantiation instead of having one definition that all instances can use - does it retain state or something?

Why not do something like:

```js
const configSchema = object({
  // config def
})

class ServiceClass implements MyService {
  constructor (components, init) {
     init = configSchema.validate(init)
     // use init
  }
}

export function myService (init: MyServiceInit = {}): (components: MyComponents) => MyService {
  return (components) => {
    return new ServiceClass(components, validatedConfig)
  }
}
```

achingbrain (changes requested):
Comment above..

maschad:
> The thing that looks wonky to me is that the class doesn't validate it's args, it's done in the factory instead, so the class has to trust whatever it's given.

> The config schema is also created for each object instantiation instead of having one definition that all instances can use - does it retain state or something?

If we continue to use factory method pattern with a 1-to-1 mapping of service to method then this way doesn't offer any benefits but also there isn't any drawback of the service object trusting the parameters it was passed (since only one method would pass those parameters in). 

> The config schema is also created for each object instantiation instead of having one definition that all instances can use - does it retain state or something?

[Yup's schema is immutable](https://github.com/jquense/yup#composition-and-reuse), meaning that modifications to the schema will result in new schema instances rather than altering the original schema. So I can see the drawback of creating one for each object instantiated.

I don't have a strong opinion on doing it this way so if the memory savings outweigh the potential flexibility to have more factory methods then I can refactor.

maschad:
@achingbrain would you be able to give this a review please?

packages/libp2p/src/circuit-relay/server/index.ts
```diff
@@ -6,13 +6,18 @@ import { RecordEnvelope } from '@libp2p/peer-record'
...
+  DEFAULT_DURATION_LIMIT,
   DEFAULT_HOP_TIMEOUT,
-  RELAY_SOURCE_TAG,
-  RELAY_V2_HOP_CODEC,
-  RELAY_V2_STOP_CODEC
+  DEFAULT_MAX_RESERVATION_CLEAR_INTERVAL,
+  DEFAULT_MAX_RESERVATION_STORE_SIZE,
+  DEFAULT_MAX_RESERVATION_TTL,
+  RELAY_SOURCE_TAG
+  , RELAY_V2_HOP_CODEC, RELAY_V2_STOP_CODEC
```
SgtPooki, line 20:
```suggestion
  RELAY_SOURCE_TAG, 
  RELAY_V2_HOP_CODEC, 
  RELAY_V2_STOP_CODEC
```

packages/libp2p/src/config/config.ts
```diff
@@ -0,0 +1,44 @@
...
+      throw new Error('identify service is required when using kadDHT, relay, or ping')
+    }
+  }
+
+  const parsedOpts = libp2pConfig.validateSync(opts)
+
+  const resultingOptions: Libp2pInit<T> = mergeOptions(DefaultConfig, parsedOpts)
+
+  return resultingOptions
+}
```
SgtPooki, line 44: I love the idea of a root config validator that explicitly fails on an invalid config instead of relying on the failure to happen in a deeply nested component where the issue may arise.

packages/libp2p/src/upnp-nat/index.ts
```diff
@@ -70,21 +71,32 @@ class UPnPNAT implements Startable {
...
-
     this.started = false
-    this.externalAddress = init.externalAddress
-    this.localAddress = init.localAddress
-    this.description = init.description ?? `${pkg.name}@${pkg.version} ${this.components.peerId.toString()}`
-    this.ttl = init.ttl ?? DEFAULT_TTL
-    this.keepAlive = init.keepAlive ?? true
-    this.gateway = init.gateway
+
+    const validIPRegex = /^(?:(?:^|\.)(?:\d|[1-9]\d|1\d{2}|2[0-4]\d|25[0-5])){4}$/
```
SgtPooki, line 83: would love to see a link to the tests around the regex similar to https://www.regextester.com/104038 or https://regex101.com/r/aL7tV3/1 or similar

SgtPooki (approved):
I don't see any major errors but I have a few questions:

1. Are the changes in the *.spec.* files fixes to incorrect configs set during testing, that were found because of these changes? If so, that's great.
2. What is the bundle size increase with this? 
3. Are there any hot-paths that we need to look out for?
   * things where instantiation happens frequently and config validation is re-run (and not needed)? If there are, I missed them.

packages/libp2p/src/upnp-nat/index.ts
maschad: That link should suffice :) Normally I would agree with you but since it's for validation as opposed to a regex used in a function, the validation is a test in and of itself imo

maschad:
> I don't see any major errors but I have a few questions:
> 
> 1. Are the changes in the `*.spec.*` files fixes to incorrect configs set during testing, that were found because of these changes? If so, that's great.

That's correct.

> 2. What is the bundle size increase with this?

Gzipped and minified yup is [12kB](https://bundlephobia.com/package/yup@1.3.2) which is negligible 

> 3. Are there any hot-paths that we need to look out for?
>    
>    * things where instantiation happens frequently and config validation is re-run (and not needed)? If there are, I missed them.

This would only instantiate one validation object per class that's already instantiated, which again is insignificant, any optimizations in that regard would be peripheral to this.

packages/libp2p/src/upnp-nat/index.ts
achingbrain, line 83: Can we use a function for validation instead of a regex?  Then reuse something like [@chainsafe/is-ip](https://www.npmjs.com/package/@chainsafe/is-ip)

packages/libp2p/test/circuit-relay/discovery.node.ts
```diff
@@ -34,6 +35,7 @@ describe('circuit-relay discovery', () => {
         mockContentRouting()
       ],
       services: {
+        identify: identifyService(),
```
achingbrain, line 38: Why was it necessary to add identify as part of config validation?  I see this change in a lot of tests here.

packages/libp2p/test/configuration/protocol-prefix.node.ts
```diff
@@ -67,7 +67,8 @@ describe('Protocol prefix is configurable', () => {
       '/ipfs/id/1.0.0',
       '/ipfs/id/push/1.0.0',
       '/ipfs/ping/1.0.0',
-      '/libp2p/fetch/0.0.1'
+      '/libp2p/fetch/0.0.1',
+      '/libp2p/circuit/relay/0.2.0/stop'
```
achingbrain, line 71: Why did the extra protocol appear?

packages/libp2p/test/connection-manager/index.node.ts
```diff
@@ -235,7 +235,9 @@ describe('libp2p.connections', () => {
           },
           connectionManager: {
             minConnections,
-            maxConnections: 1
+            maxConnections: 1,
+            inboundConnectionThreshold: 1,
```
achingbrain, line 239: Why do we need to specify the extra config now?  This seems to happen in a few places here.

packages/libp2p/.aegir.js
```diff
@@ -24,7 +24,9 @@ export default {
       const peerId = await createEd25519PeerId()
       const libp2p = await createLibp2p({
         connectionManager: {
-          inboundConnectionThreshold: Infinity,
+          inboundConnectionThreshold: 1000,
```
achingbrain, line 27: I take it yup doesn't accept `Infinity` as a number?

packages/libp2p/src/autonat/index.ts
```diff
@@ -108,14 +109,23 @@ class DefaultAutoNATService implements Startable {
   private started: boolean
 
   constructor (components: AutoNATComponents, init: AutoNATServiceInit) {
+    const validatedConfig = object({
```
achingbrain, line 112: Please have config definitions outside of the class so we don't create new validators on every object instantiation:

```js
const configValidator = object({
  protocolPrefix: string().default(PROTOCOL_PREFIX),
  // ...etc
})

class DefaultAutoNATService implements Startable {
  // ...etc

  constructor (components: AutoNATComponents, init: AutoNATServiceInit) {
    const config = configValidator.validateSync(init)
    // ...etc
  }
}
```

achingbrain (commented):
This was not ready.  I know it's taken a long time but other issues have been higher priority.  Please wait for review before merging.

achingbrain:
Reverted and re-opened as #2133

packages/libp2p/test/circuit-relay/discovery.node.ts
achingbrain: Ah, I see - it's because you've added validation to the `validateConfig` method to enforce the presence of certain service keys.

Identify is only necessary for circuit relay for ambient relay discovery (e.g. after dialing a peer we notice they are a relay and try to get a reservation if we need one).

In this test we ensure that the content routing module can find relays (e.g. via searching for the advertisement CID) so we actually don't want identify enabled.

Anyway I think enforcing service dependencies is outside of the scope of this PR, please revert in https://github.com/libp2p/js-libp2p/pull/2133.

packages/libp2p/.aegir.js
maschad: Correct

packages/libp2p/test/connection-manager/index.node.ts
maschad: Because the intention is that we should not be able to have more inbound connections than `maxConnections` as that would result in them getting pruned anyways, discussing with @wemeetagain he raised a use case where this may be applicable though, where a node may want to reach maxConnections as fast as possible and so it may accept a greater amount of inbound connections in a short space of time.

packages/libp2p/test/configuration/protocol-prefix.node.ts
maschad: Because we have configured a circuitRelayServer in the [aegir config](https://github.com/libp2p/js-libp2p/pull/1778/files/d45851d01f85b80dea647e0aaefa8371b753ec43#diff-9a79967796b063d2526c5bfdcf1b871c76f5f932863e48da9b2f8ceb59dba26eR56)

packages/libp2p/test/circuit-relay/discovery.node.ts
maschad: We can discuss the use cases [here](https://github.com/libp2p/js-libp2p/discussions/2135)

packages/libp2p/.aegir.js
```diff
@@ -24,7 +24,9 @@ export default {
...
+          maxIncomingPendingConnections: 1000,
+          maxConnections: 1000,
```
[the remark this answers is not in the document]
maschad: See my comment https://github.com/libp2p/js-libp2p/pull/1778?notification_referrer_id=NT_kwDOAJTalrI2NjAzMzk1ODMyOjk3NTUyODY#discussion_r1349153967

packages/libp2p/src/upnp-nat/index.ts
maschad: what's the benefit of using a external dependency and parsing the string over a regular expression?

Issue #1776, opened 2023-05-26
Package path . is not exported from package /home/.../node_modules/libp2p


reubenfirmin:
I cannot get to the hello world point with https://github.com/libp2p/js-libp2p/blob/master/doc/GETTING_STARTED.md.

```
rm -fr node_modules/
npm install
npm install libp2p
npm install @libp2p/websockets
npm run build
```

This yields an error:

```
Module not found: Error: Package path . is not exported from package /home/.../node_modules/libp2p (see exports field in /home/.../node_modules/libp2p/package.json)
```

It set up package.json with:

```
  "dependencies": {
    "@libp2p/websockets": "^6.0.1",
    "libp2p": "^0.45.4"
  },
```

My code is (typescript):

```
import { createLibp2p } from 'libp2p';
import {webSockets} from "@libp2p/websockets";

export class PeerSyncer {

    constructor(private identity: any) {}

    async start(): Promise<void> {

        const libp2p = await createLibp2p({
            // addresses: {
            //     listen: ['/ip4/0.0.0.0/tcp/0', '/ip4/0.0.0.0/tcp/0/ws']
            // },
            // transports: [webSockets()]
        });
    }
}
```

Commenting out the call to createLibp2p allows it to build, presumably because the tree shaker gets rid of the import.

```
npm --version
9.6.5
```

maschad:
Could you share your `tsconfig.json` and the node version as well? We define the export types [here](https://github.com/libp2p/js-libp2p/blob/master/package.json#L50)

reubenfirmin:
tsconfig:

```
{
  "compilerOptions": {
    "strict": true,
    "module": "commonjs",
    "target": "es6",
    "esModuleInterop": true,
    "sourceMap": true,
    "rootDir": "src",
    "outDir": "dist/js",
    "noEmitOnError": true,
    "typeRoots": [ "node_modules/@types" ],
    "jsx": "react"
  }
}
```

```
node --version
v18.16.0
```

achingbrain:
>  "module": "commonjs",
>  "target": "es6"

libp2p is an ESM-only module, this configuration option uses the wrong module system - please change it to use ESM:

```json
{
  "compilerOptions": {
    "module": "ES2022",
    "esModuleInterop": true,
    "target": "ES2022",
    "moduleResolution": "node"
  }
}
```

PR #1782, opened 2023-05-31
docs: add ESM section to avoid configuration issues

maschad:
Related https://github.com/libp2p/js-libp2p/issues/1776

### DIFF
--- a/doc/GETTING_STARTED.md
+++ b/doc/GETTING_STARTED.md
@@ -36,7 +36,19 @@ Since `libp2p@0.37.0` modules are now [ESM-only](https://gist.github.com/sindres
 
 ESM is the module system for JavaScript, it allows us to structure our code in separate files without polluting a global namespace.
 
-Other systems have tried to fill this gap, notably CommonJS, AMD, RequireJS and others, but ESM is [the official standard format](https://tc39.es/ecma262/#sec-modules) to package JavaScript code for reuse. For more info on enablng ES modules in Node, see [this guide](https://nodejs.org/api/esm.html).
+Other systems have tried to fill this gap, notably CommonJS, AMD, RequireJS and others, but ESM is [the official standard format](https://tc39.es/ecma262/#sec-modules) to package JavaScript code for reuse. This means that you need ensure your configuration uses the correct module system, if you are using Typescript, set the [`module` field in your tsconfig](https://www.typescriptlang.org/tsconfig#module) to your desired ESM version e.g.
+
+```json
+{
+  "compilerOptions": {
+    "module": "ES2022",
+    "esModuleInterop": true,
+    "target": "ES2022",
+    "moduleResolution": "node"
+  }
+}
+```
+For more info on enablng ES modules in Node, see [this guide](https://nodejs.org/api/esm.html).
 
 ### Basic setup
 
--- a/doc/GETTING_STARTED.md
+++ b/doc/GETTING_STARTED.md
@@ -49,6 +49,7 @@ Other systems have tried to fill this gap, notably CommonJS, AMD, RequireJS and 
 }
 
 For more info on enablng ES modules in Node, see [this guide](https://nodejs.org/api/esm.html).
+
 ### Basic setup
 
 Now that we have libp2p installed, let's configure the minimum needed to get your node running. The only modules libp2p requires are a [**Transport**][transport] and [**Crypto**][crypto] module. However, we recommend that a basic setup should also have a [**Stream Multiplexer**](streamMuxer) configured, which we will explain shortly. Let's start by setting up a Transport.
--- a/doc/GETTING_STARTED.md
+++ b/doc/GETTING_STARTED.md
@@ -47,6 +47,7 @@ Other systems have tried to fill this gap, notably CommonJS, AMD, RequireJS and 
     "moduleResolution": "node"
   }
 }
+```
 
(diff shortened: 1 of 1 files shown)


doc/GETTING_STARTED.md
```diff
@@ -29,6 +30,14 @@ npm install libp2p
 
 If you're new to libp2p, we recommend configuring your node in stages, as this can make troubleshooting configuration issues much easier. In this guide, we'll do just that. If you're more experienced with libp2p, you may wish to jump to the [Configuration readme](./CONFIGURATION.md).
 
+## ESM
```
achingbrain, line 33:
```suggestion
### ESM
```

this needs to be the same level as it’s sibling headers, otherwise “Basic Setup” etc become sub-sections of “ESM”

the TOC will also need updating.

doc/GETTING_STARTED.md
```diff
@@ -29,6 +30,14 @@ npm install libp2p
...
+
+Since `libp2p@0.37.0` is that the module is now [ESM-only](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).
```
achingbrain, line 35:
```suggestion
Since `libp2p@0.37.0` the module is now [ESM-only](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).
```

doc/GETTING_STARTED.md
```diff
@@ -29,6 +30,14 @@ npm install libp2p
...
+
+ESM is the module system for JavaScript, it allows us to structure our code in separate files without polluting a global namespace.
+
+Other systems have tried to fill this gap, notably CommonJS, AMD, RequireJS and others, but ESM is [the official standard format](https://tc39.es/ecma262/#sec-modules) to package JavaScript code for reuse.
+
```
achingbrain, line 40: I think it’s worth adding a section with the minimal typescript config for generating ESM here too

achingbrain (changes requested):

doc/GETTING_STARTED.md
maschad: Well someone could use ESM modules without Typescript but I suppose we suggest that  the `type` parameter be set to `module` as mentioned in the [node docs](https://nodejs.org/api/esm.html#enabling)

achingbrain: That is of course true, but by default TypeScript outputs CJS which then explodes at runtime.

People are then confused because since TypeScript uses `import` it looks like they are authoring ESM when in fact they are not.

A cursory search through the closed issues in js-IPFS around the time we converted to ESM will reveal that this is a very common mistake.

doc/GETTING_STARTED.md
```diff
@@ -29,6 +30,26 @@ npm install libp2p
 
 If you're new to libp2p, we recommend configuring your node in stages, as this can make troubleshooting configuration issues much easier. In this guide, we'll do just that. If you're more experienced with libp2p, you may wish to jump to the [Configuration readme](./CONFIGURATION.md).
 
+### ESM
+
+Since `libp2p@0.37.0` modules are now [ESM-only](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).
+
+ESM is the module system for JavaScript, it allows us to structure our code in separate files without polluting a global namespace.
+
+Other systems have tried to fill this gap, notably CommonJS, AMD, RequireJS and others, but ESM is [the official standard format](https://tc39.es/ecma262/#sec-modules) to package JavaScript code for reuse. This means that you need ensure your configuration uses the correct module system, if you are using Typescript, set the [`module` field in your tsconfig](https://www.typescriptlang.org/tsconfig#module) to your desired ESM version e.g.
```
achingbrain, line 39:
```suggestion
Other systems have tried to fill this gap, notably CommonJS, AMD, RequireJS and others, but ESM is [the official standard format](https://tc39.es/ecma262/#sec-modules) to package JavaScript code for reuse. This means that you need ensure your configuration uses the correct module system, if you are using Typescript, set the [`module` field in your tsconfig](https://www.typescriptlang.org/tsconfig#module) to `ES2022 ` or later e.g.
```

doc/GETTING_STARTED.md
```diff
@@ -29,6 +30,26 @@ npm install libp2p
...
+{
+  "compilerOptions": {
+    "module": "ES2022",
+    "esModuleInterop": true,
+    "target": "ES2022",
+    "moduleResolution": "node"
+  }
+}
+```
+For more info on enablng ES modules in Node, see [this guide](https://nodejs.org/api/esm.html).
```
achingbrain, line 51:
```suggestion
```

For more info on enablng ES modules in Node, see [this guide](https://nodejs.org/api/esm.html).
```

**Review by achingbrain — APPROVED** [review]

File: doc/GETTING_STARTED.md [hunk 6]
```diff
@@ -29,6 +30,25 @@ npm install libp2p
...
+  "compilerOptions": {
+    "module": "ES2022",
+    "esModuleInterop": true,
+    "target": "ES2022",
+    "moduleResolution": "node"
+  }
+}
+
+For more info on enablng ES modules in Node, see [this guide](https://nodejs.org/api/esm.html).
 ### Basic setup
```
achingbrain, line 52:
```suggestion
For more info on enablng ES modules in Node, see [this guide](https://nodejs.org/api/esm.html).

### Basic setup
```

PR #1784, opened 2023-06-01
feat: enable manual identify

wemeetagain:
In some applications, automatically dialing identify is not desired, and more manual control is useful.

To that end, this PR contains two edits that act together
- Add an additional identify init boolean option `runOnConnectionOpen`, which controls whether to listen to `'connection:open'` events to trigger identify
- Add a return value to the `identify` method


maschad (changes requested):
I see the potential benefits of having more control over when streams are opened, but could you elaborate a bit more on your use case? I don't think this is in [accordance with the spec](https://github.com/libp2p/specs/blob/master/identify/README.md#identify) since identify is used to inform known peers about changes which will happen whilst the node is running (which suggests it should always be run when opening a new connection).

wemeetagain:
The spec doesn't say anything about when the identify protocol should be dialed. I think the 'dial on connection open' behavior is just a common practice that is useful amd sufficient for most usecases. (Identify/push, a separate protocol, is used to inform peers about changes, but again, the spec gives no commands about when that protocol must or should be dialed)

In our usecase, we have another protocol that we want to dial first on connection open (We call it the 'status' protocol). Then depending on the result of that, either success or failure (note that failure can be failure to support that protocol, or responding to that protocol with a response that we don't like), the peer will either be disconnected or more protocols will be dialed, along with identify, which we use mainly for the agent version string.
We'd like to avoid the unnecessary identify in the case that the first protocol result is a failure.

maschad:
> The spec doesn't say anything about when the identify protocol should be dialed. I think the 'dial on connection open' behavior is just a common practice that is useful amd sufficient for most usecases. (Identify/push, a separate protocol, is used to inform peers about changes, but again, the spec gives no commands about when that protocol must or should be dialed)

That's true I do agree that the matter of when is more open to interpretation, so since this would be enabled by default this doesn't change anything functionally. 

> In our usecase, we have another protocol that we want to dial first on connection open (We call it the 'status' protocol). Then depending on the result of that, either success or failure (note that failure can be failure to support that protocol, or responding to that protocol with a response that we don't like), the peer will either be disconnected or more protocols will be dialed, along with identify, which we use mainly for the agent version string.
We'd like to avoid the unnecessary identify in the case that the first protocol result is a failure.

Thanks for explaining, I think this is reasonable.

maschad (approved):

src/identify/identify.ts
```diff
@@ -50,7 +50,8 @@ const defaultValues = {
   maxPushIncomingStreams: 1,
   maxPushOutgoingStreams: 1,
   maxObservedAddresses: 10,
-  maxIdentifyMessageSize: 8192
+  maxIdentifyMessageSize: 8192,
+  autodialOnConnectionOpen: true
```
achingbrain, line 54: "auto dial" is quite a loaded term, plus this service doesn't dial so perhaps `runOnConnectionOpen`?

src/identify/identify.ts
```diff
@@ -297,7 +300,7 @@ export class DefaultIdentifyService implements Startable {
     }
   }
 
-  async identify (connection: Connection, options: AbortOptions = {}): Promise<void> {
+  async identify (connection: Connection, options: AbortOptions = {}): Promise<IdentifyResult> {
```
achingbrain, line 303: Please can you create an `IdentifyService` interface in the index file that contains only this method, have the factory function export it and have this class implement it.

This looks like something that was missed, oops.

The interface method could also use a JSDoc that says something like "due to the default limits on inbound/outbound streams for this protocol, invoking this method when `runOnConnectionOpen` is true can lead to unpredictable results as streams may be closed by the local or the remote node.  Please use with caution.  If you find yourself needing to call this method to discover other peers that support your protocol, you may be better off configuring a topology to be notified instead." and linking to the appropriate types.

achingbrain (changes requested):
LGTM in theory, though I fear people will call this function without specifying the only-run-manually config option which will lead to hard-to-predict stream closures due to the default stream limits so this introduces a bit of a foot gun.

This PR is missing tests, please can you add some to prevent regressions in future.

packages/libp2p/test/connection-manager/direct.spec.ts
```diff
@@ -402,6 +403,57 @@ describe('libp2p.dialer (direct, WebSockets)', () => {
...
+          agentVersion: AGENT_VERSION,
+          timeout: 60000,
+          maxInboundStreams: 1,
+          maxOutboundStreams: 1,
+          maxPushIncomingStreams: 1,
+          maxPushOutgoingStreams: 1,
+          maxObservedAddresses: 10,
+          maxIdentifyMessageSize: 8192,
+          runOnConnectionOpen: false
+        })
```
achingbrain, line 432: This could be simpler to make it more obvious what the test is testing?

```suggestion
        identify: identifyService({
          runOnConnectionOpen: false
        })
```

packages/libp2p/test/connection-manager/direct.spec.ts
```diff
@@ -402,6 +403,57 @@ describe('libp2p.dialer (direct, WebSockets)', () => {
...
+      },
+      connectionGater: mockConnectionGater()
+    })
+
+    if (libp2p.services.identify == null) {
+      throw new Error('Identify service missing')
+    }
+
+    const identifySpy = sinon.spy(libp2p.services.identify as DefaultIdentifyService, 'identify')
```
achingbrain, line 441: Since `identify` is in the interface now, do we still need the cast?

packages/libp2p/src/identify/index.ts
```diff
@@ -60,6 +66,22 @@ export const multicodecs = {
...
+  /**
+   * due to the default limits on inbound/outbound streams for this protocol,
+   * invoking this method when runOnConnectionOpen is true can lead to unpredictable results
+   * as streams may be closed by the local or the remote node.
+   * Please use with caution. If you find yourself needing to call this method to discover other peers that support your protocol,
+   * you may be better off configuring a topology to be notified instead.
+   *
+   * @param connection
+   * @param options
+   * @returns
```
achingbrain, line 79: These jsdoc comments don't seem to add any information that isn't derivable from the interface?

```suggestion
```

packages/libp2p/src/identify/index.ts
```diff
@@ -60,6 +66,22 @@ export const multicodecs = {
...
+   */
+  identify: (connection: Connection, options: AbortOptions) => Promise<IdentifyResult>
+
+}
+
 export function identifyService (init: IdentifyServiceInit = {}): (components: IdentifyServiceComponents) => DefaultIdentifyService {
```
achingbrain, line 85: This should return the interface

```suggestion
export function identifyService (init: IdentifyServiceInit = {}): (components: IdentifyServiceComponents) => IdentifyService {
```

achingbrain (changes requested):
Nearly there, just a few small things to address.

packages/libp2p/test/connection-manager/direct.spec.ts
maschad: Unfortunately yes, I think for similar limitations raised in https://github.com/libp2p/js-libp2p/pull/1762 the compiler doesn't detect the `IdentifyService` interface.

packages/libp2p/src/identify/index.ts
maschad: This has [implications beyond the scope of this PR](https://github.com/libp2p/js-libp2p/actions/runs/5405601445/jobs/9821368599?pr=1784#step:4:316) which can be refactored in a follow up PR.

packages/libp2p/test/connection-manager/direct.spec.ts
achingbrain: I'm not sure this is correct - the `libp2p` variable is declared as `let libp2p: Libp2p<{ identify: unknown }>`.

If it was `let libp2p: Libp2p<{ identify: IndentifyService }>` the cast should be able to be removed.

achingbrain (approved):

PR #1786, opened 2023-06-02
chore: remove webkit fixes

achingbrain:
To see if this is still broken in ci.

### DIFF
--- a/test/autonat/index.spec.ts
+++ b/test/autonat/index.spec.ts
@@ -10,7 +10,7 @@ import * as lp from 'it-length-prefixed'
 import { pipe } from 'it-pipe'
 import { pushable } from 'it-pushable'
 import sinon from 'sinon'
-import { stubInterface, stubObject } from 'sinon-ts'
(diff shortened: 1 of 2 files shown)


achingbrain:
Can you hold off merging draft PRs in future please, this wasn’t done!

maschad:
Sorry about that, will do.

p-shahi:
General convention is to leave PR merges to the author, unless requested/special circumstance. I think we should follow that?

Can also think about enabling mergify which would allow us to add a `merge` or `ship-it` label )indicating a PR is ready) and let automation handle it

maschad:
> General convention is to leave PR merges to the author, unless requested/special circumstance. I think we should follow that?

Agreed. This was an particular case since we where trying to reproduce a CI issue but will adhere to that going forward.

> Can also think about enabling mergify which would allow us to add a merge or ship-it label )indicating a PR is ready) and let automation handle it

Sounds like a good idea.

PR #1798, opened 2023-06-07
fix: map ports via uPnP when addresses change

achingbrain:
Refactors the uPnP port mapper to map ports when addresses change instead of once at startup.

This is because some addresses are not open at startup and only become available some time later.


achingbrain:
Closing in favour of #2797

PR #1802, opened 2023-06-09
docs: Update README.md configure pubsub as a service module

bradwbradw:
enclose `pubsub: gossibsub(...` with `services: {}`


maschad:
Thanks for spotting fixing this!

maschad (approved):
Thanks for spotting fixing this!

PR #1803, opened 2023-06-10
docs: Update keychain info

marcus-pousette:


doc/CONFIGURATION.md
```diff
@@ -588,7 +588,9 @@ Libp2p allows you to setup a secure keychain to manage your keys. The keychain c
 | Name      | Type     | Description                                                                            |
 | --------- | -------- | -------------------------------------------------------------------------------------- |
 | pass      | `string` | Passphrase to use in the keychain (minimum of 20 characters).                          |
-| datastore | `object` | must implement [ipfs/interface-datastore](https://github.com/ipfs/interface-datastore) |
+| dek      | `DEKConfig` | [Info here](https://github.com/libp2p/js-libp2p-keychain)                            |
```
maschad, line 591: We may want to include a bit more details here such as:

```suggestion
| dek      | `DEKConfig` | the default options for generating the derived encryption key, which, along with the passphrase are input to the PBKDF2 function. For more info see: https://github.com/libp2p/js-libp2p-keychain  |
```

doc/CONFIGURATION.md
```diff
@@ -614,11 +616,12 @@ const node = await createLibp2p({
   ],
   keychain: {
     pass: 'notsafepassword123456789',
-    datastore: dsInstant,
-  }
+  },
+  datastore
```
maschad, line 620: We now use `MemoryDatastore` from the https://www.npmjs.com/package/datastore-core package, could you update this accordingly?

maschad (changes requested):
Thanks for updating this! We will soon have doc verification to catch these types of issues https://github.com/libp2p/js-libp2p/pull/1505. I have made some suggestions, almost there.

doc/CONFIGURATION.md
marcus-pousette: I changed it to `import { FsDatastore } from "datastore-fs";` as the purpose of including a datastore was to write to disc. MemoryDatastore only seem to be in memory. It that ok?

marcus-pousette:
I am not sure about the pass config, it [does not seem to be used](https://github.com/libp2p/js-libp2p-keychain/blob/89e9fa93fc333231374d831ab8d15a4e024be6f0/src/index.ts#L389) for exporting and importing peerIds ?

doc/CONFIGURATION.md
marcus-pousette: Perhaps off topic, but, what is really the the difference here, "datastore-level" will use a datastore-interface wrapper around "classic-level", which in turn is a c++  lib with node bindings. "datastore-fs" instead relies on just javascript and Node APIs, functionality wise they are the same, but they might store things differently on disc?  Why is one preferred over another?

maschad:
> I am not sure about the pass config, it [does not seem to be used](https://github.com/libp2p/js-libp2p-keychain/blob/89e9fa93fc333231374d831ab8d15a4e024be6f0/src/index.ts#L389) for exporting and importing peerIds ? I wonder whether the description would be misleading in this case, when I think about it now.

I would say the key management is separate from the exporting of peerIds associated with those keys. A password isn't required because the key isn't being exported in a format that is typically password-protected as Peer IDs are derived from public keys.

maschad (approved):

doc/CONFIGURATION.md
marcus-pousette: yeap, makes sense!

marcus-pousette:
> I would say the key management is separate from the exporting of peerIds associated with those keys. A password isn't required because the key isn't being exported in a format that is typically password-protected as Peer IDs are derived from public keys.

I see. Well I got initially confused as I assumed the password would make anything I import to the keychain, encrypted at rest with the password provided in the config. But there is a different case for these PeerIds I guess then. 

But I am ok with how it is now, if you are ok. I guess the documentation can improve further more in the future iteratively

maschad:
> Would it make sense that I create a separate issue for safely encrypting PeerIds at rest? It is kind of important if you want to protect yourself against supply chain attacks, especially in the browser.

This is a separate issue so feel free to open one, but I am not sure that encrypting PeerIds _alone_ would prevent a supply chain attack in the browser,  especially if the public key is known. I would suspect that preventing such an attack would require a more comprehensive approach that addresses the whole networking stack.

Issue #1800, opened 2023-06-09
JavaScript heap out of memory


chiro-hiro:
- **Version**:

libp2p@0.45.5

- **Platform**:

Mac Book Pro M1

```
ProductName:		macOS
ProductVersion:		13.2.1
BuildVersion:		22D68
```

- **Subsystem**:

#### Severity:

Critical - System crash, application panic.

#### Description:

Write a program with Helia and libp2p, boot up the node. The memory slowly increase to 3.5 GBs then crash.

```
<--- Last few GCs --->

[71875:0x128040000]   222428 ms: Scavenge 4016.5 (4127.8) -> 4005.5 (4128.3) MB, 10.9 / 0.2 ms  (average mu = 0.265, current mu = 0.231) task; 
[71875:0x128040000]   226939 ms: Mark-sweep 4017.8 (4128.3) -> 4004.3 (4129.5) MB, 4479.8 / 0.5 ms  (average mu = 0.218, current mu = 0.134) task; scavenge might not succeed
[71875:0x128040000]   234869 ms: Mark-sweep 4017.9 (4129.5) -> 4005.3 (4130.0) MB, 7851.2 / 0.5 ms  (average mu = 0.110, current mu = 0.010) task; scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x1050054fc node::Abort() [/Users/chiro/.nvm/versions/node/v18.16.0/bin/node]
 2: 0x1050056ec node::ModifyCodeGenerationFromStrings(v8::Local<v8::Context>, v8::Local<v8::Value>, bool) [/Users/chiro/.nvm/versions/node/v18.16.0/bin/node]
 3: 0x10515b0fc v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/Users/chiro/.nvm/versions/node/v18.16.0/bin/node]
 4: 0x105306088 v8::internal::EmbedderStackStateScope::EmbedderStackStateScope(v8::internal::Heap*, v8::internal::EmbedderStackStateScope::Origin, cppgc::EmbedderStackState) [/Users/chiro/.nvm/versions/node/v18.16.0/bin/node]
 5: 0x105309c90 v8::internal::Heap::CollectSharedGarbage(v8::internal::GarbageCollectionReason) [/Users/chiro/.nvm/versions/node/v18.16.0/bin/node]
[...]

PR #1804, opened 2023-06-11
fix: do not allow autodial to run in parallel

achingbrain:
Auto dial attempts to keep the number of active connections above
a configurable minimum, but dialling is slow so the dial queue can
grow quite large. If unchecked this can cause OOMs.

- Adds a `running` boolean to the autodialler to prevent it running
in parallel.
- Adds a configurable limit for the auto dial and circuit relay reservation queue lengths
- Adds a PeerJobQueue that extends PQueue to ensure we don't have two jobs for the same peer in a queue (e.g. when auto dialling or trying to reserve a relay slot)

Fixes https://github.com/libp2p/js-libp2p/issues/1800
Fixes https://github.com/ipfs/helia/issues/143


achingbrain:
Merging because this solves a very real problem though I'd still like a review..

PR #1828, opened 2023-06-15
chore: add non-interface modules to monorepo

achingbrain:
Adds all supported non-interface modules to the monorepo.


BigLep:
@achingbrain @p-shahi : have we lost all the git history of these files (same goes with https://github.com/libp2p/js-libp2p/pull/1823 )?  Given the age of the codebase, it seems unfortunate to lose the history of how the code evolved.  With all the repo consolidation we did with Boxo, we maintained the commit history: https://github.com/ipfs/boxo/wiki/Copied-or-Migrated-Repos-FAQ#was-the-git-history-maintained-for-repos-that-were-copied-in .

p-shahi:
Yeah it looks like this was missed, I believe history was maintained in the go-libp2p monorepo consolidation as well. If it's not too late to add it back, I would be in favor of that.
I would also like to transfer issues from archived repos into this one. I manually unarchived https://github.com/libp2p/js-libp2p-webrtc earlier today and migrated issues, it looks like not all have been archived yet though

achingbrain:
I can add it back but it'll mean force pushing to master.  I hadn't done it because I previously did for the repo consolidation of `js-ipfs` but I don't think anyone ever actually went spelunking deep in the repo history so it didn't seem worth the effort involved but from an OSS point of view it's probably a good idea to preserve contributor history if nothing else.

The simplest way to do this is probably to push straight to master rather than go via a PR.

achingbrain:
```
Your branch and 'origin/master' have diverged,
and have 3828 and 1 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)
```

😱

BigLep:
I'm going to move the conversation into the monorepo issue which is still open: https://github.com/libp2p/js-libp2p/issues/824

PR #1829, opened 2023-06-15
fix: removing stream methods from connection interface

mmsaki:
Moving the discussion from [this pull request](https://github.com/libp2p/js-libp2p-interfaces/pull/417), PR was originally opened in js-libp2p-interfaces before being moving to this monorepo

the muxers should be ultimately responsible for managing streams, that's why `addStream` and `removeStream` methods are being removed from the interface [see](https://github.com/libp2p/js-libp2p-interfaces/pull/417/files#r1230570032)

Fixes [issue libp2p/js-libp2p#1855](https://github.com/libp2p/js-libp2p/issues/1855)

### DIFF
--- a/packages/interface-connection/src/index.ts
+++ b/packages/interface-connection/src/index.ts
@@ -182,8 +182,6 @@ export interface Connection {
   streams: Stream[]
 
   newStream: (multicodecs: string | string[], options?: NewStreamOptions) => Promise<Stream>
-  addStream: (stream: Stream) => void
-  removeStream: (id: string) => void
   close: () => Promise<void>
 }
 


maschad (changes requested):
Thanks for opening this PR, really appreciate the input. Given this is an interface though, it will cause a breaking change in [any class that implements this interface](https://github.com/libp2p/js-libp2p/blob/master/packages/libp2p/src/connection/index.ts#L25) could you remove the methods there as well?

achingbrain:
It's probably worth rebasing this once #1833 is in as it'll show the full scope of the change across the stack.

maschad:
Closing in favour of https://github.com/libp2p/js-libp2p/pull/1912

PR #1847, opened 2023-06-23
feat!: muxed streams as web streams

### LINKED ISSUE #1793: Cleanly closing streams

```js

const pb = pbStream(stream)
stream.close()
```

```
... more stack trace here
```

```
-- local pb.writePB(...) --
-- local stream.close() --
  id: 0,
} +1ms
  id: 0,
} +0ms
```

```js
await delay(0)
stream.close()
```

- [ ] interfaces
- [ ] mplex
- [ ] yamux 

todo:

achingbrain:
Refactors streams from duplex async iterables:

```js
{
  source: Duplex<AsyncGenerator<Uint8Array, void, unknown>, Source<Uint8Array | Uint8ArrayList>, Promise<void>
  sink: (Source<Uint8Array | Uint8ArrayList>) => Promise<void>
}
```

to `ReadableWriteablePair<Uint8Array>`s:

```js
{
  readable: ReadableStream<Uint8Array>
  writable: WritableStream<Uint8Array>
}
```

Since the close methods for web streams are asynchronous, this lets us close streams cleanly - that is, wait for any buffered data to be sent/consumed before closing the stream.

We still need to be able abort a stream in an emergency, so streams have the following methods for graceful closing:

```js
stream.readable.cancel(reason?: any): Promise<void>
stream.writable.close(): Promise<void>

// or

stream.close(): Promise<void>
```

..and for emergency closing:

```js
stream.abort(err: Error): void
```

Connections and multiaddr connections have the same `close`/`abort` semantics, but are still Duplexes since making them web streams would mean we need to convert things like node streams (for tcp) to web streams which would just make things slower.

Transports such as WebTransport and WebRTC already deal in web streams when multiplexing so these no longer need to be converted to Duplex streams so it's win-win.

Fixes #1793


achingbrain:
Still a draft as there's more work to do - one obvious thing is this PR adds yamux to the monorepo - this is just temporary to ensure compatibility and will be removed before release and a separate PR opened with the changes there.

Opening early to get some feedback on the stream/connection interface changes.

wemeetagain:
> we need to convert things like node streams (for tcp) to web streams which would just make things slower.

I'm not convinced that they would be slower.
We already have [some conversion](https://github.com/libp2p/js-libp2p/blob/master/packages/transport-tcp/src/socket-to-conn.ts) which uses [`stream-to-it`](https://github.com/alanshaw/stream-to-it) to convert the node stream to a duplex.
We could instead use [`stream.Duplex.toWeb`](https://nodejs.org/dist/latest-v18.x/docs/api/stream.html#streamduplextowebstreamduplex) to convert the node stream to a web stream.

We would need actual benchmarks to know if it is slower or not.

achingbrain:
@wemeetagain I put a simple benchmark together here - https://gist.github.com/achingbrain/7e65ca748326d3b87b81508020a3321d

Results are on there but TLDR is node streams and web streams with `"type": "bytes"` are about the same, regular web streams are 2x slower and node streams converted to web streams are 2x slower still.

achingbrain:
On the basis of the benchmarks showing a performance degradation with the switch to web streams, is this something we actually want to do?  My gut feeling would be no, not until web streams close the performance gap to node streams/duplex iterables.

achingbrain:
Now that we have performance testing available that involves actual networks, it is worth revisiting this PR to see if the performance penalties previously still apply when networking is involved - if not then the simplification of the codebase would be worth the breaking change here.

achingbrain:
One problem with web streams is there's no way to cancel a read or a write, so we'd need to wrap everything in `raceSignal`:

```ts
import { byteStream } from 'it-byte-stream'
import { racePromise } from 'race-promise'

const signal = AbortSignal.timeout(1_000)

// old
const bytes = byteStream(stream)
await bytes.write(Uint8Array.from([0, 1, 2]), {
  signal
})

// new
const writer = stream.writable.getWriter()
await racePromise(writer.write(Uint8Array.from([0, 1, 2])), signal)
```

One thing to consider is that `racePromise` cannot cancel the onward promise, it will hang around in memory until it eventually resolves or rejects.

PR #1856, opened 2023-06-26
feat!: merge stat properties into stream/connection objects

### LINKED ISSUE #1849: Refactor Streams and Connections to include Stat properties
I think maybe it'd be better if we broke all the properties of `.stat` out from the streams and connections and just had them on the stream/connection.  It's not clear (to me) what belongs in the `.stat` property and what belongs on the stream/connection.

	It'd make it simpler to extend the interfaces and we also wouldn't need to have deeply nested property accessors then either.

@wemeetagain what do you think?

_Originally posted by @achingbrain in https://github.com/libp2p/js-libp2p-interfaces/issues/265#issuecomment-1170195006_
            

Refactor the [Stream](https://github.com/libp2p/js-libp2p-interfaces/blob/22b3051dd51cb697b6cbac57e69ce97ef40e74f1/packages/interface-connection/src/index.ts#L73-L88) and [Connection stat](https://github.com/libp2p/js-libp2p-interfaces/blob/22b3051dd51cb697b6cbac57e69ce97ef40e74f1/packages/interface-connection/src/index.ts#L19-L43) properties should be at the top level for ease of extendability.  This would enforce that negotiated streams have a protocol for instance and avoid these unnecessary types of [non-nullish assertions](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/324c69356fa928244f571af20d673c6d8e1ba2d2/src/stream.ts#L37)

achingbrain:
Moves `direction`, `timeline`, `status`, etc from `.stat` objects on `Stream` and `Connection` objects onto the root object so we don't have awkward constructs like `conn.stat.status`, instead the simpler `conn.status`.

Fixes #1849

BREAKING CHANGE: `stream.stat.*` and `conn.stat.*` properties are now accessed via `stream.*` and `conn.*`


achingbrain:
Extracted from #1847

interop/test/ping.spec.ts
```diff
@@ -140,7 +140,7 @@ describe('ping test', () => {
           options.streamMuxers = [mplex()]
           break
         case 'yamux':
-          options.streamMuxers = [yamux()]
+          options.streamMuxers = [mplex()]
```
wemeetagain, line 143: I think this crept in

achingbrain: No, this breaks the Stream interface, which`@chainsafe/yamux` has a dependency on.  `@libp2p/mplex` here is updated to the new version but `@libp2p/interface` will need a release that `@chainsafe/yamux` can be upgraded to.

Groan.

I guess I could include it in this PR as `@libp2p/yamux` or something then we can swap out `@libp2p/yamux` for `@chainsafe/yamux` after the release?

wemeetagain: Oh darn. I guess this is the issue with relying on packages outside the monorepo when doing a breaking change.

At least you could make the diff smaller with somehting like:
```ts
// TODO replace this import once @chainsafe/libp2p-yamux has been updated
import { mlex as yamux } from '@libp2p/mplex'
```

achingbrain: I've temporarily added yamux to the monorepo but with `"private": true"` in the `package.json`, this should let us build and publish the other packages without trying to publish yamux, we can then open a PR to yamux afterwards with the changes and remove it from here.

maschad (commented):
I am bit confused, what necessitates the addition of `libp2p-daemon` and `libp2p-daemon-client`  and `libp2p-daemon-server` here?

achingbrain:
The daemon has a dependency on libp2p (e.g. it references the changed stream properties), but libp2p uses the daemon in it’s interop tests so 🫠

PR #1864, opened 2023-07-03
fix!: close streams gracefully

### LINKED ISSUE #656: Full stream close guarantee
`js-libp2p` should have an utility logic regarding full stream close, where we wait for an EOF and then reset the stream if it does not close quickly enough.

Some inspiration: https://github.com/libp2p/go-libp2p-core/blob/master/helpers/stream.go#L25-L56

### LINKED ISSUE #1793: Cleanly closing streams

achingbrain:
- Refactors `.close`, `closeRead` and `.closeWrite` methods on the `Stream` interface to be async
- The `Connection` interface now has `.close` and `.abort` methods
- `.close` on `Stream`s and `Connection`s wait for the internal message queues to empty before closing
- `.abort` on `Stream`s and `Connection`s close the underlying stream immediately and discards any unsent data
- `.reset` is removed from the `Stream` interface - instead call `.abort(err)` to signal a local error
- `.reset` is still present on the `AbstractStream` class - the muxer implementation should call this to signal a remote error
- `@chainsafe/libp2p-yamux` now uses the `AbstractStream` class from `@libp2p/interface` the same as `@libp2p/mplex` and `@libp2p/webrtc` - all the logic around the \*checks notes* 17 different ways to close a stream is contained there

Follow-up PRs will be necessary to `@chainsafe/libp2p-yamux`, `@chainsafe/libp2p-gossipsub` and `@chainsafe/libp2p-noise` though they will not block the release as their code is temporarily added to this repo to let CI run.

Fixes #1793
Fixes #656

BREAKING CHANGE: the `.close`, `closeRead` and `closeWrite` methods on the `Stream` interface are now asynchronous


interop/test/ping.spec.ts
```diff
@@ -38,8 +40,11 @@ describe('ping test', () => {
     const MUXER = process.env.muxer
     const IP = process.env.ip ?? '0.0.0.0'
 
-    const options: Libp2pOptions<{ ping: PingService, identify: DefaultIdentifyService }> = {
+    const options: Libp2pOptions<{ ping: PingService, identify: unknown }> = {
```
maschad, line 43: This has been [refactored](https://github.com/libp2p/js-libp2p/pull/1784/files#diff-024eabde3750f3bf282cd9a642bd16ba0357c8c0b3110ffa4c78bfdd60f9ed2fR27)

packages/interface-compliance-tests/src/mocks/connection.ts
```diff
@@ -76,24 +74,22 @@ class MockConnection implements Connection {
       throw new Error('protocols must have a length')
     }
 
-    if (this.status !== STATUS.OPEN) {
```
maschad, line 79: Why did we remove the string constant? I see this is re-used in multiple places

packages/interface-compliance-tests/src/connection/index.ts
```diff
@@ -89,7 +89,7 @@ export default (test: TestSetup<Connection>): void => {
         await connection.close()
 
         expect(connection.timeline.close).to.exist()
-        expect(connection.status).to.equal('CLOSED')
```
maschad, line 92: Please convert to string constants

maschad (changes requested):

maschad:
I assume we had to the encryption libraries here as well for a similar reason to https://github.com/libp2p/js-libp2p/pull/1856#issuecomment-1622374976 ?

packages/interface-compliance-tests/src/mocks/connection.ts
achingbrain: For consistency with types used elsewhere, e.g.

```ts
export type Direction = 'inbound' | 'outbound'
```

```ts
export type PeerIdType = 'RSA' | 'Ed25519' | 'secp256k1'
```

```ts
export type PendingDialStatus = 'queued' | 'active' | 'error' | 'success'
```

etc

interop/README.md
```diff
@@ -9,14 +9,84 @@
...
+    - [node.js](#nodejs)
+    - [Browsers](#browsers)
+  - [Build another libp2p implementation](#build-another-libp2p-implementation)
+  - [Running Redis](#running-redis)
+  - [Start libp2p](#start-libp2p)
 - [License](#license)
 - [Contribution](#contribution)
 
-## Install
+## Usage
```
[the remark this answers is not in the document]
achingbrain: I think the audiences are different.

The README in this repo is useful for people who want to run the tests with their local copy of js-libp2p+changes against release versions of other libp2p implementations and is what the usage examples have in mind.

The users of the test plan repo are a little less well defined.

packages/interface-compliance-tests/src/mocks/connection.ts
achingbrain: If we want to reduce duplication these should probably be enums rather than strings, but that can be done in a follow-up PR.

achingbrain:
> I assume we had to the encryption libraries here as well for a similar reason to https://github.com/libp2p/js-libp2p/pull/1856#issuecomment-1622374976 ?

Yes, that's correct - they'll be removed after release.

interop/README.md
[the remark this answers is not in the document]
maschad: My understanding is that if someone wants to test their js-libp2p changes against other libp2p implementation versions, they would follow a pattern similar to [the one utilized here](https://github.com/libp2p/test-plans/pull/178) and run the changes following the [README](https://github.com/libp2p/test-plans/tree/master/multidim-interop#interoperability-test)

interop/test/ping.spec.ts
achingbrain: Rebased 👍

interop/package.json
```diff
@@ -35,7 +35,8 @@
   "scripts": {
     "start": "node index.js",
     "build": "aegir build",
-    "test:interop:multidim": "aegir test"
+    "lint": "aegir lint",
+    "test:interop:multidim": "DEBUG=libp2p:ping* aegir test"
```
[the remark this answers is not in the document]
achingbrain: Removed 👍

interop/README.md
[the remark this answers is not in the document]
achingbrain: I did not find the documentation in the test plans repo to be sufficient to actually run the tests locally.

Perhaps we can consolidate these in a future PR.

PR #1877, opened 2023-07-12
Fix a property name of `websockets` option in README.md: `filters` => `filter`

bellbind:
The libp2p transport `websockets` option's valid property name would be `filter`; not `filters`.  

I struggled how to`libp2p.dial()` from in browser to helia on nodejs, added filter**s** option for libp2p in browser  as in example of README.
Today I tried to use `filter` instead of `filters`, it can dial from browser to nodejs...


maschad (approved):
Thanks for spotting this!

Issue #1777, opened 2023-05-29
Docs: Convert Configuration examples in docs to Typescript


maschad:
#### Description:

Once https://github.com/libp2p/js-libp2p/pull/1505 is landed we should ensure all configuration examples are in Typescript.  For instance the [basic setup](https://github.com/libp2p/js-libp2p/blob/master/doc/CONFIGURATION.md#basic-setup)

maschad:
Thanks for pointing this out @DougAnderson444 

Unfortunately during the [consolidation of the interface modules](https://github.com/libp2p/js-libp2p/pull/1833) some of the `README`s where removed without some of the details being consolidated into the new amalgamated _interface_  module's `README` that you referenced.

Mind you, I've edited the description of this issue as it iss actually meant to address the [js snippets being converted to Typescript in our configuration DOCS](https://github.com/libp2p/js-libp2p/blob/master/doc/CONFIGURATION.md?plain=1#L208) and not necessarily what is being mentioned here. 

I have created these two other issues to address the problems you've raised, so we can continue the discussions there, I think there is more triaging to be done. See below:

- Broken links https://github.com/libp2p/js-libp2p/issues/1880
- Update `README` s https://github.com/libp2p/js-libp2p/issues/1881

PR #1883, opened 2023-07-20
docs: add migration guide for libp2p@0.46.x

achingbrain:
Adds migration guide detailing breaking changes in libp2p@0.46.x


doc/migrations/v0.45-v0.46.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Table of Contents <!-- omit in toc -->
+
+- [Graceful stream closing](#graceful-stream-closing)
+- [Stream/Connection stat properties](#streamconnection-stat-properties)
+- [Interface module consolidation](#interface-module-consolidation)
+
+## Graceful stream closing
+
+Streams can either be closed gracefully, where we wait for any unsent data to be sent, or aborted in which case any unsent data is discarded and a reset message is sent, notifying the remote of the abnormal termination.
```
maschad, line 13: If a stream is aborted, and the `.reset` method has been removed, why is it that a _reset_ message is sent as opposed to an _abort_ message?

doc/migrations/v0.45-v0.46.md
```diff
@@ -0,0 +1,119 @@
...
+
+From `0.46.x` the `.close`/`.closeRead`/`.closeWrite` methods on the Stream interface are now asynchronous.  `.abort` is a synchronous method that accepts an Error object.
+
+Similarly the Connection interface now has asynchronous `.close` and synchronous `.abort` methods.
+
+The `.reset` method has been removed from the Stream interface as it is only to be invoked internally by stream multiplexers when a remote stream reset has occurred.
+
+**Before**
+
+```js
```
maschad, line 27: Going forward we should really utilize Typescript examples so that they can be compiled and thus reduce the probability of doc errors.

maschad (commented):
Should we also include the option to disable `runOnConnectionOpen` for identify queries that was added in https://github.com/libp2p/js-libp2p/pull/1784 ?

doc/migrations/v0.45-v0.46.md
achingbrain: There's no such thing as an abort message.

Calling `.abort` will cause a reset message to be sent to the remote.

Receiving a reset message from the remote will cause the stream to throw.

doc/migrations/v0.45-v0.46.md
achingbrain: Probably, though it'll involve a lot of boilerplate `import`s.

doc/migrations/v0.45-v0.46.md
maschad: That's good to know, whenever working with yamux I often wondered what was the distinction between the [two](https://github.com/ChainSafe/js-libp2p-yamux/blob/dc55dcf4972c4805cebea4c3a4c18eafc9d17adc/src/constants.ts#L26-L27) albeit the spec rightly only makes reference to a stream reset.

doc/migrations/v0.45-v0.46.md
maschad: True, but otherwise we lose the benefit of the type safety. Perhaps we can begin with https://github.com/libp2p/js-libp2p/issues/1777 as a precedent.

doc/migrations/v0.45-v0.46.md
```diff
@@ -0,0 +1,234 @@
+# Migrating to libp2p@46 <!-- omit in toc -->
+
+A migration guide for refactoring your application code from libp2p v__ to v__.
```
maschad, line 3:
```suggestion
A migration guide for refactoring your application code from libp2p `v0.45` to `v0.46`.
```

doc/migrations/v0.45-v0.46.md
```diff
@@ -0,0 +1,234 @@
...
+  - [Stream/Connection stat properties](#streamconnection-stat-properties)
+  - [Interface module consolidation](#interface-module-consolidation)
+
+## New features
+
+### Manual identify
+
+The [identify protocol](https://github.com/libp2p/specs/blob/master/identify/README.md) is used by libp2p to discover which protocols a remote peer supports. By default it runs on every connection after it opens.
+
+Applications interested in peers that support certain protocols can register [topology callbacks]() to be notified when network peers that support those protocols connect or disconnect.
```
maschad, line 21: Missing link on the topology callbacks

maschad (changes requested):
We should also enable `doc-check` on this doc if we are going to have TS snippets.

doc/migrations/v0.45-v0.46.md
achingbrain: Fixed.

PR #1885, opened 2023-07-20
test: ipv6 tests can be skipped if unsupported

SgtPooki:
Allows disabling of running ipv6 tests. On machines without ipv6 support, the tests hang forever. This means users without ipv6 (I suspect most CenturyLink ISP users) could not run tests before this change.

* Adds toggle for ipv6 tests based on `DISABLE_IPV6` env var.
* CONTRIBUTING.md is also being added to mention the scenario I ran into.

**Note:** When running the tests, and aegir times out on serverAddr6.promise, it should fail, but it doesn't.

replacement for https://github.com/libp2p/js-libp2p-webtransport/pull/88


SgtPooki:
CI is failing but doesn't seem related to my changes

Lmk if theres something i need to fix

achingbrain:
I've updated the branch - the multidim tests are failing with `Error: ENOSPC: no space left on device, write`, a more recent commit switches them to run on dedicated runners which have more disk space.

maschad (approved):

p-shahi:
@SgtPooki ready to merge?

achingbrain (approved):

PR #1890, opened 2023-07-25
feat: mark connections with limits as limited

achingbrain:
Some connections have resources limits imposed on them, such as circuit relay connections.  If these limits are breached the connection will be closed by the remote.

When this is the case, the connection will have a `.limited` boolean property set to true.

By default any attempt to run a protocol over a limited connection will throw (outgoing) or be reset (incoming), this is to prevent, for example, bitswap exceeding the connection transfer limit and causing the connection to be closed by the relay server when it should be reserved for the WebRTC SDP exchange to allow incoming dials.

Protocols can opt-in to being run over limited connections by specifying a flag during `libp2p.handle` (incoming) and `connection.openStream` (outgoing).


packages/interface/src/connection/index.ts
```diff
@@ -178,6 +178,12 @@ export interface NewStreamOptions extends AbortOptions {
    * for the protocol
    */
   maxOutboundStreams?: number
+
+  /**
+   * Opt-in to running over a limited connection - one that has time/data limits
+   * placed on it.
+   */
+  allowOnLimitedConnection?: boolean
```
maschad, line 186: I think for consistency we should the terminology [transient connections](https://github.com/libp2p/go-libp2p/blob/260b9695cafdd8e35ec65b30ef153f0c15549c72/core/network/network.go#L115) to be consistent with other implementations

packages/libp2p/src/circuit-relay/transport/index.ts
```diff
@@ -380,8 +382,10 @@ class CircuitRelayTransport implements Transport {
       localAddr
     })
 
-    log('new inbound connection %s', maConn.remoteAddr)
-    await this.upgrader.upgradeInbound(maConn)
+    log('new inbound transient connection %s', maConn.remoteAddr)
```
maschad, line 385: we have a specific debug formatter for multiaddrs

```suggestion
    log('new inbound transient connection %a', maConn.remoteAddr)
```

maschad (changes requested):

packages/interface/src/connection/index.ts
achingbrain: I agree in principal, but the only [documentation](https://docs.libp2p.io/concepts/security/dos-mitigation/#transient-connections) I could find about transient connections in libp2p says:

1. A transient connection is a connection that has yet to have encryption or a muxer negotiated
2. go-libp2p calls this "transient", rust-libp2p calls this "negotiating"

So there doesn't seem to be much consistency across implementations and we're dealing with connections with data/time limits here which is different, hence the different name.

packages/libp2p/src/circuit-relay/transport/index.ts
achingbrain: Done

packages/interface/src/connection/index.ts
achingbrain: Clarification on slack: data/time limited connections are "transient" in go-libp2p, but the resource manager also has a "transient" scope for newly opened streams that have yet to have encryption or a muxer negotiated.

🫠

achingbrain: Renamed.

sudiptab2100:
@maschad 
Do you have any documentation for it? I'm confused how to setup the relay node and behind NAT nodes for this.

maschad:
Hey @sudiptab2100 I'm not sure exactly you're referring to. If you'd like more info on setting up a Relay node you can take a look [here](https://github.com/libp2p/js-libp2p/blob/main/doc/CONFIGURATION.md#setup-with-relay)

For more background you may want to read [this](ipns://docs.libp2p.io/concepts/nat/circuit-relay/)

If you are still having issues feel free to open a [discussion](https://github.com/libp2p/js-libp2p/discussions) as this PR isn't the best place to discuss

PR #1900, opened 2023-07-26
deps: upgrade aegir to 40.x.x

achingbrain:
Upgrades aegir and fixes linting errors and typedoc config.


packages/interface-internal/package.json
```diff
@@ -60,10 +60,6 @@
       "types": "./dist/src/registrar/index.d.ts",
       "import": "./dist/src/registrar/index.js"
     },
-    "./transport": {
```
maschad, line 63: Why are these types no longer exported?

interop/BrowserDockerfile
```diff
@@ -4,10 +4,11 @@ FROM mcr.microsoft.com/playwright
 
 COPY --from=node-js-libp2p-head /app/ /app/
 WORKDIR /app/interop
-RUN npx playwright install
```
maschad, line 7: Why is it we no longer need to install the playwright deps?

achingbrain: playwright-test does it automatically if needed

packages/interface-internal/package.json
achingbrain: These files don’t exist in this module

PR #1902, opened 2023-07-26
fix: ignore peers with invalid multiaddrs 

### LINKED ISSUE #1873: `'peer:discovery'` event does not always emit a valid `PeerInfo`
- **Version**:

0.45.9

- **Subsystem**:

libp2p

#### Severity:
low/medium

#### Description:
It seems that `'peer:discovery'` does not always emit a full [`PeerInfo`](https://github.com/libp2p/js-libp2p/blob/master/packages/interface/src/peer-info/index.ts).
We've seen `multiaddrs` being `undefined`, even though the type is `Multiaddr[]`.

See https://github.com/ChainSafe/lodestar/pull/5736

maschad:
[Looking through the codepath](https://github.com/libp2p/js-libp2p/blob/master/packages/peer-store/src/index.ts#L212) I don't really see a scenario where this code occur unless a consumer intentionally omits a _multiaddr_ , I also see that we already [have tests for this type of scenario](https://github.com/libp2p/js-libp2p/blob/master/packages/interface-compliance-tests/src/peer-discovery/index.ts#L41-L60)

Closes #1873


packages/libp2p/src/libp2p.ts
```diff
@@ -98,8 +98,19 @@ export class Libp2pNode<T extends ServiceMap = Record<string, unknown>> extends
 
     components.events.addEventListener('peer:update', evt => {
       // if there was no peer previously in the peer store this is a new peer
-      if (evt.detail.previous == null) {
-        this.safeDispatchEvent('peer:discovery', { detail: evt.detail.peer })
+      if (evt.detail.previous === null) {
+
+        if(evt.detail.peer.addresses.length === 0) {
+          throw new CodeError('peer must have at least one multiaddr', codes.ERR_INVALID_MULTIADDR)
```
achingbrain, line 104: This will cause an `uncaughtException` event to be emitted on the `process` object and node will crash.

If `peer.addresses` is empty it should just translate to `peerInfo.multiaddrs` being empty.

achingbrain (changes requested):

packages/libp2p/src/libp2p.ts
maschad: Why would a peer emit an update event if it has no have no addresses though? Seems a bit counter-intuitive to a discover a peer with no dial-able addresses.

achingbrain: `peer:update` happens when a peer's info is updated in the peer store.  This can be in response to any field, not just multiaddrs.

Regardless, the event callback invocation is not wrapped in a try/catch so we cannot throw here.

achingbrain: I think the problem here is that the `detail.peer` field in `peer:update` is a [Peer](https://github.com/libp2p/js-libp2p/blob/master/packages/interface/src/peer-store/index.ts#L22-L53) instance, but the detail field of `peer:discovery` should be a [PeerInfo](https://github.com/libp2p/js-libp2p/blob/master/packages/interface/src/peer-info/index.ts#L4-L8) - `Peer` doesn't have a `.multiaddrs` field so the conversion in this PR is necessary for the emitted value to conform to the `PeerInfo` interface.

packages/libp2p/src/libp2p.ts
```diff
@@ -98,8 +98,19 @@ export class Libp2pNode<T extends ServiceMap = Record<string, unknown>> extends
...
+        }
+
+        const peerInfo: PeerInfo = {
+          id: evt.detail.peer.id,
+          multiaddrs: evt.detail.peer.addresses.map(a => a.multiaddr),
+          protocols: evt.detail.peer.protocols
+        }
+
+        this.safeDispatchEvent('peer:discovery', { detail: peerInfo})
       }
```
achingbrain, line 114:
```suggestion
      if (evt.detail.previous === null) {
        const peerInfo: PeerInfo = {
          id: evt.detail.peer.id,
          multiaddrs: evt.detail.peer.addresses.map(a => a.multiaddr),
          protocols: evt.detail.peer.protocols
        }

        this.safeDispatchEvent('peer:discovery', { detail: peerInfo})
      }
```

packages/libp2p/src/libp2p.ts
```diff
@@ -98,8 +98,14 @@ export class Libp2pNode<T extends ServiceMap = Record<string, unknown>> extends
 
     components.events.addEventListener('peer:update', evt => {
       // if there was no peer previously in the peer store this is a new peer
-      if (evt.detail.previous == null) {
-        this.safeDispatchEvent('peer:discovery', { detail: evt.detail.peer })
+      if (evt.detail.previous === null) {
```
achingbrain, line 101:
```suggestion
      if (evt.detail.previous == null) {
```

maschad: the emphasis on the double equal is because it's is it possible this value could also be `undefined` ?

achingbrain: Yes, in this case == equals null or undefined, === equals only null.

achingbrain (approved):

PR #1845, opened 2023-06-21
test: temporarily remove webRTC tests 

maschad:
Following the discussion [here](https://github.com/libp2p/js-libp2p/issues/1809#issuecomment-1599357723) we need to investigate why webRTC tests are inconsistent. 

Closes #1809

PR #1905, opened 2023-07-28
feat: add node.js/electron support for webrtc transport

achingbrain:
Uses [node-datachannel](https://www.npmjs.com/package/node-datachannel) to add support for node.js and electron (main process) to @libp2p/webrtc.

Supports:

- Dialing private-to-private
- Listening private-to-private
- Dialing private-to-public

Does not support:

- Listening private-to-public

### DIFF
--- a/packages/transport-webrtc/src/private-to-private/handler.ts
+++ b/packages/transport-webrtc/src/private-to-private/handler.ts
@@ -1,3 +1,4 @@
+import { CodeError } from '@libp2p/interface/errors'
 import { logger } from '@libp2p/logger'
 import { abortableDuplex } from 'abortable-iterator'
 import { pbStream } from 'it-protobuf-stream'
@@ -27,7 +28,9 @@ export async function handleIncomingStream ({ rtcConfiguration, dataChannelOptio
     const connectedPromise: DeferredPromise<void> = pDefer()
     const answerSentPromise: DeferredPromise<void> = pDefer()
 
-    signal.onabort = () => { connectedPromise.reject() }
+    signal.onabort = () => {
+      connectedPromise.reject(new CodeError('Timed out while trying to connect', 'ERR_TIMEOUT'))
+    }
     // candidate callbacks
     pc.onicecandidate = ({ candidate }) => {
       answerSentPromise.promise.then(
@@ -39,6 +42,7 @@ export async function handleIncomingStream ({ rtcConfiguration, dataChannelOptio
         },
         (err) => {
           log.error('cannot set candidate since sending answer failed', err)
+          connectedPromise.reject(err)
         }
       )
     }
--- a/packages/transport-webrtc/src/private-to-private/handler.ts
+++ b/packages/transport-webrtc/src/private-to-private/handler.ts
@@ -3,6 +3,7 @@ import { abortableDuplex } from 'abortable-iterator'
 import { pbStream } from 'it-protobuf-stream'
 import pDefer, { type DeferredPromise } from 'p-defer'
 import { DataChannelMuxerFactory } from '../muxer.js'
+import { RTCPeerConnection, RTCSessionDescription } from '../webrtc/index.js'
 import { Message } from './pb/message.js'
 import { readCandidatesUntilConnected, resolveOnConnected } from './util.js'
 import type { DataChannelOpts } from '../stream.js'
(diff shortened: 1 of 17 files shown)


packages/transport-webrtc/src/webrtc/rtc-peer-connection.ts
```diff
@@ -0,0 +1,306 @@
...
+        const urls = (Array.isArray(server.urls) ? server.urls : [server.urls]).map(str => new URL(str))
+
+        return urls.map(url => {
+          /** @type {import('../lib/index.js').IceServer} */
+          const iceServer = {
+            hostname: url.hostname,
+            port: parseInt(url.port, 10),
+            username: server.username,
+            password: server.credential
+            // relayType - how to specify?
```
maschad, line 53: Looks like the [Relay type](https://github.com/murat-dogan/node-datachannel/blob/9e35a2040a569e61ba98231cd37e9232b95bf656/API.md?plain=1#L24-L28) is an enum specified by a string?

packages/transport-webrtc/src/webrtc/rtc-peer-connection.ts
```diff
@@ -0,0 +1,306 @@
...
+          }
+
+          return iceServer
+        })
+      })
```
maschad, line 58: Should we also allow for configuration of the other [RTCPeerConnection params](https://github.com/murat-dogan/node-datachannel/blob/9e35a2040a569e61ba98231cd37e9232b95bf656/API.md?plain=1#L13-L21) ?

packages/transport-webrtc/src/webrtc/rtc-peer-connection.ts
```diff
@@ -0,0 +1,306 @@
...
+        .flat(),
+      iceTransportPolicy: init?.iceTransportPolicy
+    })
+
+    this.#peerConnection.onStateChange(() => {
+      this.dispatchEvent(new Event('connectionstatechange'))
+    })
+    // https://github.com/murat-dogan/node-datachannel/pull/171
```
maschad, line 66: This should be in the next release 👍🏾  good job.

packages/transport-webrtc/src/webrtc/rtc-data-channel.ts
```diff
@@ -0,0 +1,142 @@
...
+    })
+
+    this.onbufferedamountlow = null
+    this.onclose = null
+    this.onclosing = null
+    this.onerror = null
+    this.onmessage = null
+    this.onopen = null
+
+    // TODO: update these
```
maschad, line 88: Missing the setters for these.

maschad (changes requested):
Great work on this! I left some comments.

packages/transport-webrtc/src/webrtc/rtc-data-channel.ts
```diff
@@ -0,0 +1,142 @@
...
+  #readyState: RTCDataChannelState
+
+  constructor (dataChannel: node.DataChannel) {
+    super()
+
+    this.#dataChannel = dataChannel
+    this.#readyState = 'connecting'
+
+    // TODO: what is the default?
+    this.#bufferedAmountLowThreshold = 0
```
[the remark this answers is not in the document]
achingbrain: It can be removed I think. It's not in the spec and we don't set it anywhere.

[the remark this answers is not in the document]
achingbrain: Because the spec lets you read the `bufferedAmountLowThreshold` property on a `RTCDataChannel` but the [DataChannel](https://github.com/murat-dogan/node-datachannel/blob/polyfill/lib/index.d.ts#L189-L205) object from `node-datachannel` only lets you set the value, not get it, so we cache it on the js side.

packages/transport-webrtc/src/webrtc/rtc-peer-connection.ts
achingbrain: Yes, but the `RTCIceServer` type from the standard doesn't have a `relayType` field so there's nothing to map this from.

packages/transport-webrtc/src/webrtc/rtc-data-channel.ts
achingbrain: These are readonly properties according to the spec.

packages/transport-webrtc/src/webrtc/rtc-peer-connection.ts
achingbrain: Probably, where we can map values from the spec config type.

achingbrain: I think all the values that the `PeerConnection` type from `node-datachannel` supports that have corresponding values in the `RTCConfigutaion` type from the spec are already mapped.

Any further params would need PRs opening against `node-datachannel`.

packages/transport-webrtc/src/webrtc/rtc-peer-connection.ts
achingbrain: Not going to block on this, since we don't use it - will uncomment this code when the next `node-datachannel` release happens..

packages/transport-webrtc/src/private-to-private/handler.ts
```diff
@@ -93,56 +100,63 @@ export async function initiateConnection ({ rtcConfiguration, dataChannelOptions
...
-    throw new Error('Failed to set localDescription')
-  })
-
-  // read answer
-  const answerMessage = await stream.read()
-  if (answerMessage.type !== Message.Type.SDP_ANSWER) {
-    throw new Error('remote should send an SDP answer')
-  }
+        .catch(err => {
+          log.error('error sending ICE candidate', err)
```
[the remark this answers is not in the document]
achingbrain: Probably, yes, although just logging the error is what the code did before this PR so maybe as a follow up.

packages/transport-webrtc/src/webrtc/rtc-data-channel.ts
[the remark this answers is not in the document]
achingbrain: Also, turns out the spec says the default is 0..

BigLep:
@achingbrain : Thanks!  Will there be a corresponding PR to add this to test-plans/interop?

BigLep:
And to https://libp2p.io/implementations/#transports ?

maschad:
> @achingbrain : Thanks! Will there be a corresponding PR to add this to test-plans/interop?

Currently we have [removed the webRTC tests from interop](https://github.com/libp2p/js-libp2p/pull/1845) due to https://github.com/libp2p/test-plans/issues/235 - which I will be resuming work on this week. Once that is resolved we can add this to suite and re-introduce the tests.

maschad (approved):
LGTM! There's some follow up work to be done given that https://github.com/murat-dogan/node-datachannel/pull/171 is merged, once `v0.4.4` is released.

p-shahi:
> And to https://libp2p.io/implementations/#transports ?

Added in https://github.com/libp2p/website/pull/180 @BigLep

AustinFoss:
I came across this while trying to figure out why my nodejs node wasn't advertising a webRTCDirect listening multiaddr. Spent too much time before learning webRTC wasn't supported in node, just the browser so far.

But I'm curious about your statements "Listening private-to-private" & "Listening private-to-public", does this mean a public node can't listen on `"/ip4/0.0.0.0/udp/0/webrtc-direct"` even with this branch's solution?

achingbrain:
No, this PR only adds support for private WebRTC addresses (e.g. the SDP handshake needs to take place via a relay server).

It should be possible to support this in future, I've opened a [draft PR here](https://github.com/libp2p/js-libp2p/pull/2027) but we need some features adding to the underlying libraries.

Sotatek-HocNguyena:
@achingbrain Does it support private (nodejs) - private (browser)? I'm trying but can't connect webrtc between nodejs (private) and browser through relay

PR #1910, opened 2023-07-31
docs: remove outdated ping README

### LINKED ISSUE #1590: Ping module has no method `mount`
- The example: https://github.com/libp2p/js-libp2p/tree/master/src/ping#usage the method `mount` is used, but upon inspection of the class, a method `mount` isn't found. Is this readme still accurate?

tabcat:
Closes #1590


maschad (changes requested):
I know that there was suggestion to [delete this](https://github.com/libp2p/js-libp2p/issues/1590#issuecomment-1463775673) but I think it may useful to have description (similar to [identify](https://github.com/libp2p/js-libp2p/blob/master/packages/libp2p/src/identify/README.md#description)) and for usage, have the API docs generated via our `docs` command.

cc @achingbrain

maschad (approved):
After discussing with @achingbrain it makes more sense to have this [as a top-level doc](https://github.com/libp2p/js-libp2p/blob/master/doc/CONFIGURATION.md) with it's own subsection for describing the service, with the associated generated API docs.

achingbrain (approved):

PR #1924, opened 2023-08-01
docs: removes redundant setup step

UsersHaveNames:
removes redundant 'npm run build' command

I have tested the guides and npm run build is not only redundant for this POC but also might make someone think that they are doing things wrong.

### DIFF
--- a/examples/chat/README.md
+++ b/examples/chat/README.md
@@ -16,7 +16,7 @@
 
 ## Setup
 
-1. Install the modules in the libp2p root directory, `npm install` and `npm run build`.
+1. Install the modules in the libp2p root directory, `npm install`.
 2. Open 2 terminal windows in the `./examples/chat/src` directory.
 
 ## Running


achingbrain:
Are you sure you only ran `npm install` in the root when you tested this?

If you've just cloned the repo and run `npm install` in the root, symlinks will be set up for dependencies like [this one](https://github.com/libp2p/js-libp2p/blob/master/examples/chat/package.json#L43) since they are monorepo siblings - they are all typescript so need to be built before the example can run.

If you've run `npm install` in the example directory OTOH you'll get the deps from npm and the extra build step isn't necessary.

UsersHaveNames:
> Are you sure you only ran `npm install` in the root when you tested this?
> 
> If you've just cloned the repo and run `npm install` in the root, symlinks will be set up for dependencies like [this one](https://github.com/libp2p/js-libp2p/blob/master/examples/chat/package.json#L43) since they are monorepo siblings - they are all typescript so need to be built before the example can run.
> 
> If you've run `npm install` in the example directory OTOH you'll get the deps from npm and the extra build step isn't necessary.

Ah you are correct :)
I downloaded that Chat application part only as a standalone thing and ran it in the example directory.

As a user I do prefer to look at examples as fully self-contained, just a thought that might be useful in the future. Especially since libp2p seems to have a reputation for modularity.

Closing the PR since the intent seems to be that examples work as part of entire repo, let me know if you changed your mind.

UsersHaveNames:
> Are you sure you only ran `npm install` in the root when you tested this?
> 
> If you've just cloned the repo and run `npm install` in the root, symlinks will be set up for dependencies like [this one](https://github.com/libp2p/js-libp2p/blob/master/examples/chat/package.json#L43) since they are monorepo siblings - they are all typescript so need to be built before the example can run.
> 
> If you've run `npm install` in the example directory OTOH you'll get the deps from npm and the extra build step isn't necessary.

Ah you are correct :)
I downloaded that Chat application part only as a standalone thing and ran it in the example directory.

As a user I do prefer to look at examples as fully self-contained, just a thought that might be useful in the future. Especially since libp2p seems to have a reputation for modularity.

Closing the PR since the intent seems to be that examples work as part of entire repo, let me know if you changed your mind, Ill be happy to help with this and other examples.

PR #1931, opened 2023-08-03
fix: crypto: limit RSA key size to <= 8192 bits

MarcoPolo:
Restrict the RSA key sizes we expect from peers. Protects us from spending a lot of compute on verifying signatures from big keys. Similar to https://github.com/libp2p/go-libp2p/pull/2454


achingbrain (approved):

achingbrain:
I made a few changes here:

- Make max key size a const to prevent it being accidentally overwritten
- Export max key size for reuse
- Add a fixture with a key bigger than max key size so we don't need to override the max key size at runtime
- Remove ability to override max key size (since if it's exported it becomes part of the documented API)
- Ensure we check the max key size before doing any async work
- Use uint8arrays/from-string to decode base64url instead of rolling our own version
- Minor edits to test names to make them consistent with the others

PR #1937, opened 2023-08-05
feat: added bigint multiformatting

maschad:
Related: https://github.com/libp2p/js-libp2p-logger/pull/37


achingbrain:
I'm not convinced this is a good idea.  I think we lose the ability to see that it's a bigint that's being logged, not a number and we lost the ability to simply copy/paste the value, we'd have to remove the formatting this PR adds.

Going to close this for now.

PR #1939, opened 2023-08-07
deps: Upgrade to `@noble/curves`

### LINKED ISSUE #1938: Security improvement: update crypto dependencies
crypto has following deps:

    "@noble/ed25519": "^1.6.0",
    "@noble/secp256k1": "^1.5.4",

They can be replaced with a consolidated [@noble/curves](https://github.com/paulmillr/noble-curves), which had an audit by Trail of Bits in Q1. That way, your bundle would use less code. See [Upgrading to curves](https://github.com/paulmillr/noble-curves#upgrading) - should be a 5-min job.

I would have submitted PR on my own, but there are build errors: after git clone, root `npm install` and `cd packages/crypto && npm install`, `npm build` gives an error: https://gist.github.com/paulmillr/8561cb26714853d1643c3087575d9bd0

maschad:
Close #1938

### DIFF
--- a/packages/crypto/benchmark/ed25519/package.json
+++ b/packages/crypto/benchmark/ed25519/package.json
@@ -9,7 +9,6 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@noble/ed25519": "^1.3.0",
     "@stablelib/ed25519": "^1.0.2",
     "benchmark": "^2.1.4",
     "ed25519": "^0.0.5",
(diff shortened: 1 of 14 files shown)


achingbrain (approved):
LGTM.

I think we now have a load of async functions that don't do any async work - we should remove the async keyword from them to prevent promises being created unecessarily but that can be done in a follow-up.

Issue #1899, opened 2023-07-25
feat: autoDialer backoff implementation


SgtPooki:
autoDialer should support configurable exponential backoff, or retry policies, for multiaddr that have consistently failed to establish a connection:

related to 
* https://github.com/libp2p/js-libp2p/issues/744#issuecomment-1650555616
* https://github.com/libp2p/js-libp2p/issues/1897
* https://github.com/libp2p/js-libp2p/issues/1898

maschad:
Discussion from Open Maintainers call on 1/08/2023

Potential solution:

Penalize peers / multiaddrs via decreasing their [tag value](https://github.com/libp2p/js-libp2p/blob/b3c87b6dc14192ad4fe24b51738b4867a2177d1d/packages/interface/src/peer-store/index.ts#L116-L121) if dials are [failing / rejected dials](https://github.com/libp2p/js-libp2p/blob/b3c87b6dc14192ad4fe24b51738b4867a2177d1d/packages/libp2p/src/connection-manager/dial-queue.ts#L500) so that they are [moved lower in the dial-queue](https://github.com/libp2p/js-libp2p/blob/b3c87b6dc14192ad4fe24b51738b4867a2177d1d/packages/libp2p/src/connection-manager/auto-dial.ts#L179-L209).

cc @achingbrain

PR #1943, opened 2023-08-08
feat(libp2p): add autodial retry threshold config option

achingbrain:
When auto-dialing peers, if we have failed to dial them recently there's little point in redialing them as it's probably just going to fail again which consumes a slot in the dial queue and other resources.

Adds a `autoDialPeerRetryThreshold` config key to the connection manager which is a value in ms.

If we have attempted to dial a peer and that dial attempt failed but we are under our min connection count, do not auto dial the peer within the retry threshold.

Defaults to 1 minute.

Closes https://github.com/libp2p/js-libp2p/issues/1899


packages/libp2p/src/connection-manager/dial-queue.ts
```diff
@@ -230,9 +232,22 @@ export class DialQueue {
...
         signal.clear()
       })
-      .catch(err => {
+      .catch(async err => {
         log.error('dial failed to %s', pendingDial.multiaddrs.map(ma => ma.toString()).join(', '), err)
 
+        if (peerId != null) {
+          // record the last failed dial
+          try {
+            await this.peerStore.patch(peerId, {
```
maschad, line 241: Should this peer also be tagged with a lower value? Let's say the node discovers a new Peer that it has not yet dialled, and this _bad_ peer's threshold has just expired (meaning it's available for a re-dial), shouldn't the newly discovered peer be given a preference in the dial-queue?

packages/libp2p/test/connection-manager/auto-dial.spec.ts
```diff
@@ -224,4 +226,56 @@ describe('auto-dial', () => {
...
+
+    await pWaitFor(() => {
+      return connectionManager.openConnection.callCount === 1
+    })
+    await delay(1000)
+
+    expect(connectionManager.openConnection.callCount).to.equal(1)
+    expect(connectionManager.openConnection.calledWith(matchPeerId(peerWithAddress.id))).to.be.true()
+    expect(connectionManager.openConnection.calledWith(matchPeerId(undialablePeer.id))).to.be.false()
+  })
```
maschad, line 280: I think we should add another test to ensure that the _undialable_ peer is eventually dialled once the threshold has expired.

packages/libp2p/test/connection-manager/auto-dial.spec.ts
```diff
@@ -224,4 +226,56 @@ describe('auto-dial', () => {
...
+      minConnections: 10,
+      autoDialInterval: 10000
+    })
+    autoDialler.start()
+    void autoDialler.autoDial()
+
+    await pWaitFor(() => {
+      return connectionManager.openConnection.callCount === 1
+    })
+    await delay(1000)
```
maschad, line 275: If the `autoDialInterval` is set `10_000` then we wouldn't expect another peer to be dialled after `1000` regardless of the `autoDialPeerRetryThreshold` , correct?

maschad (changes requested):
Looks promising, I have a few comments.

packages/libp2p/test/connection-manager/auto-dial.spec.ts
achingbrain: Yeah, that should be the retry threshold being set, not the auto dial interval..

packages/libp2p/src/connection-manager/dial-queue.ts
maschad: I agree but also I cannot foresee a scenario where a component would want to prioritize dialling a peer that is currently unreachable, so it should at least be penalized.

Perhaps if there is exists `LAST_DIAL_FAILURE_KEY` in that `peerStore`, it should be sorted below a `peerStore` that doesn't have one (even if the threshold value for the key is expired).

achingbrain: > so it should at least be penalized

The problem is the connection manager doesn't know what it means to penalise a peer, because it doesn't know what the value of a tag added by one component means compared to a tag added by another so it cannot alter the tag values without possibly compromising the functioning of the components doing the tagging.

That's why it sets a metadata value to just not dial the peer if it's failed recently.

maschad: > The problem is the connection manager doesn't know what it means to penalise a peer, because it doesn't know what the value of a tag added by one component means compared to a tag added by another so it cannot alter the tag values without possibly compromising the functioning of the components doing the tagging.

That's why I suggested that we augment the sorting to consider if that key has been added previously so that those peers are de-prioritized.

```typescript
const sortedPeers = shuffledPeers.sort((a, b) => {
      const peerAValue = peerValues.get(a.id) ?? 0
      const peerBValue = peerValues.get(b.id) ?? 0
      
      const peerAhasFailedRecently = a.metadata.has(LAST_DIAL_FAILURE_KEY)
      const  peerBhasFailedRecently = b.metadata.has(LAST_DIAL_FAILURE_KEY)
      
      if (peerAValue > peerBValue && !peerAhasFailedRecently) {
        return -1
      }
      
      if (peerAValue < peerBValue && !peerBhasFailedRecently) {
        return 1
      }
      return 0
})
```

Even after all tag values have been added, it's still _less likely_ that a successful dial would result from a peer that was dialled recently that and failed, than from a newly discovered peer which has not been dialled yet - This assumption is based on the premise that if a peer recently failed to establish a connection, there might be underlying issues causing the failure that haven't been resolved yet (may take longer than 10 minutes) as opposed to when you discover a peer through mechanisms like peer discovery protocols or peer routing, it _should_ indicate that the peer is reachable and active within the network.

achingbrain: Ah, right - when you said "Should this peer also be tagged with a lower value?" I thought you meant changing tag values.

The change to the sort function seems reasonable, but I think if you have a small number of peers to choose from (such as during initial startup when you only have bootstrap nodes), you might end up dialling peers you've just failed to dial?  This is expensive and is what this PR is attempting to address.

We might need the opposite though - if we've successfully dialled the peer recently, but we are no longer connected to them, to make sure they bubble up to the top of the autodial queue?

> when you discover a peer through mechanisms like peer discovery protocols or peer routing, it should indicate that the peer is reachable and active within the network.

I don't think this is something we can assume - we might discover peers via a DHT query for example - all that means is they were in the routing tables of another peer, not that they are diallable or even online.

maschad: > I think if you have a small number of peers to choose from (such as during initial startup when you only have bootstrap nodes), you might end up dialling peers you've just failed to dial? This is expensive and is what this PR is attempting to address.

That's reasonable. I think we can discuss this more on a separate issue.

packages/libp2p/src/connection-manager/index.ts
```diff
@@ -66,6 +66,12 @@ export interface ConnectionManagerInit {
    */
   autoDialMaxQueueLength?: number
 
+  /**
+   * When we've failed to dial a peer, do not autodial them again within this
+   * number of ms. (default: 10 minutes)
+   */
+  autoDialPeerRetryThreshold?: number
```
SgtPooki, line 73: honestly, even 1 minute would significantly help from my testing experience.

Once we add in a badPeer/bad multiaddr, it will free up time to check others instead of constantly spamming the same addresses.

packages/libp2p/src/connection-manager/auto-dial.ts
```diff
@@ -207,9 +212,26 @@ export class AutoDial implements Startable {
...
+      }
+
+      const when = parseInt(uint8ArrayToString(lastDialFailure))
+
+      if (isNaN(when)) {
+        return true
+      }
+
+      // only dial if the time since the last failure is above the retry threshold
+      return Date.now() - when > this.autoDialPeerRetryThresholdMs
```
SgtPooki, line 229:
```suggestion
      const lastDialFailuretimestamp = parseInt(uint8ArrayToString(lastDialFailure))

      if (isNaN(lastDialFailuretimestamp)) {
        return true
      }

      // only dial if the time since the last failure is above the retry threshold
      return Date.now() - lastDialFailuretimestamp > this.autoDialPeerRetryThresholdMs
```

packages/libp2p/src/connection-manager/constants.ts
```diff
@@ -53,6 +53,11 @@ export const AUTO_DIAL_PRIORITY = 0
  */
 export const AUTO_DIAL_MAX_QUEUE_LENGTH = 100
 
+/**
+ * @see https://libp2p.github.io/js-libp2p/interfaces/libp2p.index.unknown.ConnectionManagerInit.html#autoDialPeerRetryThreshold
+ */
+export const AUTO_DIAL_PEER_RETRY_THRESHOLD = 1000 * 60 * 10
```
SgtPooki, line 59: I think we could safely lower this and still achieve the goals of not spamming likely-to-fail multiaddrs, while also not blocking users with a 10 minute wait (or forcing a refresh in the browser)

1-2 minutes would be a significant improvement and minor inconvenience in the case of internet outages.

A good approach could be determining how long it takes(D) to dial typical size of peers(N) and setting the value to some ratio of that number. If we expect users to typically find 100-200 peers per session (in the browser that's likely), and it takes 5 minutes to dial them all, then 5 minutes should be default.. but I doubt it needs to be that high.

packages/libp2p/src/connection-manager/constants.ts
```diff
@@ -62,3 +67,13 @@ export const INBOUND_CONNECTION_THRESHOLD = 5
  * @see https://libp2p.github.io/js-libp2p/interfaces/index._internal_.ConnectionManagerConfig.html#maxIncomingPendingConnections
  */
 export const MAX_INCOMING_PENDING_CONNECTIONS = 10
+
+/**
+ * Store as part of the peer store metadata for a given peer, the value for this
+ * key is a stringified number representing the timestamp of the last time a
```
SgtPooki, line 73:
```suggestion
 * key is a timestamp representing the last time a
```

packages/libp2p/src/connection-manager/constants.ts
```diff
@@ -62,3 +67,13 @@ export const INBOUND_CONNECTION_THRESHOLD = 5
...
+ * dial attempted failed with the relevant peer.
+ *
+ * Used to insure we do not endlessly try to auto dial peers we have recently
+ * failed to dial.
```
SgtPooki, line 77: nit, but also a learning op on semantics for me: are the dial attempts failures, or are the connections failures? 

```suggestion
 * Used to insure we do not endlessly try to auto dial peers we have recently
 * failed to connect to.
```

packages/libp2p/src/connection-manager/dial-queue.ts
```diff
@@ -230,9 +232,22 @@ export class DialQueue {
         // clean up abort signals/controllers
         signal.clear()
       })
-      .catch(err => {
+      .catch(async err => {
```
SgtPooki, line 235: if i'm not mistaken, this dial-queue is also used for direct dials, should we discern between an auto-dial triggered dial vs a manually/user requested dial?

packages/libp2p/test/connection-manager/auto-dial.spec.ts
```diff
@@ -224,4 +226,56 @@ describe('auto-dial', () => {
...
+    autoDialler = new AutoDial({
+      peerStore,
+      connectionManager,
+      events
+    }, {
+      minConnections: 10,
+      autoDialInterval: 10000
+    })
+    autoDialler.start()
+    void autoDialler.autoDial()
```
SgtPooki, line 270: See my other comment as well, but do we need all the test scaffolding for auto-dialer or can we just attempt to dial the peer directly?

packages/libp2p/src/connection-manager/dial-queue.ts
```diff
@@ -230,9 +232,22 @@ export class DialQueue {
...
         log.error('dial failed to %s', pendingDial.multiaddrs.map(ma => ma.toString()).join(', '), err)
 
+        if (peerId != null) {
```
SgtPooki, line 238:
```suggestion
        if (peerId != null && priority < someVal) {
```

to address manual dials? or do we want a specific dial option to flag whether we will update the LAST_DIAL_FAILURE_KEY for a peer that we only use with auto-dialer?

SgtPooki (changes requested):
some comments / questions. Action items i would love to see addressed prior to merging:

1. lower the default value
2. don't set LAST_DIAL_FAILED if not an auto-dial dial

packages/libp2p/test/connection-manager/auto-dial.spec.ts
SgtPooki: regardless of the answer here, we may want another test regarding direct-dials covering the expected cases (dial them again even if failed auto dial previously, because it's a direct dial request)

packages/libp2p/src/connection-manager/constants.ts
achingbrain: An attempt to open a connection is a "dial"?

achingbrain:
> don't set LAST_DIAL_FAILED if not an auto-dial dial

Any reason why?  The value is only used by the auto-dialer to exclude peers from the list to dial, direct dials don't use this.

packages/libp2p/test/connection-manager/auto-dial.spec.ts
achingbrain: The last-failed-dial key is only used by the auto-dialer to exclude those peers from autodialing, it's not used by explicit dials.

maschad (approved):
Looks good, I would add another test to ensure that the _undiallable_ peer is eventually dialled once the threshold has expired, but my opinion is strong on that.

maschad:
> I added this check into the existing test

Ah, [I see now](https://github.com/libp2p/js-libp2p/pull/1943/files#diff-4e316fbf927d458976e70ce5db11c866e957eea55fe88cf84b26dac4f58683d8R282-R292), great.

packages/libp2p/src/connection-manager/constants.ts
SgtPooki: I think the existing comment is wordier than needs to be

packages/libp2p/src/connection-manager/auto-dial.ts
SgtPooki: a more intentional variable name here would be nice

SgtPooki (approved):

packages/libp2p/test/connection-manager/auto-dial.spec.ts
SgtPooki: I see that now, thanks

PR #1947, opened 2023-08-09
fix(@libp2p/webtransport): maximum call stack size exceeded on abort

achingbrain:
Fixes bug in webtransport where the `.abort` method on the stream class was calling `stream.abort` and leading to a call stack overflow.

image

Also splits the code into multiple files to make it a bit easier to see what is doing what.


packages/transport-webtransport/src/index.ts
```diff
@@ -17,300 +18,6 @@ declare global {
...
-    id: streamId,
-    status: 'open',
-    writeStatus: 'ready',
-    readStatus: 'ready',
-    abort (err: Error) {
-      if (!writerClosed) {
-        writer.abort()
-        writerClosed = true
-      }
-      stream.abort(err)
```
achingbrain, line 103: This is where the call stack overflow was happening.

maschad:
@achingbrain I am not familiar with this commit convention, where is it derived from?

achingbrain:
Do you mean the `(@libp2p/webtransport)` bit?  It's a [scope](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-scope).  Just an experiment - I figure it might make our release notes a bit more readable as it'll indicate which module is affected by which PR.

SgtPooki:
It's hard to fit the scope into the git commit and still be under 50 characters, but with modern technology, I prefer scoped commits rather than short git commit messages.

packages/transport-webtransport/src/utils/is-subset.ts
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Determines if `maybeSubset` is a subset of `set`. This means that all byte
+ * arrays in `maybeSubset` are present in `set`.
+ */
+export function isSubset (set: Uint8Array[], maybeSubset: Uint8Array[]): boolean {
+  const intersection = maybeSubset.filter(byteArray => {
+    return Boolean(set.find((otherByteArray: Uint8Array) => uint8ArrayEquals(byteArray, otherByteArray)))
+  })
+  return (intersection.length === maybeSubset.length)
+}
```
SgtPooki, line 12: do we need this to be Uint8array or can it be generic?

PR #1957, opened 2023-08-11
fix(@libp2p/utils): switch to @chainsafe/is-ip

### LINKED ISSUE #1926: switch to @chainsafe/is-ip (away from @achingbrain/ip-address)
Is there a reason why we're still using `@achingbrain/ip-address`, or even ip-address, when it's outdated? There seem to be better and more up to date alternatives: https://npmtrends.com/ip-vs-ip-address-vs-ipaddr.js. Though maybe they don't do everything we need them to?

One of `ip-address` 's dependencies, `jsbn`, is very stale and very underused in comparison to it's competitors: https://npmtrends.com/bignumber.js-vs-bn-vs-bn.js-vs-jsbn

There is also the native BigInt which has been around for a while: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

We only seem to have a very small surface area for each here: 

* ip-address - https://github.com/search?q=repo%3Alibp2p%2Fjs-libp2p+ip-address&type=code
* jsbn - https://github.com/search?q=repo%3Alibp2p%2Fjs-libp2p+jsbn&type=code

I also seem to remember some concerns around node-forge being used in js-libp2p and this may be another place where we could reduce our ties to it.

---

I'm not aware of the nuances but generally if a project leaves a PR that we need (https://github.com/beaugunderson/ip-address/pull/148) untouched for over a year (it's going on 2), and we still don't have a permanent place to modify code for our patched version, we should probably think about migrating, or publishing our version.

cc @achingbrain. related to https://github.com/ipfs/helia/issues/73#issuecomment-1644024103

achingbrain:
The `@achingbrain/ip-address` module is a fork of `ip-address` which seems to have dependencies on older modules that break modern runtimes like react native.

`@Chainsafe/is-ip` does not have these limitations so switch to that.

Fixes #1926

### DIFF
--- a/packages/utils/src/ip-port-to-multiaddr.ts
+++ b/packages/utils/src/ip-port-to-multiaddr.ts
@@ -1,4 +1,4 @@
-import { Address4, Address6 } from '@achingbrain/ip-address'
+import { isIPv4, isIPv6 } from '@chainsafe/is-ip'
 import { CodeError } from '@libp2p/interface/errors'
 import { logger } from '@libp2p/logger'
 import { type Multiaddr, multiaddr } from '@multiformats/multiaddr'
@@ -27,21 +27,15 @@ export function ipPortToMultiaddr (ip: string, port: number | string): Multiaddr
     throw new CodeError(`invalid port provided: ${port}`, Errors.ERR_INVALID_PORT_PARAMETER)
   }
 
-  try {
-    // Test valid IPv4
-    new Address4(ip) // eslint-disable-line no-new
+  if (isIPv4(ip)) {
     return multiaddr(`/ip4/${ip}/tcp/${port}`)
-  } catch {}
-
-  try {
-    // Test valid IPv6
-    const ip6 = new Address6(ip)
-    return ip6.is4()
-      ? multiaddr(`/ip4/${ip6.to4().correctForm()}/tcp/${port}`)
-      : multiaddr(`/ip6/${ip}/tcp/${port}`)
-  } catch (err) {
-    const errMsg = `invalid ip:port for creating a multiaddr: ${ip}:${port}`
-    log.error(errMsg)
-    throw new CodeError(errMsg, Errors.ERR_INVALID_IP)
   }
+
+  if (isIPv6(ip)) {
+    return multiaddr(`/ip6/${ip}/tcp/${port}`)
+  }
+
+  const errMsg = `invalid ip:port for creating a multiaddr: ${ip}:${port}`
+  log.error(errMsg)
+  throw new CodeError(errMsg, Errors.ERR_INVALID_IP)
 }
--- a/packages/utils/test/ip-port-to-multiaddr.spec.ts
+++ b/packages/utils/test/ip-port-to-multiaddr.spec.ts
@@ -19,7 +19,7 @@ describe('IP and port to Multiaddr', () => {
   it('creates multiaddr from valid IPv4 in IPv6 IP and port', () => {
     const ip = '0:0:0:0:0:0:101.45.75.219'
     const port = '9090'
-    expect(ipPortToMultiaddr(ip, port).toString()).to.equal(`/ip4/101.45.75.219/tcp/${port}`)
+    expect(ipPortToMultiaddr(ip, port).toString()).to.equal(`/ip6/::652d:4bdb/tcp/${port}`)
   })
 
   it('creates multiaddr from valid IPv6 IP and port', () => {
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -85,7 +85,7 @@
     "test:electron-main": "aegir test -t electron-main"
   },
   "dependencies": {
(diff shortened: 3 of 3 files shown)


packages/utils/src/ip-port-to-multiaddr.ts
```diff
@@ -27,21 +27,15 @@ export function ipPortToMultiaddr (ip: string, port: number | string): Multiaddr
...
-    return ip6.is4()
-      ? multiaddr(`/ip4/${ip6.to4().correctForm()}/tcp/${port}`)
-      : multiaddr(`/ip6/${ip}/tcp/${port}`)
-  } catch (err) {
-    const errMsg = `invalid ip:port for creating a multiaddr: ${ip}:${port}`
-    log.error(errMsg)
-    throw new CodeError(errMsg, Errors.ERR_INVALID_IP)
   }
+
+  if (isIPv6(ip)) {
```
achingbrain, line 34: `@chainsafe/is-ip` doesn't support detecting [IPv4 in IPv6 addresses](https://www.ibm.com/docs/en/ts4500-tape-library?topic=functionality-ipv4-ipv6-address-formats) so we lose support for returning an `IPv4` address in this case.

Node.js recognises these sorts of addresses as `IPv6` so I don't know if this is a problem or not.

E.g.:


```console
% node
Welcome to Node.js v18.14.0.
Type ".help" for more information.
> const net = require('net')
undefined
> net.isIP('0:0:0:0:0:0:101.45.75.219')
6
```

packages/utils/test/ip-port-to-multiaddr.spec.ts
```diff
@@ -19,7 +19,7 @@ describe('IP and port to Multiaddr', () => {
   it('creates multiaddr from valid IPv4 in IPv6 IP and port', () => {
     const ip = '0:0:0:0:0:0:101.45.75.219'
     const port = '9090'
-    expect(ipPortToMultiaddr(ip, port).toString()).to.equal(`/ip4/101.45.75.219/tcp/${port}`)
+    expect(ipPortToMultiaddr(ip, port).toString()).to.equal(`/ip6/::652d:4bdb/tcp/${port}`)
```
achingbrain, line 22: `@chainsafe/is-ip` doesn't support detecting [IPv4 in IPv6 addresses](https://www.ibm.com/docs/en/ts4500-tape-library?topic=functionality-ipv4-ipv6-address-formats) so we lose support for returning an `IPv4` address in this case.

Node.js recognises these sorts of addresses as `IPv6` so I don't know if this is a problem or not.

E.g.:


```console
% node
Welcome to Node.js v18.14.0.
Type ".help" for more information.
> const net = require('net')
undefined
> net.isIP('0:0:0:0:0:0:101.45.75.219')
6
```

packages/utils/src/ip-port-to-multiaddr.ts
```diff
@@ -27,21 +27,15 @@ export function ipPortToMultiaddr (ip: string, port: number | string): Multiaddr
...
+    return multiaddr(`/ip6/${ip}/tcp/${port}`)
+  }
+
+  const errMsg = `invalid ip:port for creating a multiaddr: ${ip}:${port}`
+  log.error(errMsg)
+  throw new CodeError(errMsg, Errors.ERR_INVALID_IP)
```
SgtPooki, line 40: This is so much cleaner than the previous code

packages/utils/test/ip-port-to-multiaddr.spec.ts
SgtPooki: We should be able to detect dual ipv6 addresses by simply testing for octects (or their delimiter, a period)

SgtPooki: I need to test the browser usecase. if this is showing as ipv6, we should be able to just wrap in brackets for URL construction.

SgtPooki (commented):
Need to confirm dualIpv6 support in browser before approving

packages/utils/test/ip-port-to-multiaddr.spec.ts
SgtPooki:
```bash
> new URL('http://[::101.45.75.219]')
URL {
  href: 'http://[::652d:4bdb]/',
  origin: 'http://[::652d:4bdb]',
  protocol: 'http:',
  username: '',
  password: '',
  host: '[::652d:4bdb]',
  hostname: '[::652d:4bdb]',
  port: '',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

SgtPooki (approved):

SgtPooki:
I gave approval, but I expect this will break one scenario: consumers of dualIpv6 addresses that don't support ipv6 used to get an ipv4 multiaddr they could use; they won't anymore.

achingbrain:
It seems weird to me that we’d accept a v6 address and return a v4 one, not what I would expect.

PR #1960, opened 2023-08-11
fix(@libp2p/protocol-perf): ensure only client calls measure performance

maschad:
According to the [spec](https://github.com/libp2p/specs/blob/master/perf/perf.md) perf is a client-driven set of benchmarks., therefore the server need not call the _measurePerfomance_ method. This PR fixes two things:

- Only client calls the `measurePerformance` method
- The server node must remaining i.e. not stopped as there are multiple perf runs against a single server process. The [test runner](https://github.com/libp2p/test-plans/blob/master/perf/runner/src/index.ts#L128-L133) will end the server process.


MarcoPolo (approved):

achingbrain:
Please can you check the state of the build before merging PRs.

The previous commit merged [would have performed a release](https://github.com/libp2p/js-libp2p/pull/1941).  One of the interop tests [timed out](https://github.com/libp2p/js-libp2p/actions/runs/5858114184/job/15881582843) which means the release didn't happen.

Now this commit is the head of master so release-please will not (may not?) perform the release since it's no longer the tip of the tree.

In general please do not merge to master when master is red.

maschad:
My apologies, noted. Perhaps another way to articulate it is we should ensure our releases are successful before merging to master.

PR #1961, opened 2023-08-11
feat: returned peer multiaddrs include peerIds

### LINKED ISSUE #1959: feat: peerStore returns multiaddrs for a peer with peerIds postfixed where necessary.
When we get a peer from the peer store, we should be able to assume the multiaddrs need no modification. However, this is only sometimes true.

Some cases where it's not true:
* p2p-circuit multiaddrs

References: 
* https://github.com/libp2p/js-libp2p/pull/1928#issuecomment-1672474715
* https://github.com/multiformats/js-multiaddr-matcher/pull/2
* https://github.com/multiformats/js-multiaddr-matcher/issues/3

SgtPooki:
This PR addresses some concerns brought up in #1959 that I discussed with @achingbrain.

> We should have a firm distinction between the types [Peer](https://github.com/libp2p/js-libp2p/blob/master/packages/interface/src/peer-store/index.ts#L22-L53) and [PeerInfo](https://github.com/libp2p/js-libp2p/blob/master/packages/interface/src/peer-info/index.ts#L4-L8)

> peerStore.get returns Peer, this is the fully-featured version with peer ids appended to the multiaddrs (which we would do prior to returning them), along with tags, metadata, etc, then add a new peerStore.getInfo method that returns a [PeerInfo](https://docs.libp2p.io/concepts/fundamentals/peers/#peer-info) - this just has the peer id and the multiaddrs without the peer id and is the structure you'd expect to exchange with other implementations.

fixes #1959


packages/peer-store/src/utils/address-to-fqmultiaddr.ts
```diff
@@ -0,0 +1,27 @@
...
+export function addressToFqMultiaddr (peerId: PeerId, addresses: Address[]): Address[] {
+  return addresses.map(({ multiaddr, isCertified }) => {
+    if (!isMultiaddr(multiaddr)) {
+      throw new CodeError('Multiaddr was invalid', codes.ERR_INVALID_PARAMETERS)
+    }
+
+    // is checking for null enough or should we check if peerId.toString() is a mismatch?
+    if (multiaddr.getPeerId() !== peerId.toString()) {
+      multiaddr = multiaddr.encapsulate(`/p2p/${peerId.toString()}`)
+    }
```
SgtPooki, line 20: this was a `== null` check in @achingbrain 's code. I'm wondering if we want to ensure addresses being saved for a particular peer are returning the expected address for that peer.

lmk and I can change back if desired.

packages/peer-store/src/utils/to-peer-pb.ts
```diff
@@ -133,7 +134,7 @@ export async function toPeerPB (peerId: PeerId, data: Partial<PeerData>, strateg
   }
 
   const output: PeerPB = {
-    addresses: await dedupeFilterAndSortAddresses(peerId, options.addressFilter ?? (async () => true), addresses),
+    addresses: await dedupeFilterAndSortAddresses(peerId, options.addressFilter ?? (async () => true), addressToFqMultiaddr(peerId, addresses)),
```
SgtPooki, line 137: as far as I could tell, this `to-peer-pb` is called prior to saving/merging/etc.. so it seemed like the best place to add this if we _do_ want the store to have the peer Id. If we don't, either src/index or src/store would be the best place. lmk.

packages/peer-store/test/save.spec.ts
```diff
@@ -67,12 +67,13 @@ describe('save', () => {
...
         isCertified: false,
-        multiaddr
+        multiaddr: multiaddr.encapsulate(`/p2p/${otherPeerId.toString()}`)
       }))
     )
     expect(previous).to.be.undefined()
   })
 
+  // todo: this test needs fixed: it's not testing what it says it is; or it's passing when it shouldn't
   it('emits on set if not storing the exact same content', async () => {
```
SgtPooki, line 77: I left this comment in the code to trigger a discussion. This test says it's emitting when not storing the same content, but I see we have `patch` and other methods now. There are two potential actions here:

1. rename the test if not fully covered by others (I believe it is)
2. figure out why the `// set 2` is not throwing (was patch added, and this is no longer the expectation?)
3. remove the test entirely

SgtPooki (commented):
self review

packages/peer-store/test/save.spec.ts
SgtPooki: Moved this to https://github.com/libp2p/js-libp2p/issues/1962

packages/peer-store/src/utils/address-to-fqmultiaddr.ts
achingbrain: If there's a peer id present and it's not the peer id we expect maybe we should filter it out?

Otherwise we could end up returning `/ip4/.../tcp/.../p2p/QmFoo/p2p/QmBar` which isn't correct.

Off the top of my head the only time there will be a peer id in the multiaddr that isn't for the peer is circuit relay addresses and since `@multiformats/multaddr@12.1.5` the target peer's address is returned from `getPeerId` rather than the relay.

packages/peer-store/src/utils/to-peer-pb.ts
achingbrain: I don't think we should store the multiaddrs with peer ids in the datastore, the reason being some users will have existing datastore which will then have some peers with multiaddrs with peer ids and some without, so we'd need to do a datastore migration to get everything consistent.

Instead we should have a firm disctinction between the types `Peer` and `PeerInfo`

`peerStore.get` returns `Peer`, this is the fully-featured version with peer ids appended to the multiaddrs (which we would do prior to returning them), along with tags, metadata, etc, then add a new `peerStore.getInfo` method that returns a [PeerInfo](https://docs.libp2p.io/concepts/fundamentals/peers/#peer-info) - this just has the peer id and the multiaddrs without the peer id and is the structure you'd expect to exchange with other implementations.

patches/@libp2p+interop+9.0.0.patch
```diff
@@ -0,0 +1,33 @@
+diff --git a/node_modules/@libp2p/interop/src/dht/peer-routing.ts b/node_modules/@libp2p/interop/src/dht/peer-routing.ts
```
achingbrain, line 1: See the other comment about `Peer` vs `PeerInfo`, the DHT `.findPeer` method returns a `PeerInfo` which doesn't have the peer id appended so this patch shouldn't be necessary.

packages/peer-store/src/utils/address-to-fqmultiaddr.ts
```diff
@@ -0,0 +1,27 @@
...
+import { isMultiaddr } from '@multiformats/multiaddr'
+import { codes } from '../errors.js'
+import type { PeerId } from '@libp2p/interface/peer-id'
+import type { Address } from '@libp2p/interface/peer-store'
+
+/**
+ * Converts {@link Address}[] for a peer that may not contain the {@link PeerId} to {@link Address}[]
+ * that do contain the {@link PeerId}.
+ */
+export function addressToFqMultiaddr (peerId: PeerId, addresses: Address[]): Address[] {
```
achingbrain, line 11: A nit, but the output type is `Address[]` not `Multiaddr` and they are already fully qualified (assuming that's what `fq` refers to here) as you don't necessarily need the peer id to dial them so this could probably have a better name like `addressesWithPeerIds` or something?

SgtPooki (commented):
self review

Issue #1949, opened 2023-08-09
refactor: Remove unnecessary async methods in `crypto` package


maschad:
I think we now have a load of async functions that don't do any async work - we should remove the async keyword from them to prevent promises being created unnecessarily but that can be done in a follow-up.

_Originally posted by @achingbrain in https://github.com/libp2p/js-libp2p/pull/1939#pullrequestreview-1567754174_

tabcat:
Seems like any async functions with `// eslint-disable-line require-await` are candidates to be changed?

maschad:
That's a good place to start but not all async functions with that comment are necessarily sync such as the [aes-gcm encrypt function](https://github.com/libp2p/js-libp2p/blob/master/packages/crypto/src/ciphers/aes-gcm.browser.ts#L35) which incorrectly has this comment.

PR #1963, opened 2023-08-12
refactor!: remove unnecessary async from crypto methods

tabcat:
Removed async keyword from function which don't:

1) do any async work
2) return a promise
3) have an async browser equivalent

Closes: #1949


maschad (changes requested):
Left a comment. I think there are some other functions such as [encrypt](https://github.com/libp2p/js-libp2p/blob/master/packages/crypto/src/ciphers/aes-gcm.ts#L34) and [create](https://github.com/libp2p/js-libp2p/blob/master/packages/crypto/src/aes/index.ts#L9) which could also be made synchronous.

Majority of the functions addressed in https://github.com/libp2p/js-libp2p/commit/78b008c72d301de38dbe9e32196c03676c8b6db8 would have been async without an _await_ so I think many could be made synchronous

tabcat:
The create function is already made sync. I chose to keep the encrypt function returning a promise because that is what the browser equivalent does. Can change if it's fine to do this.

The other places where async is still used without await still return promises. With current linter rules any function returning a promise must be an async function.

tabcat:
> The premise was that given the functions are not doing asynchronous work, they should not return a promise. So they need to be refactored to not create promises.

https://github.com/libp2p/js-libp2p/blob/124947511512d9e49a9e9304886fa3b1a117c5c7/packages/crypto/src/keys/importer.ts#L9-L13

`cipher.decrypt` returns a promise so `importer` must be async. in node `cipher.decrypt` could be made synchronous, but in the browser using webcrypto this is not possible.

There may be some situations where the function being returned could be refactored to not return a promise. I will check for this.

tabcat:
Ah, hmac's create function for the browser cannot be made sync. Did you talk with @achingbrain yet about if they have to match? Seems like they should, especially if part of public API.

tabcat:
I think would be best for the node/browser mirror methods to keep parity wrt their interfaces. One returning a value while the other returns a Promise<value> looks like trouble.

There are cases (Ed25519 keys) where both node and browser methods could be synchronous. Since this is not a public API/not exported from package and gets turned into a promise via the public API methods, changing these seems ok.
Same thing with Secp256k1 keys except that there is no node/browser separation here.

maschad:
> i did some benchmarking with synchronous ed25519 in node and the async api is much faster.
also checked over all the async function and i think i covered them all.

Could you share the benchmark results here, I find it strange that's the case.

tabcat:
I've made the ed25519 methods all synchronous, this doesn't change public api since they are later wrapped in a promise inside of keys/index.ts

use the refactor semantic commit tag but maybe chore would be better?

achingbrain (approved):

achingbrain:
Thanks!

Issue #1896, opened 2023-07-25
WebTransport session establishment failed. Too many pending WebTransport sessions (64)


SgtPooki:
Sometimes I see thousands instances of this warning in Chrome:

`WebTransport session establishment failed. Too many pending WebTransport sessions (64)`

![image](https://github.com/libp2p/js-libp2p/assets/1173416/e01eb92e-866e-4196-9a17-7ea8655a71aa)

This module may need some sort of dial queue to ensure it doesn't open too many connections and trigger this error.

ported over from https://github.com/libp2p/js-libp2p-webtransport/issues/64

SgtPooki:
I believe this is a blocker for https://github.com/ipfs/helia/issues/182 because:

1. webtransport is one of the few consistent ways we can connect to other nodes from the browser.
2. A Helia nodes' connectivity become unstable once we have too many pending dials.

@maschad are you actively working on this?

maschad:
I'm not actively working on it at the moment @SgtPooki although I think @achingbrain 's PR https://github.com/libp2p/js-libp2p/pull/1947 may be related.

achingbrain:
I think #1947 will help but I can't help but wonder if there's some cleanup we need to do that we're missing.

achingbrain:
This may be a bug in Chrome.

When we forcibly close WebTransport connections whose `.ready` promise hasn't resolved within the connection timeout (the peer has gone away, is on a slow connection, is overloaded, is firewalled, etc), Chrome may not be cleaning up properly in the background.

More details here: https://bugs.chromium.org/p/chromium/issues/detail?id=1473980

achingbrain:
I've tried to add a global count to the WebTransport transport to ensure we don't go over 64 "pending" connections, taking "pending" as meaning "has yet to resolve/reject the `.ready`/`.closed` promises" but it doesn't solve the problem.

Counting the various WebTransport sessions that have been opened and what happened to them, it seems sessions that reject* their `.ready`/`.closed` promises are still counted as "pending".

Therefore regardless of any limit we set on how many connections we open simultaneously, once the number of errored connections plus the number of yet-to-resolve/reject connections reaches 65 no further WebTransport connections can be opened.

This is bad news and needs a browser fix because once 65 connections have errored it's essentially game over until the page is reloaded.

I've updated the chromium bug report with this information.

----

\* = The rejection reasons are normal network things - an unreachable host, a handshake timeout, etc.

achingbrain:
[...]

PR #1969, opened 2023-08-15
fix(@libp2p/webtransport): be more thorough about closing sessions

achingbrain:
Refactors session closing to happen in one function and call that function when the session has closed or failed to init.

Doesn't quite solve the "Too many pending WebTransport Sessions" problem but does slow it down a little bit.

Refs: #1896

### DIFF
--- a/packages/transport-webtransport/src/index.ts
+++ b/packages/transport-webtransport/src/index.ts
@@ -16,6 +16,16 @@ declare global {
   var WebTransport: any
 }
 
+// https://www.w3.org/TR/webtransport/#web-transport-close-info
+interface WebTransportCloseInfo {
+  closeCode: number
+  reason: string
+}
+
+interface WebTransportSessionCleanup {
+  (closeInfo?: WebTransportCloseInfo): void
+}
+
 const log = logger('libp2p:webtransport')
 
 export interface WebTransportInit {
@@ -42,6 +52,8 @@ class WebTransportTransport implements Transport {
   readonly [symbol] = true
 
   async dial (ma: Multiaddr, options: DialOptions): Promise<Connection> {
+    options?.signal?.throwIfAborted()
+
     log('dialing %s', ma)
     const localPeer = this.components.peerId
     if (localPeer === undefined) {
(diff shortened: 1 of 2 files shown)


packages/transport-webtransport/src/index.ts
```diff
@@ -52,60 +64,122 @@ class WebTransportTransport implements Transport {
...
+            reason: err.message
+          })
+        },
+        remoteAddr: ma,
+        timeline: {
+          open: Date.now()
+        },
+        // This connection is never used directly since webtransport supports native streams.
+        ...inertDuplex()
+      }
```
SgtPooki, line 164: if we make this a property on the class we can split out this dial function a bit

packages/transport-webtransport/src/index.ts
```diff
@@ -52,60 +64,122 @@ class WebTransportTransport implements Transport {
...
+        .catch((err: Error) => {
+          log.error('error on remote wt session close', err)
+          maConn?.abort(err)
+        })
+        .finally(() => {
+          // if we never got as far as creating the maConn, just clean up the session
+          if (maConn == null) {
+            cleanUpWTSession()
+          }
+        })
```
SgtPooki, line 128: I believe this should cover the case where `await wt.ready` throws, but from my testing, I'm not confident the `.closed` catch always handles errors thrown from `wt.ready`.

SgtPooki (approved):
definitely an improvement

packages/transport-webtransport/src/index.ts
```diff
@@ -52,60 +64,122 @@ class WebTransportTransport implements Transport {
...
     try {
+      let closed = false
+      const wt = new WebTransport(`${url}/.well-known/libp2p-webtransport?type=noise`, {
+        serverCertificateHashes: certhashes.map(certhash => ({
+          algorithm: 'sha-256',
+          value: certhash.digest
+        }))
+      })
+
+      cleanUpWTSession = (closeInfo?: WebTransportCloseInfo) => {
```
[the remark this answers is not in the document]
achingbrain: It needed access to a lot of scoped variables, and it's passed to the stream muxer so it got a bit fiddly.

PR #1970, opened 2023-08-15
fix(libp2p): reduce dialer activity in browsers

achingbrain:
Use a different set of defaults for browsers vs node - from testing we never really see more than 20x connections so there's no point having the auto-dialler trying to reach 50x, also reduce the number of concurrent dials and how many peers we try to autodial at once.

### DIFF
--- a/packages/libp2p/package.json
+++ b/packages/libp2p/package.json
@@ -191,6 +191,7 @@
     "sinon-ts": "^1.0.0"
   },
   "browser": {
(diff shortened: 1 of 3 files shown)


maschad (approved):
lgtm

SgtPooki (approved):
lgtm, but is this actually reducing dialer activity in browsers, or just changing constants that still need consumed?

achingbrain:
> but is this actually reducing dialer activity in browsers, or just changing constants that still need consumed?

Yes, this will reduce dialer activity in browsers.

Issue #1895, opened 2023-07-25
WebRTC Transport creating large number of DataChannels when used with Multiplexer


justin0mcateer:
- **Version**:
0.45.9

- **Platform**:

- **Subsystem**:
WebRTC Transport
Yamux / Mplex

#### Severity:
Medium

#### Description:
When connecting two browsers directly via a CircuitRelayV2 server, I am seeing stable signaling (a single WebRTC connection 'stays up'), but a large number (~60) of DataChannels are created, each only sending a very small number of bytes (<100 bytes in nearly all cases). As an additional bit of information, Yamux opens 60 DataChannels and Mplex seems to open 20, but they both continue adding new DataChannels every five minutes. I don't think this would create an real issue in the short-term, but for a long running application (like ours) it will almost certainly be problematic fairly quickly.

According to this article the number of DataChannels supported by the browser are much, much less than the theoretical maximum (as little as 128) for FireFox as of ~2 years ago. Assuming nothing has changed, this would exhaust the number of available DataChannels in several hours.
https://tensorworks.com.au/blog/webrtc-stream-limits-investigation/

~60 DataChannels, mostly created at once:
![image](https://github.com/libp2p/js-libp2p/assets/2607229/7ad1cf75-e309-40b2-b938-ec5db7aee02b)

<100 Bytes sent on each DataChannel:
![image](https://github.com/libp2p/js-libp2p/assets/2607229/ac5de6c2-6d84-430e-99ec-fcb9642c5bb0)

The number of DataChannels continues to grow, in my configuration a new DataChannel is opened approximately every five minutes. All the DataChannels appear to remain open (at least in excess of 15 minutes)

New DataChannels opened every five minutes (exactly):
![image](https://github.com/libp2p/js-libp2p/assets/2607229/1eeb8e52-d7c7-43e7-8c9a-1927fb012e49)

Old DataChannels stay open:
![image](https://github.com/libp2p/js-libp2p/assets/2607229/025624ea-cd13-4264-b2c0-272b4f3e90f3)

#### Steps to reproduce the error:
- Establish a Browser to Browser WebRTC connection with the WebRTC Transport
- Open the 'brave://webrtc-internals' (or 'chrome://webrtc-internals')
- Observe the large and ever growing number of DataChannels

justin0mcateer:
Just as a follow up, left to run overnight, this does hit the maximum number of DataChannels Chrome will allow (512).

Marcel-G:
[...]

Issue #1835, opened 2023-05-15
Connection is not shutting down properly, which makes reconnects within ~10 seconds impossible


marcus-pousette:
I noticed an issue when I have to nodes in two browser tabs, and I refresh on of the tabs. The connection is able to resume immediately, but 

1. After a few seconds the WebRTC connection closes in the tab that did not refresh (even though the client in both tabs are live and connected)
2. The tab that refreshes still thinks the connection is live which makes messages not delivering

To reproduce the issues partially: 

Put a log here
https://github.com/libp2p/js-libp2p-webrtc/blob/efe5ce70bdffdb67c237cfb4688afd562528a050/src/maconn.ts#LL68C9-L68C9

and run the browser-to-browser example. 

Node 1 connects to the relay, and Node 2 connects to Node 1 relayed address. 

After connection, close Node 2 and observe that the state change does not appear *immediately* in the logs. 

--

If Node 2 would restart with the same PeerId and connects to Node 1 before the state change has happened, Node 1 will terminate Node 2 wrongfully as

https://github.com/libp2p/js-libp2p-webrtc/blob/efe5ce70bdffdb67c237cfb4688afd562528a050/src/maconn.ts#L70

will bubble up. 

Since the browser-to-browser example is using new PeerIds each time the issue becomes less severe, but in real world scenarios where you run nodes with the same PeerId you will have nodes running with broken connections, as one side of the connection has wrongfully terminated it due to the reason outlined above.

maschad:
> If Node 2 would restart with the same PeerId and connects to Node 1 before the state change has happened, Node 1 will terminate Node 2 wrongfully as

[libp2p/js-libp2p-webrtc@efe5ce7/src/maconn.ts#L70](https://github.com/libp2p/js-libp2p-webrtc/blob/efe5ce70bdffdb67c237cfb4688afd562528a050/src/maconn.ts#L70)

will bubble up.

I am not sure if this is still an issue as should the redial occur from the [auto-dialler](https://github.com/libp2p/js-libp2p/blob/70fe699c424c22f81e39c47631fda73abc1cae9e/packages/libp2p/src/connection-manager/auto-dial.ts#L35) (given the previous connection was closed) then a new connection will be created, so it would have been correctly terminated given that Node B was in fact shutdown.

Please re-open if you feel this is still an issue using our [issue template](https://github.com/libp2p/js-libp2p/blob/master/.github/ISSUE_TEMPLATE/open_an_issue.md)

[...]

PR #1976, opened 2023-08-16
fix(@libp2p/webrtc): close data-channel on muxer stream end

Marcel-G:
https://github.com/libp2p/js-libp2p/issues/1895

## Problem

`RTCDataChannel`s are not properly closed after the stream ends. They stay in an open state and eventually the browser will not be able to create more `RTCDataChannel`s with the following error:

```
Uncaught (in promise) DOMException: Failed to execute 'createDataChannel' on 'RTCPeerConnection': RTCDataChannel creation failed
```

## Solution

Close data channels when the stream ends.


achingbrain:
CI has become very sad - could this be because the WebRTC data channel is being closed before any queued data has been sent?

marcus-pousette:
Would be great if this could be part of next release. This would perhaps also solve a related [issue](https://github.com/libp2p/js-libp2p/issues/1835)

achingbrain (approved):

achingbrain:
This PR needs a followup:

1. Perhaps close the datachannel in the receiver as well for safety
2. Add tests that assert we can open/close a large amount of streams
3. Maybe figure out how to assert that WebRTC DataChannels are closed correctly - they are scoped to the calling context so they'll need to be exposed externally somehow and cleaned up from however they are exposed

PR #1981, opened 2023-08-18
docs(libp2p): fix typo in circuit-relay debug log

tabcat:


maschad (approved):
Thanks!

maschad:
I changed the commit type to `docs` because a consumer would see this change whilst debugging (thus it's a user-facing change) and so it's not a `chore` in my opinion (as `chore`s often refer to maintenance changes such as build tasks that would not be consumer facing).

PR #1987, opened 2023-08-24
fix(@libp2p/interface-compliance-tests): make ageir a dep

### LINKED ISSUE #1974: bug: broken @libp2p/interface-compliance-tests
- **Version**: 0.46.3

- **Platform**: macos

- **Subsystem**: compliance tests

#### Description:
on the upgrade of `@libp2p/interface-peer-discovery-compliance-tests` to `@libp2p/interface-compliance-tests@4.0.2` the package breaks because `aegir` is installed as a dev dependency on the latter while it's needed to run `tests` which was installed as a `dependency` on the former package.

the solution can be simply moving `aegir` to `dependency` or to install `chai` as a dependency and using that instead of `aegir/chai`?

achingbrain:
The interface compliance tests depend on aegir to run the tests so make it a dep instead of a dev dep.

Fixes #1974

### DIFF
--- a/packages/interface-compliance-tests/package.json
+++ b/packages/interface-compliance-tests/package.json
@@ -111,6 +111,7 @@
     "@libp2p/peer-id-factory": "^3.0.3",
     "@multiformats/multiaddr": "^12.1.5",
     "abortable-iterator": "^5.0.1",
+    "aegir": "^40.0.8",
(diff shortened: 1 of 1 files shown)


achingbrain:
Closing in favour of #1983

PR #1988, opened 2023-08-24
refactor(libp2p): remove unnecessary any-signal

achingbrain:
Removes use of `any-signal` where we only have one signal.


maschad (approved):
lgtm

achingbrain:
Thanks for updating the tag used in the commit message.

For this project the permissible tags are found in the [main.yml GH actions file](https://github.com/libp2p/js-libp2p/blob/master/.github/workflows/main.yml#L235-L241).  "refactor" isn't in there so it either needs to be added or an existing one chosen, perhaps "chore"?

maschad:
Good spot @achingbrain , perhaps we can merge https://github.com/libp2p/js-libp2p/pull/1990 first then this.

SgtPooki (approved):

PR #1992, opened 2023-08-25
fix(@libp2p/protocol-perf): use noise for encryption

### LINKED ISSUE #1991: fix(protocol/perf): consider using tls or noise instead of plaintext
All other [libp2p perf protocol implementations](https://github.com/libp2p/test-plans/tree/master/perf/impl) use some mechanism of encryption. For a better comparison, consider using TLS or Noise in the js-libp2p perf implementation as well.

https://github.com/libp2p/js-libp2p/blob/89778624908a536e3253ee4fe1a0d287e1aad2e9/packages/protocol-perf/src/main.ts#L59-L61

//CC @maschad 

Related pull request adding js-libp2p to libp2p/test-plans/perf https://github.com/libp2p/test-plans/pull/244.

maschad:
Closes #1991

### DIFF
--- a/packages/protocol-perf/package.json
+++ b/packages/protocol-perf/package.json
@@ -47,6 +47,7 @@
     "renderResults": "node dist/src/renderResults.js"
   },
   "dependencies": {
+    "@chainsafe/libp2p-noise": "^13.0.0",
     "@chainsafe/libp2p-yamux": "^5.0.0",
     "@libp2p/crypto": "^2.0.3",
     "@libp2p/interface": "^0.1.2",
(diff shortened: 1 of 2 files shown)


achingbrain (changes requested):
@mxinden's comment in #1991 was:

> For a better comparison, consider using TLS or Noise in the js-libp2p perf implementation as well.

This is "instead of" rather than "as well", is that correct?

mxinden (commented):
Correct. Noise instead of plaintext. Opposite to the interop tests, the perf benchmarks can not (yet) select a security protocol.

PR #1998, opened 2023-08-29
fix: remove relay:removed event listener after relay is removed

### LINKED ISSUE #1944: Possible EventTarget memory leak detected.
- **Version**:
"libp2p": "^0.45.9"

- **Platform**:
Linux nf 6.2.0-26-generic #26~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Jul 13 16:27:29 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

- **Subsystem**:
Dialer

#### Severity:
Medium

#### Description:

I have a setup with multiple nodes connected to relay. Usually on start works with no problems after 1 day or so of runtime. We usually experience this warning an all of the nodes and we see a spike of 100% cpu constant usage after that, which slows down dramatically other running processes. (I'm not sure those are related, but everything seems to happen after that warning log). 
Also, we experienced that nodes lose pubsub messages after that as well.  

It's the Dialer Relay that logs that warning. 

### Error: 
(node:2880541) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 11 relay:removed listeners added to ReservationStore. Use ev ents.setMaxListeners() to increase limit

**Relay Setup:**
```javascript
async function main () {
    peerId: await createFromJSON(rid),
    addresses: {
      listen: ['/ip4/127.0.0.1/tcp/5040/ws'],
      announce: ['/dns4/relay.smoothly.money/tcp/443/wss/']
    },
    transports: [
      webSockets(),
    ],
    connectionEncryption: [
      noise()
    ],
    streamMuxers: [
      yamux({
        maxMessageSize: 1 << 20 // 1 MB
      })
    ],
    services: {
      identify: identifyService(),
      relay: circuitRelayServer({
        reservations: {
          defaultDataLimit: BigInt(1 << 20), // the default maximum number of bytes that can be transferred over a relayed connection
        }
      }),
      pubsub: gossipsub({ allowPublishToZeroPeers: true })
    },
    peerDiscovery: [
      pubsubPeerDiscovery({
        interval: 1000
      })
    ]
  })

  console.log(`Node started with id ${node.peerId.toString()}`)
  console.log('Listening on:')
  node.getMultiaddrs().forEach((ma) => console.log(ma.toString()))
  node.addEventListener('peer:discovery', async (evt) => {
        const { id } = evt.detail;
        console.log("Found:", id);
        await node.dial(id);
  })
}

main()

```
**Dialer Setup:**
```javascript
  async createNode(): Promise<void> {
    try {
        transports: [
          webSockets(),
          circuitRelayTransport({
            discoverRelays: 1
          })
        ],
        streamMuxers: [
          yamux({
            maxMessageSize: 1 << 20
          })
        ],
        connectionEncryption: [
          noise()
        ],
        services: {
          pubsub: gossipsub({ allowPublishToZeroPeers: true }),
          identify: identifyService()
        },
        peerDiscovery: [
          bootstrap({
            list: this.bootstrapers
          }),
          pubsubPeerDiscovery({
            interval: 1000,
          })
        ],
      });

      // Personal_id channel
      node.services.pubsub.subscribe(`${node.peerId.toString()}`)
      // Checkpoint check
      node.services.pubsub.subscribe('checkpoint')

      // Track peers
      node.addEventListener('peer:discovery', async (evt) => {
        const { addresses, id } = evt.detail as any;

        // Avoids relay nodes
        if(addresses.length > 0) {
          const peer: Peer = {
            address: addresses[0].multiaddr,
            id: id
          };

          // Exists peer?
          if(this._findPeer(peer.id) === null) {
            this.peers.push(peer);
            console.log("Discovered Peer:", id.toString(), "total:", this.peers.length);
          }
        }
      })

      // Handle pubsub messages
      node.services.pubsub.addEventListener('message', (evt) => {
        const { from } = evt.detail as any;
        if(evt.detail.topic === `${node.peerId.toString()}`) {
          const data = Buffer.from(evt.detail.data).toString();
          if(data === 'sync') {
            const peer = this._findPeer(from);
            if(peer) {
              this.dialPeerSync(peer.address);
            }
          }
        } else if(evt.detail.topic === 'checkpoint'){
          const { root, epoch } = JSON.parse(uint8ArrayToString(evt.detail.data));
          console.log('checkpoint:', from, root, epoch);
          this.consensus.addVote(from, root, epoch);
        }
      })

      // Handle stream muxing from peer:sync 
      node.handle('/sync:peer', async ({ stream }) => {
        let db: DB = this.db;
        await pipe(
          stream,
          async function (source) {
            // Add data
              let validator = JSON.parse(uint8ArrayToString(msg.subarray()));
              await db.insert(validator.index, validator);
            }
          }
        )
        stream.close();
      })

      await node.start();
      this.node = node;
    } catch(err: any) {
      console.log(err);
    }
  }

```

maschad:
If a transport has multiple listeners (as is often the case with CircuitRelay) this allows the closing of a specific listener per transport. Also prevents a potential memory leak in circuit relay transport listener.

Closes #1944

Closes https://github.com/libp2p/js-libp2p/issues/2106


packages/libp2p/src/circuit-relay/transport/index.ts
```diff
@@ -133,6 +135,13 @@ class CircuitRelayTransport implements Transport {
     this.maxOutboundStopStreams = init.maxOutboundStopStreams ?? defaults.maxOutboundStopStreams
     this.stopTimeout = init.stopTimeout ?? defaults.stopTimeout
 
+    this.shutdownController = new AbortController()
```
achingbrain, line 138: This controller seems to be created only so we can increase the max number of listeners on it's signal.

The signal is not passed into any other context that might actually add some event listeners to it - how does this solved the problem?

achingbrain (changes requested):

packages/libp2p/src/circuit-relay/transport/index.ts
maschad: Right, good catch. It turns out it was actually the transport listener not removing the [relayStore event listener](https://github.com/libp2p/js-libp2p/blob/9a69e6f76018d2df5c6e51de930557d4b60b2fca/packages/libp2p/src/circuit-relay/transport/listener.ts#L34) when closing the connection as we discussed.

I decided against adding a test as the [relay spec already tests for connection close events](https://github.com/libp2p/js-libp2p/blob/9a69e6f76018d2df5c6e51de930557d4b60b2fca/packages/libp2p/test/circuit-relay/relay.spec.ts#L103) and so given the connection is closed via the [safeDispatch](https://github.com/libp2p/js-libp2p/blob/9a69e6f76018d2df5c6e51de930557d4b60b2fca/packages/libp2p/src/circuit-relay/transport/listener.ts#L104) then the `close` method is called (which then emits the close connection event). 

Seperately I can see the event being removed in the logs
image

achingbrain (changes requested):
A test for this could set up a node listening on two relay servers, it could use the [listenerCount method](https://github.com/libp2p/js-libp2p/blob/master/packages/interface/src/events.ts#L21) on EventEmitter to assert that the event listeners were set up, then shut one relay down which should cause the transport listener to close, then assert that one of the event listeners was removed.

packages/libp2p/test/circuit-relay/relay.node.ts
```diff
@@ -675,6 +675,26 @@ describe('circuit-relay', () => {
       expect(events[1].detail.remotePeer.toString()).to.equal(relay1.peerId.toString())
     })
 
+    // Not sure why this test isn't passing
+    it.skip('should remove the relay event listener when the relay stops', async () => {
```
maschad, line 679: I am not sure why there aren't any listeners being attached to the relay as the logs are saying otherwise, but `listenerCount` is returning 0.

packages/libp2p/test/circuit-relay/relay.node.ts
```diff
@@ -669,6 +673,42 @@ describe('circuit-relay', () => {
...
+        const circuitMultiaddrs = listener.getAddrs().filter(ma => Circuit.matches(ma))
+        return circuitMultiaddrs.length > 0
+      })
+
+      expect(circuitListener[0].relayStore.listenerCount('relay:removed')).to.equal(2)
+
+      // remove one listener
+      await relay1.stop()
+
+      await notUsingAsRelay(local, relay1)
```
SgtPooki, line 704:
```suggestion
      await usingAsRelayCount(local, [relay2], 1) // we ARE using relay2 as a relay
      await usingAsRelayCount(local, [remote, relay1, relay2, relay3], 1) // we are not using any others as a relay
```

packages/libp2p/src/transport-manager.ts
```diff
@@ -262,27 +257,26 @@ export class DefaultTransportManager implements TransportManager, Startable {
...
   }
 
+  async removeListener (ma: Multiaddr): Promise<void> {
+    log('removing listener for %a', ma)
+
+    for (const listeners of this.listeners.values()) {
+      const listener = listeners.get(ma)
+      if (listener != null) {
+        await listener.close()
+        listeners.delete(ma)
```
wemeetagain, line 275: Is there a situation where a single multiaddr can be attached to multiple listeners? I think no.
In that case we can `return` here after removing/closing the first match.

packages/libp2p/src/circuit-relay/transport/reservation-store.ts
```diff
@@ -287,23 +289,44 @@ export class ReservationStore extends EventEmitter<ReservationStoreEvents> imple
...
-  #removeRelay (peerId: PeerId): void {
+  async #removeRelay (peerId: PeerId): Promise<void> {
     const existingReservation = this.reservations.get(peerId)
 
     if (existingReservation == null) {
       return
     }
 
-    log('connection to relay %p closed, removing reservation from local store', peerId)
+    log.trace('connection to relay %p closed, removing reservation from local store', peerId)
```
SgtPooki, line 319: I have seen Alex mention standards for jslibp2p logging (trace for happy path, log for important entry/announce things, error for errors) but do we have this in docs somewhere?

packages/libp2p/src/transport-manager.ts
```diff
@@ -262,27 +257,27 @@ export class DefaultTransportManager implements TransportManager, Startable {
...
-      const listener = listeners.pop()
-
-      if (listener == null) {
-        continue
-      }
-
-      tasks.push(listener.close())
+    for (const listener of this.listeners.get(key)?.values() ?? []) {
+      await listener.close()
     }
```
SgtPooki, line 263: Do we want to map over these so we can wait for multiple listener.close() in parallel instead;  Similar to how we’re doing in .stop()

packages/libp2p/src/circuit-relay/transport/reservation-store.ts
maschad: Not to my knowledge, but in any event seems like an appropriate place to put a trace log given that rubric. 

May be worthwhile starting a discussion thread for that.

packages/libp2p/src/transport-manager.ts
```diff
@@ -29,7 +29,8 @@ export interface DefaultTransportManagerComponents {
 export class DefaultTransportManager implements TransportManager, Startable {
   private readonly components: DefaultTransportManagerComponents
   private readonly transports: Map<string, Transport>
-  private readonly listeners: Map<string, Listener[]>
+  /** Map<TransportKey, Map<Multiaddr Listener>> */
+  private readonly listeners: Map<string, Map<Multiaddr, Listener>>
```
achingbrain, line 33: I may be missing something, but this change seems to have been made just to support the `removeListener` method, itself only added to be invoked in a test.

Can this be reverted?

achingbrain (changes requested):
Still a bit of work to do, I think the changes to `DefaultTransportManager` can be reverted, all this PR really needs is the `'should remove the relay event listener when the relay stops'` test and the code to make it pass.

packages/libp2p/src/transport-manager.ts
```diff
@@ -262,27 +257,27 @@ export class DefaultTransportManager implements TransportManager, Startable {
...
+    for (const listener of this.listeners.get(key)?.values() ?? []) {
+      await listener.close()
     }
-    await Promise.all(tasks)
-
     this.transports.delete(key)
     this.listeners.delete(key)
   }
 
+  async removeListener (ma: Multiaddr): Promise<void> {
```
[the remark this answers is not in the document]
wemeetagain: I thought this is what circuit relay would call to remove a disconnected relay?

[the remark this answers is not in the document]
maschad: @achingbrain is right, this [test](https://github.com/libp2p/js-libp2p/blob/master/packages/libp2p/test/circuit-relay/relay.node.ts#L188) ensures the listener is removed in that case. 

@wemeetagain I think I was mislead in the debugging process as the event listener was being removed, and I didn't see where the transport's listener was being explicitly removed, leading me to believe it was not being removed.

As @achingbrain explained [here](https://github.com/libp2p/js-libp2p/pull/1998#discussion_r1356687919) the _close_ event being emitted by the Listener is what causes it to be removed from the Transport Manager's listeners.

packages/libp2p/src/circuit-relay/transport/index.ts
```diff
@@ -296,7 +296,7 @@ class CircuitRelayTransport implements Transport {
   /**
    * Create a listener
    */
-  createListener (options: CreateListenerOptions): Listener {
+  createListener (): Listener {
```
wemeetagain, line 299: Is this change needed?

achingbrain: Doesn't look like it, no.

achingbrain (approved):

packages/libp2p/src/circuit-relay/transport/index.ts
maschad: The argument is not being used so I am not sure why it's being passed.

PR #2013, opened 2023-08-31
fix(libp2p): update peer store with supported protocols after unhandle

achingbrain:
When unhandling a protocol, update the peer store entry for the current peer with the correct set of protocols.


SgtPooki (approved):
Lgtm

PR #2014, opened 2023-09-01
fix(@libp2p/mdns): do not send TXT records that are too long

### LINKED ISSUE #2012: mDNS peer discovery stops working once a certain threshold is reached
mDNS peer discovery works by making a broadcast query for `_p2p._udp.local` services.  Peers respond with their multiaddrs.

The number of multiaddrs a peer has affects the size of the UDP message being sent as the mDNS response.

Once this message hits a certain size the `multicast-dns` module fails to decode the message.  The message size this occurs at is well under the maximum size for a mDNS response message.

Refs: https://github.com/mafintosh/multicast-dns/issues/85

achingbrain:
The data field of mDNS TXT records has a [hard limit of 255 characters](https://agnat.github.io/node_mdns/user_guide.html#txt_records) - some multiaddrs can be longer than this so filter them out before sending, otherwise remote peers can fail to parse our mDNS response.

We should also only be sending [link-local addresses](https://github.com/libp2p/specs/blob/master/discovery/mdns.md#issues) in mDNS responses so filter any non-link-local addresses out too, though go-libp2p includes loopback addresses even though it probably shouldn't (according to the mDNS spec which conflicts with our mDNS Peer Discovery spec) so we continue to do so as well.

Fixes #2012

### DIFF
--- a/packages/peer-discovery-mdns/src/index.ts
+++ b/packages/peer-discovery-mdns/src/index.ts
@@ -6,6 +6,7 @@ import * as query from './query.js'
 import { stringGen } from './utils.js'
 import type { PeerDiscovery, PeerDiscoveryEvents } from '@libp2p/interface/peer-discovery'
 import type { PeerInfo } from '@libp2p/interface/peer-info'
+import type { Startable } from '@libp2p/interface/src/startable.js'
 import type { AddressManager } from '@libp2p/interface-internal/address-manager'
 
 const log = logger('libp2p:mdns')
@@ -23,7 +24,7 @@ export interface MulticastDNSComponents {
   addressManager: AddressManager
 }
 
-class MulticastDNS extends EventEmitter<PeerDiscoveryEvents> implements PeerDiscovery {
+class MulticastDNS extends EventEmitter<PeerDiscoveryEvents> implements PeerDiscovery, Startable {
   public mdns?: multicastDNS.MulticastDNS
 
   private readonly broadcast: boolean
@@ -50,9 +51,10 @@ class MulticastDNS extends EventEmitter<PeerDiscoveryEvents> implements PeerDisc
     this.port = init.port ?? 5353
     this.components = components
     this._queryInterval = null
-    this._onPeer = this._onPeer.bind(this)
     this._onMdnsQuery = this._onMdnsQuery.bind(this)
     this._onMdnsResponse = this._onMdnsResponse.bind(this)
+    this._onMdnsWarning = this._onMdnsWarning.bind(this)
+    this._onMdnsError = this._onMdnsError.bind(this)
   }
 
   readonly [peerDiscovery] = this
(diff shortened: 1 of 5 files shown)


packages/peer-discovery-mdns/src/index.ts
```diff
@@ -113,14 +117,12 @@ class MulticastDNS extends EventEmitter<PeerDiscoveryEvents> implements PeerDisc
     }
   }
 
-  _onPeer (evt: CustomEvent<PeerInfo>): void {
-    if (this.mdns == null) {
-      return
-    }
+  _onMdnsWarning (err: Error): void {
+    log.error('mdns warning', err)
```
SgtPooki, line 121: Should this be .error or .info?

packages/peer-discovery-mdns/src/query.ts
```diff
@@ -92,20 +103,45 @@ export function gotQuery (qry: QueryPacket, mdns: MulticastDNS, peerName: string
...
+      // mDNS requires link-local addresses only
+      // https://github.com/libp2p/specs/blob/master/discovery/mdns.md#issues
+      .filter(isLinkLocal)
+      .forEach((addr) => {
+        const data = 'dnsaddr=' + addr.toString()
+
+        // TXT record fields have a max data length of 255 bytes
+        // see 6.1 - https://www.ietf.org/rfc/rfc6763.txt
+        if (data.length > 255) {
+          log('multiaddr %a is too long to use in mDNS query response', addr)
```
SgtPooki, line 116: This will likely lead to a lot of circuit relay webtransport multiaddrs failing, won't it?

packages/peer-discovery-mdns/test/multicast-dns.spec.ts
```diff
@@ -141,15 +142,6 @@ describe('MulticastDNS', () => {
     await stop(mdnsC)
   })
 
-  it('should start and stop with go-libp2p-mdns compat', async () => {
```
SgtPooki, line 144: Why remove this test?

packages/peer-discovery-mdns/test/multicast-dns.spec.ts
```diff
@@ -219,4 +211,80 @@ describe('MulticastDNS', () => {
...
+
+    const { detail: { id, multiaddrs } } = await new Promise<CustomEvent<PeerInfo>>((resolve) => {
+      mdnsA.addEventListener('peer', resolve, {
+        once: true
+      })
+    })
+
+    expect(pB.toString()).to.eql(id.toString())
+
+    ;[
```
SgtPooki, line 267: Pre-colon? 🤢

SgtPooki (approved):
The tests are hard to follow but the src code changes make sense.

packages/peer-discovery-mdns/test/multicast-dns.spec.ts
achingbrain: Historically go-libp2p-mdns didn't speak spec-compliant mDNS so we had a compatibility layer.

go-libp2p-mdns has since been updated to behave better so the compatibility layer was removed, this test should have gone too.

PR #2023, opened 2023-09-05
deps(@multidim-interop): add aegir to deps

maschad:
This whole package should have `aegir` as a dep since it's core functionality is built around testing js-libp2p with `aegir` via the [test command](https://github.com/libp2p/js-libp2p/blob/master/interop/package.json#L49) but especially In order to utilize `aegir` for browser multidim interop testing we should add `aegir` as a dependency here.

Related: https://github.com/libp2p/test-plans/pull/288


p-shahi (approved):
rubber stamping

Issue #2010, opened 2023-08-31
Introduce smart dialing in js-libp2p


p-shahi:
## Goal

Reduce the number of outbound spurious dials that result in failed connections. This is especially a problem when dialing WebTransport addresses in Chromium where we often hit the 64 session limit https://github.com/libp2p/js-libp2p/issues/1896

Ref: notes on smart dialing in go-libp2p https://github.com/libp2p/go-libp2p/releases/tag/v0.29.0
```[tasklist]
### Tasks
- [x] fix(libp2p): only dial one address at a time for peers #2028
- [x] fix(libp2p): sort addresses to dial as public, then relay #2031
- [x] Increase no-redial threshold in browsers to greater than the WebTransport error timeout (fixed in #2028)
- [ ] https://github.com/libp2p/js-libp2p/pull/2033
- [ ] https://github.com/libp2p/js-libp2p/issues/2114
- [ ] https://github.com/libp2p/js-libp2p/issues/2150
```

maschad:
It would be good to dog-feed this in https://github.com/libp2p/universal-connectivity to see how effective our dialling is after this refactor

dhuseby:
Done!

PR #2028, opened 2023-09-07
fix(libp2p): only dial one address at a time for peers

achingbrain:
The observation for dialing is that for a given peer with multiple addresses, for a given class of address (e.g. public or private), either the first dial succeeds or they all fail, so there's not much point dialing all addresses together since in either case you waste resources either cancelling redundant dials or dialing addresses that fail.

We can't do much about dialing addresses that fail but we can reduce the number of redundant dials so with this change we only ever dial one multiaddr at a time.

Refs: #2010

### DIFF
--- a/packages/libp2p/src/connection-manager/constants.browser.ts
+++ b/packages/libp2p/src/connection-manager/constants.browser.ts
@@ -1,11 +1,6 @@
 export * from './constants.defaults.js'
 
 /**
- * @see https://libp2p.github.io/js-libp2p/interfaces/index._internal_.ConnectionManagerConfig.html#maxParallelDials
- */
-export const MAX_PARALLEL_DIALS = 10
-
-/**
  * @see https://libp2p.github.io/js-libp2p/interfaces/index._internal_.ConnectionManagerConfig.html#minConnections
  */
 export const MIN_CONNECTIONS = 5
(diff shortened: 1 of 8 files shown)


packages/libp2p/src/connection-manager/index.ts
```diff
@@ -257,6 +257,7 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
       addressSorter: init.addressSorter ?? publicAddressesFirst,
       maxParallelDials: init.maxParallelDials ?? MAX_PARALLEL_DIALS,
       maxPeerAddrsToDial: init.maxPeerAddrsToDial ?? MAX_PEER_ADDRS_TO_DIAL,
+      maxParallelDialsPerPeer: init.maxParallelDialsPerPeer ?? MAX_PARALLEL_DIALS_PER_PEER,
```
maschad, line 260: Where is this value being used?

packages/libp2p/src/connection-manager/constants.browser.ts
```diff
@@ -16,6 +11,11 @@ export const MIN_CONNECTIONS = 5
...
- * @see https://libp2p.github.io/js-libp2p/interfaces/index._internal_.ConnectionManagerConfig.html#autoDialConcurrency
+ * @see https://libp2p.github.io/js-libp2p/interfaces/index._internal_.ConnectionManagerConfig.html#maxParallelDials
+ */
+export const MAX_PARALLEL_DIALS = 50
+
+/**
+ * @see https://libp2p.github.io/js-libp2p/interfaces/libp2p.index.unknown.ConnectionManagerInit.html#autoDialPeerRetryThreshold
  */
-export const AUTO_DIAL_CONCURRENCY = 10
+export const AUTO_DIAL_PEER_RETRY_THRESHOLD = 1000 * 60 * 7
```
maschad, line 21: Why is this being increased to 7 minutes?

maschad (changes requested):
I am not sure how this ensures we are only dialling one multiaddr at time, wouldn't there need to be some changes to the dial queue?

packages/libp2p/src/connection-manager/constants.browser.ts
achingbrain: To be slightly longer than the WebTransport error timeout - see https://bugs.chromium.org/p/chromium/issues/detail?id=1473980

Granted this could be a separate PR.

packages/libp2p/src/connection-manager/index.ts
achingbrain: https://github.com/libp2p/js-libp2p/blob/master/packages/libp2p/src/connection-manager/dial-queue.ts#L428

achingbrain:
> wouldn't there need to be some changes to the dial queue?

No, we already have the ability to control how many parallel dials we do per peer as a parameter so all we need to do here is default it to 1.

maschad:
I see, my understanding was that we would remove parallel dialling per peer all together, but this offers better flexibility.

maschad (approved):

PR #2030, opened 2023-09-08
fix(libp2p): prioritize dialling peers that have been recently connected

### LINKED ISSUE #1953: feat: Prioritize peers based on previous dial succession / failure
Following from the conversation [here](https://github.com/libp2p/js-libp2p/pull/1943#discussion_r1288617849), 

If a peer recently failed to establish a connection, there might be underlying issues causing the failure that haven't been resolved yet (may take longer than 10 minutes), and so it's far less likely that a successful connection would result from dialling that peer, as opposed to a peer which was recently successfully dialled but for whatever a reason that connection was closed. 

Modifying the tag value is not ideal. The problem is the connection manager doesn't know what it means to penalize a peer, because it doesn't know what the value of a tag added by one component means compared to a tag added by another so it cannot alter the tag values without possibly compromising the functioning of the components doing the tagging.

So  if we've successfully dialled the peer recently, but we are no longer connected to them, we should prioritize those peers and conversely if we have dialled a peer but failed recently, those should be de-prioritized. 

cc @achingbrain @wemeetagain @p-shahi

### LINKED ISSUE #1955: Circuit relay - reconnect to peers we have reservations on
When we have a reservation on a relay, and we lose connectivity to that relay, we should make an effort to reconnect to the same relay before searching for other relays to make a reservation on.

This is because making a new reservation will change our multiaddrs which impacts things like DHT provider records which will now be out of date just because our WiFi was a bit flaky for a time or we closed our laptop lid.

maschad:
if we've successfully dialled the peer recently, but we are no longer connected to them, we should prioritize dialling those peers.

Closes #1953
Closes #1955


maschad:
I think we should land https://github.com/libp2p/js-libp2p/pull/2028 first given that will impact the dialling behaviour, primarily [this test](https://github.com/libp2p/js-libp2p/pull/2030/files#diff-4e316fbf927d458976e70ce5db11c866e957eea55fe88cf84b26dac4f58683d8R296)

maschad:
Closing in favour of https://github.com/libp2p/js-libp2p/pull/2033

PR #2033, opened 2023-09-08
feat: track failed/successful dials per-address

achingbrain:
Instead of tracking which peer we failed to dial, track dial success or failure on a per-address basis.

This will let us detect when IPv6 is unsupported or certain transports (e.g. UDP ones) are blocked.

Refs: 2010

### DIFF
--- a/packages/libp2p/src/connection-manager/dial-queue.ts
+++ b/packages/libp2p/src/connection-manager/dial-queue.ts
@@ -464,10 +464,54 @@ export class DialQueue {
               // update dial status
               pendingDial.status = 'active'
 
-              const conn = await this.transportManager.dial(addr, {
-                ...options,
-                signal
-              })
+              let conn: Connection
+
+              try {
+                conn = await this.transportManager.dial(addr, {
+                  ...options,
+                  signal
+                })
+
+                const peerData = await this.peerStore.get(conn.remotePeer)
+                const addresses = peerData.addresses.map((address) => {
+                  if (address.multiaddr.equals(addr)) {
+                    return {
+                      ...address,
+                      lastSuccess: Date.now()
+                    }
+                  }
+
+                  return address
+                })
+
+                // mark multiaddr dial as successful
+                await this.peerStore.merge(conn.remotePeer, {
+                  addresses
+                })
+              } catch (err: any) {
+                if (pendingDial.peerId != null) {
+                  // mark multiaddr dial as failure
+                  const peerData = await this.peerStore.get(pendingDial.peerId)
+                  const addresses = peerData.addresses.map((address) => {
+                    if (address.multiaddr.equals(addr)) {
+                      return {
+                        ...address,
+                        lastFailure: Date.now()
+                      }
+                    }
+
+                    return address
+                  })
+
+                  // mark multiaddr dial as failed
+                  await this.peerStore.merge(pendingDial.peerId, {
+                    addresses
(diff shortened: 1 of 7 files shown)


achingbrain:
Not ready for merge, just for discussion/experimentation for now.

packages/peer-store/test/merge.spec.ts
```diff
@@ -244,4 +241,53 @@ describe('merge', () => {
     expect(updated).to.have.property('tags').that.deep.equals(original.tags)
     expect(updated).to.have.property('protocols').that.deep.equals(original.protocols)
   })
+
+  it('merges addresses', async () => {
```
SgtPooki, line 245:
```suggestion
  it('merges lastFailure & lastSuccess on peer addresses', async () => {
```

packages/libp2p/src/connection-manager/dial-queue.ts
```diff
@@ -464,10 +456,25 @@ export class DialQueue {
...
+                  signal
+                })
+
+                // mark multiaddr dial as successful
+                await this._updateAddressStatus(conn.remotePeer, addr, true)
+              } catch (err: any) {
+                if (pendingDial.peerId != null) {
+                  // mark multiaddr dial as failure
+                  await this._updateAddressStatus(pendingDial.peerId, addr, false)
+                }
```
SgtPooki, line 473: This check implies that we may also attempt to dial peers where peerId == null. Can we log when that is the case, and that we're not marking peer as having a failed dial?

SgtPooki (commented):
just looking through the code.. some minor improvements and a question but in general it looks reasonable. 

There doesn't seem to be any code utilizing the lastSuccess or lastFailed yet, but from the title of the PR that seems intentional.

packages/peer-store/src/utils/dedupe-addresses.ts
```diff
@@ -26,26 +26,51 @@ export async function dedupeFilterAndSortAddresses (peerId: PeerId, filter: Addr
...
+    if (addr.isCertified === true) {
+      existingAddr.isCertified = true
+    }
+
+    if (addr.lastFailure != null) {
+      existingAddr.lastFailure = Number(addr.lastFailure)
+    }
+
+    if (addr.lastSuccess != null) {
+      existingAddr.lastSuccess = Number(addr.lastSuccess)
```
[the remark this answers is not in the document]
achingbrain: The value is a timestamp, the max value of which can exceed 32 bits so we need to use a 64 bit number in the protobuf to store it, which means we need to represent it as a BigInt, but this is overkill for actual time values.

We can refactor this once https://github.com/ipfs/protons/issues/112 is implemented to have protons serialize/deserialize to `Number` instead of `BigInt`.

packages/peer-store/test/merge.spec.ts
achingbrain: This test also ensures `multiaddr` and `isCertified` values are merged and not lost so they are missing from the test name if you want it to be exhaustive, or it can remain as it is.

maschad (commented):
Following from previous conversations I think this is the right direction, is this ready for review @achingbrain ?

maschad:
LGTM but we should perhaps merge https://github.com/libp2p/js-libp2p/issues/2150 into this before merging this as it effectively adds no enhancement without the sorting functionality.

PR #2037, opened 2023-09-09
docs(@libp2p/mdns): update mdns defaults in README.md

futpib:
It's clear from the code that `broadcast` and `serviceTag` defaults are not what docs said https://github.com/libp2p/js-libp2p/blob/5294f14caa314bb150554afff3a7ff45d2bf17ba/packages/peer-discovery-mdns/src/index.ts#L42-L44


maschad (approved):
Thanks!

achingbrain (approved):

PR #2042, opened 2023-09-12
fix(@libp2p/webrtc): use stream logger instead of global logger

achingbrain:
Streams should use the `this.log` property for logging to ensure the log messages are scoped to an invidual stream.


maschad:
It's less obvious than if it was initialized in the constructor 🤷🏾

achingbrain:
My opinion is not very strong, but as it is it's consistent with yamux and mplex and WebTransport doesn't have per-stream loggers.  This can be done as a follow up PR if necessary, it is out of scope here.

maschad (approved):

PR #2048, opened 2023-09-13
fix(@libp2p/webrtc): delay closing datachannels

achingbrain:
Tactical fix until we can update the spec to coordinate closing datachannels.

Calling `.close` on a datachannel causes a stream reset - depending on the implementation there's a good chance any data waiting to be transmitted will be lost.

As a workaround when closing a datachannel wait until it's `.bufferedAmount` property is `0`, then wait a bit longer before closing the channel.

This is because although Firefox will have sent all the data by this point, Chrome may have not.

This should be removed when the spec is updated.

Refs: https://github.com/libp2p/specs/issues/575


packages/transport-webrtc/src/muxer.ts
```diff
@@ -9,6 +9,7 @@ import type { Source, Sink } from 'it-stream-types'
 import type { Uint8ArrayList } from 'uint8arraylist'
 
 const PROTOCOL = '/webrtc'
+const DEFAULT_CHANNEL_CLOSE_DELAY = 30000
```
maschad, line 12: How'd you arrive at this time? 30 seconds seems a bit long...

packages/transport-webrtc/src/util.ts
```diff
@@ -1,8 +1,59 @@
...
+      }, channelCloseDelay)
+
+      if (timeout.unref != null) {
+        timeout.unref()
+      }
+
+      await deferred.promise
+    })
+    .catch(err => {
+      log.error('error closing outbound stream', err)
```
maschad, line 57: Do you think we should bubble this error up? i.e. throw an `ERR_DATA_CHANNEL` ?

maschad (commented):
Looks good, I have two questions. Also I do wonder how frequently this is happening, do you think it would make sense to instrument this to measure that?

achingbrain:
> I do wonder how frequently this is happening

Pretty frequently, it turns out.  It's what causes [this kind](https://github.com/libp2p/js-libp2p/actions/runs/6151079913/job/16690547506?pr=1998) of CI failure.

packages/transport-webrtc/src/util.ts
achingbrain: I thought about it, but it would make the stream `onEnd` callback need to be async which seems wrong.

From testing the only time this throws is when the datachannel is already closed but this PR guards on the `readyState` before closing so the `catch` is really just here to catch edge cases.

packages/transport-webrtc/src/muxer.ts
achingbrain: I plucked it out of thin air.  It is configurable if users want something shorter.

maschad:
> Pretty frequently, it turns out. It's what causes [this kind](https://github.com/libp2p/js-libp2p/actions/runs/6151079913/job/16690547506?pr=1998) of CI failure.

Have you ran that test a few times on CI to ensure that this fix solves it?

packages/transport-webrtc/src/muxer.ts
maschad: Gotcha, I suppose it will vary based on use case, and this is a temp fix anyway.

maschad:
As discussed in triage, this may actually be an issue with Chrome, @achingbrain will file a report.

maschad:
@achingbrainI I am re-opening this as it will unblock CI without us having to skip the test entirely, until https://bugs.chromium.org/p/chromium/issues/detail?id=1484907 is resolved.

maschad:
Closing in favour of https://github.com/libp2p/js-libp2p/pull/2073

PR #2049, opened 2023-09-13
fix: opt-in to toplogy notifications on transient connections

### LINKED ISSUE #2036: How to use topology changes with WebRTC connections?
- **Version**:

0.46.9 

- **Platform**:

- **Subsystem**:

Registrar. WebRTC transport. 

#### Severity:

Medium

#### Description:

I am making another attempt to make the WebRTC transport to work again with my project..

But I have a problem: 

In my protocol handler I listen for topology changes and look for new connection and open streams to them. 

```
this.components.registrar.register(multicodec, {
	onConnect: (....) => { open stream }
})

```
With WebRTC transport, I will receive 1 signaling connection that is transient which I am ignoring in my protocol handler because that one will close once a direct WebRTC connection has been established (expected).

However, once this direct WebRTC connection has been established (verified by logging and using a debugger in the browser)  I expected the topology again to emit an onConnect event with the non transient connection. However it seems like the change detection  in [registrar](https://github.com/libp2p/js-libp2p/blob/5294f14caa314bb150554afff3a7ff45d2bf17ba/packages/libp2p/src/registrar.ts#L28) does not take into account that the last connection was transient. I never see a `onConnect` event for the WebRTC connection (perhaps since the signaling connection is closing after the WebRTC connection has started?).  (unexpected). 

- What events can I use to effectively listen for new incoming connections that are non-transient for my protocol handler? The topology changes does work when working with other transports but in this case I can no longer rely on it, or there is a bug/unintended behaviour. I could listen to the "connection:open" but that might lead to flakyiness as I don't yet perhaps know if the remote peer is supporting the protocol handler.

achingbrain:
Adds a `notifyOnTransient` option when registering a network topology to opt-in to being notified when peers that support the registered protocol connect over transient connections.  False by default.

The logic has been switched to check each connection's transient property and only notify if the user has opted in.

The side effect here is that if `notifyOnTransient` is true, and the peer ends up opening a direct connection (for example they dial us via circuit relay, open a stream to do the WebRTC SDP exchange, then open a WebRTC connection), identify will run on the second connection so the topology will receive two notifications.

This is not a breaking change since the previous behaviour would have been to only notify on the initial transient connection, which you can't do data-heavy things like bitswap over, or long-lived things like GossipSub so is probably a bug.

Fixes #2036


maschad (approved):

achingbrain:
Still needs a bit of work, it's missing the case where a peer opens a transient connection, then opens a second non-transient connection.

joeltg:
Hate to nag but is there an ETA on this? GossipSub over WebRTC is completely broken so this is blocking new releases for us 😖

achingbrain:
> Still needs a bit of work, it's missing the case where a peer opens a transient connection, then opens a second non-transient connection.

I've added a test for this.  The topology will get a notification on each connection that is opened and it'll be down the implementation to not duplicate peer operations, etc.  It'll only be notified on transient connections if it ops in to notifications, otherwise it'll just get them for non-transient connections.

maschad (approved):

joeltg:
I've been testing this branch locally and I think there's still something missing.

Currently, the registrar only notifies topologies of new connections from within the `Registrar._onPeerUpdate` handler, which diffs the updated protocols with the existing stored protocols to get two filtered protocol arrays `added` and `removed`. Then, for each topology registered for the protocols in `added`, it calls the topology's `onConnect` callback with every open connection, skipping transient connections that aren't opted-into.

The problem with this is with the protocol diff / filtering at the beginning. For example, when I try to use GossipSub over WebRTC in the browser, the following happens:

1. peer A opens a transient connection to peer B via circuit-relay
2. both peers perform `identify` over the circuit-relay connection, exchanging all registered protocols and updating their peer store. the gossipsub topology is NOT notified because the connection is transient
3. the peers establish a WebRTC connection
4. both peers perform `identify` again, over the WebRTC connection, exchanging the same set of protocols
5. nothing happens because `added` is empty

joeltg:
I don't think there's not a straightforward one- or two-line fix for this. I have a patch https://github.com/joeltg/js-libp2p/commit/bef1927d1e0b71b58986e48fadd82a250ba2d74d that adds a `peerSets: Map<string, PeerSet>` to the registrar (one for every topology) and maintains those to tell whether it should notify peers or not, and this works for me in my app. Feel free to PR/cherrypick it in if you want (it's one commit).

I seem to remember that `Topology` used to be a concrete class, not just an interface, with a `PeerSet` attached to it? Maybe that's worth revisiting since it looks like that peer set is something both the registrar and external users want access to.

joeltg:
Awesome, this works locally for me! Thanks so much

Issue #2003, opened 2023-08-30
discrepancies in webrtc mesage lengths


DougAnderson444:
Hello,

I am implementing WebRTC in https://github.com/libp2p/rust-libp2p/pull/4248 and noticed a discrepancy between the Rust impl and the JS Impl, specifically when it comes to `PROTO_OVERHEAD` length

In Rust we have these values:

```rust
// https://github.com/libp2p/rust-libp2p/blob/96213b37c6d7b8ea998d652d245f51b920bb4852/transports/webrtc/src/tokio/substream.rs#L50-L56

const MAX_MSG_LEN: usize = 16384; // 16kiB
/// Length of varint, in bytes.
const VARINT_LEN: usize = 2;
/// Overhead of the protobuf encoding, in bytes.
const PROTO_OVERHEAD: usize = 5;
/// Maximum length of data, in bytes.
const MAX_DATA_LEN: usize = MAX_MSG_LEN - VARINT_LEN - PROTO_OVERHEAD;
```

With the rationale [here in the spec](https://github.com/libp2p/specs/blob/master/webrtc/README.md#faq).

Whereas the javascript is:

https://github.com/libp2p/js-libp2p/blob/9a69e6f76018d2df5c6e51de930557d4b60b2fca/packages/transport-webrtc/src/stream.ts#L43

https://github.com/libp2p/js-libp2p/blob/9a69e6f76018d2df5c6e51de930557d4b60b2fca/packages/transport-webrtc/src/stream.ts#L280

I am looking for the spec so that I can resolve this discrepancy, can you please send me what was used for `js-libp2p`?  Am I missing something as to why there is a difference between rust and js implementations?

maschad:
JS uses the same spec. The discrepancy is likely an oversight, would you like to open a PR to fix this?

PR #2050, opened 2023-09-13
fix(@libp2p/webrtc): set max message size in alignment with spec

maschad:
According to the [webRTC spec](https://github.com/libp2p/specs/blob/master/webrtc/README.md#faq) The Protobuf framing adds an overhead of 5 bytes. The unsigned-varint prefix adds another 2 bytes. This takes that into account.

Closes https://github.com/libp2p/js-libp2p/issues/2003


packages/transport-webrtc/test/stream.spec.ts
```diff
@@ -23,7 +23,8 @@ const MAX_MESSAGE_SIZE = 16 * 1024
 describe('Max message size', () => {
   it(`sends messages smaller or equal to ${MAX_MESSAGE_SIZE} bytes in one`, async () => {
     const sent: Uint8ArrayList = new Uint8ArrayList()
-    const data = new Uint8Array(MAX_MESSAGE_SIZE - 5)
+    const PROTOBUF_OVERHEAD = 5
```
achingbrain, line 26: Maybe this should be exported from `src/stream.ts`, then it wouldn't need to be duplicated here?

achingbrain (approved):
LGTM.  Small nit with duplicating the constants, comment inline.

achingbrain (approved):

PR #2054, opened 2023-09-15
fix(@libp2p/webtransport): handle dialing circuit addresses

achingbrain:
We should be able to dial WebRTC addresses over Circuit Relay over WebTransport but the `parseMultiaddr` function was rejecting them.

Simplifies the parsing function to handle this and future address variants and adds tests.


maschad (approved):

SgtPooki (commented):
👍

Issue #440, opened 2018-04-23
[switch] start and stop accepting connections


pgte:
For each different transport individually, in order to do connection throttling, it's important to be able to start and stop the listener by command of the user.

PR #2058, opened 2023-09-18
fix: ensure all listeners are properly closed on tcp shutdown

nazarhussain:
The lister is implemented the logic for the server to operate in multiple states. During the connection limits, it stops and starts again. We can consider these stats as pause and resume. 

But due to the way `TransportManager` is implemented, as soon the listener is stopped it's been removed from the transport manager and if during the same time the listener connection limit is changed then it's started listening again. 

Because the listener was already removed from the `TransportManager`, it's been been closed when we close the libp2p instance. 

And this behavior can trigger everyone when we close two libp2p instances which are connected to each other with the connection limits of 1. As soon the first instance is disconnected, the second instance connection count reaches to 0 and it closes the listener and transport manager remove the listener, but because 0 is the lower limit to start as well so the listener start listening again. 

Resolves https://github.com/libp2p/js-libp2p/issues/440


nazarhussain:
@achingbrain Yes definitly. I opened it to get initial feedback from team members and will add tests to make sure that the edge case is covered.

packages/transport-tcp/src/listener.ts
```diff
@@ -144,6 +157,8 @@ export class TCPListener extends EventEmitter<ListenerEvents> implements Listene
         this.dispatchEvent(new CustomEvent<Error>('error', { detail: err }))
       })
       .on('close', () => {
+        if(this.status.code === TCPListenerStatusCode.PAUSED) return
+
         this.metrics?.status.update({
           [this.addr]: SERVER_STATUS_DOWN
         })
```
wemeetagain, line 164: Can you move L160 and L161 after this metrics call? Or else add a metric to know when the listener gets paused / unpaused.

wemeetagain (commented):
The approach LGTM

L160-161 is the important change. Before, this listener emitted the `'close'` event when it hit its connection limit. If libp2p is closed at that point, the listener is never cleaned up.

Now, thanks to differentiating paused vs closed listener status, we can only emit the `'close'` event when the listener is actually closed. And the listener will always be cleaned up.

wemeetagain (approved):

wemeetagain:
@achingbrain can we cut a release with this fix? This is causing some chaos in our CI pipeline

packages/transport-tcp/src/listener.ts
```diff
@@ -46,17 +46,25 @@ interface Context extends TCPCreateListenerOptions {
...
 }
 
-type Status = { started: false } | {
-  started: true
+enum TCPListenerStatusCode {
+  /**
+   * When server object is initialized but we don't know the listening address yet or
+   * the server object is stopped manually, can be resumed only by calling listen()
+   **/
+  DOWN = 0,
```
maschad, line 60: Not directly related, but I think these should be renamed to `ACTIVE` , `INACTIVE` or `PAUSED` for better clarity.

packages/transport-tcp/src/listener.ts
```diff
@@ -191,7 +208,7 @@ export class TCPListener extends EventEmitter<ListenerEvents> implements Listene
               // another process during the time the server if closed. In that case there's not much
               // we can do. netListen() will be called again every time a connection is dropped, which
```
maschad, line 209: This comment needs to be updated.

packages/transport-tcp/test/connection-limits.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+    await assertConnectedSocket(1)
+    await assertConnectedSocket(2)
+    await assertConnectedSocket(3)
+    await assertServerConnections(listener, 3)
+
+    // Limit reached, server should be closed but should not emit "close" event
+    expect(closeEventCallCount).equals(0)
+  })
+
+  it('should not emit "listening" event when server is resumed due to "listenBelow" limit', async () => {
```
maschad, line 176: This test is a bit misleading, if we wanted to ensure that a listening event is not emitted when the server resumes, wouldn't we allow the server to resume and ensure that the `listeningEventCallCount` remains unchanged? 

Here that event is being incremented twice in this test

packages/transport-tcp/test/connection-limits.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+  for (let i = 0; i < 100; i++) {
+    // eslint-disable-next-line @typescript-eslint/dot-notation
+    if (listener['connections'].size === connections) {
+      return
+    } else {
+      await promisify(setTimeout)(10)
+    }
+  }
+  // eslint-disable-next-line @typescript-eslint/dot-notation
+  expect(listener['connections'].size).equals(connections, 'Wrong server connections')
```
maschad, line 73:
```suggestion
  expect(listener['connections'].size).equals(connections, 'invalid amount of server connections')
```

packages/transport-tcp/src/listener.ts
```diff
@@ -264,35 +283,42 @@ export class TCPListener extends EventEmitter<ListenerEvents> implements Listene
...
   }
 
   async close (): Promise<void> {
     await Promise.all(
       Array.from(this.connections.values()).map(async maConn => { await attemptClose(maConn) })
     )
 
-    // netClose already checks if server.listening
-    this.netClose()
+    await this.pause(true)
```
maschad, line 314: We should handle any error thrown here

packages/transport-tcp/src/listener.ts
```diff
@@ -145,13 +153,22 @@ export class TCPListener extends EventEmitter<ListenerEvents> implements Listene
...
+        // and the transport manager will not be able to close the server
+        if (this.status.code !== TCPListenerStatusCode.PAUSED) {
+          this.dispatchEvent(new CustomEvent('close'))
+        }
       })
   }
 
   private onSocket (socket: net.Socket): void {
+    if (this.status.code === TCPListenerStatusCode.DOWN) {
+      throw new Error('Server is is not listening yet')
```
maschad, line 170: You can use the [CodeError](https://github.com/libp2p/js-libp2p/blob/1af6add837bff88e923faa4ffe4ad1af8e8783bd/packages/interface/src/errors.ts#L21) class

maschad (changes requested):

packages/transport-tcp/test/connection-limits.spec.ts
nazarhussain: Standard copy paste error. I wanted to "should emit" in this test case.

maschad:
@nazarhussain [this test](https://github.com/libp2p/js-libp2p/actions/runs/6314620990/job/17146786713?pr=2058#step:5:1960) has failed quite a few times, is it passing for you locally?

nazarhussain:
> @nazarhussain [this test](https://github.com/libp2p/js-libp2p/actions/runs/6314620990/job/17146786713?pr=2058#step:5:1960) has failed quite a few times, is it passing for you locally?

Some wiered behavior. 

- If I run whole test suite with `npm run test:node` that runs fine but stuck at the end and never exit though that test seems to pass fine. 
- If I run that only that single file it passes fine and exit fine. 
- If I the whole test suite with `.only` for that test case it passes fine and exits fine.  
- If I do same with `.skip` for the test case it passes but stuck the exit. 

I assume there some setup in any other file which is not cleanup properly. So need to figure which code is the culprit for that behavior.

```
> npm run test:node

> libp2p@0.46.11 test:node
> aegir test -t node -f "./dist/test/**/*.{node,spec}.js" --cov

  circuit-relay
    flows with 1 listener
      ✔ should ask if node supports hop on protocol change (relay protocol) and add to listen multiaddrs
      ✔ should only add discovered relays relayed addresses (71ms)
      ✔ should not listen on a relayed address after we disconnect from peer (1036ms)
      ✔ should try to listen on other connected peers relayed address if one used relay disconnects (84ms)
      ✔ should try to listen on stored peers relayed address if one used relay disconnects and there are not enough connected (109ms)
      ✔ should not fail when trying to dial unreachable peers to add as hop relay and replaced removed ones (62ms)
      ✔ should announce new addresses when using a peer as a relay (1021ms)
      ✔ should announce new addresses when using no longer using peer as a relay

351 passing (45s)
  2 pending
```

```
> npx aegir test -t node -f dist/test/circuit-relay/relay.node.js

build

> libp2p@0.46.11 build
> aegir build

[10:55:02] tsc [started]
[10:55:07] tsc [completed]
[10:55:07] esbuild [started]
[10:55:07] esbuild [completed]
test node.js

....
....
....

26 passing (8s)
```

nazarhussain:
These are the open handlers when tests stuck, none of these are related to the code change in this PR. Trying to debug further. 

```
- File descriptors: (note: stdio always exists)
  - fd 1 (tty) (stdio)
  - fd 0 (tty)
  - fd 2 (tty) (stdio)
- Child processes
  - PID 54509
    - Entry point: node:internal/child_process:255
- Servers:
  - :::55637 (HTTP)
    - Listeners:
      - request: proxy @ file:///js-libp2p/node_modules/it-ws/dist/src/server.js:70
- Intervals:
  - (300000 ~ 5 min) (anonymous) @ file:///js-libp2p/packages/libp2p/dist/src/circuit-relay/server/reservation-store.js:30
```

nazarhussain:
@maschad Please trigger the workflow, hope everything works fine now. The transport manager `remove` logic was different than the similar in the `close`, which was causing not all listeners to close for a transport.

nazarhussain:
That is strange, now everything passing for me locally. 

```
~/projects/js-libp2p/packages/libp2p on nh/listener-bug *1 ......................................... at 16:11:06
> npm run test:node

> libp2p@0.46.11 test:node
> aegir test -t node -f "./dist/test/**/*.{node,spec}.js" --cov

build

> libp2p@0.46.11 build
> aegir build

[16:11:18] tsc [started]
[16:11:25] tsc [completed]
[16:11:25] esbuild [started]
[16:11:26] esbuild [completed]
test node.js


  Address Manager
    ✔ should not need any addresses
    ✔ should return listen multiaddrs on get
    ✔ should return announce multiaddrs on get
    ✔ should add observed addresses
    ✔ should allow duplicate listen addresses
    ✔ should dedupe added observed addresses
    ✔ should only set addresses once (1504ms)
    ✔ should strip our peer address from added obs

....
....
....
  peer job queue
    ✔ should have jobs


  351 passing (43s)
  2 pending

~/projects/js-libp2p/packages/libp2p on nh/listener-bug *1 .............................. took 1m 4s at 16:12:16
```

wemeetagain:
I ran this locally
```bash
~/Code/js-libp2p/packages/libp2p$ npx aegir test -t node -f dist/test/circuit-relay/relay.node.js
build

> libp2p@0.46.11 build
> aegir build

[23:44:40] tsc [started]
[23:44:44] tsc [completed]
[23:44:44] esbuild [started]
[23:44:44] esbuild [completed]
test node.js


  circuit-relay
    flows with 1 listener
      ✔ should ask if node supports hop on protocol change (relay protocol) and add to listen multiaddrs (368ms)
      ✔ should only add discovered relays relayed addresses (382ms)
      1) should not listen on a relayed address after we disconnect from peer


  2 passing (1m)
  1 failing

  1) circuit-relay
       flows with 1 listener
         should not listen on a relayed address after we disconnect from peer:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/cayman/Code/js-libp2p/packages/libp2p/dist/test/circuit-relay/relay.node.js)
      at listOnTimeout (node:internal/timers:573:17)
      at processTimers (node:internal/timers:514:7)

```

nazarhussain:
**Update:** Verified that it's the issue between Mac/Linux, same test is passing on Mac but not on Linux for the same version of Node. 

Debugging it further.

nazarhussain:
That is really strange, all test passing (except for one due to VPS network) for me on the Linux VPS. 

```
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.2 LTS
Release:        22.04
Codename:       jammy
```

```
:~/libp2p$ node -v
v18.12.1

:~/libp2p$ npm -v
8.19.2
```

```
:~/libp2p$ npm run test:node

> libp2p@0.46.11 test:node
> aegir test -t node -f "./dist/test/**/*.{node,spec}.js" --cov

!!! Temporarily disabling code coverage

build


> libp2p@0.46.11 build
> aegir build

[09:48:30] tsc [started]
[09:48:33] tsc [completed]
[09:48:33] esbuild [started]
[09:48:33] esbuild [completed]
test node.js
....
....
....
  267 passing (2m)
  2 pending
  1 failing

  1) peer discovery scenarios
       MulticastDNS should discover all peers on the local network:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/devops/nazar-dev/js-libp2p/packages/libp2p/dist/test/peer-discovery/index.node.js)
      at listOnTimeout (node:internal/timers:564:17)
      at processTimers (node:internal/timers:507:7)

```

packages/libp2p/src/transport-manager.ts
```diff
@@ -262,12 +262,22 @@ export class DefaultTransportManager implements TransportManager, Startable {
...
   async remove (key: string): Promise<void> {
-    log('removing %s', key)
+    const listeners = this.listeners.get(key) ?? []
+    log('removing transport %s', key)
 
     // Close any running listeners
-    for (const listener of this.listeners.get(key) ?? []) {
-      await listener.close()
+    const tasks = []
+    log('closing listeners for %s', key)
```
maschad, line 270:
```suggestion
    log.trace('closing listeners for %s', key)
```

wemeetagain (approved):
LGTM

packages/libp2p/src/transport-manager.ts
```diff
@@ -262,12 +262,22 @@ export class DefaultTransportManager implements TransportManager, Startable {
    * If a transport has any running listeners, they will be closed.
    */
   async remove (key: string): Promise<void> {
-    log('removing %s', key)
+    const listeners = this.listeners.get(key) ?? []
+    log('removing transport %s', key)
```
maschad, line 266: I don't think we need these logs enabled by default.
```suggestion
    log.trace('removing transport %s', key)
```

maschad (approved):
LGTM, ~~left some minor suggestions~~. Thanks again for these improvements!

wemeetagain:
Awesome, can we get this cut in a release? This bug is currently breaking some Lodestar CI tests

nazarhussain:
@maschad @wemeetagain Thanks for reviewing and merging this PR.

maschad:
@wemeetagain @nazarhussain master has been [released](https://github.com/libp2p/js-libp2p/pull/2068) it should appear in libp2p `v0.46.12` and `@libp2p/tcp` `v8.0.8`

wemeetagain:
:pray:

PR #2059, opened 2023-09-18
deps(@libp2p/websockets): upgrade multiaddr-to-uri to parse tls addrs

### LINKED ISSUE #2024: bug: missing support for `/tls` multiaddr
- **Version**: 0.46.x
- **Platform**: macos
- **Subsystem**: Dialer

#### Description:
js-libp2p throws an error `Unsupported protocol tsl` when a multiaddr `/ip4/192.168.68.58/tcp/433/tls/ws/p2p/16Uiu2HAkyZuLQy4sPRte7khemYHmkQewzjdTPRXZcxPH6NgjhR97` is dialed

The expectation is that a secure websocket multiaddr would have `/wss/` but apparently `wss` was replaced by `/tls/ws`:
- https://github.com/multiformats/go-multiaddr/pull/153
- https://github.com/multiformats/multiaddr/pull/109
- https://github.com/multiformats/multiaddr/issues/163

Try to dial `/ip4/192.168.68.58/tcp/433/tls/ws/p2p/16Uiu2HAkyZuLQy4sPRte7khemYHmkQewzjdTPRXZcxPH6NgjhR97` through js-libp2p

danisharora099:
Closes #2024


achingbrain:
These are all in-range semver updates, if you are not seeing these versions when you install `@libp2p/websockets`, please delete your lockfiles.

maschad:
Whilst I agree @achingbrain I still think it's valid to set `@multiformats/multiaddr-to-uri` as a version higher than `9.0.5` given [the fix](https://github.com/multiformats/js-multiaddr-to-uri/pull/120) that introduced support for `tls` addrs.

Consumers who upgrade to `@libp2p/websockets` `7.0.7` or later will not have to be told to delete their lockfile, and it's routine to upgrade a package when encountering an issue.

achingbrain:
Updating minors for every dependency is busywork that creates needless release churn.

This is precisely what semver ranges are supposed to prevent.

maschad:
Actually this PR is upgrading  patch versions -  but `@multiformats/multiaddr-to-uri` should have had a **minor** upgrade i.e. from `9.0.4` -> `9.1.0` since that fix [added new backwards compatible functionality](https://semver.org/#spec-item-7) to `multiaddrToUri`,  but that was an oversight, and so this is an extenuating circumstance. I agree creating PRs to upgrade patch versions should not be a standard practice.

achingbrain:
> but @multiformats/multiaddr-to-uri should have had a minor upgrade i.e. from 9.0.4 -> 9.1.0

This would still have been an in-range semver update.

maschad:
That's true, I forgot that Caret Ranges only presume breaking changes in minor upgrades for packages versioned less than `1.0.0 `

PR #2061, opened 2023-09-19
feat: collect dial/listen metrics in webrtc and webtransport

achingbrain:
Use the documented libp2p metrics system to collect metrics for the webrtc and webtransport transports

### DIFF
--- a/packages/transport-webtransport/src/index.ts
+++ b/packages/transport-webtransport/src/index.ts
@@ -14,7 +14,7 @@ import type { Source } from 'it-stream-types'
 import type { MultihashDigest } from 'multiformats/hashes/interface'
 
 interface WebTransportSessionCleanup {
-  (closeInfo?: WebTransportCloseInfo): void
+  (metric: string): void
 }
 
 const log = logger('libp2p:webtransport')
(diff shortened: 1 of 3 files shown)


maschad (approved):

packages/transport-webtransport/src/index.ts
```diff
@@ -93,15 +110,32 @@ class WebTransportTransport implements Transport {
             return
           }
 
-          wt.close(closeInfo)
+          this.metrics?.dialerEvents.increment({ close: true })
+          wt.close()
```
SgtPooki, line 114: Whyd we get rid of closeInfo

SgtPooki (approved):
LGTM with question

packages/transport-webtransport/src/index.ts
achingbrain: We use it as a cheap error object, but that's not what it does and it affects the WebTransport after closing.

From step 9 in the [cleanup procedure](https://www.w3.org/TR/webtransport/#webtransport-procedures):

> If closeInfo is given, then set transport.[[[State]]](https://www.w3.org/TR/webtransport/#dom-webtransport-state-slot) to "closed". Otherwise, set transport.[[[State]]](https://www.w3.org/TR/webtransport/#dom-webtransport-state-slot) to "failed".

SgtPooki: ok makes sense, good to know

PR #2067, opened 2023-09-21
feat!: measure transfer perf over time

### LINKED ISSUE #2064: feat(@libp2p/perf): Continously measure Performance on a Single Connection
The motivation for the larger effort is documented in https://github.com/libp2p/test-plans/issues/261. 

Instead of downloading an arbitrary amount of  MB multiple times, each on a new connection, establish a single connection and continuously measure the throughput for a fixed duration (60s by default). Perf `stdout` output would be refactored as such : 

 ``` typescript
               interface Data {
                 type: "intermediary" | "final";
                 timeSeconds: number;
                 uploadBytes: number;
                 downloadBytes: number;
               }
```

where every second the client must print the current progress to `stdout`. See example below. Note the `type: "intermediary"`.
         
``` json
               {
                 "type": "intermediary",
                 "timeSeconds": 1.004957645,
                 "uploadBytes": 73039872,
                 "downloadBytes": 0
               }
```
          
Before terminating, the client must print a final summary. See example below. Note the `type: "final"`. Also note that the measurement includes the time to (1) establish the connection, (2) upload the bytes and (3) download the bytes.
      
``` json
               {
                 "type": "final",
                 "timeSeconds": 60.127230659,
                 "uploadBytes": 4382392320,
                 "downloadBytes": 0
               }
```

maschad:
Closes #2064

BREAKING CHANGE:  `measurePerformance` now returns an async generator that yields `PerfOutput`s and no longer accepts the `startTime` parameter


packages/protocol-perf/src/index.ts
```diff
@@ -52,12 +52,19 @@ import type { AbortOptions } from '@libp2p/interfaces'
 const log = logger('libp2p:perf')
 
 export const defaultInit: PerfServiceInit = {
-  protocolName: '/perf/1.0.0',
+  protocolName: '/perf/1.2.0',
```
mxinden, line 55: There is no change on the network layer, right? Why bump the minor version?

packages/protocol-perf/src/index.ts
```diff
@@ -52,12 +52,19 @@ import type { AbortOptions } from '@libp2p/interfaces'
...
 export interface PerfService {
-  measurePerformance: (startTime: number, connection: Connection, sendBytes: bigint, recvBytes: bigint, options?: AbortOptions) => Promise<number>
+  measurePerformance: (connection: Connection, sendBytes: bigint, recvBytes: bigint, options?: AbortOptions) => Promise<PerfOutput>
+}
+
+export interface PerfOutput {
+  type: 'intermediary' | 'final'
+  timeSeconds: number
+  uploadBytes: string
+  downloadBytes: string
```
mxinden, line 67: Why represent these as `string` and not `number`? Wouldn't `number` suffice in precision?

mxinden (commented):
Thank you for tackling this!

packages/protocol-perf/src/index.ts
maschad: Also this interface exists primarily for output, so it's [only instantiated](https://github.com/libp2p/js-libp2p/pull/2067/files#diff-8bb96d7a36e8cb630880dd75531f3c4517659bf5912d4faaaf2f03f7757a2472R221-R222)f for the purposes of outputting to `stdout` , otherwise the values are being incremented as `bigint`

packages/protocol-perf/src/index.ts
maschad: I am wondering if [the spec](https://github.com/libp2p/specs/blob/master/perf/perf.md#protocol) should be amended to address the scenario where we are not able to send over the `MAX_INTEGER` amount of bytes in the 60s timeframe.

Step 2 in the client workflow states:

```
2. Tell the server how many bytes we want the server to send us as a single big-endian uint64 number. Zero is a valid number, so is the max uint64 value.
```

In this case we are requesting an _unlimited server response_

Step 5 states:

```
5. Read from the read side of the stream. This should be the same number of bytes as we told the server in step 2.
```

If we had requested an _infinite_ amount of bytes, would read side match that amount requested?

mxinden: > Step 2 in the client workflow states:
> 
> ```
> 2. Tell the server how many bytes we want the server to send us as a single big-endian uint64 number. Zero is a valid number, so is the max uint64 value.
> ```
> 
> In this case we are requesting an _unlimited server response_

Correct. In other words the server keeps sending data for max uint64 and given that that is such a large number, it never finishes.

> Step 5 states:
> 
> ```
> 5. Read from the read side of the stream. This should be the same number of bytes as we told the server in step 2.
> ```
>
> If we had requested an infinite amount of bytes, would read side match that amount requested?

Given that max uint64 is so large, the transfer will not finish within 20s, the client will be killed (through Linux `timeout` tool) and thus the connection reset on the server side. Neither the client nor the server is aware of the timeout (20s), but the client is simply killed by the benchmark runner.

Does that help?

packages/protocol-perf/src/index.ts
mxinden: > “Number” would limit us to representing 2^53 bytes (maximum integer in JS, 9007199254740990) which would prevent us from representing values greater than roughly 9KB.

I am confused. `2^53 - 1` which is `9007199254740990` allows us to represent 8 PB (`2^53 ÷ 1024 ÷ 1024 ÷ 1024 ÷ 1024`). What am I missing @SgtPooki?

I am basing the above on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

> Also, a number in json would be converted back to a Number in JS, losing precision in those cases.

Given that we only use JS parsers (NodeJS on both ends) and given that JSON itself does not define a `number` size limit, shouldn't precision be fine until `2^53`? Also note, that high precision is not relevant here.



> Also this interface exists primarily for output, so it's [only instantiated](https://github.com/libp2p/js-libp2p/pull/2067/files#diff-8bb96d7a36e8cb630880dd75531f3c4517659bf5912d4faaaf2f03f7757a2472R221-R222) for the purposes of outputting to `stdout` , otherwise the values are being incremented as `bigint`s

Fine by me. As long as I can parse the values as `number` on the benchmark runner side.

packages/protocol-perf/src/index.ts
maschad: > Given that max uint64 is so large, the transfer will not finish within 20s, the client will be killed (through Linux timeout tool) and thus the connection reset on the server side. Neither the client nor the server is aware of the timeout (20s), but the client is simply killed by the benchmark runner.

I understand this execution flow., but my question is does Step 5 in the spec contradict this? Because the client is killed before the _infinite_ amount of bytes would be read.

mxinden: In the throughput benchmark the client is killed before step 5, i.e. it is killed before it can finish reading all data from the stream and before it can check whether it received enough data. Thus the client will never print the `final` output, but only the `intermediary` output. That is fine as the throughput benchmark only needs the latter, not the former.

For the latency benchmark, the client is not killed, but able to finish the whole flow, including step 5, given that only a small amount (1 byte) is uploaded and downloaded. Thus in this benchmark the client will print the `final` output.

Does that make sense @maschad?

maschad: That makes sense. Thanks for clarifying!

packages/protocol-perf/src/index.ts
maschad: The output will is as a result of the `toString()` of a `bigint` and the perf test runner is currently using [MAX_SAFE_NUMBER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) which is a static property of the JS number object, so you will be fine parsing this string as a number @mxinden

mxinden:
Friendly ping @maschad. Anything blocking this pull request?

maschad:
Thanks for following up @mxinden I made the changes suggested in the feedback and this is ready for review now.

packages/protocol-perf/src/index.ts
```diff
@@ -126,19 +133,54 @@ class DefaultPerfService implements Startable, PerfService {
       throw new Error('bytesToSendBack was not set')
     }
 
+    let lastAmountOfBytesSent = BigInt(0)
+    let lastReportedTime = Date.now()
+    let totalBytesSent = BigInt(0)
+
+    const initialStartTime = Date.now()
```
SgtPooki, line 140:
```suggestion
    let totalBytesSent = BigInt(0)

    const initialStartTime = Date.now()
    let lastReportedTime = initialStartTime
```

packages/protocol-perf/src/index.ts
```diff
@@ -126,19 +133,54 @@ class DefaultPerfService implements Startable, PerfService {
...
+            uploadBytes: BigInt(0).toString(),
+            downloadBytes: lastAmountOfBytesSent.toString()
+          }
+
+          // eslint-disable-next-line no-console
+          console.log(JSON.stringify(output))
+
+          lastReportedTime = Date.now()
+          lastAmountOfBytesSent = BigInt(0)
+        }
```
SgtPooki, line 165:
```suggestion
        if (Date.now() - lastReportedTime > 1000) {
          const newReportedTime = Date.now()
          const output: PerfOutput = {
            type: 'intermediary',
            timeSeconds: (newReportedTime - lastReportedTime) / 1000,
            uploadBytes: BigInt(0).toString(),
            downloadBytes: lastAmountOfBytesSent.toString()
          }

          // eslint-disable-next-line no-console
          console.log(JSON.stringify(output))

          lastReportedTime = newReportedTime
          lastAmountOfBytesSent = BigInt(0)
        }
```

move up lastReportedTime date so it doesn't have microsecond lag from console.log+stringify

SgtPooki: idk why the diff here is showing so many changes. Only changed line 153, 156, & 164

packages/protocol-perf/src/index.ts
```diff
@@ -147,23 +189,48 @@ class DefaultPerfService implements Startable, PerfService {
...
           if (toSend > sendBytes) {
             toSend = sendBytes
           }
           sendBytes = sendBytes - toSend
           yield uint8Buf.subarray(0, Number(toSend))
+
+          if (Date.now() - lastReportedTime > 1000) {
+            const output: PerfOutput = {
+              type: 'intermediary',
+              timeSeconds: (Date.now() - lastReportedTime) / 1000,
```
SgtPooki, line 220:
```suggestion
            const newReportedTime = Date.now()
            const output: PerfOutput = {
              type: 'intermediary',
              timeSeconds: (newReportedTime - lastReportedTime) / 1000,
```

packages/protocol-perf/src/index.ts
```diff
@@ -147,23 +189,48 @@ class DefaultPerfService implements Startable, PerfService {
...
+              uploadBytes: lastAmountOfBytesSent.toString(),
+              downloadBytes: BigInt(0).toString()
+            }
+
+            lastReportedTime = Date.now()
```
SgtPooki, line 225:
```suggestion
            lastReportedTime = newReportedTime
```

packages/protocol-perf/src/index.ts
```diff
@@ -147,23 +189,48 @@ class DefaultPerfService implements Startable, PerfService {
 
     const stream = await connection.newStream([this.protocol], options)
 
+    let lastAmountOfBytesSent = BigInt(0)
+    let lastReportedTime = Date.now()
+    let totalBytesSent = BigInt(0)
+
+    const initialStartTime = Date.now()
```
SgtPooki, line 196:
```suggestion
    const initialStartTime = Date.now()
    
    let lastReportedTime = initialStartTime
    let totalBytesSent = BigInt(0)
```

initial and lastReportedTime should be the same for first response to ensure the most accuracy

SgtPooki (approved):
a few comments, but looks mostly good to me. How could we run this locally to confirm? Can we add that info to README.md ?

packages/protocol-perf/src/index.ts
SgtPooki: I get that this doesn't fix a bug or issue or anything but will set up future changes to not bite us if we do end up doing other things

packages/protocol-perf/src/index.ts
```diff
@@ -126,19 +133,54 @@ class DefaultPerfService implements Startable, PerfService {
...
+
+    const finalOutput: PerfOutput = {
+      type: 'final',
+      timeSeconds: (Date.now() - initialStartTime) / 1000,
+      uploadBytes: BigInt(0).toString(),
+      downloadBytes: totalBytesSent.toString()
+    }
+
+    // eslint-disable-next-line no-console
+    console.log(JSON.stringify(finalOutput))
```
wemeetagain, line 180: Why are we printing this to console? Should we use a logger? should this be emitted as an event?

packages/protocol-perf/src/index.ts
```diff
@@ -177,15 +244,21 @@ class DefaultPerfService implements Startable, PerfService {
...
     } finally {
       log('performed %s to %p', this.protocol, connection.remotePeer)
       await stream.close()
     }
 
-    // Return the latency
-    return Date.now() - startTime
+    const finalOutput: PerfOutput = {
+      type: 'final',
+      timeSeconds: (Date.now() - initialStartTime) / 1000,
```
achingbrain, line 256: `timeSeconds` should be calculated after the final byte has been read - if we calculate it here it includes the time taken to gracefully close the stream (e.g. the `await stream.close()`) in the `finally` block above.

packages/protocol-perf/src/index.ts
```diff
@@ -147,23 +189,48 @@ class DefaultPerfService implements Startable, PerfService {
...
+              timeSeconds: (Date.now() - lastReportedTime) / 1000,
+              uploadBytes: lastAmountOfBytesSent.toString(),
+              downloadBytes: BigInt(0).toString()
+            }
+
+            lastReportedTime = Date.now()
+            lastAmountOfBytesSent = BigInt(0)
+
+            // eslint-disable-next-line no-console
+            console.log(JSON.stringify(output))
```
achingbrain, line 229: Same as https://github.com/libp2p/js-libp2p/pull/2067/files#r1373864207 - we should not be writing anything out to the console during a measured benchmark.

packages/protocol-perf/src/index.ts
```diff
@@ -57,7 +57,14 @@ export const defaultInit: PerfServiceInit = {
 }
 
 export interface PerfService {
-  measurePerformance(startTime: number, connection: Connection, sendBytes: bigint, recvBytes: bigint, options?: AbortOptions): Promise<number>
+  measurePerformance(connection: Connection, sendBytes: bigint, recvBytes: bigint, options?: AbortOptions): Promise<PerfOutput>
```
achingbrain, line 60: `BigInt` arithmetic is much slower than `Number` arithmetic in JavaScript - can `sendBytes` and `recvBytes` become `Number`s?

achingbrain (changes requested):

packages/protocol-perf/src/index.ts
maschad: Measurement output is printed to `stdout` , see https://github.com/libp2p/test-plans/tree/master/perf#adding-a-new-implementation-or-a-new-version for more details

mxinden: To add some more context, the test runner (i.e. the one starting the various implementation binaries) parses the test results off of a binaries stdout.

I can not judge whether `console.log` is the best way to print to stdout in JS.

packages/protocol-perf/src/index.ts
mxinden: Referencing comment above: https://github.com/libp2p/js-libp2p/pull/2067#discussion_r1377845934

packages/protocol-perf/src/index.ts
mxinden: We are currently testing single connection single stream throughput. Thus, I don't expect this benchmark to be CPU bound, but network bound. I don't expect the above optimization to make a difference. My intuition might be off.

packages/protocol-perf/src/index.ts
achingbrain: We can presumably buffer the output up and log it all at the end without affecting the test measurements?  As in, the runner just collects the results rather than measuring how long it takes the implementation to deliver the results?

In JS `console.log` is blocking so very slow.

packages/protocol-perf/src/index.ts
maschad: Either way the largest value we are accepting is [still of type Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) so it's a worthwhile refactor imo.

packages/protocol-perf/src/index.ts
maschad: Not sure what you mean @achingbrain the test runner relies on the output (which to my knowledge the best way to be outputted in a cross-environment fashion is via `console.log`) in order to aggregate the various results. Where should the output be buffered to?

packages/protocol-perf/src/index.ts
maschad: As @mxinden [mentioned](https://github.com/libp2p/js-libp2p/pull/2067#discussion_r1377848293) we are not benchmarking CPU performance but just the amount of bytes that are being uploaded vs downloaded so I am not sure if this is a worthwhile optimization.

packages/protocol-perf/src/index.ts
achingbrain: `console.log` is a blocking operation that involves I/O, it's really slow and should not be used during timed sections of benchmarks.

Here is a simplistic example:

```js

function measureConsole () {
  const start = Date.now()

  for (let i = 0; i < 1_000_000; i++) {
    console.info('hello world')
  }

  return Date.now() - start
}

function measureStdOut () {
  const start = Date.now()

  for (let i = 0; i < 1_000_000; i++) {
    process.stdout.write('hello world\n')
  }

  return Date.now() - start
}

function measureBuffer () {
  const start = Date.now()
  const output = []

  for (let i = 0; i < 1_000_000; i++) {
    output.push('hello world')
  }

  const done = Date.now() - start

  console.info(JSON.stringify(output))

  return done
}

const cons = measureConsole()
const stdout = measureStdOut()
const buf = measureBuffer()

console.info('console.info', cons, 'ms')
console.info('process.stdout.write', stdout, 'ms')
console.info('buffered output', buf, 'ms')
```

Result:

```
console.info 4747 ms
process.stdout.write 2242 ms
buffered output 16 ms
```

> Where should the output be buffered to?

To a variable, nothing exiting.  See the `measureBuffer` function above.

The point is, buffer the measurement data, and write it to the console after the timed section of the test.

achingbrain: Ok, so the perf README says:

> Every second the client must print the current progress to stdout

I've adjusted the timing sections so we aren't measuring anything while this happens.

achingbrain (approved):

PR #2071, opened 2023-09-23
fix(libp2p): update circuit relay and upgrader logs

achingbrain:
Updates a couple of log lines to include more information.


packages/libp2p/src/upgrader.ts
```diff
@@ -513,6 +513,7 @@ export class DefaultUpgrader implements Upgrader {
               }
             } catch (err: any) {
               log.error(err)
+              connection.abort(err)
```
achingbrain, line 516:
```suggestion
```

PR #2073, opened 2023-09-23
fix: close webrtc streams

achingbrain:
Depends on:

- [ ] #2069
- [ ] #2070
- [ ] #2071
- [ ] #2072

PR #2074, opened 2023-09-23
fix: close webrtc streams

achingbrain:
Chrome does not always send all messages after closing a datachannel even if `bufferedAmount` is `0` before closing.

This PR adds a `FIN_ACK` message that is sent in reply to a `FIN` message - because all messages are send in-order, when this is received we know the remote has recieved all of our data messages and it's safe to close the channel.

- Depends on
- [ ] #2073
- [ ] https://github.com/libp2p/specs/pull/582


achingbrain:
The spec changes have been merged so I'm going to merge this into the other WebRTC PR so we can land both together.

PR #2085, opened 2023-09-27
fix!: remove dialler language

wemeetagain:
## Description

- Remove the confusing "dialler" word in metrics, replacing with "dial_queue"

### DIFF
--- a/packages/libp2p/test/connection-manager/auto-dial.spec.ts
+++ b/packages/libp2p/test/connection-manager/auto-dial.spec.ts
@@ -22,7 +22,7 @@ import type { PeerStore, Peer } from '@libp2p/interface/peer-store'
 import type { ConnectionManager } from '@libp2p/interface-internal/connection-manager'
 
 describe('auto-dial', () => {
-  let autoDial: AutoDial
+  let autoDialer: AutoDial
   let events: EventEmitter<Libp2pEvents>
   let peerStore: PeerStore
   let peerId: PeerId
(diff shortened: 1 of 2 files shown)


packages/libp2p/test/connection-manager/auto-dial.spec.ts
```diff
@@ -22,7 +22,7 @@ import type { PeerStore, Peer } from '@libp2p/interface/peer-store'
 import type { ConnectionManager } from '@libp2p/interface-internal/connection-manager'
 
 describe('auto-dial', () => {
-  let autoDialler: AutoDial
+  let autoDial: AutoDial
```
maschad, line 25: I think `autoDialler` is the correct language. Generally it's more appropriate to name components with nouns since as opposed to a verb.

maschad (changes requested):
II can agree to the metric name change but I left a comment re: the variable name changes

packages/libp2p/test/connection-manager/auto-dial.spec.ts
wemeetagain: Ok, will replace with `autoDialer` to be consistent with "dialer" used elsewhere.

maschad (approved):

achingbrain:
Can you hold off merging breaking changes for the time being please, we’ll get these all in for v1

achingbrain:
I'm going to revert this so we can get the bug fix in #2139 out without needing a major release.

achingbrain:
Re-opened here: https://github.com/libp2p/js-libp2p/pull/2143

PR #2109, opened 2023-10-02
chore: add root tsconfig

### LINKED ISSUE #2095: Linting support in vscode
- **Version**:
latest master

- **Platform**:
linux

- **Subsystem**:
IDE support

#### Severity:
?

#### Description:
My linter is not properly configured, so I can't see linter errors in vscode.
Does anyone else have it working? Are there any specific steps I need to get it working?

- Open root js-libp2p directory in vscode
- Open a typescript file, see this:
Screenshot from 2023-09-30 15-04-08

wemeetagain:
## Description

- Resolves #2095 

## Change checklist

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,6 @@
+// This root tsconfig allows the linter to work properly in vscode
+// It is not used to build subpackages
+{
+  "extends": "aegir/src/config/tsconfig.aegir.json",
+  "include": ["packages/**/*.ts"]
+}


maschad (approved):
Thanks!

achingbrain:
This is the wrong solution.

- The VSCode eslint plugin is now loading config from the root of the repo, this means any per-package linting config is ignored (not that we should have any, but still).
- It only lints `/packages/**/*.ts`, this ignores the examples and the multidim tests.

The proper solution is to add `"project": true` to `eslintConfig.parseOptions` in each `package.json` - this will cause the eslint plugin to search for the closest `tsconfig.json` to the file being linted instead of using a global one in the root of the project.

Refs:

- https://github.com/typescript-eslint/typescript-eslint/issues/101
- https://github.com/typescript-eslint/typescript-eslint/pull/6084
- https://typescript-eslint.io/blog/parser-options-project-true

achingbrain:
See: https://github.com/libp2p/js-libp2p/pull/2122

Issue #2096, opened 2023-09-30
docs: Refactor documented functions to be methods rather than properties


maschad:
In order for methods to not be rendered as properties on the API docs e.g. the [Libp2p Event Docs](https://libp2p.github.io/js-libp2p/interfaces/_libp2p_interface.index.Libp2pEvents.html) they should be refactored as such.

e.g.

```ts
/**
 * @packageDocumentation
 *
 * Exports a `Libp2p` type for modules to use as a type argument.
 *
 * @example
 *
 * ```typescript
 * import type { Libp2p } from '@libp2p/interface'
 *
 * function doSomethingWithLibp2p (node: Libp2p) {
 *   // ...
 * }
 * ```
 */
interface MyInterface {
  /**
   * This is a method description.
   */
  myMethod(arg1: string, arg2: number): boolean
}
```

as opposed to 

```ts
interface MyInterface {
  /**
   * This is a method description.
   */
  myMethod: (arg1: string, arg2: number) => boolean
}

```

wemeetagain:
This should be enforced when we upgrade aegir.
I have a local branch, but got distracted with #2095

DarhkVoyd:
@maschad Hey, is this issue still valid? I'm eager to help with this, I'm new to this project so please let me know how I can get started. Thanks!

wemeetagain:
@DarhkVoyd feel free to take on this issue.
You should be able to resolve this in an automated way via a new linter rule introduced in aegir v41.0.0
(After you update the aegir dependency across the monorepo, you should be able to run `npm run lint -- -- --fix` from the project root to apply linter fixes everywhere)

DarhkVoyd:
@wemeetagain Hey, after updating aegir to 41.0.0 and running `npm run lint -- -- --fix`, the terminal keeps on throwing linting error one after fixing another all over the codebase such as `@typescript-eslint/strict-boolean-expressions` and `@typescript-eslint/await-thenable`. So, should I keep on resolving all these errors along the way until the lint runs smoothly or am I doing something wrong?

wemeetagain:
Yup, keep resolving all the errors you run into. You may have to resolve the linter errors manually.

maschad:
@DarhkVoyd the latest master should help make things a bit easier in your code editor since the [tsconfig is now adapted](https://github.com/libp2p/js-libp2p/pull/2109) for the monorepo

DarhkVoyd:
[...]

PR #2127, opened 2023-10-04
docs: Refactor documented functions to be methods rather than properties

DarhkVoyd:
## Description
This PR aims to close the issue for methods to not be rendered as properties on the API docs.
Closes: #2096 

## Notes & open questions
The issue was label good first issue, since I am new to this project and unfamiliar with the codebase, I wanted review that if I am on the correct path cause the linting issue don't seem to decrease instead increase with more complex understanding of the codebase.

## Change checklist


maschad (changes requested):
Many of these changes shouldn't be necessary since [aegir v41.0.0](https://github.com/ipfs/aegir/releases/tag/v41.0.0) only introduced the enforcement of  method signature style rather than property function style. 

Please revert all non-package.json changes, then run `npm run clean` at the root, then `npm i` and then `npm run lint`-  you should only see errors [like this](https://github.com/libp2p/js-libp2p/actions/runs/6426521544/job/17450915059?pr=2132#step:5:17) which will indicate that you need to refactor some functions signatures.

as @wemeetagain [mentioned](https://github.com/libp2p/js-libp2p/issues/2096#issuecomment-1743499169) running  `npm run lint -- -- --fix` at the root should resolve all / most of these.

packages/libp2p/src/circuit-relay/transport/index.ts
```diff
@@ -285,7 +285,7 @@ class CircuitRelayTransport implements Transport {
       })
 
       log('new outbound transient connection %a', maConn.remoteAddr)
-      return await this.upgrader.upgradeOutbound(maConn, {
+      return this.upgrader.upgradeOutbound(maConn, {
```
achingbrain, line 288: Was this flagged by the linter?  This sort of return function will introduce bugs:

```js
try {
  return await fn()
} catch (err) {
  // if fn() throws it will be caught here
}
```

vs:

```js
try {
  return fn()
} catch (err) {
  // if fn() throws it will NOT be caught here
}
```

achingbrain:
@DarhkVoyd I'm really sorry - #2096 should not have been tagged as a good first issue.  As you've found it's results in an awful lot of changes to the codebase, some of which have non-obvious knock-on effects.

DarhkVoyd:
@achingbrain Hey there, so what do you suggest I do now?

maschad:
@DarhkVoyd it's unfortunate that these issues came up, I am not sure why as running `npm run lint -- -- --fix` didn't resolve the issues on your branch as it did on https://github.com/libp2p/js-libp2p/pull/2132 , regardless https://github.com/libp2p/js-libp2p/pull/2137 has been merged.

PR #2132, opened 2023-10-06
deps: upgrade aegir to 41.0.0

maschad:
Do Not Merge

Fixes: #2096

### DIFF
--- a/doc/package.json
+++ b/doc/package.json
@@ -28,7 +28,7 @@
     "@libp2p/mplex": "^9.0.0",
     "@libp2p/prometheus-metrics": "^2.0.0",
     "@libp2p/tcp": "^8.0.0",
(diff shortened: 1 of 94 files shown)


achingbrain:
I do wonder why dependabot didn't open a PR for this.  The dependabot yaml here is identical to other monorepos we have but it doesn't seem to be running 🤔

achingbrain:
Huh weird, this was opened a few days ago - https://github.com/libp2p/js-libp2p/pull/2116

maschad:
My browser extension had obscured the dependabot PR so I didn't see it either 😅  - I had originally opened this as a CI check to assist in pointing out the expected linter errors as a reference for https://github.com/libp2p/js-libp2p/pull/2127

PR #2133, opened 2023-10-06
feat: configuration validation

### LINKED ISSUE #1757: Undefined services causes exception

### LINKED ISSUE #1903: Swallowed Error in CircuitRelayV2 if reservations.defaultDataLimit is set to number

achingbrain:
Closes https://github.com/libp2p/js-libp2p/issues/1573
Closes #1757 
Closes #1903 

I decided to go with [yup](https://github.com/jquense/yup) for a few reasons:

- Integration was more seamless
- Is used by substantially more projects ( 545k vs 105k)
- Preferred the schema building via chaining as opposed to defining complex functions (especially for defining integers, mins and defaults)
- Was more actively maintained ( most recent [commit 3 weeks ago](https://github.com/jquense/yup/commit/e593f8f72e7195cf0ac48fa8e1cd82d95c1e6bb5) as opposed to [last year November](https://github.com/ianstormtaylor/superstruct/commit/03d65bd37d505d934db7cb48eed691fc41ecba7c)
- More downloads ( 4,201,246 vs 889,754 weekly downloads)


packages/libp2p/src/config/helpers.ts
```diff
@@ -0,0 +1,12 @@
+import { multiaddr } from '@multiformats/multiaddr'
+
+export const validateMultiaddr = (value: Array<string | undefined> | undefined): boolean => {
+  value?.forEach((addr) => {
+    try {
+      multiaddr(addr)
+    } catch (err) {
+      throw new Error(`invalid multiaddr: ${addr}`)
```
wemeetagain, line 8: Probably should throw a CodeError
And bubble up the error message

packages/libp2p/src/autonat/index.ts
```diff
@@ -108,14 +118,16 @@ class DefaultAutoNATService implements Startable {
   private started: boolean
 
   constructor (components: AutoNATComponents, init: AutoNATServiceInit) {
+    const validatedConfig = configValidator.validateSync(init)
```
wemeetagain, line 121: This feels quite verbose and could just be called `config`.
Applies to all places where xxx.validateSync is used.

packages/libp2p/src/circuit-relay/server/index.ts
```diff
@@ -95,9 +101,19 @@ export interface RelayServerEvents {
   'relay:advert:error': CustomEvent<Error>
 }
 
-const defaults = {
-  maxOutboundStopStreams: MAX_CONNECTIONS
-}
+const configValidator = object({
+  hopTimeout: number().min(0).integer().default(DEFAULT_HOP_TIMEOUT),
```
wemeetagain, line 105: probably should pick an order and stick with it.
Either `number().min(x).integer()`
or
`number().integer().min(x)`

my preference to the latter

packages/libp2p/src/connection-manager/utils.ts
```diff
@@ -69,7 +73,26 @@ export function combineSignals (...signals: Array<AbortSignal | undefined>): Cle
...
     setMaxListeners?.(Infinity, signal)
-  } catch {}
+  } catch { }
 
   return signal
 }
+
+export const validateConnectionManagerConfig = (opts: ConnectionManagerInit): ObjectSchema<Record<string, unknown>> => {
+  return object({
+    maxConnections: number().min(opts?.minConnections ?? MIN_CONNECTIONS, `maxConnections must be greater than the min connections limit: ${opts?.minConnections}`).integer().default(MAX_CONNECTIONS),
```
wemeetagain, line 83:
```suggestion
    maxConnections: number().min(opts?.minConnections ?? MIN_CONNECTIONS, `maxConnections must be greater than the min connections limit: ${opts?.minConnections ?? MIN_CONNECTIONS}`).integer().default(MAX_CONNECTIONS),
```

packages/libp2p/src/connection-manager/utils.ts
```diff
@@ -69,7 +73,26 @@ export function combineSignals (...signals: Array<AbortSignal | undefined>): Cle
...
+    minConnections: number().min(0).integer().max(opts?.maxConnections ?? MAX_CONNECTIONS, `minConnections must be less than the max connections limit: ${opts?.maxConnections}`).default(MIN_CONNECTIONS),
```
wemeetagain, line 84:
```suggestion
    minConnections: number().min(0).integer().max(opts?.maxConnections ?? MAX_CONNECTIONS, `minConnections must be less than the max connections limit: ${opts?.maxConnections ?? MAX_CONNECTIONS}`).default(MIN_CONNECTIONS),
```

packages/libp2p/src/connection-manager/utils.ts
wemeetagain, line 83:
```suggestion
    maxConnections: number().min(opts?.minConnections ?? MIN_CONNECTIONS, `maxConnections must be greater than minConnections: ${opts?.minConnections}`).integer().default(MAX_CONNECTIONS),
```

I think these error messages should use the names of the related options rather than a description of them, eg: minConnections, rather than 'the min connections limit'.
Applies to all of these error messages

wemeetagain, line 83:
```suggestion
    maxConnections: number().min(opts?.minConnections ?? MIN_CONNECTIONS, `maxConnections must be greater than or equal to the min connections limit: ${opts?.minConnections}`).integer().default(MAX_CONNECTIONS),
```
need to be precise about gt vs gte (greater than vs greater than or equal)

packages/libp2p/src/connection-manager/utils.ts
```diff
@@ -69,7 +73,26 @@ export function combineSignals (...signals: Array<AbortSignal | undefined>): Cle
...
+    autoDialInterval: number().min(0).integer().default(AUTO_DIAL_INTERVAL),
+    autoDialConcurrency: number().min(0).integer().default(AUTO_DIAL_CONCURRENCY),
+    autoDialPriority: number().min(0).integer().default(AUTO_DIAL_PRIORITY),
+    maxParallelDials: number().min(0).integer().default(MAX_PARALLEL_DIALS),
+    maxParallelDialsPerPeer: number().max(opts?.autoDialConcurrency ?? AUTO_DIAL_CONCURRENCY, `maxParallelDialsPerPeer must be less than the min auto dial conccurency limit: ${opts?.autoDialConcurrency}`).default(MAX_PARALLEL_DIALS_PER_PEER),
+    maxPeerAddrsToDialed: number().min(0).integer().default(MAX_PEER_ADDRS_TO_DIAL),
+    dialTimeout: number().min(0).integer().default(DIAL_TIMEOUT),
+    inboundUpgradeTimeout: number().integer().default(INBOUND_UPGRADE_TIMEOUT),
```
wemeetagain, line 92: some of these are missing `min(x)`

packages/libp2p/src/dcutr/dcutr.ts
```diff
@@ -24,14 +25,14 @@ const MAX_DCUTR_MESSAGE_SIZE = 1024 * 4
...
+const configValidator = object({
   // https://github.com/libp2p/go-libp2p/blob/8d2e54e1637041d5cf4fac1e531287560bd1f4ac/p2p/protocol/holepunch/holepuncher.go#L27
-  timeout: 5000,
+  timeout: number().integer().default(5000).min(1),
   // https://github.com/libp2p/go-libp2p/blob/8d2e54e1637041d5cf4fac1e531287560bd1f4ac/p2p/protocol/holepunch/holepuncher.go#L28
-  retries: 3,
-  maxInboundStreams: 1,
-  maxOutboundStreams: 1
-}
+  retries: number().integer().default(3).min(1),
```
wemeetagain, line 32: min(0) ? Same with some of the other options here

packages/libp2p/src/fetch/index.ts
```diff
@@ -83,6 +82,13 @@ export interface FetchService {
   unregisterLookupFunction(prefix: string, lookup?: LookupFunction): void
 }
 
+const configValidator = object({
+  protocolPrefix: string().default('libp2p'),
+  timeout: number().integer().default(TIMEOUT),
```
wemeetagain, line 87: min(0)

packages/libp2p/src/identify/identify.ts
```diff
@@ -31,23 +42,19 @@ import type { IncomingStreamData, Registrar } from '@libp2p/interface-internal/r
...
-  maxPushOutgoingStreams: 1,
-  maxObservedAddresses: 10,
-  maxIdentifyMessageSize: 8192,
-  runOnConnectionOpen: true,
-  runOnTransientConnection: true
-}
+const configValidator = object({
+  protocolPrefix: string().default(PROTOCOL_PREFIX),
+  agentVersion: string().default(AGENT_VERSION),
+  timeout: number().integer().default(TIMEOUT),
```
wemeetagain, line 48: min(0)

packages/libp2p/src/ping/index.ts
```diff
@@ -37,6 +38,14 @@ export interface PingServiceComponents {
   connectionManager: ConnectionManager
 }
 
+const configValidator = object({
+  protocolPrefix: string().default(PROTOCOL_PREFIX),
+  timeout: number().integer().default(TIMEOUT),
```
wemeetagain, line 43: min(0)

packages/libp2p/src/upnp-nat/index.ts
```diff
@@ -64,27 +65,40 @@ export interface UPnPNATComponents {
   addressManager: AddressManager
 }
 
+const validIPRegex = /^(?:(?:^|\.)(?:\d|[1-9]\d|1\d{2}|2[0-4]\d|25[0-5])){4}$/
+
+const configValidator = object({
+  externalAddress: string().matches(validIPRegex, 'Invalid IP address'),
+  localAddress: string().matches(validIPRegex, 'Invalid IP address'),
+  description: string(),
+  ttl: number().integer().default(DEFAULT_TTL),
```
wemeetagain, line 74: min(0)

wemeetagain (changes requested):

packages/libp2p/src/circuit-relay/server/index.ts
achingbrain: The latter reads better to me.

packages/libp2p/src/autonat/index.ts
maschad: I think `validatedConfig` var name implies that all `undefined` properties of the interface passed in via the constructor will now set to default values,  so I think the name is appropriate, not strongly opinionated but I think the rename has value.

wemeetagain: > I think validatedConfig var name implies that all undefined properties of the interface passed in via the constructor will now set to default values

I don't think the name implies that. It implies that its a configuration that is validated. Filling in default values is not necessarily a standard step in validation. Validation, imo, speaks to making assertions about unsanitized user inputs.

Moreover, the fact that it is validated or not is not relevant, to the rest of the constructor. We can tell that the config variable is validated because its being assigned to the output of a function called validateSync. We can also tell by its type that its basically a `Required<XXXInit>`. (We could always make that explicit with a type declaration `const config: Required<XXXInit> = xxx.validateSync(init)`)

packages/libp2p/src/connection-manager/utils.ts
```diff
@@ -65,3 +69,21 @@ export function combineSignals (...signals: Array<AbortSignal | undefined>): Cle
...
+
+export const validateConnectionManagerConfig = (opts: ConnectionManagerInit): ObjectSchema<Record<string, unknown>> => {
+  return object({
+    maxConnections: number().integer().min(opts?.minConnections ?? MIN_CONNECTIONS, `maxConnections must be greater than or equal to minConnections: ${opts?.minConnections ?? MIN_CONNECTIONS}`).default(MAX_CONNECTIONS),
+    minConnections: number().integer().min(0).max(opts?.maxConnections ?? MAX_CONNECTIONS, `minConnections must be less than or equal to maxConnections : ${opts?.maxConnections ?? MAX_CONNECTIONS}`).default(MIN_CONNECTIONS),
+    autoDialInterval: number().integer().min(0).default(AUTO_DIAL_INTERVAL),
+    autoDialConcurrency: number().integer().min(0).default(AUTO_DIAL_CONCURRENCY),
+    autoDialPriority: number().integer().min(0).default(AUTO_DIAL_PRIORITY),
+    maxParallelDials: number().integer().min(0).default(MAX_PARALLEL_DIALS),
+    maxPeerAddrsToDialed: number().min(0).integer().default(MAX_PEER_ADDRS_TO_DIAL),
```
wemeetagain, line 81: Inconsistent ordering

packages/libp2p/src/address-manager/utils.ts
```diff
@@ -11,3 +16,12 @@ export function debounce (func: () => void, wait: number): () => void {
     timeout = setTimeout(later, wait)
   }
 }
+
+export function validateAddressManagerConfig (opts: AddressManagerInit): ObjectSchema<Record<string, unknown>> {
```
wemeetagain, line 20: If opts isn't used, maybe it shouldn't be passed in?

packages/libp2p/src/config/config.ts
```diff
@@ -0,0 +1,36 @@
...
+import mergeOptions from 'merge-options'
+import { object } from 'yup'
+import { validateAddressManagerConfig } from '../address-manager/utils.js'
+import { validateConnectionManagerConfig } from '../connection-manager/utils.js'
+import type { AddressManagerInit } from '../address-manager'
+import type { ConnectionManagerInit } from '../connection-manager/index.js'
+import type { Libp2pInit } from '../index.js'
+import type { ServiceMap, RecursivePartial } from '@libp2p/interface'
+
+const DefaultConfig: Partial<Libp2pInit> = {
```
wemeetagain, line 13: Should this be capitalized?

packages/protocol-dcutr/src/dcutr.ts
```diff
@@ -20,14 +21,14 @@ const MAX_DCUTR_MESSAGE_SIZE = 1024 * 4
 // ensure the dial has a high priority to jump to the head of the dial queue
 const DCUTR_DIAL_PRIORITY = 100
 
-const defaultValues = {
+const configValidator = object({
   // https://github.com/libp2p/go-libp2p/blob/8d2e54e1637041d5cf4fac1e531287560bd1f4ac/p2p/protocol/holepunch/holepuncher.go#L27
-  timeout: 5000,
+  timeout: number().integer().default(5000).min(0),
```
wemeetagain, line 26: Inconsistent ordering

packages/protocol-dcutr/src/dcutr.ts
```diff
@@ -20,14 +21,14 @@ const MAX_DCUTR_MESSAGE_SIZE = 1024 * 4
...
   // https://github.com/libp2p/go-libp2p/blob/8d2e54e1637041d5cf4fac1e531287560bd1f4ac/p2p/protocol/holepunch/holepuncher.go#L28
-  retries: 3,
-  maxInboundStreams: 1,
-  maxOutboundStreams: 1
-}
+  retries: number().integer().default(3).min(0),
```
wemeetagain, line 28: Inconsistent ordering

packages/libp2p/src/connection-manager/utils.ts
```diff
@@ -65,3 +69,21 @@ export function combineSignals (...signals: Array<AbortSignal | undefined>): Cle
 
   return signal
 }
+
+export const validateConnectionManagerConfig = (opts: ConnectionManagerInit): ObjectSchema<Record<string, unknown>> => {
+  return object({
+    maxConnections: number().integer().min(opts?.minConnections ?? MIN_CONNECTIONS, `maxConnections must be greater than or equal to minConnections: ${opts?.minConnections ?? MIN_CONNECTIONS}`).default(MAX_CONNECTIONS),
```
wemeetagain, line 75: Do we need to check against user-configured values here? If so, the same approach should be done elsewhere consistently, eg, the fetch config.

packages/libp2p/src/config/helpers.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const validateMultiaddr = (value: Array<string | undefined> | undefined): boolean => {
+  value?.forEach((addr) => {
+    try {
+      multiaddr(addr)
+    } catch (err) {
+      throw new CodeError(`invalid multiaddr: ${addr}`, codes.ERR_INVALID_MULTIADDR)
+    }
+  })
+  return true
```
achingbrain, line 13: If `value` is `undefined` this will return true.  Was this intentional?

packages/transport-circuit-relay-v2/src/server/index.ts
```diff
@@ -93,9 +100,19 @@ export interface RelayServerEvents {
...
 
-const defaults = {
-  maxOutboundStopStreams: MAX_CONNECTIONS
-}
+const configValidator = object({
+  hopTimeout: number().integer().min(0).default(DEFAULT_HOP_TIMEOUT),
+  reservations: object({
+    maxReservations: number().integer().min(0).default(DEFAULT_MAX_RESERVATION_STORE_SIZE),
+    reservationClearInterval: number().integer().min(0).default(DEFAULT_MAX_RESERVATION_CLEAR_INTERVAL),
+    applyDefaultLimit: boolean().default(false),
```
achingbrain, line 108: I think this needs to default to true, otherwise all js nodes will be open relays by default.

achingbrain: Can we not reuse the config validator from the reservation store here?

packages/transport-circuit-relay-v2/src/server/reservation-store.ts
```diff
@@ -50,12 +59,14 @@ export class ReservationStore implements Startable {
...
-    this.defaultDurationLimit = options.defaultDurationLimit ?? DEFAULT_DURATION_LIMIT
-    this.defaultDataLimit = options.defaultDataLimit ?? DEFAULT_DATA_LIMIT
+    const config = configValidator.validateSync(options)
+
+    this.maxReservations = config.maxReservations
+    this.reservationClearInterval = config.reservationClearInterval
+    this.applyDefaultLimit = config.applyDefaultLimit
+    this.reservationTtl = config.reservationTtl
+    this.defaultDurationLimit = config.defaultDurationLimit
+    this.defaultDataLimit = config.defaultDataLimit as bigint
```
achingbrain, line 69: Why is the cast necessary?  Does `yup` not handle `BigInt`s properly?

packages/transport-circuit-relay-v2/src/transport/reservation-store.ts
```diff
@@ -75,6 +76,13 @@ export interface ReservationStoreEvents {
   'relay:removed': CustomEvent<PeerId>
 }
 
+const configValidator = object({
+  discoverRelays: number().integer().min(0).default(0),
+  maxReservationQueueLength: number().integer().min(0).default(100),
+  reservationCompletionTimeout: number().integer().min(0).default(10000),
+  reservationConcurrency: number().integer().min(0).default(DEFAULT_RESERVATION_CONCURRENCY)
```
achingbrain, line 83: We have some defaults defined as constants, some as literals - can it be one or the other please, probably named constants.

packages/upnp-nat/src/upnp-nat.ts
```diff
@@ -14,29 +15,42 @@ function highPort (min = 1024, max = 65535): number {
...
+})
+
 export class UPnPNAT implements Startable {
   private readonly components: UPnPNATComponents
   private readonly externalAddress?: string
   private readonly localAddress?: string
   private readonly description: string
   private readonly ttl: number
-  private readonly keepAlive: boolean
+  private readonly keepAlive?: boolean
```
achingbrain, line 35: Why has this become optional?

packages/transport-circuit-relay-v2/src/server/reservation-store.ts
achingbrain: TLDR: it doesn't but we can remove the cast with `mixed<bigint>()...`

maschad (approved):

maschad (approved):

achingbrain:
Did a bit of bundle analysis of different config validators.

## Zod

👍 Great TS support, can derive output types with no tweaking
👍  Big user base - 10M downloads/week
👎  Tree shaking doesn't work so is very large

```ts
import { object, array, function as func, custom } from 'zod'

const validateMultiaddr = (value?: string): boolean => {
  try {
    multiaddr(value ?? '')
  } catch (err) {
    return false
  }

  return true
}

const multiaddrStringArray = array(custom<string>(validateMultiaddr)).default(() => [])
const multiaddrArray = array(custom<Multiaddr>(validateMultiaddr)).default(() => [])

const configValidator = object({
  listen: multiaddrStringArray,
  announce: multiaddrStringArray,
  noAnnounce: multiaddrStringArray,
  announceFilter: func().args(multiaddrArray).returns(multiaddrArray).default(() => defaultAddressFilter)
})
```

image

## Yup

😐 TS support not amazing (see use of `unknown` in return types)
😐 Reasonable user base - 6M downloads/week
👎 Tree shaking doesn't work so is very large

```ts
import { object, array, string, mixed } from 'yup'

const validateMultiaddr = (value?: Array<string | undefined>): boolean => {
  try {
    (value ?? []).forEach(value => multiaddr(value ?? ''))
  } catch (err) {
    return false
  }

  return true
}

const multiaddrStringArray = array().of(string()).test('is multiaddr', validateMultiaddr).default([])

const configValidator = object({
  listen: multiaddrStringArray,
  announce: multiaddrStringArray,
  noAnnounce: multiaddrStringArray,
  announceFilter: mixed().default(() => defaultAddressFilter)
})
```

image

## Valibot

👍  Tiny addition to the bundle
👍 Great TS support, can derive output types with no tweaking
👎 Small user base - 200k downloads/week
👎 Parts of API missing (function validation, for example)

```ts
import * as v from 'valibot'

const validateMultiaddr = (value?: any): boolean => {
  try {
    multiaddr(value ?? '')
  } catch {
    return false
  }

  return true
}

const multiaddrStringArray = v.optional(v.array(v.custom<string>(validateMultiaddr)), [])

const configValidator = v.object({
  listen: multiaddrStringArray,
  announce: multiaddrStringArray,
  noAnnounce: multiaddrStringArray,
  announceFilter: v.optional(v.any(), defaultAddressFilter)
})
```

image

## Superstruct

👍  Tiny addition to the bundle
👍 Great TS support, can derive output types with no tweaking
😐 Reasonable user base - 1.3m downloads/week

```ts
import { define, object, array, func, defaulted, assert } from 'superstruct'

const validateMultiaddr = (value?: any): boolean => {
  try {
    multiaddr(value ?? '')
  } catch {
    return false
  }

  return true
}

const multiaddrString = define('MultiaddrString', validateMultiaddr)
const multiaddrStringArray = defaulted(array(multiaddrString), () => [])

const configValidator = object({
  listen: multiaddrStringArray,
  announce: multiaddrStringArray,
  noAnnounce: multiaddrStringArray,
  announceFilter: defaulted(func(), defaultAddressFilter)
})
```

image

-----

Zod is working on v4 which will improve tree shaking.  Valibot is still under active development but it's early days yet.

I think revisit this in a few months time, or after Zod v4 is released and reassess.

maschad:
Good call @achingbrain I agree, let's revisit once Zod has greater stability.

PR #2139, opened 2023-10-07
fix(circuit-relay): respect applyDefaultLimit when it is false

achingbrain:
## Description

We were accidentally applying the default limits if no limits were set, which is intentionally the case when the `applyDefaultLimit` config option is set to `false`.

Also simplifies the relay test setup to create nodes with just config overrides to remove a bit of the copy/paste.

## Change checklist


packages/libp2p/src/circuit-relay/utils.ts
```diff
@@ -13,6 +12,8 @@ import type { Uint8ArrayList } from 'uint8arraylist'
 const log = logger('libp2p:circuit-relay:utils')
 
 async function * countStreamBytes (source: Source<Uint8Array | Uint8ArrayList>, limit: { remaining: bigint }): AsyncGenerator<Uint8Array | Uint8ArrayList, void, unknown> {
```
maschad, line 14: we probably should change the signature of this function to

```suggestion
async function * countStreamBytes (source: Source<Uint8Array | Uint8ArrayList>, limitBytes: bigint): AsyncGenerator<Uint8Array | Uint8ArrayList, void, unknown> {
```

and replace all the `limit.remaining`

packages/libp2p/src/circuit-relay/utils.ts
```diff
@@ -47,25 +48,33 @@ const doRelay = (src: Stream, dst: Stream, abortSignal: AbortSignal, limit: Requ
...
   let srcDstFinished = false
   let dstSrcFinished = false
 
-  const dataLimit = {
-    remaining: limit.data
+  let dataLimit: { remaining: bigint } | undefined
+
+  if (limit?.data != null) {
+    dataLimit = {
+      remaining: limit.data
```
maschad, line 66: See comment above, If the function signature is changed then creating this object is unnecessary

maschad (changes requested):
Good catch, I think we can do some refactors and then land it.

packages/libp2p/src/circuit-relay/utils.ts
achingbrain: Unfortunately this doesn't work and would introduce a new bug.

The data limit applies in both directions (e.g. `A->Relay->B` and `B->Relay->A`) so both ends of the stream need to decrement the same `remaining` variable.

Passing an object copies by reference, passing a number would copy by value so you'd end up with separate limits for each direction.

maschad (approved):

PR #2107, opened 2023-10-02
refactor!: move autonat into separate package

wemeetagain:
## Description
- This is an attempt to start #1913, see what kind of issues arise from it
- Move autonat into separate package
- Move error codes and messages into @libp2p/interface/errors

## Notes & open questions
- @achingbrain  How does the "Browser <script> tag" section of the readme work? is it stale?

## Change checklist

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation if necessary (this includes comments as well)
- [ ] I have added tests that prove my fix is effective or that my feature works

Issue #2104, opened 2023-10-02
release js-libp2p v1


SgtPooki:
discussed in https://pl-strflt.notion.site/Js-libp2p-v1-0-19d1a9724167469badaad4bf736cd7d6?pvs=4

```[tasklist]
### Tasks
- [ ] https://github.com/libp2p/js-libp2p/pull/2144
- [ ] https://github.com/libp2p/blog/issues/115
- [ ] https://github.com/libp2p/js-libp2p/pull/2107
- [ ] https://github.com/libp2p/js-libp2p/pull/2084
- [ ] https://github.com/libp2p/js-libp2p/issues/1913
- [ ] https://github.com/multiformats/js-multiformats/issues/249
- [ ] https://github.com/libp2p/js-libp2p/pull/2143
- [ ] Additional metadata/context on why multiaddrs are being listened on
```

### For blog post: 
- Why now
- Reminder of what all is in js-libp2p
- Significant accomplishments over the last year
    - New connectivity
    - Transports
    - Monorepo
    - Cherry pick things from the release notes

achingbrain:
I'd like to get the WebRTC stream fixes released before doing this, but @maschad can get started on the blog post before then.

I've added some PRs that will target v1 to the task list above.

achingbrain:
I have created a release-v1.0 branch that we can use to aggregate breaking changes ahead of release while not blocking bug fixes for `0.46.x`.

On merge day I will rebase this on top of master and push locally to preserve individual PR commits.

achingbrain:
I have created a release-v1.0 branch that we can use to aggregate breaking changes ahead of release while not blocking bug fixes for `0.46.x`.

On merge day I will rebase this on top of master and push locally to preserve individual PR commits.

cc @maschad @wemeetagain

achingbrain:
This has shipped!

PR #2140, opened 2023-10-09
refactor: move error codes to @libp2p/interface/errors

wemeetagain:
## Description

- move `codes` from `libp2p/src/errors.js` to `@libp2p/interface/errors`
- from discussion of #2107, we decided that moving error codes should be done here in a separate PR

## Change checklist


wemeetagain:
@maschad this is needed for #2107, which is a prereq for #2104

maschad:
That's correct @wemeetagain , my apologies as I should clarify - we should resolve how we are going to synchronize all of the PRs which contain breaking changes such that they are all in released in v1 as opposed to any bug fix releases which we may need to do before v1.

We can discuss more on the Open Maintainers call on 10-10-2023

achingbrain:
Is this actually breaking?  Do we export these error codes?

wemeetagain:
> Is this actually breaking? Do we export these error codes?

I guess not. Still would like to target v1 branch tho

wemeetagain:
closing this since we decided to move over error codes only one by one as needed.

PR #2144, opened 2023-10-10
chore: release v1.0.0

achingbrain:
This PR is not for merging, just to surface the changes that will go into that release.

When we are ready to release, I will merge this locally, rebase on top of master & push.  This should preserve each PR commit that targets this branch.


achingbrain:
These commits are all in master now.

PR #2145, opened 2023-10-10
refactor!: remove isStarted method from Startable

achingbrain:
## Description

We have an `isStarted` method on the `Startable` interface but we only really use it in tests. Our implementations tend to guard on being started twice so it just adds noise to every implementation.

BREAKING CHANGE: the `isStarted` method has been removed from the `Startable` interface

## Change checklist


achingbrain:
Draft because some tests are failing.

wemeetagain (commented):
I don't see where the method is removed from the interface

achingbrain:
Looks like I didn't manage to push all the changes, oops.  Should be ok now.

wemeetagain (approved):
Does it need a migration note?

achingbrain:
I don't *think* so because people by and large shouldn't be starting/stopping components behind libp2p's back?

Issue #2152, opened 2023-10-10
test: Ensure js-libp2p is compatible with TS Multiformats 


maschad:
Following the refactor of [js-multiformats to Typescript](https://github.com/multiformats/js-multiformats/pull/251) we need to ensure js-libp2p is compatible.

maschad:
Closed by #2153

PR #2153, opened 2023-10-10
deps: use TS multiformats 

maschad:
do not merge

Related https://github.com/libp2p/js-libp2p/issues/2152


maschad:
@wemeetagain I am getting some module not found errors despite them appearing on the file system, I am not sure if it's related to how [the types are exported](https://github.com/multiformats/js-multiformats/pull/251/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L30-L32)

achingbrain:
I think `multiformats` is not being built on install?  It's a GH branch dependency here so it's coming down as raw typescript, it needs to be transpiled to js before it'll work - the multiformats module needs a [prepare](https://blog.jim-nielsen.com/2018/installing-and-building-an-npm-package-from-github/) script or smth.

Just, we need to be careful with `prepare` because it runs during `npm i` locally so if there's some ts breakage in `dom.d.ts` or whatever it'll break installing deps when doing local multiformats development 😩 

It also pulls down all dev deps when used as a GH dependency (vs npm dependency) so can be absurdly heavy, and mean multiple versions of aegir being installed and other such 🤮 .

maschad:
Good call @achingbrain the tests are passing now. 

@wemeetagain I've updated https://github.com/multiformats/js-multiformats/pull/251

PR #2154, opened 2023-10-11
refactor: move setMaxListeners to @libp2p/interface

wemeetagain:
## Description

- Discussed in https://github.com/libp2p/js-libp2p/discussions/2138
- Move `setMaxListeners` handling into a single wrapped function exported from `@libp2p/interface/events`

## Notes

- Instead of setting `"browser": {"events": false}` as was discussed in #2138, an actual file is given. Without doing this, an annoying warning is produced in this package and every consumer package in the monorepo!

## Change checklist

### DIFF
--- a/packages/interface/src/events.ts
+++ b/packages/interface/src/events.ts
@@ -1,3 +1,5 @@
+import { setMaxListeners as nodeSetMaxListeners } from 'events'
+
 export interface EventCallback<EventType> { (evt: EventType): void }
 export interface EventObject<EventType> { handleEvent: EventCallback<EventType> }
 export type EventHandler<EventType> = EventCallback<EventType> | EventObject<EventType>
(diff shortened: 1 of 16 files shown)


packages/interface/src/events.ts
```diff
@@ -98,3 +100,13 @@ class CustomEventPolyfill<T = any> extends Event {
 }
 
 export const CustomEvent = globalThis.CustomEvent ?? CustomEventPolyfill
+
+// create a setMaxListeners that doesn't break browser usage
+
+export const setMaxListeners: typeof nodeSetMaxListeners = (n, ...eventTargets) => {
+  try {
+    nodeSetMaxListeners(n, ...eventTargets)
+  } catch (e) {
```
achingbrain, line 109:
```suggestion
  } catch {
```

achingbrain (changes requested):
LGTM.

Just needs the `"events"` polyfill modules removing from `kad-dht` and `libp2p`.

We can release this as a patch to master as it's non-breaking.

achingbrain (approved):

Issue #2130, opened 2023-10-05
interop: rename interop transport test script


maschad:
~~With the rename that took place in https://github.com/libp2p/test-plans/pull/308 we should deprecate the previous [npm package](https://www.npmjs.com/package/@libp2p/multidim-interop) and create a new one, as well as~~ [rename the test script](https://github.com/libp2p/js-libp2p/blob/master/interop/package.json?rgh-link-date=2023-10-04T02%3A00%3A37Z#L49)

achingbrain:
Would it be possible to instead have one interop test module that contains multiple tests rather than multiple modules with a single test each?

It's a lot of boilerplate otherwise.

maschad:
Good idea, I suppose we only need rename the script then.

PR #2162, opened 2023-10-17
refactor: rename multidim interop test script 

maschad:
## Title

rename multidim interop test script 

## Description

Closes https://github.com/libp2p/js-libp2p/issues/2130 as we are now more specific in what we are testing

## Change checklist

### DIFF
--- a/interop/package.json
+++ b/interop/package.json
@@ -47,6 +47,7 @@
     "start": "node index.js",
     "build": "aegir build",
     "lint": "aegir lint",
+    "test:interop:multidim": "aegir test",
     "test:transport:interop": "aegir test"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


interop/package.json
```diff
@@ -47,7 +47,8 @@
     "start": "node index.js",
     "build": "aegir build",
     "lint": "aegir lint",
-    "test:interop:multidim": "aegir test"
+    "test:interop:multidim": "aegir test",
```
wemeetagain, line 50: Do we still need this line?

maschad: yes we do, in order to be backwards compatible until https://github.com/libp2p/test-plans/pull/312 is merged

Issue #2090, opened 2023-09-28
refactor: Remove parallel dialling 


maschad:
As it stands we currently have parallel dials per peer [disabled by default](https://github.com/libp2p/js-libp2p/blob/f09ac4a7704070fd92bae8d4482d06eac45ddd2c/packages/libp2p/src/connection-manager/constants.defaults.ts#L19), post https://github.com/libp2p/js-libp2p/issues/2010 there is currently no use case for parallel dials per peer given that if we have ranked multiaddrs based on likelihood of success, it would be inefficient to dial multiaddrs that could potentially fail.

PR #2163, opened 2023-10-18
refactor!: remove per-peer parallel dialling

maschad:
## Description

Since we are now dialling one peer at a time as a means of smarter dialling, we no longer need the option to control the concurrency of parallel dials across the multiaddrs of a single peer.

Closes https://github.com/libp2p/js-libp2p/issues/2090

## Change checklist


doc/migrations/v0.46-v1.0.0.md
```diff
@@ -63,3 +64,8 @@ The following metrics were renamed:
 
 `libp2p_dialler_pending_dials` => `libp2p_dial_queue_pending_dials`
 `libp2p_dialler_in_progress_dials` => `libp2p_dial_queue_in_progress_dials`
+
+
+## Connection Manager
+
+We now only dial one peer at a time as this is a more efficient use of resources.  This means that the `maxParallelDialsPerPeer` option has been removed.
```
wemeetagain, line 71: I don't think this is right. We're removing parallel dials per peer, but not across peers.

doc/migrations/v0.46-v1.0.0.md
```diff
@@ -77,3 +78,7 @@ The following metrics were renamed:
 
 `libp2p_dialler_pending_dials` => `libp2p_dial_queue_pending_dials`
 `libp2p_dialler_in_progress_dials` => `libp2p_dial_queue_in_progress_dials`
+
+## Connection Manager
+
+We now only dial one peer at a time as this is a more efficient use of resources.  This means that the `maParallelDialsPerPeer` option has been removed.
```
achingbrain, line 84:
```suggestion
The observed behavior of dialing peers has been that given a list of supported addresses, if any one routable address would succeed then all would succeed and that if any routable address would fail then all would fail.

Consequently the previous dial behaviour of dialing all available addresses (up to a concurrency limit) and cancelling any in-flight dials when the first succeeds was a very inefficient use of resources.

Since `libp2p@0.46.10` we have only dialed one address at a time for each peer by setting the default value of the `ConnectionManager`'s `maxParallelDialsPerPeer` option to `1`.  As of `libp2p@1.0.0` this option has been removed.
```

achingbrain (approved):

doc/migrations/v0.46-v1.0.0.md
achingbrain: I've updated the note, hopefully it's clearer now.

Issue #2150, opened 2023-10-10
feat: Ensure dials are sorted by success rate


maschad:
once https://github.com/libp2p/js-libp2p/pull/2033 is merged we can dial by success rate

Issue #2114, opened 2023-10-02
fix: replace autodial retry threshold in peers to be per multiaddr


maschad:
Given it is more useful to track failure on a per-address basis, we should add a retry mechanism that exponentially backoffs of dialling failed addresses per peer. 

Related https://github.com/libp2p/js-libp2p/pull/1943

Depends on https://github.com/libp2p/js-libp2p/pull/2033

PR #2164, opened 2023-10-20
feat: ensure dials are sorted based on their recent success

maschad:
## Title

feat: ensure dials are sorted based on their recent success

## Description

Given that we are now tracking dials based on their `lastSuccess` and `lastFailure` we should ensure that they the are properly ordered before dialling, and subsequently update the peerStore with that metadata dependent on the outcome of that dial.

This PR also removes the `AUTO_DIAL_PEER_RETRY_THRESHOLD` since we are no longer sorting dials on a per peer basis but rather on a peer multiaddr basis. This adds a retry mechanism that exponentially backoffs of dialling failed addresses per peer. 

Closes https://github.com/libp2p/js-libp2p/issues/2150
Closes https://github.com/libp2p/js-libp2p/issues/2114

Related https://github.com/libp2p/js-libp2p/issues/2090

## Change checklist


packages/libp2p/src/connection-manager/dial-queue.ts
```diff
@@ -457,6 +484,17 @@ export class DialQueue {
               pendingDial.status = 'active'
 
               let conn: Connection
+              let peer: Peer | undefined
+
+              try {
+                if (pendingDial.peerId != null) {
+                  peer = await this.peerStore.get(pendingDial.peerId)
```
wemeetagain, line 491: maybe this peerStore.get should happen in _updateAddressStatus and leave _updateAddressStatus signature alone?

I think it tracks more nicely to have the get+update in the "same place".
Also thats one less promise that has to be fulfilled before we get to dial.

packages/libp2p/src/connection-manager/dial-queue.ts
```diff
@@ -550,15 +588,28 @@ export class DialQueue {
...
+    }
+
     if (success) {
       addr.lastSuccess = Date.now()
     } else {
-      addr.lastFailure = Date.now()
+      if (addr.lastFailure !== undefined) {
+        addr.lastFailure *= 2
+      } else {
+        addr.lastFailure = Date.now()
```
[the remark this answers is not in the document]
maschad: The `lastFailure` value is a unix timestamp but it's being utilized as a value for ordering the list of multiaddrs, therefore exponentially increasing it so that it will be penalized even further to the bottom of the list seems a valid approach i.e. it could be any unique incremental numerical value as long as it is sequentially increasing to create an order within the list of addrs.

packages/libp2p/src/connection-manager/dial-queue.ts
maschad: I take your point but when refactored that way a potential `ERR_NOT_FOUND` would be propagated in lieu of a consequent and perhaps more meaningful error such as a `NO_RESERVATION` error.

Issue #1913, opened 2023-07-31
refactor: move services to separate packages


maschad:
There are a few internal services such as [identify](https://github.com/libp2p/js-libp2p/blob/master/packages/libp2p/src/identify/README.md) , [Address Manager](https://github.com/libp2p/js-libp2p/blob/master/packages/libp2p/src/address-manager/README.md), [Pnet](https://github.com/libp2p/js-libp2p/blob/master/packages/libp2p/src/pnet/README.md) and others in the [libp2p package](https://github.com/libp2p/js-libp2p/tree/master/packages/libp2p/src) which have their own `README` that contains descriptions about how to configure the services. Because they are buried a few subdirectories down in the libp2p package they have low discoverability and so that information should be extracted to the [top level docs](https://github.com/libp2p/js-libp2p/blob/master/doc/CONFIGURATION.md) as well as their usage should be generated by our [`docs` tooling](https://github.com/ipfs/aegir#clihttps://github.com/ipfs/aegir#cli). This also unnecessarily bloates the libp2p npm package.

Related: https://github.com/libp2p/js-libp2p/pull/1910#pullrequestreview-1555165302

Packages:

```[tasklist]
### Tasks
- [ ] AutoNAT - #2107
- [ ] CircuitRelayServer #2222
- [ ] DCUtR #2220
- [ ] Identify #2219
- [ ] Fetch #2223
- [ ] Insecure / plaintext #2221
- [ ] Ping #2218 
- [ ] UnPnP #2217
- [ ] Pnet - https://github.com/libp2p/js-libp2p/pull/2165
```

achingbrain:
This has been completed in the v1.0.0 release branch.

PR #2165, opened 2023-10-20
refactor!: move pnet into separate package

wemeetagain:
## Description

- Move pnet module to a separate package
- Related to #1913

## Change checklist


packages/libp2p/test/connection-manager/direct.node.ts
```diff
@@ -496,9 +493,11 @@ describe('libp2p.dialer (direct, TCP)', () => {
   })
 
   it('should use the protectors when provided for connecting', async () => {
-    const protector: ConnectionProtector = preSharedKey({
-      psk: swarmKeyBuffer
-    })()
+    const protector: ConnectionProtector = {
+      async protect (connection) {
```
SgtPooki, line 497: since this is just a fake, now, this is a more isolated test to connection-manager's use of protectors, which is good.

do we need to test that the arguments passed to connection protector are correct?

does the protector have tests to confirm it is doing the right thing since we're no longer confirming that as a side-effect of this test?

packages/pnet/package.json
```diff
@@ -0,0 +1,67 @@
...
+    }
+  },
+  "eslintConfig": {
+    "extends": "ipfs",
+    "parserOptions": {
+      "sourceType": "module"
+    }
+  },
+  "scripts": {
+    "start": "node dist/src/main.js",
```
SgtPooki, line 35: `npm start` is usually an immediately available command after `npm i`, should this have `npm run build` in it?

packages/pnet/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@libp2p/pnet",
+  "version": "1.0.0",
+  "description": "Implementation of Connection protection management via a shared secret",
+  "license": "Apache-2.0 OR MIT",
+  "homepage": "https://github.com/libp2p/js-libp2p/tree/master/packages/pnet#readme",
```
SgtPooki, line 6: ✔️

packages/pnet/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "test:chrome": "aegir test -t browser --cov",
+    "test:chrome-webworker": "aegir test -t webworker",
+    "test:firefox": "aegir test -t browser -- --browser firefox",
+    "test:firefox-webworker": "aegir test -t webworker -- --browser firefox",
+    "test:node": "aegir test -t node --cov",
+    "dep-check": "aegir dep-check"
+  },
+  "dependencies": {
+    "@libp2p/crypto": "^2.0.5",
+    "@libp2p/interface": "^0.1.3",
```
SgtPooki, line 49: i cant wait for libp2p and similar to all be 1.0.0.. feels so odd depending on 0.n.n in a n.0.0 package

packages/pnet/src/errors.ts
```diff
@@ -3,3 +3,4 @@ export const INVALID_PSK = 'Your private shared key is invalid'
 export const NO_LOCAL_ID = 'No local private key provided'
 export const NO_HANDSHAKE_CONNECTION = 'No connection for the handshake provided'
 export const STREAM_ENDED = 'Stream ended prematurely'
+export const ERR_INVALID_PARAMETERS = 'ERR_INVALID_PARAMETERS'
```
SgtPooki, line 6: should this and others be in `@libp2p/interface` ?

I know we had a discussion about this somewhere..

packages/pnet/src/index.ts
```diff
@@ -81,7 +80,7 @@ class PreSharedKeyConnectionProtector implements ConnectionProtector {
     }
 
     if (connection == null) {
-      throw new CodeError(Errors.NO_HANDSHAKE_CONNECTION, codes.ERR_INVALID_PARAMETERS)
+      throw new CodeError(Errors.NO_HANDSHAKE_CONNECTION, Errors.ERR_INVALID_PARAMETERS)
```
SgtPooki, line 83: nit: we should normalize on `errors|codes.ACTUAL_ERR` and always import a singular object or always do `ACTUAL_ERR` 

also(prob for outside this PR but..), where does CodeError come from, and can we enforce the error types there? if we did, we could use generics on modules and ensure each module has their possible error types enforced too (makes it easier for devs to know what errors can be thrown)

packages/protocol-autonat/test/index.spec.ts
```diff
@@ -42,33 +39,30 @@ const defaultInit: AutoNATServiceInit = {
...
-    peerStore = stubInterface<PeerStore>()
 
-    components = defaultComponents({
+    components = {
       peerId: await createEd25519PeerId(),
       peerRouting,
       registrar,
       addressManager,
       connectionManager,
-      transportManager,
```
SgtPooki, line 69: the preference for no trailing commas makes this diff so bad...

SgtPooki (approved):
lgtm, some comments mostly for my learning

packages/protocol-autonat/test/index.spec.ts
wemeetagain: *cries in needless diff lines*

cc @achingbrain

packages/pnet/package.json
achingbrain: This should have been removed.

packages/pnet/package.json
achingbrain: Soon (tm)

packages/pnet/src/errors.ts
achingbrain: Yes, some generic codes will reduce duplication but no need to block on this.

packages/pnet/src/index.ts
achingbrain: > where does CodeError come from

```
import { CodeError } from '@libp2p/interface/errors'
```

> we could use generics

Maybe, I think this might get quite unwieldy.  Feel free to open a PR to discuss?

achingbrain (approved):

PR #2166, opened 2023-10-24
fix!: remove protocols from PeerInfo

achingbrain:
A [PeerInfo](https://docs.libp2p.io/concepts/fundamentals/peers/#peer-info) is a libp2p object that combines a PeerID and some Multiaddrs.

We also add a list of protocols. This was a mistake because protocols are exchanged during Identify but the PeerInfo object is used for peer discovery.

This is evident because we set the protocol list to an empty array everywhere.

PeerInfo is useful for exchanging peer data with other nodes, if we need a more fleshed-out peer representation we'd use the `Peer` interface from the peer store.

BREAKING CHANGE: the `.protocols` property has been removed from the `PeerInfo` interface

## Change checklist

### DIFF
--- a/packages/interface-compliance-tests/src/mocks/peer-discovery.ts
+++ b/packages/interface-compliance-tests/src/mocks/peer-discovery.ts
@@ -49,8 +49,7 @@ export class MockDiscovery extends EventEmitter<PeerDiscoveryEvents> implements 
           this.safeDispatchEvent<PeerInfo>('peer', {
             detail: {
               id: peerId,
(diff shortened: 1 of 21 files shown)


wemeetagain (approved):

maschad:
We should probably add this to the migration doc in case a consumer may have been updating their peerStore with the protocols, as was done [in previous versions of libp2p](https://github.com/libp2p/js-libp2p/blob/master/doc/migrations/v0.27-v0.28.md#adding-a-peer)

PR #2169, opened 2023-10-24
fix: allow DHT self-query to time out

achingbrain:
When we run our first self-query we first check the routing table to see if it has peers.

If not we wait for some peers to be discovered before continuing.

The change here is to use the general query AbortSignal to abort waiting for peers to be discovered.  If they are not discovered allow other queries to continue and fail with eaiser to understand error messages.

Also fixes some logging typos and makes error messages more understandable.

## Change checklist

### DIFF
--- a/packages/kad-dht/src/dual-kad-dht.ts
+++ b/packages/kad-dht/src/dual-kad-dht.ts
@@ -103,7 +103,7 @@ function multiaddrIsPublic (multiaddr: Multiaddr): boolean {
 
   // dns4 or dns6 or dnsaddr
   if (tuples[0][0] === DNS4_CODE || tuples[0][0] === DNS6_CODE || tuples[0][0] === DNSADDR_CODE) {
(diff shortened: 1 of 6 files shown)


packages/kad-dht/src/dual-kad-dht.ts
```diff
@@ -103,7 +103,7 @@ function multiaddrIsPublic (multiaddr: Multiaddr): boolean {
 
   // dns4 or dns6 or dnsaddr
   if (tuples[0][0] === DNS4_CODE || tuples[0][0] === DNS6_CODE || tuples[0][0] === DNSADDR_CODE) {
-    log('%m is public %s', multiaddr, true)
+    log('%a is public %s', multiaddr, true)
```
maschad, line 106: Do we need these messages for general debugging or should they be trace logs?

achingbrain: They can probably be removed

PR #2173, opened 2023-10-25
fix: rename event emitter class

achingbrain:
We add types to a `EventEmitter` class which extends the plain js `EventTarget`.

The name is chosen to give familiarity to node developers but it's not generally a good idea.

This refactor:

1. Exports a new `TypedEventTarget` interface which adds types to `EventTarget`
2. Renames `EventEmitter` to `TypedEventEmitter` to draw a distinction between them
3. Re-exports `TypedEventEmitter` as `EventEmitter` to preserve backwards compatibility

In the future all consuming code should rely on the `TypedEventTarget` interface while implementing code uses `TypedEventEmitter` as an implementation of a `TypedEventEmitter`.

By depending on the interface and not the implementation consuming code is isolated from problems that arise when two versions of `@libp2p/interface` is in the dependency tree and the event subsystems would otherwise be compatible due to type overlap.

This manifests as an error message about incompatible implementations of the private `#listeners` field in EventEmitter.

Backwards compatibility is achieved by exporting the `TypedEventEmitter` class as the older `EventEmitter` name.  This should be removed in a future PR to the v1.0 branch once external modules (e.g. Gossipsub) have been updated to use the new `TypedEventEmitter` symbol.

## Change checklist


achingbrain:
The original version of this PR went out with the `refactor:` tag which `release-please` didn't translate into modules needing a release so `@libp2p/interface` had the `TypedEventEmitter` class added but it was not released.

Hold tight, a new version should be out shortly that fixes the issue.

achingbrain:
@MysticalDragon98 this should be fixed now, sorry for the disruption.  Please delete any lockfiles & reinstall deps.

PR #2180, opened 2023-10-30
docs: move docs to packageDocumentation

achingbrain:
Since ageir 41.1.0 the contents of the `@packageDocumentation` tag in the `/src/index.ts` file of a given module will be copied to the README for that repo under the `About` section, so move all hand-written README data into `@pacakgeDocumentation` tags so we can keep the docs up to date in just one place.


packages/peer-discovery-mdns/README.md
```diff
@@ -1,30 +1,17 @@
...
+## Example
 
-```console
-$ npm i @libp2p/mdns
-```
-
-## Usage
-
-```Typescript
+```ts
```
maschad, line 14: You've chosen to keep the Typescript example here (in the README) but moved it to the `@packageDocumentation` in other instances, is this intentional?

packages/kad-dht/README.md
```diff
@@ -40,66 +23,17 @@ Loading this module through a script tag will make it's exports available as `Li
...
-
-The libp2p-kad-dht module offers 3 APIs: Peer Routing, Content Routing and Peer Discovery.
-
-### Custom secondary DHT in libp2p
-
-```js
-import { createLibp2pNode } from 'libp2p'
-import { kadDHT } from '@libp2p/kad-dht'
-
-const node = await createLibp2pNode({
```
maschad, line 61: Where was this moved to?

maschad (commented):
My main aversion to this is that we lose the ability to lint the TS examples as the [typescript-docs-verifier](https://github.com/bbc/typescript-docs-verifier) runs on the markdown, I suppose this could be run when the docs are being generated from the `packageDocumentation` though.

PR #2181, opened 2023-10-30
fix(transport-webrtc): link encoding in readme

kumavis:
link encoding typo fix


achingbrain:
Good spot, thanks for opening this.  We've actually just reformatted a lot of the READMEs and that section has been removed so this change is no longer necessary, but please open PRs for anything else you spot!

PR #2182, opened 2023-11-01
fix: append peer id to connection remote addr if not present

achingbrain:
If we dial an address like `/ip4/123.123.123.123/tcp/123` we'll get a peer id back after the connection upgrade - append it to the remote addr to make it easier to test if we already have a connection to a given multiaddr.

## Change checklist


achingbrain:
Srsly, what is happening with the pubsub interop tests?!

maschad (approved):

PR #2183, opened 2023-11-01
chore: try exit after cov build

achingbrain:
The code coverage still causes builds to hang occasionally.

Try forcing exit through mocha options.

## Change checklist


achingbrain:
Doesn't fix the problem.

PR #2196, opened 2023-11-03
fix!: remove event emitter type from interfaces

achingbrain:
All code should be using `TypedEventEmitter` instead of `EventEmitter` so remove the re-export from the interfaces package.
    
BREAKING CHANGE: removed EventEmitter re-export - please use TypedEventEmitter instead

## Change checklist


achingbrain:
Blocked on https://github.com/ChainSafe/js-libp2p-gossipsub/pull/470

maschad (approved):

wemeetagain:
> Blocked on https://github.com/ChainSafe/js-libp2p-gossipsub/pull/470

should be unblocked now

Issue #2105, opened 2023-10-02
refactor: logging improvements - all components should receive a logger component


SgtPooki:
We discussed in https://pl-strflt.notion.site/Mob-Coding-session-2f0cdd202cdf4b6faec03bd87ca15703?pvs=4 how having a logger component that includes the last 5 characters of the peerId could be incredibly useful when running multiple nodes locally

One way we discussed for accomplishing this is to pass a logger component to all components similar to how other components reference each other.
```[tasklist]
### Tasks
- [ ] https://github.com/libp2p/js-libp2p/issues/378
- [ ] https://github.com/libp2p/js-libp2p/issues/2207
```

achingbrain:
Developer ergonomics should be considered.

Given the logger `libp2p:yamux` you can have:

```
libp2p:yamux hello
libp2p:yamux:error oh no an error
libp2p:yamux:trace really detailed trace log
```

Logging is disabled by default:

```console
$ libp2p
* no output *
```

Even when logging is enabled trace logging is still disabled by default:

```console
$ DEBUG=libp2p:* libp2p
libp2p:yamux hello
libp2p:yamux:error oh no an error
* no trace logging*
```

Turning on trace logging:

```console
$ DEBUG=libp2p:*,*:trace libp2p
libp2p:yamux hello
libp2p:yamux:error oh no an error
libp2p:yamux:trace really detailed trace log
```

Prefixing with the PeerId will result in:

```
123Foo:libp2p:yamux hello
123Foo:libp2p:yamux:error oh no an error
123Foo:libp2p:yamux:trace really detailed trace log
```

Turning on logging might then be like:

```console
$ DEBUG=*:libp2p:* libp2p
123Foo:libp2p:yamux hello
123Foo:libp2p:yamux:error oh no an error
```

Turning on trace logging is the same as before:

```console
$ DEBUG=*:libp2p:*,*:trace libp2p
123Foo:libp2p:yamux hello
123Foo:libp2p:yamux:error oh no an error
123Foo:libp2p:yamux:trace really detailed trace log
```

maschad:
> Developer ergonomics should be considered.


I agree but are these examples to suggest that it would be too noisy? Because the current alternative to this is to have the peerID appended such as:

```console
$ DEBUG=*:libp2p:*,*:trace libp2p
libp2p:yamux hello by peer 123Foo3252352QwmsaIksFN...
[...]

PR #2198, opened 2023-11-04
feat: add component logger

achingbrain:
When running multiple libp2p instances simultaneously it's often hard to work out which log messages come from which instance. This PR solves the problem.

- Adds a `ComponentLogger` interface to libp2p which is an object that lets a component get a logger specific to that component.
- Adds an optional `logger` config key to libp2p which is an implementation of `ComponentLogger`.
- Adds three implementations of `ComponentLogger`
  - `defaultLogger` - this preserves existing logging behaviour
  - `peerLogger` - this prefixes all log lines with a truncated peer id
  - `prefixLogger` - this prefixes all log lines with an arbitrary string

`libp2p` uses the passed logger if one is configured, otherwise it uses `defaultLogger` to make this change backwards compatible.

The logger is added to the `components` map so libp2p components can accept it as part of their required components and use the `forComponent` function to create a component-specific logger.

It will take a bit of time to thread the use of `ComponentLogger` throughout the monorepo, but this can be done in follow-up PRs since the existing logging behaviour is maintained by the changes here.

For v1.x we may switch to `peerLogger`.

Refs #2105

## Change checklist

### DIFF
--- a/packages/libp2p/src/libp2p.ts
+++ b/packages/libp2p/src/libp2p.ts
@@ -5,7 +5,7 @@ import { TypedEventEmitter, CustomEvent, setMaxListeners } from '@libp2p/interfa
 import { peerDiscovery } from '@libp2p/interface/peer-discovery'
 import { type PeerRouting, peerRouting } from '@libp2p/interface/peer-routing'
 import { DefaultKeyChain } from '@libp2p/keychain'
-import { logger } from '@libp2p/logger'
+import { defaultLogger } from '@libp2p/logger'
 import { PeerSet } from '@libp2p/peer-collections'
 import { peerIdFromString } from '@libp2p/peer-id'
 import { createEd25519PeerId } from '@libp2p/peer-id-factory'
@@ -28,7 +28,7 @@ import { DefaultTransportManager } from './transport-manager.js'
 import { DefaultUpgrader } from './upgrader.js'
 import type { Components } from './components.js'
 import type { Libp2p, Libp2pInit, Libp2pOptions } from './index.js'
-import type { Libp2pEvents, PendingDial, ServiceMap, AbortOptions } from '@libp2p/interface'
+import type { Libp2pEvents, PendingDial, ServiceMap, AbortOptions, ComponentLogger, Logger } from '@libp2p/interface'
 import type { Connection, NewStreamOptions, Stream } from '@libp2p/interface/connection'
 import type { KeyChain } from '@libp2p/interface/keychain'
 import type { Metrics } from '@libp2p/interface/metrics'
@@ -39,8 +39,6 @@ import type { Topology } from '@libp2p/interface/topology'
 import type { StreamHandler, StreamHandlerOptions } from '@libp2p/interface-internal/registrar'
 import type { Datastore } from 'interface-datastore'
 
-const log = logger('libp2p')
-
 export class Libp2pNode<T extends ServiceMap = Record<string, unknown>> extends TypedEventEmitter<Libp2pEvents> implements Libp2p<T> {
   public peerId: PeerId
   public peerStore: PeerStore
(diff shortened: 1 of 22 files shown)


packages/logger/src/index.ts
```diff
@@ -93,6 +96,100 @@ function createDisabledLogger (namespace: string): debug.Debugger {
...
+  suffixLength: number
+}
+
+/**
+ * Create a component logger that will prefix any log messages with a truncated
+ * peer id.
+ *
+ * @example
+ *
+ * ```js
```
maschad, line 110: we really should have these examples as TS and then add the capability for aegir to lint them.

packages/logger/src/index.ts
```diff
@@ -93,6 +96,100 @@ function createDisabledLogger (namespace: string): debug.Debugger {
...
+ *
+ * const peerId = peerIdFromString('12D3FooBar')
+ * const logger = peerLogger(peerId)
+ *
+ * const log = logger.forComponent('my-component')
+ * log.info('hello world')
+ * // logs "12D3…oBar:my-component hello world"
+ * ```
+ */
+export function peerLogger (peerIdOrString: PeerId, options: Partial<ComponentLoggerOptions> = {}): ComponentLogger {
```
maschad, line 122: Let's add some tests for these

packages/libp2p/test/connection-manager/index.spec.ts
```diff
@@ -365,7 +366,8 @@ describe('Connection Manager', () => {
       peerStore: stubInterface<PeerStore>(),
       transportManager: stubInterface<TransportManager>(),
       connectionGater: stubInterface<ConnectionGater>(),
-      events: new TypedEventEmitter()
+      events: new TypedEventEmitter(),
+      logger: defaultLogger()
```
maschad, line 370: shouldn't it default to the `defaultLogger` as opposed to passing it each time?

packages/libp2p/src/libp2p.ts
```diff
@@ -49,9 +47,11 @@ export class Libp2pNode<T extends ServiceMap = Record<string, unknown>> extends
   public keychain: KeyChain
   public metrics?: Metrics
   public services: T
+  public logger: ComponentLogger
 
   public components: Components
   #started: boolean
+  readonly #log: Logger
```
maschad, line 54: Any particular rationale that this is a private field as opposed to the components? or should they be private as well?

maschad (changes requested):
very useful feature, great work. I left some comments and suggestions.

packages/libp2p/test/connection-manager/index.spec.ts
achingbrain: Could probably remove a bit of the duplication, but the component interface for the connection manager requires a logger so it doesn't need to worry about defaulting it.

packages/libp2p/src/libp2p.ts
achingbrain: Anything not in the libp2p interface should be private.

packages/logger/src/index.ts
achingbrain: They're marked as ts now 👍

packages/logger/src/index.ts
achingbrain: I've added a test for the peer id truncation, not sure what else there is to test here.

packages/libp2p/src/libp2p.ts
achingbrain: Component log fields are private now.

packages/libp2p/test/connection-manager/index.spec.ts
achingbrain: I've deduped the component setup.

maschad (approved):

PR #2218, opened 2023-11-08
refactor!: extract ping service into separate module

achingbrain:
BREAKING CHANGE: imports from `libp2p/ping` must be updated to `@libp2p/ping`

## Change checklist


packages/protocol-ping/package.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "@libp2p/ping",
+  "version": "0.0.0",
```
wemeetagain, line 3: :eyes:

achingbrain: Yes, this is a breaking change in this module so publishing will give us v1

Issue #2136, opened 2023-10-06
React Native support


achingbrain:
We'd love to officially support running on React Native.

The requirements for this would be:

1. The unit test suite running in CI
2. Lightweight - e.g. the developer should not have to install extra applications or sign EULAs if possible

There's been some preliminary support added to aegir for running tests under Android but I'm not sure what the current state of the art is in automated testing.

A preliminary pass might just be to get a simple demo react-native app running libp2p working.  This might involve a list of required changes to modules to polyfill or replace APIs that do not exist/are buggy/incomplete in React Native.

If it helps to reduce the scope, perhaps a stand-alone repo could be tackled first to prove it out instead of the js-libp2p monorepo.

This was discussed briefly at the js-colo in NYC last week.

Refs:
- https://github.com/ipfs/aegir/issues/437
- https://github.com/ipfs/aegir/pull/777

Demo:

- https://github.com/maschad/libp2p-react-native

maschad:
I've been putting together a [demo](https://github.com/maschad/libp2p-react-native) that hopefully we can showcase at [libp2p Day](https://github.com/libp2p/specs/discussions/591#discussioncomment-7492763) there's still some work to do, as @achingbrain mentioned a bulk of the work has been polyfill-ing and adapting modules. 

the goal is to hopefully have two mobile nodes connected via websockets and/or a laptop as well for bonus.

anistark:
Is this supported now for React Native in iOS ?

maschad:
Not yet @anistark I will update this issue once either platform is supported, for now you can track progress at https://github.com/maschad/libp2p-react-native

achingbrain:
The latest patch release of js-libp2p & supporting modules have shipped with some fixes for running under react-native.

There's a new demo app at https://github.com/ipfs-shipyard/js-libp2p-react-native which uses Expo to build a simple app with a running libp2p node.

It currently supports WebSocket and tcp transports.  WebRTC should also work via [react-native-webrtc](https://www.npmjs.com/package/react-native-webrtc) but I'm having some problems - `WebRTCModule` being passed to `new NativeEventEmitter` [here](https://github.com/react-native-webrtc/react-native-webrtc/blob/871c465c84c7ebfcc304f6f2df4d770dbb78e945/src/EventEmitter.ts#L9) is null which probably means the demo is missing some config - PRs are gratefully accepted.

Please try it out and let us know how you get on.

achingbrain:
[...]

PR #2229, opened 2023-11-10
fix: react native adjustments

maschad:
- [ ] https://github.com/PeculiarVentures/webcrypto/pull/67
- [ ] https://github.com/alanshaw/it-ws/pull/97

Related https://github.com/libp2p/js-libp2p/issues/2136


achingbrain:
This is the app, very basic so far: https://github.com/ipfs-shipyard/js-libp2p-react-native

achingbrain:
Actually it sort of does but generators don't appear to have the `Symbol.asyncIterator` property.

achingbrain:
The crypto situation is pretty dire, from what I can gather there's no [SubtleCrypto](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) support and Facebook's response to questions about adding it were [lukewarm](https://github.com/facebook/react-native/issues/22933).

This rules out simply following the browser path in modules with node/browser support since they depend on SubtleCrypto.

We could shim in pure-js versions of everything but crypto is such a bottleneck for us that doing everything in js on your phone is likely to melt it.

The [community response](https://github.com/react-native-community/discussions-and-proposals/issues/83) yielded [react-native-webcrypto](https://www.npmjs.com/package/react-native-webcrypto) but it's incomplete and only works on Android.

[react-native-webview-crypto](https://www.npmjs.com/package/react-native-webview-crypto) is an interesting hack though the [library it depends on](https://github.com/webview-crypto/webview-crypto) has a showstopping bug that's only had a [PR to fix it](https://github.com/webview-crypto/webview-crypto/pull/6) open for three years.  It also involves stringifying and copying all data to be encrypted/decryped to/from a webview - I haven't measured it but common sense says that will not be a performant approach.

We could patch the library to fix the bug and [transfer](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects) the `Uint8Array`s but that will break [clever memory optimisations elsewhere](https://github.com/ChainSafe/js-libp2p-noise/blob/master/src/crypto/streaming.ts#L51-L55) because the underlying `Uint8Array`s will become unusable after the transfer.

----

The other approach is to follow the node path instead and use [rn-nodeify](https://www.npmjs.com/package/rn-nodeify) to shim node core modules in where necessary.

This opens up the possibility of using [react-native-quick-crypto](https://www.npmjs.com/package/react-native-quick-crypto) which does crypto in C/C++ so would be preferable.

Unlike most of the other crypto libraries here, quick-crypto appears to [still be under development](https://github.com/margelo/react-native-quick-crypto/pull/200) so that's heartening.

achingbrain:
Getting there, it can dial the bootstrappers now:

image

achingbrain:
Circuit relay works:

image

Issue #2238, opened 2023-11-17
bug: webrtc-signaling proto ID does not match spec


SgtPooki:
https://github.com/libp2p/js-libp2p/blob/9c67c5b3d0ab63c7a1a62f363ae732b300ef6b87/packages/transport-webrtc/src/private-to-private/transport.ts#L25

The spec says to use `/webrtc-signaling`. see https://github.com/libp2p/specs/blob/master/webrtc/webrtc.md#signaling-protocol

Discovered this with @sukunrt when doing some webrtc transport stuff locally. 

### Some notes:

He updated his proto_id to include the `0.0.1` and we were able to fully connect (using local [helia-identify](https://helia-identify.on.fleek.co/)) ran on his address of `/ip6/<blah>/udp/4001/quic-v1/webtransport/certhash/<blah>/certhash/<blah>/p2p/<blah>/p2p-circuit/webrtc/p2p/<blah>`

but we need to find agreement on the protocol id name

achingbrain:
The addition of the version number here seems to be a mistake, but removing it will break compatibility with any deployed js-libp2p nodes.

Given that this is the only shipping implementation of WebRTC private to private, perhaps the simpler solution would be to add the version number [to the spec](https://github.com/libp2p/specs/blob/master/webrtc/webrtc.md#signaling-protocol)?

maschad:
We decided updating the spec would cause less friction https://github.com/libp2p/specs/pull/598

PR #2239, opened 2023-11-17
fix!: remove version number from webrtc signalling protocol 

maschad:
## Title

fix!: remove version number from webrtc signalling protocol

## Description

BREAKING CHANGE: I don't see a rationale for having a version number at the moment and the [spec](https://github.com/libp2p/specs/blob/master/webrtc/webrtc.md#signaling-protocol) doesn't recommend this.

Closes https://github.com/libp2p/js-libp2p/issues/2238 

## Notes & open questions

Should there be a reason to include a version number a discussion could be opened on the spec

## Change checklist

### DIFF
--- a/packages/transport-webrtc/src/private-to-private/transport.ts
+++ b/packages/transport-webrtc/src/private-to-private/transport.ts
@@ -22,7 +22,7 @@ import type { TransportManager } from '@libp2p/interface-internal/transport-mana
 
 const WEBRTC_TRANSPORT = '/webrtc'
 const CIRCUIT_RELAY_TRANSPORT = '/p2p-circuit'
-export const SIGNALING_PROTO_ID = '/webrtc-signaling/0.0.1'
+export const SIGNALING_PROTO_ID = '/webrtc-signaling'
 const INBOUND_CONNECTION_TIMEOUT = 30 * 1000
 
 export interface WebRTCTransportInit {


achingbrain:
Granted the spec doesn't have a version number for the protocol so the addition of the version number seems to be a mistake, but this change will break compatibility with any deployed js-libp2p nodes.

Given that this is the only shipping implementation of WebRTC private to private, perhaps the simpler solution would be to add the version number [to the spec](https://github.com/libp2p/specs/blob/master/webrtc/webrtc.md#signaling-protocol)?

achingbrain (changes requested):
I'd like to explore the idea that adding a version number to the spec may be a less disruptive change than removing it here.

This change will break compatibility with any currently deployed js-libp2p nodes.

maschad:
> Granted the spec doesn't have a version number for the protocol so the addition of the version number here seems to be a mistake, but this change will break compatibility with any deployed js-libp2p nodes.
> 
> Given that this is the only shipping implementation of WebRTC private to private, perhaps the simpler solution would be to add the version number [to the spec](https://github.com/libp2p/specs/blob/master/webrtc/webrtc.md?rgh-link-date=2023-11-17T17%3A48%3A02Z#signaling-protocol)?

That's a valid point but despite this being a breaking change I don't think that we should introduce versioning at the spec level just because this implementation was mistakenly versioned. I would be interested to see what other parties think though so perhaps open a discussion on specs?

achingbrain:
Plenty of our other specs have versions so I don't think it's a massive ask, plus they are useful for future proofing ourselves.

We need to value reducing disruption to users above our own wish to have everything just so.

wemeetagain:
I think the version in a protocol string can be valuable and help provide a clear upgrade path.  Also tend to think of protocols without explicit versions as an implicit 1.0.
When inevitably things change, and a new protocol spec supercedes an old one, the version helps aid in backwards compatibility decisions and, more subjectively, how to understand the evolution of a protocol (ie from v1 to v2).

maschad:
Closing in favour of https://github.com/libp2p/specs/pull/598

PR #2241, opened 2023-11-18
feat: support default options for tcp sockets/servers

achingbrain:
Allow specifying default socket options to `net.connect` and `net.createServer` - this allows users to turn keep alive, tcp delays, etc on/off as per their individual requirements.

```js
tcp({
  socket: {
    // net.connect options here
  },
  server: {
    // net.createServer options here
  }
})

```

## Change checklist

### DIFF
--- a/packages/transport-tcp/src/utils.ts
+++ b/packages/transport-tcp/src/utils.ts
@@ -22,7 +22,7 @@ export function multiaddrToNetConfig (addr: Multiaddr, config: NetConfig = {}): 
   }
 
   // tcp listening
-  return { ...addr.toOptions(), ...config }
+  return { ...config, ...addr.toOptions() }
 }
 
 export function getMultiaddrs (proto: 'ip4' | 'ip6', ip: string, port: number): Multiaddr[] {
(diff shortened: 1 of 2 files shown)


packages/transport-tcp/src/utils.ts
```diff
@@ -22,7 +22,7 @@ export function multiaddrToNetConfig (addr: Multiaddr, config: NetConfig = {}):
   }
 
   // tcp listening
-  return { ...addr.toOptions(), ...config }
+  return { ...config, ...addr.toOptions() }
```
achingbrain, line 25: Ensure we take the target host/port from the multiaddr and not the passed options.

wemeetagain (approved):

achingbrain:
Do you think the option names are obvious enough?  `"socket"` is maybe a bit too generic?  `client`/`server`, `listener`/`dialer`, `listenTcpOpts`/`dialTcpOpts`?

The Node.js type name is `TcpSocketConnectOpts` if that helps.

wemeetagain:
I think listenOpts / dialOpts sounds good

PR #2248, opened 2023-11-20
docs: update tcp socket options references

maschad:
## Title

docs: update tcp socket options references

## Description

The previous urls did not link to the appropriate docs.

## Change checklist


packages/transport-tcp/src/index.ts
```diff
@@ -118,17 +118,17 @@ export interface TCPOptions {
  */
 export interface TCPSocketOptions extends AbortOptions {
   /**
-   * @see https://nodejs.org/dist/latest-v18.x/docs/api/net.html#serverlisten
+   * @see https://nodejs.org/dist/latest-v18.x/docs/api/net.html#socketconnectoptions-connectlistener
```
achingbrain, line 121: Maybe we should link to the latest docs here too instead of node 18?

```suggestion
   * @see https://nodejs.org/dist/latest/docs/api/net.html#socketconnectoptions-connectlistener
```

achingbrain (approved):

PR #2250, opened 2023-11-21
chore: re-enable doc-check for Typescript snippets

maschad:
## Title

chore: enable typescript docs verification for example snippets

## Description

This had been disabled leading to some erroneous Typescript snippets in the examples

## Change checklist


doc/package.json
```diff
@@ -19,6 +19,7 @@
       "sourceType": "module"
     }
   },
+  "exports": {},
```
maschad, line 22: This is a hack for now given that [typescript docs verifier](https://github.com/bbc/typescript-docs-verifier/blob/30a2d634b7cae88b68d8cbcbcb873b222ea3b732/src/PackageInfo.ts#L69) relies on the exports for its check

doc/METRICS.md
```diff
@@ -86,8 +86,11 @@ A tracked metric can be created by calling either `registerMetric` on the metric
 ```ts
 import type { Metrics } from '@libp2p/interface/metrics'
 import { prometheusMetrics } from '@libp2p/prometheus-metrics'
+import { defaultLogger } from '@libp2p/logger'
 
-const metrics: Metrics = prometheusMetrics()()
+const metrics: Metrics = prometheusMetrics()({
+  logger: defaultLogger()
```
achingbrain, line 92: I understand why this is necessary but I don't think makes for the clearest documentation.  Perhaps these examples can be refactored to be a generic libp2p service that receives a `Metrics` object and adds  metrics to it?

Then we wouldn't need to get loggers involved.

achingbrain (changes requested):
There are some good doc improvements in here so I'd like to merge this but where is the bit of the PR that re-enables doc-check for TypeScript snippets like the title says?

maschad:
Strange, I had assumed [CI was running the command](https://github.com/libp2p/js-libp2p/actions/runs/7018198824/job/19093168666?pr=2250) and that the actual `doc-check` command was a no-op, which is what this PR addressed. Turns out the CI was also a no-op as well, so I pushed a commit to fix in de72f20.

It also turns out [a recent typescript upgrade](https://github.com/microsoft/TypeScript/issues/56508) broke [ts-node](https://github.com/TypeStrong/ts-node/issues/2076#issuecomment-1823221840) which `typescript-docs-verifier` depends on so I am using a fixed Typescript version until that gets resolved and the ts-node fix should bubble upstream.

doc/package.json
```diff
@@ -30,7 +31,8 @@
     "@libp2p/tcp": "^9.0.0",
     "aegir": "^41.0.2",
     "libp2p": "^1.0.0",
-    "prom-client": "^15.0.0"
+    "prom-client": "^15.0.0",
+    "typescript": "5.2.2"
```
achingbrain, line 35: Let's wait for the fix to bubble up instead of trying to manage multiple versions of typescript in the monorepo.

wemeetagain (approved):

achingbrain (approved):

PR #2251, opened 2023-11-21
refactor!: remove per-peer parallel dialling (#2090)

achingbrain:
Since we are now dialling one peer at a time as a means of smarter dialling, we no longer need the option to control the concurrency of parallel dials across a single peers' multiaddrs.

Closes https://github.com/libp2p/js-libp2p/issues/2090

## Change checklist


achingbrain:
Recreation of #2163 that somehow got omitted from the main branch.

Issue #378, opened 2019-07-14
Prefix debug messages per node with peerID


mkg20001:
- **Version**: *
- **Platform**: *
- **Subsystem**: *

#### Description: Prefix debug messages per node with peerID

Let's take this log as an example:

```
  libp2p:switch:dialer create: 8 peer limit, 30000 dial timeout +0ms
  libp2p:connection-manager options: {"maxPeers":null,"minPeers":25,"maxData":null,"maxSentData":null,"maxReceivedData":null,"maxEventLoopDelay":null,"pollInterval":2000,"movingAverageInterval":60000,"defaultPeerValue":1,"maxPeersPerProtocol":{}} +0ms
  libp2p:switch:transport adding Circuit +0ms
  libp2p:switch:dialer create: 8 peer limit, 30000 dial timeout +13ms
  libp2p:connection-manager options: {"maxPeers":null,"minPeers":25,"maxData":null,"maxSentData":null,"maxReceivedData":null,"maxEventLoopDelay":null,"pollInterval":2000,"movingAverageInterval":60000,"defaultPeerValue":1,"maxPeersPerProtocol":{}} +10ms
  libp2p:switch:transport adding Circuit +9ms
  libp2p:switch:dialer create: 8 peer limit, 30000 dial timeout +14ms
  libp2p:connection-manager options: {"maxPeers":null,"minPeers":25,"maxData":null,"maxSentData":null,"maxReceivedData":null,"maxEventLoopDelay":null,"pollInterval":2000,"movingAverageInterval":60000,"defaultPeerValue":1,"maxPeersPerProtocol":{}} +14ms
  libp2p:switch:transport adding Circuit +14ms
  libp2p libp2p is starting +0ms
  libp2p:switch The switch is starting +0ms
  libp2p:switch The switch has started +4ms
  libp2p:switch:transport adding WebSockets +17ms
  libp2p libp2p has started +9ms
  libp2p libp2p is starting +1ms
  libp2p:switch The switch is starting +3ms
  libp2p:switch The switch has started +1ms
  libp2p:switch:transport adding WebSockets +3ms
  libp2p libp2p has started +2ms
  libp2p libp2p is starting +0ms
  libp2p:switch:transport adding WebSockets +2ms
  libp2p:switch The switch is starting +2ms
  libp2p:switch The switch has started +11ms
  libp2p libp2p has started +14ms
  libp2p:exchange:rendezvous:server server started +0ms
[...]

PR #2257, opened 2023-11-26
feat: use peer logger by default

achingbrain:
Switches the default logger to one that prefixes log lines with a truncated version of the node's peer id.

This can be overridden at configuration time, eg:

```js
const node = await createLibp2p()

// e.g.
// 12…tsSg:libp2p:component hello world
```

```js
import { defaultLogger } from '@libp2p/logger'

  logger: defaultLogger()
})

// e.g.
// libp2p:component hello world
```

```js
import { prefixLogger } from '@libp2p/logger'

  logger: prefixLogger('custom-prefix')
})

// e.g.
// custom-prefix:libp2p:component hello world
```

Refs: #2105
Refs: #2198
Refs: #378

## Change checklist


wemeetagain (approved):

achingbrain:
Do we want this? I know we said we did, it makes life easier when you have two nodes running in the same process but it’s not really a real-world setup.

wemeetagain:
Its also may be a problem that no longer exists now that the logger is centrally configured and passed thru all the components.

achingbrain:
Going to close this, it's easy enough to configure if the user wants it.

PR #2289, opened 2023-12-02
fix: remove duplicate autodial from startup

achingbrain:
There's no need to set a timeout for autodialing during `.start` since we autodial during `.afterStart` which will then set the timeout for the next dial if one is required.

We guard on running twice so this isn't a problem, it's just unecessary.

## Change checklist

### DIFF
--- a/packages/libp2p/src/connection-manager/auto-dial.ts
+++ b/packages/libp2p/src/connection-manager/auto-dial.ts
@@ -101,12 +101,6 @@ export class AutoDial implements Startable {
   }
 
   start (): void {
-    this.autoDialInterval = setTimeout(() => {
-      this.autoDial()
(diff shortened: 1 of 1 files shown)


packages/libp2p/src/connection-manager/auto-dial.ts
```diff
@@ -126,28 +120,27 @@ export class AutoDial implements Startable {
...
-    if (!this.started) {
+    if (!this.started || this.running) {
       return
     }
 
     const connections = this.connectionManager.getConnectionsMap()
     const numConnections = connections.size
 
-    // Already has enough connections
+    // already hava enough connections
```
achingbrain, line 130:
```suggestion
    // already have enough connections
```

PR #2299, opened 2023-12-05
fix: WebRTC transport unhandled promise rejection during connect

achingbrain:
Simplifies the connection logic to just use the abort signal to abort the dial instead of the abort signal and multiple deferred promises.

Fixes a crash in node where the abort signal passed into a WebRTC private to private dial can cause an unhandled promise rejection due to a race condition between the abort firing and awaiting the `connectedPromise` promise:

```
file:///Users/alex/Documents/Workspaces/ipfs/helia-http-gateway/node_modules/@libp2p/webrtc/dist/src/private-to-private/initiate-connection.js:42
            connectedPromise.reject(new CodeError('SDP handshake aborted', 'ERR_SDP_HANDSHAKE_ABORTED'));
                                    ^

CodeError: SDP handshake aborted
    at EventTarget.sdpAbortedListener (file:///Users/alex/Documents/Workspaces/ipfs/helia-http-gateway/node_modules/@libp2p/webrtc/dist/src/private-to-private/initiate-connection.js:42:37)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:807:20)
    at EventTarget.dispatchEvent (node:internal/event_target:742:26)
    at abortSignal (node:internal/abort_controller:369:10)
    at AbortController.abort (node:internal/abort_controller:391:5)
    at EventTarget.onAbort (file:///Users/alex/Documents/Workspaces/ipfs/helia-http-gateway/node_modules/any-signal/dist/src/index.js:8:20)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:807:20)
    at EventTarget.dispatchEvent (node:internal/event_target:742:26)
    at abortSignal (node:internal/abort_controller:369:10)
    at AbortController.abort (node:internal/abort_controller:391:5) {
  code: 'ERR_SDP_HANDSHAKE_ABORTED',
  props: {}
}

Node.js v20.8.0
```

## Change checklist

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation if necessary (this includes comments as well)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2302, opened 2023-12-05
fix: WebRTC uncaught promise rejection on incoming connection

achingbrain:
We catch WebRTC errors, log them and throw normalised errors to ensure they are the same across WebRTC stacks (FF, Chrome, node).

#2299 made one of these operations async which would cause an UHPR if thrown.

## Change checklist

### DIFF
--- a/packages/transport-webrtc/src/private-to-private/signaling-stream-handler.ts
+++ b/packages/transport-webrtc/src/private-to-private/signaling-stream-handler.ts
@@ -1,7 +1,6 @@
 import { CodeError } from '@libp2p/interface'
 import { multiaddr, type Multiaddr } from '@multiformats/multiaddr'
 import { pbStream } from 'it-protobuf-stream'
-import pDefer, { type DeferredPromise } from 'p-defer'
 import { type RTCPeerConnection, RTCSessionDescription } from '../webrtc/index.js'
 import { Message } from './pb/message.js'
 import { readCandidatesUntilConnected } from './util.js'
@@ -20,8 +19,6 @@ export async function handleIncomingStream ({ peerConnection, stream, signal, co
   const messageStream = pbStream(stream).pb(Message)
 
   try {
-    const answerSentPromise: DeferredPromise<void> = pDefer()
-
     // candidate callbacks
     peerConnection.onicecandidate = ({ candidate }) => {
       // a null candidate means end-of-candidates, an empty string candidate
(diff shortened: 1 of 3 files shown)


packages/transport-webrtc/src/private-to-private/signaling-stream-handler.ts
```diff
@@ -96,11 +87,6 @@ export async function handleIncomingStream ({ peerConnection, stream, signal, co
       signal,
       log
     })
-
-    log.trace('recipient connected, closing signaling stream')
-    await messageStream.unwrap().unwrap().close({
-      signal
-    })
```
achingbrain, line 103: This method is called from `_onProtocol` in `transport.ts` which handles closing/aborting the stream so there's no need to do it here.

packages/transport-webrtc/src/private-to-private/signaling-stream-handler.ts
```diff
@@ -96,11 +87,6 @@ export async function handleIncomingStream ({ peerConnection, stream, signal, co
       signal,
       log
     })
```
SgtPooki, line 89: should we still `log.trace('recipient connected, closing signaling stream')` ?

packages/transport-webrtc/src/private-to-private/signaling-stream-handler.ts
```diff
@@ -78,16 +74,11 @@ export async function handleIncomingStream ({ peerConnection, stream, signal, co
       signal
     })
 
-    peerConnection.setLocalDescription(answer).then(() => {
-      answerSentPromise.resolve()
-    }, err => {
+    await peerConnection.setLocalDescription(answer).catch(err => {
```
SgtPooki, line 77: :+1:

SgtPooki (approved):

packages/transport-webrtc/src/private-to-private/signaling-stream-handler.ts
achingbrain: I would have thought not, since this function isn't closing the signalling stream any more.

SgtPooki: I was thinking that we're still waiting for `readCandidatesUntilConnected` and the last log was "reading until connected", so in logs it could look like you're never done reading, nor connected, unless we see a "done reading" type message"

maybe my original comment should have asked for keeping `log.trace('recipient connected, stopped reading')`

PR #2303, opened 2023-12-05
feat!: add private key to libp2p components

wemeetagain:
## Description

- Add `privateKey` to `Components` and use it where possible
- Update pubsub, identify, circuit-relay v2, record envelope sealing to use `privateKey` instead of `peerId`

## Change checklist

BREAKING CHANGE:
- `IdentifyComponents` now includes a `privateKey`
- `PubSubComponents` now includes a `privateKey`
- `CircuitRelayServerComponents` now includes a `privateKey`
- `RecordEnvelope.seal` now has a `privateKey` as the second param rather than a `peerId`
- `RecordEnvelope` constructor now takes a `publicKey`


achingbrain:
Is there a way we can do this in stages to prevent it causing libp2p v2?  Perhaps just adding the private key in the first instance, then releasing it, then upgrading noise etc to use it?

dhuseby:
Blocked waiting for the next breaking release

achingbrain:
Superseded by #2660

PR #2304, opened 2023-12-05
fix!: remove localPeer from secureInbound and secureOutbound

wemeetagain:
## Description

- Remove `localPeer: PeerId` first parameter from `secureInbound` and `secureOutbound` in `ConnectionEncrypter`

Requiring the local peer id in connection encryption appears to be historical cruft. The more current approach would be to have a `ConnectionEncrypter` be specific to a single peer id, passed in during instantiation in its Components.

## Change checklist


dhuseby:
Blocked waiting for the next breaking change milestone

achingbrain:
Depends on:

- [ ] https://github.com/ChainSafe/js-libp2p-noise/pull/440

packages/connection-encrypter-tls/test/index.spec.ts
```diff
@@ -46,8 +47,8 @@ describe('tls', () => {
     })
 
     await Promise.all([
-      encrypter.secureInbound(remotePeer, inbound),
-      encrypter.secureOutbound(localPeer, outbound, wrongPeer)
+      encrypter.secureInbound(inbound, remotePeer),
+      encrypter.secureOutbound(outbound, wrongPeer)
```
SgtPooki, line 51: Curious, do we need an `encrypterRemote` here like we have in [packages/connection-encrypter-plaintext/test/index.spec.ts](https://github.com/libp2p/js-libp2p/pull/2304/files#diff-b48cf156af4dc44ee636b9b7f7f19255ef04ce7ceb471c7e7c68dfb8d88ef6ae) ?

If we can do either/or, it would be nice to use the same setup to not cause confusion and questions like this one :)

If we can't, maybe a comment explaining why? (maybe only for libp2p noobs like myself)

packages/interface-compliance-tests/src/connection-encryption/index.ts
```diff
@@ -47,8 +54,8 @@ export default (common: TestSetup<ConnectionEncrypter>): void => {
         inboundResult,
         outboundResult
       ] = await Promise.all([
-        crypto.secureInbound(remotePeer, localConn),
-        crypto.secureOutbound(localPeer, remoteConn, remotePeer)
+        cryptoRemote.secureInbound(localConn),
+        crypto.secureOutbound(remoteConn, remotePeer)
```
SgtPooki, line 58: similar question about dissimilarities between encrypters we create&call for `secureInbound` and `secureOutbound` calls

SgtPooki (commented):
Looks good to me, but I added a few questions for the tests. Much nicer to have each encrypter have the peerId within them.

approved, but not approving because I believe we're still waiting on another package

packages/connection-encrypter-tls/test/index.spec.ts
achingbrain: It would make the test easier to read but this encrypter is stateless which is why it works as-is.

achingbrain (approved):

PR #2319, opened 2023-12-17
chore: release main

achingbrain:
:robot: I have created a release *beep* *boop*
---

<details><summary>@libp2p/bootstrap: 10.0.8</summary>

### Dependencies

* The following workspace dependencies were updated
  * devDependencies
    * @libp2p/interface-compliance-tests bumped from ^5.0.7 to ^5.0.8
</details>

<details><summary>circuit-relay-v2: 1.0.8</summary>

### Bug Fixes

* query for circuit relays after start ([#2309](https://github.com/libp2p/js-libp2p/issues/2309)) ([dc56856](https://github.com/libp2p/js-libp2p/commit/dc56856f3d1d7603c3b0cc79afea7eef36a323c9))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @libp2p/peer-record bumped from ^7.0.1 to ^7.0.2
    * @libp2p/utils bumped from ^5.0.3 to ^5.1.0
  * devDependencies
    * @libp2p/interface-compliance-tests bumped from ^5.0.7 to ^5.0.8
</details>

<details><summary>@libp2p/floodsub: 9.0.4</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @libp2p/pubsub bumped from ^9.0.2 to ^9.0.3
  * devDependencies
    * @libp2p/interface-compliance-tests bumped from ^5.0.7 to ^5.0.8
</details>

<details><summary>@libp2p/identify: 1.0.7</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @libp2p/peer-record bumped from ^7.0.1 to ^7.0.2
</details>

<details><summary>@libp2p/interface-compliance-tests: 5.0.8</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @libp2p/multistream-select bumped from ^5.0.2 to ^5.0.3
    * @libp2p/utils bumped from ^5.0.3 to ^5.1.0
</details>

<details><summary>@libp2p/kad-dht: 11.0.8</summary>

### Dependencies

* The following workspace dependencies were updated
  * devDependencies
    * @libp2p/interface-compliance-tests bumped from ^5.0.7 to ^5.0.8
    * @libp2p/peer-store bumped from ^10.0.2 to ^10.0.3
</details>

<details><summary>@libp2p/mdns: 10.0.8</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
[...]

PR #2322, opened 2023-12-18
feat!: use single DHT only by default

achingbrain:
We have a weird dual-DHT setup whereby we have a DHT for public addresses (amino, `/ipfs/kad/1.0.0`) and a DHT for private addresess (`/ipfs/lan/kad/1.0.0`).

This is an artefact of the previous libp2p setup whereby there could only be a single DHT implementation at once.

Now we have services we can configure an amino DHT service and if desired an additional lan-only DHT, or the user can just use amino and not pay the overhead of running an extra DHT.

Most content is resolved via the public DHT so running the lan-only DHT gives the user no benefit.

BREAKING CHANGE: the `kadDHT` function returns a single DHT - see the readme for how to configure amino/lan as before

## Change checklist


wemeetagain:
👍

achingbrain:
Going to merge this after https://github.com/libp2p/js-libp2p/pull/2319 because https://github.com/libp2p/js-libp2p/pull/2313 should allow us to do a breaking change to kad-dht without revving libp2p to v2

wemeetagain (approved):

PR #2323, opened 2023-12-18
docs: use ts in examples, ensure package docs are up to date

achingbrain:
## Change checklist

Closes https://github.com/libp2p/js-libp2p/issues/1777


maschad (changes requested):
We should wait until `doc-check` is re-enabled in https://github.com/libp2p/js-libp2p/pull/2250 before merging this to ensure they are up to date.

achingbrain:
AFAIK #2250 is blocked on https://github.com/TypeStrong/ts-node/issues/2076 making it into a release so let's not wait for that here.

PR #2332, opened 2023-12-20
docs: add note about persisting peer ids in the keychain

achingbrain:
This was missed from the release notes.

## Change checklist

### DIFF
--- a/doc/migrations/v0.46-v1.0.0.md
+++ b/doc/migrations/v0.46-v1.0.0.md
@@ -20,6 +20,7 @@ A migration guide for refactoring your application code from libp2p `v0.46` to `
   - [Misc](#misc)
     - [KeyChain](#keychain)
   - [Pnet](#pnet)
+- [Keeping the same PeerId after a restart](#keeping-the-same-peerid-after-a-restart)
 - [Metrics](#metrics)
 - [Connection Manager](#connection-manager)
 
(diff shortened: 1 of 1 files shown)


doc/migrations/v0.46-v1.0.0.md
```diff
@@ -365,6 +366,52 @@ const node = await createLibp2p({
 })
 ```
 
+## Keeping the same PeerId after a restart
+
+The libp2p PeerId is a public/private keypair. libp2p uses the keychain to encrypt it at rest when the node is not running.
```
maschad, line 371:
```suggestion
Given the libp2p `PeerId` is a public/private keypair, it is encrypted by libp2p's `Keychain` whenever the node is not running.
```

doc/migrations/v0.46-v1.0.0.md
```diff
@@ -365,6 +366,52 @@ const node = await createLibp2p({
...
+
+As of `libp2p@1.0.0` the keychain is no longer part of the default bundle so to replicate this behaviour you need to use a temporary keychain instance to load the peer id from the datastore before starting your node.
```
maschad, line 373:
```suggestion
As of `libp2p@1.0.0` the keychain is no longer part of the default bundle so to replicate this behaviour you need to use the`Keychain` module to load a peer id from the datastore before starting your node.
```

doc/migrations/v0.46-v1.0.0.md
```diff
@@ -365,6 +366,52 @@ const node = await createLibp2p({
 })
 ```
 
+## Keeping the same PeerId after a restart
```
maschad, line 369:
```suggestion
### Keeping the same PeerId after a restart
```

I would suggest moving this section directly below the `###Keychain` section as a subsection of `##Misc`

maschad (changes requested):
Comments inline

doc/migrations/v0.46-v1.0.0.md
achingbrain: My opinion is not terribly strong but `Misc` is under `Extracted modules` so I think it might mess up the hierarchy a little.

The keychain module was extracted, but I feel this is important enough to have it's own section?

doc/migrations/v0.46-v1.0.0.md
maschad: I can appreciate the familiarity with the terminology but at the same time I think " node not running" implies at rest which is why I wanted to remove the redundancy. 

```suggestion
Given the libp2p `PeerId` is a public/private keypair, it is encrypted at rest by libp2p's `Keychain`
```

doc/migrations/v0.46-v1.0.0.md
maschad: If the keychain module is now an extracted module and this relates to the keychain doesn't that suggest it should be associated with the Keychain subsection `Misc` section?

achingbrain: It felt a bit buried down there.

PR #2338, opened 2023-12-30
feat: add tracked list to utils

achingbrain:
To allow tracking sizes of internal lists using libp2p metrics, add a tracked list type simmilar to tracked map.

## Change checklist

### DIFF
--- a/packages/utils/src/tracked-list.ts
+++ b/packages/utils/src/tracked-list.ts
@@ -1,10 +1,5 @@
 import type { Metrics } from '@libp2p/interface'
 
-export interface TrackedListInit {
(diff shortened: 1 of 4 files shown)


packages/utils/src/tracked-list.ts
```diff
@@ -0,0 +1,31 @@
+import type { Metrics } from '@libp2p/interface'
+
+export interface TrackedListInit {
```
wemeetagain, line 3: What is this interface used for?

achingbrain: Nothing, I've removed it

PR #2339, opened 2023-12-30
fix: replace p-queue with less restrictive queue

achingbrain:
Adds a `Queue` class to `@libp2p/utils` modelled on p-queue with a few key differences:

1. The queue is externally accessible so we can modify it before jobs run
2. It can be turned into an async generator
3. Jobs remain in the queue while they are executing for better introspection
4. It integrates with libp2p metrics, if desired

The dial queue has been replaced with this new queue class, this means we don't need to maintain a separate internal queue for pending dials since the dial queue itself is accessible.

## Change checklist


packages/utils/src/queue/index.ts
```diff
@@ -0,0 +1,377 @@
...
+  async onIdle (): Promise<void> {
+    // Instantly resolve if none pending and if nothing else is queued
+    if (this.pending === 0 && this.size === 0) {
+      return
+    }
+
+    await this.onEvent('idle')
+  }
+
+  private async onEvent (event: keyof QueueEvents<JobReturnType>, filter?: () => boolean): Promise<void> {
```
wemeetagain, line 276: Does this need to be abortable?
I guess not if clear/abort will allow all pending promises (from public methods, onEmpty, onSizeLessThan, onIdle, toGenerator) to resolve.

achingbrain: It's not abortable in p-queue, though it makes sense

achingbrain: They are abortable now

packages/utils/package.json
```diff
@@ -76,9 +76,13 @@
...
-    "./peer-job-queue": {
-      "types": "./dist/src/peer-job-queue.d.ts",
-      "import": "./dist/src/peer-job-queue.js"
+    "./queue": {
+      "types": "./dist/src/queue/index.d.ts",
+      "import": "./dist/src/queue/index.js"
+    },
+    "./peer-queue": {
+      "types": "./dist/src/peer-queue.d.ts",
+      "import": "./dist/src/peer-queue.js"
```
akim-bow, line 85: @achingbrain This PR introduces BREAKING CHANGE and actually broke many libp2p projects. Why not to bump major version if you break package API? Libp2p packages use caret ranges and automatically will try to update this package when installed.

PR #2340, opened 2023-12-30
fix: remove extra deps

achingbrain:
Remove extra/add missing deps shown by https://github.com/ipfs/aegir/pull/1426/

## Change checklist


packages/utils/package.json
```diff
@@ -112,28 +112,26 @@
   "dependencies": {
     "@chainsafe/is-ip": "^2.0.2",
     "@libp2p/interface": "^1.1.0",
-    "@libp2p/peer-collections": "^5.1.2",
```
maschad, line 115: Why was this dependency removed? It's used in the [peer-job-queue](https://github.com/libp2p/js-libp2p/blob/main/packages/utils/src/peer-job-queue.ts#L4)

maschad (changes requested):
Perhaps we should wait till https://github.com/ipfs/aegir/pull/1426 is merged and aegir is upgraded so that we can re-run the _fixed_ `dep-check` on CI before merging.

packages/utils/package.json
achingbrain: It was removed in #2339, this PR depended on that one, I thought I'd mentioned that in the description but I guess I missed it..

achingbrain:
Aegir is upgraded.

PR #2348, opened 2024-01-08
feat: add private key to libp2p components

wemeetagain:
## Description
- Add `privateKey` to `Components`
- Non-breaking version of https://github.com/libp2p/js-libp2p/pull/2303
- 
## Change checklist


packages/libp2p/src/libp2p.ts
```diff
@@ -391,6 +392,7 @@ export class Libp2pNode<T extends ServiceMap = Record<string, unknown>> extends
  */
 export async function createLibp2pNode <T extends ServiceMap = Record<string, unknown>> (options: Libp2pOptions<T> = {}): Promise<Libp2pNode<T>> {
   options.peerId ??= await createEd25519PeerId()
+  options.privateKey ??= await unmarshalPrivateKey(options.peerId.privateKey as Uint8Array)
```
achingbrain, line 395: I think we can skip the cast and make it a bit safer with:

```suggestion
  const peerId = options.peerId ??= await createEd25519PeerId()
  
  if (peerId.privateKey == null) {
    throw new CodeError('peer id was missing private key', 'ERR_MISSING_PRIVATE_KEY')
  }
  
  options.privateKey ??= await unmarshalPrivateKey(peerId.privateKey)
```

This will catch people doing something like this which won't be caught by TypeScript:

```js
  peerId: peerIdFromString('QmFoo')
})
```

achingbrain (changes requested):
Small suggestion but otherwise LGTM

achingbrain (approved):

achingbrain:
Just needs green CI before merge.

maschad (approved):

Issue #2086, opened 2023-09-27
feat: reduce scope of crypto module 


SgtPooki:
Discussed during https://pl-strflt.notion.site/js-libp2p-get-rid-of-node-forge-331d2e274b5e46e6be3f697dcf7b2c4d?pvs=4

create a new `/packages/crypto` folder where all the stuff that we don't use internally in libp2p 

* There are a lot of RSA keys on the network (e.g. peerId string "Qm...")
* We can't remove import/export of the keys yet (i.e. go from pkcs1 to jwk, or pkix to jwk)
* internal format is pwk, 
* external pkix for public keys and pkcs1 for private
* we should move some stuff out that we can, which reduces effort later

Refs:
 - https://github.com/libp2p/js-libp2p-crypto/issues/220

```[tasklist]
### Tasks
- [ ] remove the encryption stuff?
- [ ] replace node-forge with noble's pbkdf2 implementation (there is a PR open for this already, not added because we haven't got rid of node-forge yet)
```

PR #2355, opened 2024-01-11
fix!: remove node-forge dependency from @libp2p/crypto

achingbrain:
TLDR: the bundle size has been reduced by about 1/3rd

- parsing/creating PEM/pkix/pkcs1 files is now done by asn1.js
- Streaming AES-CTR ciphers are now in [@libp2p/aes-ctr](https://github.com/libp2p/js-libp2p-aes-ctr)
- RSA encryption/decryption and PEM import/export are now in [@libp2p/rsa](https://github.com/libp2p/js-libp2p-rsa)

## AES-CTR

WebCrypto [doesn't support streaming ciphers](https://github.com/w3c/webcrypto/issues/73).

We have a node-forge-backed shim that allows using streaming AES-CTR in browsers but we don't use it anywhere, so this has been split out into it's own module as `@libp2p/aes-ctr`.

## RSA encrypt/decrypt

This was added to `@libp2p/crypto` to [support webrtc-stardust](https://github.com/libp2p/js-libp2p-crypto/pull/125#issuecomment-465713400) but that effort didn't go anywhere and we don't use these methods anywhere else in the stack.

For reasons lost to the mists of time, we chose to use a [padding algorithm](https://github.com/libp2p/js-libp2p-crypto/blob/3d0fd234deb73984ddf0f7c9959bbca92194926a/src/keys/rsa.ts#L59) that WebCrypto doesn't support so node-forge (or some other userland implemenation) will always be necessary in browsers, so these ops have been pulled out into `@libp2p/rsa` which people can use if they need it.

This is now done by manipulating the asn1 structures directly.

## PEM/pkix/pkcs1

The previous PEM import/export is also ported to `@libp2p/crypto-rsa` because it seems to handle more weird edge cases introduced by OpenSSL.

These could be handled in `@libp2p/crypto` eventually but for now it at least supports round-tripping it's own PEM files.

Fixes https://github.com/libp2p/js-libp2p/issues/2086

BREAKING CHANGE: Legacy RSA operations are now in @libp2p/rsa, streaming AES-CTR ciphers are in @libp2p/aes-ctr

## Change checklist


wemeetagain (approved):
lgtm

PR #2356, opened 2024-01-11
fix: replace rate-limiter

achingbrain:
[rate-limiter-flexible](https://npmjs.com/package/rate-limiter-flexible) is a CJS module with a single export of all it's various implementations.

This defeats tree shaking resulting in the addition of 42KB to the bundle size.

https://github.com/animir/node-rate-limiter-flexible/issues/249

This PR brings the source & tests for the in-memory rate limiter into `@libp2p/utils` which reduces the bundle size increase to a few KBs.

## Change checklist


wemeetagain (approved):
LGTM

maschad (approved):

PR #2363, opened 2024-01-16
docs: updated README to include Open Maintainer call schedule

maschad:
## Title

docs: updated README to include discussions info

## Description

This includes Open Maintainer call info + filing bugs & security reports.

## Change checklist


wemeetagain (changes requested):
As its currently formatted it feels awkward to me.

I think these should be moved to a specific section, perhaps contributing.
Or else, if we want some sort of 'quickstart' section, then there should be more formatting to support that.

maschad:
Fair, it turns out the luma already contained the Open Maintainer's call link, hopefully that's not too buried for others to find 🤔

wemeetagain (approved):

PR #2377, opened 2024-01-25
feat: add tls connection encrypter

achingbrain:
Encrypts connections using TLS 1.3.

Spec: https://github.com/libp2p/specs/blob/master/tls/tls.md

This connection encrypter primarily targets TCP connections on Node.js/Electron/React Native and is notable for having fewer round trips during connection establishment than Noise.

The Node.js TLS module has a fair few more users than Noise so it will be interesting to see how streaming performance compares between the two.

## Change checklist

## TODO

- [x] Compliance tests passing
- [x] Test fixtures as per the spec
- [x] Add interop tests - https://github.com/libp2p/interop/pull/145
- [x] Integrate with streaming multi-dim interop test


achingbrain:
One final compliance test is not passing - the server side of the connection doesn't notice the client closing the connection so it remains open and the test hangs until it times out.

achingbrain:
Streaming performance is comparable to noise, connection establishment is slower, which is unexpected: https://observablehq.com/@libp2p-workspace/performance-dashboard?branch=83b4e5da914c6dbcc8a04d23a8dca4df3419f33e

PR #2402, opened 2024-02-09
docs: add Peerbit logo

marcus-pousette:
## Title

Add Peerbit logo to "Used by" section

## Description
Add Peerbit logo to "Used by" section

## Notes & open questions
From suggestion by @achingbrain. 

When adding the logo I realized that there are two READMEs that are very similiar and both have this "Used by" section, so I added it to both. But if it make sense to only add to one of them, please let me know. 

Also spent some time trying to align the Lodestar logo, but it Github will not allow me to really.

## Change checklist


achingbrain (approved):

achingbrain:
Thanks!

achingbrain:
Agree on the Lodestar logo, it'd be nicer if it was formatted to be the same size as the others.  What do you think @wemeetagain?

Issue #2371, opened 2024-01-24
@libp2p/webtransport Correct multiaddr format


sachaw:
Hi, I have just started a project that requires the use of @libp2p/webtransport
However none of the documented multiaddr formats are working as a listen address.

I have seen the following formats:
/ip4/0.0.0.0/udp/0/quic/webtransport
/ip4/0.0.0.0/udp/0/quic-v1/webtransport

achingbrain:
The WebTransport transport is dial-only at the moment.  When QUIC lands in node.js we'll be able to add it but until then listening is not supported.

I did have a [PR](https://github.com/libp2p/js-libp2p-webtransport/pull/31) that added listener support via [@fails-components/webtransport](https://www.npmjs.com/package/@fails-components/webtransport) but it was never really production-grade.

Download:
@achingbrain I made a PR that adds a note to the Webtransport docs about them supporting dial only. I am currently in the process of learning libp2p and I find that although there is a lot of documentation, it often misses critical details such as these or is hard to find. So I am trying to contribute by making small PRs with small updates to the docs as I go. Hope you find it helpful.

achingbrain:
@Download thanks, yes - it's very helpful!

achingbrain:
Fixed by #2411

PR #2408, opened 2024-02-11
docs: add note about webtransport only supporting dial

Download:
## Title
docs: add note about webtransport only supporting dial

## Description
This adds a note at the top of the WebTransport `README.md` to remind people of the fact that currently, the WebTransport implementation in js-libp2p only supports dialing out to other nodes and does not actually support listening for incoming dials because js-libp2p depends on Node JS and QUIC support has not landed in Node JS yet.

Related https://github.com/libp2p/js-libp2p/issues/2371

## Notes & open questions
-

## Change checklist

### DIFF
--- a/packages/transport-webtransport/README.md
+++ b/packages/transport-webtransport/README.md
@@ -9,6 +9,18 @@
 
 A [libp2p transport](https://docs.libp2p.io/concepts/transports/overview/) based on [WebTransport](https://www.w3.org/TR/webtransport/).
 
+>
+> ⚠️ **Note**
+>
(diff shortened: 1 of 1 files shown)


achingbrain (changes requested):
Thanks for opening this.

This section of the readme is generated from the `@packageDocumentation` tag in [src/index.ts](https://github.com/libp2p/js-libp2p/blob/main/packages/transport-webtransport/src/index.ts#L1-L22) so the place to make this change is there, otherwise this'll get removed next time the docs are generated.

You can check it formats in the readme correctly by running `npm run docs` from the root of the repo after making your changes.

PR #2446, opened 2024-03-25
fix: add name property to exported interface errors

2color:
## Description

As discussed in https://github.com/ipfs/helia-verified-fetch/issues/25#issuecomment-2018003309, this PR adds the name property to the AbortError type. 

This has a couple of benefits:
- Sticks the common convention for Error types (all built in Error types have the name property set to the class name)
- Allows checking the error type at runtime in a TS codebase without importing the type

## Notes & open questions

- Should we also set the name property on UnexpectedPeerError, InvalidCryptoExchangeError and InvalidCryptoTransmissionError?

## Change checklist

### DIFF
--- a/packages/interface/src/errors.ts
+++ b/packages/interface/src/errors.ts
@@ -9,6 +9,7 @@ export class AbortError extends Error {
 
   constructor (message: string = 'The operation was aborted') {
     super(message)
+    this.name = 'AbortError'
     this.code = AbortError.code
     this.type = AbortError.type
   }
--- a/packages/interface/src/errors.ts
+++ b/packages/interface/src/errors.ts
@@ -55,6 +55,7 @@ export class UnexpectedPeerError extends Error {
 
   constructor (message = 'Unexpected Peer') {
     super(message)
+    this.name = 'UnexpectedPeerError'
     this.code = UnexpectedPeerError.code
   }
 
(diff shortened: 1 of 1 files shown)


SgtPooki:
https://github.com/nodejs/node/issues/40692#issuecomment-956658594 has some relevant discussions about custom AbortError's too

SgtPooki (approved):

achingbrain:
> Should we also set the name property on UnexpectedPeerError, InvalidCryptoExchangeError and InvalidCryptoTransmissionError?

Sorry, missed this - yes, I think we should.

2color:
> > Should we also set the name property on UnexpectedPeerError, InvalidCryptoExchangeError and InvalidCryptoTransmissionError?
> 
> 
> 
> Sorry, missed this - yes, I think we should.

Ok I'll update the pr

achingbrain:
Just noticed that [DOMException](https://developer.mozilla.org/en-US/docs/Web/API/DOMException) officially doesn't have a `.stack` property, and technically neither does [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error), though every JS runtime adds it - there's a [tc39 proposal to add it](https://github.com/tc39/proposal-error-stacks) that's been around for, oh, 8 years 🤦

Issue #2452, opened 2024-03-27
libp2p.components not included in typescript types.


MichaelJCole:
- **Version**:
    "libp2p": "^1.2.3",

- **Platform**:
Linux

- **Subsystem**:

#### Severity:
  High - The main functionality of the application does not work, API breakage, repo format breakage, etc.

#### Description:

Separating types into an interface creates *a lot* of extra work.  

The point of typescript is to have the code, types, and documentation in the same place.  The types are the documentation.

So the [interface](https://github.com/libp2p/js-libp2p/blob/main/packages/interface/src/index.ts#L330) doesn't include components, but the implementation makes [components public](https://github.com/libp2p/js-libp2p/blob/main/packages/libp2p/src/libp2p.ts#L39).

It's very difficult to contribute to this project because the code is spread all over the place.

It's not clear to me how to fix it in the interface package.

`Components` type can be exported from libp2p allowing for workarounds in the apps that use libp2p.

#### Steps to reproduce the error:
Try to get the results of what AutoNAT is doing.

MichaelJCole:
To access the address manager, here is a workaround: 

```
import { DefaultAddressManager } from 'libp2p/dist/src/address-manager'

// @ts-ignore broken libp2p types
const am = libp2p.components.addressManager as DefaultAddressManager
```

achingbrain:
This is by design to simplify the public API and allow internal refactors without breaking that public API.

Typically internal components are only accessed by services, which are the way to extend the node with your own custom functionality which can be fully typed as expected.

PR #2453, opened 2024-03-27
fix(libp2p): Export Components type for use in apps that use libp2p

MichaelJCole:
## Title
fix(libp2p): Export Components type for use in apps that use libp2p

## Description
See https://github.com/libp2p/js-libp2p/issues/2452

## Notes & open questions

See https://github.com/libp2p/js-libp2p/issues/2452

## Change checklist

### DIFF
--- a/packages/libp2p/src/index.ts
+++ b/packages/libp2p/src/index.ts
@@ -16,13 +16,14 @@
 
 import { createLibp2pNode } from './libp2p.js'
 import type { AddressManagerInit } from './address-manager/index.js'
-import type { Components } from './components.js'
 import type { ConnectionManagerInit } from './connection-manager/index.js'
 import type { TransportManagerInit } from './transport-manager.js'
 import type { Libp2p, ServiceMap, RecursivePartial, ComponentLogger, NodeInfo, ConnectionProtector, ConnectionEncrypter, ConnectionGater, ContentRouting, Metrics, PeerDiscovery, PeerId, PeerRouting, StreamMuxerFactory, Transport, PrivateKey } from '@libp2p/interface'
 import type { PersistentPeerStoreInit } from '@libp2p/peer-store'
 import type { DNS } from '@multiformats/dns'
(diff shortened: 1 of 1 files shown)


achingbrain:
Thanks for opening this.

I'm hesitant to export this interface because it encourages the sort of code where everything depends on everything else.  This is why the `components` field is not part of the [libp2p interface](https://github.com/libp2p/js-libp2p/blob/main/packages/interface/src/index.ts#L330) and is only used internally within the `libp2p` module.

Could you speak a bit more to your use-case please?

MichaelJCole:
Hi, I wrote a bit about this in https://github.com/libp2p/js-libp2p/pull/2449#issuecomment-2023373834.

I wrote a bit about the use case in https://github.com/libp2p/js-libp2p/issues/2454

More or less, my use case is an electron app with a configuration similar to Vuze, utorrent, and other peer-to-peer desktop applications.  Those applications set expectations for how configurable they should be.  Also, I need to help users around NATs and firewalls and etc.  Having access to the addressManager is part of it.  

I mentioned in the PR, how Typescript only enforces these limitations at compile time.  And they can be trivially overridden:

```
    // @ts-ignore missing components in libp2p types
    const am = libp2p.components.addressManager as DefaultAddressManager
```

So not including these types is really just making it harder to use.  Javascript runtimes don't enforce private members unless you're compiling to ES2021+ and use # at the start of the names.  The _name convention is well understood.  IMHO, separating the interfaces from classes and putting them into their own package really just makes things harder for everyone.

achingbrain:
I'm closing this issue because I think the underlying problem you mention in #2454 is that your observed addresses are not being confirmed by peers.

Exposing libp2p internals like this will not solve your problem.

If you really, really want to do this, you can `@ts-ignore` the types, or create a custom service that exposes the methods you want to call, but it seems that your particular network setup is giving AutoNAT problems, or your node hasn't been running for long enough to receive enough confirmations from a diverse enough set of peers, or perhaps both.

PR #2457, opened 2024-03-28
fix: reduce TLS cert validity time

achingbrain:
`@peculiar/x509` cannot represent dates as far in advance as go-libp2p so the certs are rejected.

Reduce the maximum validity of certificates to keep them under 2050 - this can be re-evaluated after https://github.com/PeculiarVentures/x509/issues/73 is fixed.

## Change checklist


SgtPooki (commented):
sgtm

PR #2459, opened 2024-03-31
fix: tls serial number causes illegal padding error

achingbrain:
This is a hack to work around https://github.com/PeculiarVentures/x509/issues/74 until it is addressed upstream.

It seems serial numbers starting with `80` cause `@peculiar/x509` to generate invalid certifiates that Node's `TLSSocket` then fails to parse, throwing an `ERR_OSSL_ASN1_ILLEGAL_PADDING` error, so the hack is to generate serial numbers until we get one that doesn't start with `80`.

This can be reverted when the upstream issue is fixed.

## Change checklist


achingbrain:
I have been running connection tests with this patch against go-libp2p for 24 hours without a single error.

maschad (approved):

PR #2471, opened 2024-04-07
fix: decrease default dial timeout

achingbrain:
30s is far too long for a default dial timeout and can cause dialing unresponsive peers with multiple valid addresses to block the dial queue. Decrease to 2s.

## Change checklist

### DIFF
--- a/packages/libp2p/src/connection-manager/constants.defaults.ts
+++ b/packages/libp2p/src/connection-manager/constants.defaults.ts
@@ -1,12 +1,12 @@
 /**
  * @see https://libp2p.github.io/js-libp2p/interfaces/index._internal_.ConnectionManagerConfig.html#dialTimeout
  */
-export const DIAL_TIMEOUT = 30e3
+export const DIAL_TIMEOUT = 2e3
 
 /**
  * @see https://libp2p.github.io/js-libp2p/interfaces/index._internal_.ConnectionManagerConfig.html#inboundUpgradeTimeout
  */
-export const INBOUND_UPGRADE_TIMEOUT = 30e3
+export const INBOUND_UPGRADE_TIMEOUT = 1e3
 
 /**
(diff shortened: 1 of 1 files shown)


packages/libp2p/src/connection-manager/constants.defaults.ts
```diff
@@ -1,12 +1,12 @@
 /**
  * @see https://libp2p.github.io/js-libp2p/interfaces/index._internal_.ConnectionManagerConfig.html#dialTimeout
  */
-export const DIAL_TIMEOUT = 30e3
+export const DIAL_TIMEOUT = 2e3
```
achingbrain, line 4:
```suggestion
export const DIAL_TIMEOUT = 5e3
```

packages/libp2p/src/connection-manager/constants.defaults.ts
```diff
@@ -1,12 +1,12 @@
...
 
 /**
  * @see https://libp2p.github.io/js-libp2p/interfaces/index._internal_.ConnectionManagerConfig.html#inboundUpgradeTimeout
  */
-export const INBOUND_UPGRADE_TIMEOUT = 30e3
+export const INBOUND_UPGRADE_TIMEOUT = 1e3
```
achingbrain, line 9:
```suggestion
export const INBOUND_UPGRADE_TIMEOUT = 2e3
```

packages/libp2p/src/connection-manager/constants.defaults.ts
```diff
@@ -1,12 +1,12 @@
 /**
  * @see https://libp2p.github.io/js-libp2p/interfaces/index._internal_.ConnectionManagerConfig.html#dialTimeout
  */
-export const DIAL_TIMEOUT = 30e3
+export const DIAL_TIMEOUT = 5e3
```
SgtPooki, line 4: FYI: you mentioned in PR that you were decreasing the timeout to 2s, but this is 5s

PR #2478, opened 2024-04-11
fix: close MultiaddrConnection once

### LINKED ISSUE #2477: Memory leak issue
- **Version**:
v1.1.1

- **Platform**:
`Linux 5.15.0-79-generic #86-Ubuntu SMP Mon Jul 10 16:07:21 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`

- **Subsystem**:

#### Severity:
Medium

#### Description:
Refer to the profile taken by a lodestar beacon-node [here](https://github.com/ChainSafe/lodestar/pull/6556#issuecomment-2043250725)

There are > 15k of objects retained belonging to these classes:
- Socket
- ConnectionImpl
- Timeout
- XXHandShake
- MPlexStreamMuxer

Run a lodestar mainnet node and leave it for a long time, then monitor heap size of network thread

tuyennhv:
## Title
Close MultiaddrConnection once

## Description

- It's designed to close MultiaddrConnection once, however in case there are some writable data remaining, socket is not destroyed yet https://github.com/libp2p/js-libp2p/blob/c5003d40207bc3be15645b5af93f8dec1fd5d55b/packages/transport-tcp/src/socket-to-conn.ts#L177 and when there is a called to `close()` again it'll run the function again
- Instead use a variable to control it like other places

## Notes & open questions
this could fix #2477 see https://github.com/libp2p/js-libp2p/issues/2477#issuecomment-2049121628

## Change checklist

- [x] No need documentation as the change is straightforward and small, the same pattern is applied in a lot of places
- [x] I thought about unit tests however not sure how to do it. This change is quite straightforward and scanned through the current tests through

### DIFF
--- a/packages/transport-tcp/src/socket-to-conn.ts
+++ b/packages/transport-tcp/src/socket-to-conn.ts
@@ -199,7 +199,10 @@ export const toMultiaddrConnection = (socket: Socket, options: ToConnectionOptio
     abort: (err: Error) => {
       log('%s socket abort due to error', lOptsStr, err)
 
-      socket.destroy(err)
+      // the abortSignalListener may already destroyed the socket with an error
+      if (!socket.destroyed) {
+        socket.destroy(err)
+      }
 
       if (maConn.timeline.close == null) {
         maConn.timeline.close = Date.now()
--- a/packages/transport-tcp/src/socket-to-conn.ts
+++ b/packages/transport-tcp/src/socket-to-conn.ts
@@ -23,6 +23,7 @@ interface ToConnectionOptions {
  * https://github.com/libp2p/interface-transport#multiaddrconnection
  */
 export const toMultiaddrConnection = (socket: Socket, options: ToConnectionOptions): MultiaddrConnection => {
+  let status: 'open' | 'closing' | 'closed' = 'open'
   const log = options.logger.forComponent('libp2p:tcp:socket')
   const metrics = options.metrics
   const metricPrefix = options.metricPrefix ?? ''
@@ -126,10 +127,11 @@ export const toMultiaddrConnection = (socket: Socket, options: ToConnectionOptio
     timeline: { open: Date.now() },
 
     async close (options: AbortOptions = {}) {
-      if (socket.destroyed) {
-        log('%s socket was already destroyed when trying to close', lOptsStr)
+      if (status === 'closed' || status === 'closing' || socket.destroyed) {
+        log('%s socket was already destroyed or closed or closing', lOptsStr)
         return
       }
+      status = 'closing'
 
       if (options.signal == null) {
         const signal = AbortSignal.timeout(closeTimeout)
@@ -153,6 +155,7 @@ export const toMultiaddrConnection = (socket: Socket, options: ToConnectionOptio
             // socket completely closed
             log('%s socket closed', lOptsStr)
 
+            status = 'closed'
             resolve()
           })
           socket.once('error', (err: Error) => {
@@ -163,6 +166,7 @@ export const toMultiaddrConnection = (socket: Socket, options: ToConnectionOptio
               maConn.timeline.close = Date.now()
             }
 
+            status = 'closed'
             reject(err)
           })
 
--- a/packages/transport-tcp/src/socket-to-conn.ts
+++ b/packages/transport-tcp/src/socket-to-conn.ts
@@ -128,7 +128,7 @@ export const toMultiaddrConnection = (socket: Socket, options: ToConnectionOptio
 
     async close (options: AbortOptions = {}) {
       if (status === 'closed' || status === 'closing' || socket.destroyed) {
-        log('%s socket was already destroyed or closed or closing', lOptsStr)
+        log('The %s socket is either closed, closing, or already destroyed', lOptsStr)
         return
       }
       status = 'closing'
--- a/packages/transport-tcp/test/socket-to-conn.spec.ts
+++ b/packages/transport-tcp/test/socket-to-conn.spec.ts
@@ -292,7 +292,7 @@ describe('socket-to-conn', () => {
     ({ server, clientSocket, serverSocket } = await setup())
     // proxyServerSocket.writableLength returns 100 which cause socket cannot be destroyed immediately
     const proxyServerSocket = new Proxy(serverSocket, {
-      get(target, prop, receiver) {
+      get (target, prop, receiver) {
         if (prop === 'writableLength') {
           return 100
         }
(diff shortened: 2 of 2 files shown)


SgtPooki (approved):
Lgtm

packages/transport-tcp/src/socket-to-conn.ts
```diff
@@ -126,10 +127,11 @@ export const toMultiaddrConnection = (socket: Socket, options: ToConnectionOptio
     timeline: { open: Date.now() },
 
     async close (options: AbortOptions = {}) {
-      if (socket.destroyed) {
-        log('%s socket was already destroyed when trying to close', lOptsStr)
+      if (status === 'closed' || status === 'closing' || socket.destroyed) {
+        log('%s socket was already destroyed or closed or closing', lOptsStr)
```
maschad, line 131:
```suggestion
	log('The %s socket is either closed, closing, or already destroyed', lOptsStr)
```

packages/transport-tcp/test/socket-to-conn.spec.ts
```diff
@@ -287,6 +289,52 @@ describe('socket-to-conn', () => {
     expect(serverSocket.destroyed).to.be.true()
   })
 
+  it('should not close MultiaddrConnection twice', async () => {
+    const loggerStub = Sinon.stub()
+    const logger: ComponentLogger = {
+      forComponent: () => loggerStub as unknown as Logger
+    };
```
achingbrain, line 296: Will use default logger, no need for stubbing:

```suggestion
```

packages/transport-tcp/test/socket-to-conn.spec.ts
```diff
@@ -287,6 +289,52 @@ describe('socket-to-conn', () => {
...
+    ({ server, clientSocket, serverSocket } = await setup())
+
+    // promise that is resolved when our outgoing socket is closed
+    const serverClosed = defer<boolean>()
+
+    const inboundMaConn = toMultiaddrConnection(serverSocket, {
+      socketInactivityTimeout: 100,
+      socketCloseTimeout: 10,
+      logger
```
achingbrain, line 305: Use default logger:

```suggestion
      logger: defaultLogger()
```

packages/transport-tcp/test/socket-to-conn.spec.ts
```diff
@@ -287,6 +289,52 @@ describe('socket-to-conn', () => {
...
   })
 
+  it('should not close MultiaddrConnection twice', async () => {
+    const loggerStub = Sinon.stub()
+    const logger: ComponentLogger = {
+      forComponent: () => loggerStub as unknown as Logger
+    };
+    ({ server, clientSocket, serverSocket } = await setup())
+
+    // promise that is resolved when our outgoing socket is closed
```
achingbrain, line 299: Add a spy on `.destroy()` so we can assert that the socket was only closed once:

```suggestion

    // spy on `.destroy()` invocations
    const serverSocketDestroySpy = Sinon.spy(serverSocket, 'destroy')

    // promise that is resolved when our outgoing socket is closed
```

packages/transport-tcp/test/socket-to-conn.spec.ts
```diff
@@ -287,6 +289,52 @@ describe('socket-to-conn', () => {
...
+    expect(loggerStub.calledWithMatch('socket is either closed, closing, or already destroyed')).to.be.true()
+
+    // server socket was closed for reading and writing
+    await expect(serverClosed.promise).to.eventually.be.true()
+
+    // the connection closing was recorded
+    expect(inboundMaConn.timeline.close).to.be.a('number')
+
+    // server socket is destroyed
+    expect(serverSocket.destroyed).to.be.true()
```
achingbrain, line 335: Assert that we only close the socket once:

```suggestion
    expect(serverSocket.destroyed).to.be.true()

    // the server socket was only closed once
    expect(serverSocketDestroySpy.callCount).to.equal(1)
```

packages/transport-tcp/test/socket-to-conn.spec.ts
```diff
@@ -287,6 +289,52 @@ describe('socket-to-conn', () => {
...
+    serverSocket.write('goodbye')
+
+    // the 2nd call should return immediately
+    await Promise.all([
+      inboundMaConn.close(),
+      inboundMaConn.close()
+    ])
+
+    expect(loggerStub.calledWithMatch('socket is either closed, closing, or already destroyed')).to.be.true()
+
```
achingbrain, line 327: Please don't assert on logging - just because this assertion passes it doesn't mean that the functionality works as intended.  It also introduces fragility as someone could "improve" the logging message later which would break this test.

```suggestion
```

packages/transport-tcp/test/socket-to-conn.spec.ts
```diff
@@ -1,8 +1,10 @@
 import { createServer, Socket, type Server, type ServerOpts, type SocketConstructorOpts } from 'net'
 import os from 'os'
 import { defaultLogger } from '@libp2p/logger'
+import { type ComponentLogger, type Logger } from '@libp2p/logger'
```
achingbrain, line 4: Will use default logger, no need for import:

```suggestion
```

achingbrain (changes requested):
I've suggested some improvements to the test so that it actually asserts that the socket is not closed twice.

The problem is, it still passes even if I remove the edits to `socket-to-conn.ts` so I don't think it exercises the bug that this PR is supposed to fix, though it's possible I've missed the point.

#2478 implies maconns are not being closed, rather than being closed multiple times so I guess the question is; under what circumstances are sockets not being closed?

Could you please improve the test so that it shows the bug being addressed?

tuyennhv:
@achingbrain thanks for the suggestions, now I improved the test and it failed on `main`. I did a small modification on the `maConn.abort()` function not to destroy the socket if it's already destroyed to make sure the `socket.destroy()` is called once

this is all about `maConn.close()` being called multiple times and we want it to close once, this happens when `socket.writableLength > 0` which is reflected in the test

achingbrain (approved):
Thanks for updating the test.  I wonder if we could make a further improvement as a follow up.

If two call site invoke `.close`, they may perform clean up operations after the promise from that operation resolves.  If it's the second invocation, as implemented the promise will resolve immediately even though the first operation may not have completed - so the socket is still open.

This will likely lead to subtle bugs that are hard to diagnose.

Instead perhaps there should be an internal deferred promise that can be returned to the second call site, and resolved by the invocation from the first.  It would be rejected by the call to `.abort` and the socket closed immediately if not already destroyed.

This kind of behaviour would benefit all transports so it could be pushed up in to a superclass similar to `AbstractStream` and tests added to the interface suite.

PR #2491, opened 2024-04-17
fix: upgrader error msg to show current/max

SgtPooki:
## Title

fix: upgrader error msg shows current/max

## Description

Provides more information in the error message so users can better understand the error state.

## Notes & open questions

> CodeError: Too many outbound protocol streams for protocol "/ipfs/id/push/1.0.0" - limit 1

This error message is not super helpful for debugging. What's the overage amount?

Also, the check is checking `>=`, and if your limit is 1, that will prevent ANY outbound protocol streams. Is this intended? should we be comparing with only `>` ?

## Change checklist

### DIFF
--- a/packages/libp2p/src/upgrader.ts
+++ b/packages/libp2p/src/upgrader.ts
@@ -471,7 +471,7 @@ export class DefaultUpgrader implements Upgrader {
           const streamCount = countStreams(protocol, 'outbound', connection)
 
           if (streamCount >= outgoingLimit) {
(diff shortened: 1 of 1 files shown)


SgtPooki (commented):
self review

maschad (approved):

achingbrain (approved):

achingbrain:
> Also, the check is checking >=, and if your limit is 1, that will prevent ANY outbound protocol streams. Is this intended? should we be comparing with only > ?

No, because the stream has been opened already so at this point we are above the limit.

That is, when we carry out multistream select, and we support multiple stream protocols, we don't know ahead of time which one will be selected.  We have to let the stream open, negotiate a protocol, then carry out the check.  We could compare with `===` instead, but the `>=` is slightly safer.

PR #2496, opened 2024-04-23
feat: extensible peer ids

achingbrain:
Sometimes a peer will have a non-cryptographic identifier such as a URL.

Update the PeerId type to be an interface so we can use arbitrary data to identify other peers, not just cryptographic keys.

## Change checklist

### DIFF
--- a/packages/crypto/src/keys/index.ts
+++ b/packages/crypto/src/keys/index.ts
@@ -76,7 +76,7 @@ export async function generateKeyPairFromSeed <T extends KeyTypes> (type: T, see
 /**
  * Converts a protobuf serialized public key into its representative object
  */
-export function unmarshalPublicKey (buf: Uint8Array): PublicKey<KeyTypes> {
+export function unmarshalPublicKey <T extends KeyTypes>(buf: Uint8Array): PublicKey<T> {
   const decoded = keysPBM.PublicKey.decode(buf)
   const data = decoded.Data ?? new Uint8Array()
 
@@ -104,7 +104,7 @@ export function marshalPublicKey (key: { bytes: Uint8Array }, type?: string): Ui
 /**
  * Converts a protobuf serialized private key into its representative object
  */
-export async function unmarshalPrivateKey (buf: Uint8Array): Promise<PrivateKey<KeyTypes>> {
+export async function unmarshalPrivateKey <T extends KeyTypes>(buf: Uint8Array): Promise<PrivateKey<T>> {
   const decoded = keysPBM.PrivateKey.decode(buf)
   const data = decoded.Data ?? new Uint8Array()
 
@@ -134,7 +134,7 @@ export function marshalPrivateKey (key: { bytes: Uint8Array }, type?: string): U
  *
  * Supported formats are 'pem' (RSA only) and 'libp2p-key'.
  */
-export async function importKey (encryptedKey: string, password: string): Promise<PrivateKey<KeyTypes>> {
+export async function importKey <T extends KeyTypes> (encryptedKey: string, password: string): Promise<PrivateKey<T>> {
   try {
     const key = await importer(encryptedKey, password)
     return await unmarshalPrivateKey(key)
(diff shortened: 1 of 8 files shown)


packages/crypto/src/keys/index.ts
```diff
@@ -79,7 +76,7 @@ export async function generateKeyPairFromSeed (type: KeyTypes, seed: Uint8Array,
 /**
  * Converts a protobuf serialized public key into its representative object
  */
-export function unmarshalPublicKey (buf: Uint8Array): PublicKey {
+export function unmarshalPublicKey (buf: Uint8Array): PublicKey<KeyTypes> {
```
SgtPooki, line 79: Should we?

```suggestion
export function unmarshalPublicKey <T extends KeyTypes>(buf: Uint8Array): PublicKey<T> {
```

packages/crypto/src/keys/index.ts
```diff
@@ -107,7 +104,7 @@ export function marshalPublicKey (key: { bytes: Uint8Array }, type?: string): Ui
 /**
  * Converts a protobuf serialized private key into its representative object
  */
-export async function unmarshalPrivateKey (buf: Uint8Array): Promise<PrivateKey> {
+export async function unmarshalPrivateKey (buf: Uint8Array): Promise<PrivateKey<KeyTypes>> {
```
SgtPooki, line 107:
```suggestion
export async function unmarshalPrivateKey <T extends KeyTypes>(buf: Uint8Array): Promise<PrivateKey<T>> {
```

packages/crypto/src/keys/index.ts
```diff
@@ -137,7 +134,7 @@ export function marshalPrivateKey (key: { bytes: Uint8Array }, type?: string): U
  *
  * Supported formats are 'pem' (RSA only) and 'libp2p-key'.
  */
-export async function importKey (encryptedKey: string, password: string): Promise<PrivateKey> {
+export async function importKey (encryptedKey: string, password: string): Promise<PrivateKey<KeyTypes>> {
```
SgtPooki, line 137:
```suggestion
export async function importKey <T extends KeyTypes>(encryptedKey: string, password: string): Promise<PrivateKey<T>> {
```

packages/crypto/src/keys/index.ts
achingbrain: Maybe, but I think we might want to constrain it to the cryptographic types?

maschad (approved):

packages/interface/src/peer-id/index.ts
```diff
@@ -1,36 +1,34 @@
 import type { CID } from 'multiformats/cid'
 import type { MultihashDigest } from 'multiformats/hashes/interface'
 
-export type PeerIdType = 'RSA' | 'Ed25519' | 'secp256k1'
+export type PeerIdType = 'RSA' | 'Ed25519' | 'secp256k1' | string
```
SgtPooki, line 4:
```suggestion
import { KeyType } from '../keys/index.js'

export type PeerIdType = KeyType | string
```

SgtPooki (approved):
two minor issues in comments. otherwise looks pretty straightforward.

packages/crypto/src/keys/index.ts
```diff
@@ -55,11 +55,8 @@ function typeToKey (type: string): typeof RSA | typeof Ed25519 | typeof Secp256k
 
 /**
  * Generates a keypair of the given type and bitsize
- *
- * @param type
- * @param bits -  Minimum of 1024
  */
-export async function generateKeyPair (type: KeyTypes, bits?: number): Promise<PrivateKey> {
+export async function generateKeyPair <T extends KeyTypes> (type: T, bits?: number): Promise<PrivateKey<T>> {
```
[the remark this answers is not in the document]
achingbrain: There's 100% type overlap so we can do, yes.

packages/interface/src/peer-id/index.ts
achingbrain: Done!

PR #2499, opened 2024-04-23
fix: return closest peers from FIND_NODE

### LINKED ISSUE #2450: kad: FIND_NODE(self) should return multiple peers
- **Version**:
1.0.0

- **Platform**:
all

- **Subsystem**:
kad-dht

#### Severity:
Low

#### Description:
When receiving a kademlia `FIND_NODE` request for its own peer id, a js-libp2p node will respond with its own peer record only. According to libp2p kademlia [spec](https://github.com/libp2p/specs/blob/master/kad-dht/README.md), it should reply with the `k` closest nodes.

> The libp2p Kademlia DHT offers the following types of operations:
> * Peer routing
>     * Finding the closest nodes to a given key via FIND_NODE.

> 3. Upon a response:
>     1. If successful the response will contain the k closest nodes the peer knows to the key Key. Add them to the candidate list Pn, except for those that have already been queried.

Depending on https://github.com/libp2p/specs/pull/609, nodes shouldn't even include their own peer records in the response (already known to the requester).

achingbrain:
The `FIND_NODE` DHT operation should return the closest peers the node knows to the value.

It does not need to return itself in the list because the calling peer already knows about it.

Fixes #2450

## Change checklist

### DIFF
--- a/packages/kad-dht/test/kad-dht.spec.ts
+++ b/packages/kad-dht/test/kad-dht.spec.ts
@@ -799,6 +799,37 @@ describe('KadDHT', () => {
 
       expect(res).to.not.be.empty()
     })
+
+    it('should not include itself in getClosestPeers PEER_RESPONSE', async function () {
+      this.timeout(240 * 1000)
+
+      const nDHTs = 30
+      const dhts = await Promise.all(
+        new Array(nDHTs).fill(0).map(async () => tdht.spawn())
+      )
+
+      const connected: Array<Promise<void>> = []
+
+      for (let i = 0; i < dhts.length - 1; i++) {
+        connected.push(tdht.connect(dhts[i], dhts[(i + 1) % dhts.length]))
+      }
+
+      await Promise.all(connected)
+
+      const res = await all(dhts[1].getClosestPeers(dhts[2].components.peerId.toBytes()))
+      expect(res).to.not.be.empty()
+
+      // no peer should include itself in the response, only other peers that it
+      // knows who are closer
+      for (const event of res) {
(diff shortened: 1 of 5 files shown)


guillaumemichel:
Looks good, thanks @achingbrain !

packages/kad-dht/test/rpc/handlers/find-node.spec.ts
```diff
@@ -44,21 +38,33 @@ describe('rpc - handlers - FindNode', () => {
     })
   })
 
-  it('returns self, if asked for self', async () => {
+  it('returns nodes close to self but not self, if asked for self', async () => {
```
maschad, line 41: This is a bit difficult to read
```suggestion
  it('returns nodes close to self but excludes self, if asked for self', async () => {
```

packages/kad-dht/test/kad-dht.spec.ts
```diff
@@ -799,6 +799,37 @@ describe('KadDHT', () => {
 
       expect(res).to.not.be.empty()
     })
+
+    it('should not include itself in getClosestPeers PEER_RESPONSE', async function () {
+      this.timeout(240 * 1000)
```
maschad, line 804: Just curious, why are we using a 240 second timeout?

maschad (approved):
Looks good, left some comments in line

packages/kad-dht/test/kad-dht.spec.ts
achingbrain: No specific reason, it was copied from the preceding test.

PR #2507, opened 2024-04-26
feat: add bloom filter

achingbrain:
Adds a bloom filter implementation to `@libp2p/utils` for use in libp2p components.

## Change checklist


packages/utils/src/bloom-filter.ts
```diff
@@ -0,0 +1,141 @@
+// ported from xxbloom - https://github.com/ceejbot/xxbloom/blob/master/LICENSE
```
SgtPooki, line 1: why xxbloom instead of the more popular and recently updated `bloom-filters`? https://npmtrends.com/bloom-filter-vs-bloom-filters-vs-bloomfilter-vs-xxbloom

achingbrain: `bloom-filters` is quite a big dependency and it's published as CJS.

We can swap it out if we need to, but it seemed like overkill.

achingbrain: Bleh, maybe we do need more fully featured filters, turns out being able to remove a value from a filter is necessary.

I've opened an issue on `bloom-filters` to see if the maintainers would be amenable to making it more browser friendly.

PR #2508, opened 2024-04-26
feat: add peer filter

achingbrain:
To make filtering peers in a memory-efficient way possible, add a peer filter based on a bloom filter using the binary representation of the peer id.

## Change checklist


packages/peer-collections/src/filter.ts
```diff
@@ -0,0 +1,26 @@
+import { BloomFilter } from '@libp2p/utils/bloom-filter'
+import type { PeerId } from '@libp2p/interface'
+
+/**
+ * Uses a Bloom filter to implement a mechansim for deduplicatign PeerIds in a
```
SgtPooki, line 5:
```suggestion
 * Uses a Bloom filter to implement a mechansim for deduplicating PeerIds in a
```

SgtPooki (approved):
noice

PR #2514, opened 2024-04-30
fix: WebTransport stream now extends abstract stream

achingbrain:
The PR pulls all of the non-`@fails/webtransport` parts out of #2422

There's a lot of work that's been done to re-use existing libp2p code such as the abstract stream class which handles a lot more closing scenarios than the existing implementation so it would be good to get that in.

## Change checklist


packages/transport-webtransport/.aegir.js
```diff
@@ -1,14 +1,18 @@
...
 export default {
   test: {
-    async before() {
+    async before () {
+      const main = os.platform() === 'win32' ? 'main.exe' : 'main'
+
       if (!existsSync('./go-libp2p-webtransport-server/main')) {
         await new Promise((resolve, reject) => {
-          exec('go build -o main main.go',
+          exec(`go build -o ${main} main.go`,
```
SgtPooki, line 15: I don't think these build files are being cleaned by `npm run clean`, which they probably should be

packages/transport-webtransport/src/index.ts
```diff
@@ -175,15 +196,19 @@ class WebTransportTransport implements Transport {
           cleanUpWTSession('remote_close')
         })
 
-      if (!await this.authenticateWebTransport(wt, localPeer, remotePeer, certhashes)) {
+      if (!await raceSignal(this.authenticateWebTransport(wt, localPeer, remotePeer, certhashes), options.signal)) {
         throw new Error('Failed to authenticate webtransport')
       }
 
+      if (options?.signal?.aborted === true) {
+        throw new AbortError()
```
SgtPooki, line 204: maybe i'm misunderstanding things but looking at https://www.npmjs.com/package/race-signal, if the signal is aborted, `options?.signal?.aborted === true` will never be truthy. Should we check for aborted signal inside the `!await raceSignal` block, or wrap in try {} with a check there?

packages/transport-webtransport/src/muxer.ts
```diff
@@ -0,0 +1,95 @@
...
+        //! TODO unclear how to add backpressure here?
+        while (true) {
+          const { done, value: wtStream } = await reader.read()
+
+          if (done) {
+            break
+          }
+
+          if (activeStreams.length >= config.maxInboundStreams) {
+            log('too many inbound streams open, closing new incoming stream')
```
SgtPooki, line 36: we should log activeStreams.length and maxInboundStreams

packages/transport-webtransport/src/stream.ts
```diff
@@ -1,184 +1,110 @@
...
+          init.log('remote closed write')
+          return
+        }
 
-    async closeWrite (options?: AbortOptions) {
-      if (!writerClosed) {
-        writerClosed = true
+        if (result.value != null) {
+          this.sourcePush(new Uint8ArrayList(result.value))
+        }
```
SgtPooki, line 31: Can value exist and the read be done? we should check for value first.

SgtPooki (commented):
I'm not gonna pretend like my review is good enough to validate much, but I left some comments

packages/transport-webtransport/src/stream.ts
achingbrain: Not exactly - if `value` exists and `done` is false, then `value` is the return value of the iterator.

See the iterator protocol docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols

packages/transport-webtransport/.aegir.js
achingbrain: TBH I want to remove this transport server entirely and just rely on precomplied binaries in the interop tests.

packages/transport-webtransport/src/stream.ts
SgtPooki: ah. then i've incorrectly implemented iterators previously, or consumed some incorrectly implemented ones, because i've seen a sequence value and a done=true... good to know that's not a correct implementation.

achingbrain: 😬

packages/transport-webtransport/src/index.ts
achingbrain: Letting `raceSignal` throw if the signal aborts is enough, I've updated it.

packages/transport-webtransport/src/muxer.ts
achingbrain: Done!

PR #2520, opened 2024-05-02
fix: close connections after use

achingbrain:
Historically we've not tried to close connections after they've been opened because we let the connection pruner do that when it's required.

This leads to high resource usage and requires the pruner to be able to select the best connections for pruning which without the context of their intended use is a guessing game.

A distinction is drawn between protocols that have a definite purpose (e.g. kad-dht) and "speculative" protocols that are run on newly opened connections for other purposes, e.g. identify, auto nat, etc.

For example if kad-dht opens a connection to send a message, we can close that connection if the only protocol running on it is identify.

- Where protocols open connections that are not expected to be long-lived, try to close them if we are done
- Check the number of open streams on a connection before closing, if there are none or they are only speculative protocols, close the connection
- Sort connections by number of open streams in the pruner before pruning

## Change checklist


achingbrain:
Closing as the connection pruner now closes connections without streams first.

PR #2521, opened 2024-05-02
fix: prune connections based on stream counts and direction

achingbrain:
Changes how we sort connections to prune to be more than just tags and age.

Now we'll sort by tags, the number of open streams, direction and age.

This should choose idle connections without streams to close over those that are in use.

## Change checklist

### DIFF
--- a/packages/libp2p/test/connection-manager/connection-pruner.spec.ts
+++ b/packages/libp2p/test/connection-manager/connection-pruner.spec.ts
@@ -1,0 +1,98 @@
+import { TypedEventEmitter } from '@libp2p/interface'
+import { defaultLogger } from '@libp2p/logger'
+import { PeerMap } from '@libp2p/peer-collections'
+import { createEd25519PeerId } from '@libp2p/peer-id-factory'
+import { expect } from 'aegir/chai'
+import { stubInterface } from 'sinon-ts'
+import { ConnectionPruner } from '../../src/connection-manager/connection-pruner.js'
+import type { Libp2pEvents, PeerStore, Stream, TypedEventTarget, Connection } from '@libp2p/interface'
+import type { ConnectionManager } from '@libp2p/interface-internal'
+import type { StubbedInstance } from 'sinon-ts'
+
+interface ConnectionPrunerComponents {
+  connectionManager: StubbedInstance<ConnectionManager>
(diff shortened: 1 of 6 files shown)


packages/libp2p/test/connection-manager/connection-pruner.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+            )
+          }
+        }
+      }
+    }
+
+    // priority is:
+    // 1. tagged peers
+    // 2. connections with streams
+    // 3. outbound connections
```
maschad, line 76: What's the rationale behind outbound connections being given greater priority over longer-lived connections?

packages/libp2p/test/connection-manager/connection-pruner.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+      events: new TypedEventEmitter()
+    }
+
+    pruner = new ConnectionPruner({
+      ...components,
+      logger: defaultLogger()
+    }, {})
+  })
+
+  it('should sort connections for pruning', async () => {
```
maschad, line 35: I like the succinctness of this test but I do feel for readability it is at times better to be more explicit (at the expense of repetition). So a test would probably read as:

_older connections with streams should be closed before tagged connections_ for example. 

Otherwise debugging the sorting algorithm could become a bit cumbersome.

maschad (changes requested):
I presume that this insight came from interaction with other helia nodes, nevertheless would like to hear the rationale behind the changes.

achingbrain:
> would like to hear the rationale behind the changes.

If we do something like a DHT query, we open a connection to a peer, open a `/ipfs/kad/1.0.0` stream, send a message and read the response.  We do this repeatedly and eventually hit the max connections limit.  The pruner tries to close some connections to bring us back under the limit while we're trying to continue the query.

The older connections are less useful now, since that peer plays no further part in the query.  The more useful connections are the ones we're trying to open to continue the query - these are the newest connections so right now would be the ones that get closed by the pruner, almost immediately after being opened.  Oops.

Instead, we should look for connections that have no open streams - these are safe(r) to close because no protocol is using them.

maschad:
Thanks for the clarification

maschad (approved):

PR #2523, opened 2024-05-05
fix: pass autodial config thorugh in connection manager init

oott123:
## Title

fix: options in connection manager init should be passed to autodial
## Description

`autoDialPeerRetryThreshold` and `autoDialDiscoveredPeersDebounce` should be passed from connection manager to auto dial, otherwise it won't work

## Change checklist


achingbrain (approved):

achingbrain:
Thanks!

PR #2538, opened 2024-05-10
fix: use xor-compare for finding closer peers

achingbrain:
Switches from array compare to xor compare as this is how we're supposed to measure distance between two peers.

Also stores `PeerInfo`s in peer distance list instead of `PeerId`s so to not lose the peers found in a query that are not in the peer store yet.

## Change checklist


packages/kad-dht/src/peer-list/peer-distance-list.ts
```diff
@@ -1,9 +1,11 @@
-import { convertPeerId, getDistance } from '../utils.js'
-import type { PeerId } from '@libp2p/interface'
+import { xor as uint8ArrayXor } from 'uint8arrays/xor'
+import { xorCompare as uint8ArrayXorCompare } from 'uint8arrays/xor-compare'
```
SgtPooki, line 2: for future reference, this was added in https://github.com/achingbrain/uint8arrays/pull/89

SgtPooki (approved):

packages/kad-dht/src/peer-list/peer-distance-list.ts
```diff
@@ -1,9 +1,11 @@
...
+import { convertPeerId } from '../utils.js'
+import type { PeerId, PeerInfo } from '@libp2p/interface'
 
 interface PeerDistance {
-  peerId: PeerId
-  distance: bigint
+  peer: PeerInfo
+  distance: Uint8Array
```
maschad, line 8: If this ends up getting [converted to a `bigint`](https://github.com/libp2p/js-libp2p/blob/main/packages/kad-dht/src/peer-list/peer-distance-list.ts#L99) what's the rationale behind changing these to `Uint8Array` ?

packages/kad-dht/src/peer-routing/index.ts
```diff
@@ -195,17 +195,17 @@ export class PeerRouting {
   }
 
   /**
-   * Kademlia 'node lookup' operation on a key, which could be a the
-   * bytes from a multihash or a peer ID
+   * Kademlia 'node lookup' operation on a key, which could be a the bytes from
```
maschad, line 198:
```suggestion
   * Kademlia 'node lookup' operation on a key, where the key could be the bytes from
```

maschad (changes requested):
I agree with the change to _XOR comparison_ overall but left some comments.

packages/kad-dht/src/peer-list/peer-distance-list.ts
achingbrain: You're linking to main there - it no longer gets converted to a bigint, instead the `Uint8Array`s are the xor distances that get compared using `xorCompare`.

Issue #1808, opened 2023-06-13
test: Add WebKit to interop tests


p-shahi:
Work is being done to add WebKit to interop tests here https://github.com/libp2p/js-libp2p/pull/1799 however per https://github.com/libp2p/js-libp2p/issues/2007 this effort is currently blocked

dhuseby:
Webkit still doesn't have full support for WebRTC. Until then, this is blocked.

SgtPooki:
<details>
<summary>JS Colo 2024 discussion</summary>

we need a PR to test this so we know when things are actually unblocked without needing a lot of manual interaction
</details>

### Action item

1. get out draft PR for adding webkit to test runner

maschad:
@SgtPooki isn't that the purpose of https://github.com/libp2p/js-libp2p/pull/2541 ?

Issue #2007, opened 2023-08-30
WebkitGTK implementations don't currently support WebRTC


maschad:
I did some more research/testing and found that WebkitGTK (the full-featured port of the WebKit rendering engine on Linux-based OS's) does not _yet_ support the [PeerConnection API](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection). Igalia is currently working on [the implementation](https://base-art.net/Articles/webrtc-in-webkitgtk-and-wpe-status-updates-part-i/) but this is obviously a blocker for Playwright supporting this interface as [per the docs](https://webkitgtk.org/reference/webkit2gtk/stable/index.html) 

Thanks to @MarcoPolo for pointing me in this direction.

_Originally posted by @maschad in https://github.com/libp2p/js-libp2p-webrtc/issues/177#issuecomment-1589529908_
         
      
I suppose this is why this [it marked as experimental](https://github.com/WebKit/WebKit/blob/03608538134cc6a2bf14ca5359ad1d6ae020fcf6/Source/cmake/OptionsGTK.cmake#L155)

dhuseby:
Still blocked due to WebKit not fully supporting WebRTC

achingbrain:
The tests pass now, all good.

PR #1832, opened 2023-06-16
test: run tests on webkit for webRTC

maschad:

PR #2541, opened 2024-05-11
chore: enable webkit tests in interop

maschad:
## Title

ci: enable webkit tests in interop and webRTC
## Description

We had issues in [the past](https://github.com/libp2p/js-libp2p/pull/1832) running these tests which lead to the discovery of [inadequate support in WebkitGTK](https://github.com/libp2p/js-libp2p/issues/2007) 

Closes https://github.com/libp2p/js-libp2p/issues/1808

## Change checklist


achingbrain (changes requested):
The `test:webkit` script needs to be added to the root `package.json` as well otherwise these tests won't run.

achingbrain (approved):

PR #2544, opened 2024-05-14
feat: add optional topology filter

achingbrain:
Adds a `filter` option to topologies to allow filtering out duplicate notifications.

## Change checklist


maschad (commented):
Looks good, should we add tests for `TopologyFilter` ?

achingbrain:
The registrar test suite has new `should use a filter...` tests that exercise the code paths, using a `PeerFilter` which has type overlap with `TopologyFilter` so I think we're good there.

PR #2547, opened 2024-05-15
fix: Transient connection logic judgment

wlynxg:
## Description
In go-libp2p, transient connections are judged based on whether the relay node has restrictions on the current connection. Reference code: https://github.com/libp2p/go-libp2p/blob/fccfbe7a4b6841325100038b83f86cd4e097b9a1/p2p/protocol/circuitv2/client/dial.go#L181

This submission aligns js-libp2p logic with go-libp2p logic.

## Change checklist

- [ ✔ ] I have performed a self-review of my own code
- [ ✔ ] I have made corresponding changes to the documentation if necessary (this includes comments as well)
- [ ✔ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/doc/package.json
+++ b/doc/package.json
@@ -24,7 +24,7 @@
     "doc-check": "aegir doc-check"
   },
   "devDependencies": {
-    "aegir": "^42.2.11"
+    "aegir": "^43.0.1"
   },
   "private": true
 }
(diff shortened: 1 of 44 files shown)


wlynxg:
> Thanks for opening this. Please can you add a test to prevent future regressions and ensure linting passes.

Sorry, I'm not very familiar with js, I can't write this test😟

achingbrain:
Hmm, not sure why that closed this PR.  Continued in #2575

PR #2551, opened 2024-05-17
feat: add devtools metrics

achingbrain:
Adds a metrics implementation that is scrapable by a browser DevTools plugin.

## Change checklist


maschad (approved):
Awesome 🚀

SgtPooki (approved):
this is so great. We should have a demo site/video linked in README so we can show how this works.

achingbrain:
> Do you know what might be the reason?

I think you may be using the placeholder version published to npm.  You'll need to replace the `src` and `dist` folders in `/node_modules/libp2p/devtools-metrics` with versions built from this PR branch for it to work.

PR #2580, opened 2024-06-06
added exportToJSON

st3b1t:
if exists an `createFromJSON()` method we need of an `exportToJSON()` that use same encoding


achingbrain:
Thanks for opening this.  The `@libp2p/peer-id-factory` module was removed in the recent 2.x.x release.

The `createFromJSON` function itself was only used to load test fixtures.  It's not used here any more, if you need this in your own project you can define it there.

Issue #2581, opened 2024-06-07
Tracking Issue - WebRTC direct in Node.js


2color:
## Background

[WebRTC-direct](https://github.com/libp2p/specs/blob/master/webrtc/webrtc-direct.md) is currently only supported in browsers [as mentioned in package](https://github.com/libp2p/js-libp2p/tree/main/packages/transport-webrtc#support). 

This is a tracking issue for adding support for WebRTC-direct in Node.js, which would allow browsers to dial Node.js directly (without circuit relay and signalling) without needing for a TLS certificate as in the case of WebSockets. 

## Dependencies 

This require changes to both https://github.com/murat-dogan/node-datachannel and https://github.com/paullouisageneau/libdatachannel to allow SDP Munging.

Related PRs/issues: 

```[tasklist]
### Tasks
- [ ] https://github.com/paullouisageneau/libjuice/pull/243
- [ ] https://github.com/paullouisageneau/libdatachannel/pull/1172
- [ ] https://github.com/paullouisageneau/libdatachannel/pull/1201
- [ ] https://github.com/paullouisageneau/libdatachannel/pull/1204
- [ ] https://github.com/murat-dogan/node-datachannel/pull/257
- [ ] https://github.com/murat-dogan/node-datachannel/pull/256
- [ ] https://github.com/libp2p/go-libp2p/issues/2827
```

achingbrain:
There is good news when all of the above PRs are applied:

<img width="569" alt="image" src="https://github.com/libp2p/js-libp2p/assets/665810/0f2daded-1782-48e0-b9fd-7d2c5df0a895">

christroutner:
I am trying to use webRTC in an exclusive node.js environment (no browser nodes). I created issue #2702 and wanted to link to it here, as it might be relevant to this discussion.

achingbrain:
Closed by #2583

PR #2583, opened 2024-06-07
feat: WebRTC-Direct support for Node.js

achingbrain:
Supports listening and dialing WebRTC Direct multiaddrs in Node.js.

Depends on:

- [x] https://github.com/libp2p/go-libp2p/issues/2827
- [x] https://github.com/paullouisageneau/libjuice/pull/248
- [x] https://github.com/paullouisageneau/libdatachannel/pull/1206
- [x] https://github.com/paullouisageneau/libdatachannel/pull/1207
- [x] https://github.com/murat-dogan/node-datachannel/pull/258
- [x] https://github.com/murat-dogan/node-datachannel/pull/259
- [ ] https://github.com/murat-dogan/node-datachannel/pull/260
- [ ] https://github.com/murat-dogan/node-datachannel/pull/256
- [ ] https://github.com/paullouisageneau/libdatachannel/pull/1211
- [ ] https://github.com/paullouisageneau/libjuice/pull/292

Closes:

- https://github.com/libp2p/js-libp2p/issues/2581

## Change checklist


xicilion:
I read the code a bit, in this implementation when Peer B listens, it first listens for incoming stun packets using dgram, extracts the ufrag, and then establishes a connection with Peer A from another port. This works fine in a direct network connection.
But if Peer A is behind NAT, this type of connection will fail because the NAT detects incoming packets from another port, which was not accessed from inside before.

achingbrain:
I haven't got to the point of testing NAT traversal yet, still a lot of PRs to land in the deps.

I was wondering if each `RTCPeerConnection` could negotiate it's own NAT traversal via STUN/TURN but it would probably be less resource intensive to have UDPMUX on a single port, I agree.

xicilion:
> I was wondering if each `RTCPeerConnection` could negotiate it's own NAT traversal via STUN/TURN.

From a logic perspective, even though multiple RTCPeerConnections share a single UDP socket in UDPMUX mode, each of them manages their own status separately and there is no difference compared to regular RTCPeerConnections. This means that the negotiation process for them is also no different from regular RTCPeerConnections.

achingbrain:
I have forked `node-datachannel` as `@ipshipyard/node-datachannel` with all of the above patches applied.

Now CI is passing I'm going to merge this to allow further experimentation.

PR #2587, opened 2024-06-12
docs: add doc on configuring custom services

achingbrain:
Writes up the perceived wisdom on service implementation.

## Change checklist

### DIFF
--- a/doc/SERVICES.md
+++ b/doc/SERVICES.md
@@ -131,7 +131,14 @@ class MyService {
   }
 
   saySomething (): string {
-    return this.message.replace('{}', this.components.connectionManager.getDialQueue().length)
+    return this.message.replace('{}', `${this.components.connectionManager.getDialQueue().length}`)
+  }
+}
+
+// a function that returns a factory function
+function myService (init: MyServiceInit) {
+  return (components: MyServiceComponents) => {
+    return new MyService(components, init)
   }
 }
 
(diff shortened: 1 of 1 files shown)


doc/SERVICES.md
```diff
@@ -0,0 +1,227 @@
...
+    myService: myService()
+  }
+})
+
+console.info(node.services.myService.saySomething()) // 'Hello, my peer id is 123Koo....'
+```
+
+### Using classes
+
+Use a class for your service and returning an instance will allow you to keep your code manageable.
```
maschad, line 125: I'm not sure it's appropriate to suggest a particular programming paradigms here, it seems beyond the scope of documenting configuration.

doc/SERVICES.md
```diff
@@ -0,0 +1,227 @@
...
+    })
+  }
+})
+
+console.info(node.services.myService.saySomething()) // 'World'
+```
+
+### Accessing libp2p components
+
+Services can access internal libp2p components such as the address manger and connection manager by accepting an argument to the returned function.
```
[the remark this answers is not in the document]
achingbrain: Sounds good.  I wanted to do something simple but I guess this gives a clue of where to find the internal interface definitions.

doc/SERVICES.md
achingbrain: Most people will just copy/paste so I think it's fair to show them the "most idiomatic" way of doing things, though you are right it's a bit weird to call this out specifically.

I've updated the examples.

maschad (commented):
Nice!

2color:
This is really nice. Just saw this now!❤️

PR #2590, opened 2024-06-14
feat: allow passing a function for rtcConfiguration

### LINKED ISSUE #2554: Update RTCConfig for WebRTCTransport
- **Version**:
"libp2p": "^1.2.4",
"@libp2p/webrtc": "^4.0.22",

- **Platform**:
`Linux master 6.8.6-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 13 Apr 2024 14:42:24 +0000 x86_64 GNU/Linux`

- **Subsystem**:
WebRTCTransport

#### Severity:
Medium

#### Description:
Currently, the RTC configuration can only be set at the time of construction of the WebRTCTransport. While this is typically fine for STUN servers and the like, TURN servers often have credentials which are limited in their lifetime. For libp2p peers which are long lived, the peer (and thus the Transport instance) may outlive the validity of the TURN credential.

Since the initially provided credential is used each time the 'dial' method is called on the transport, eventually the Transport will not be able to request TURN candidates.

1) Leverage any common TURN service such as Twilio, Cloudflare, Metered.ca, etc
2) Request a credential with a short lifetime (eg 180 seconds)
3) Create a libp2p instance with a WebRTCTransport including the provided TURN credentials
4) Observe that TURN cadidates are no longer available after the libp2p instance lifetime exceeds the credential lifetime

achingbrain:
In order to allow refreshing STUN/TURN credentials between dials, allow passing a function that returns rtc config.

Fixes #2554

## Change checklist

### DIFF
--- a/packages/transport-webrtc/src/private-to-private/transport.ts
+++ b/packages/transport-webrtc/src/private-to-private/transport.ts
@@ -19,7 +19,7 @@ export const SIGNALING_PROTO_ID = '/webrtc-signaling/0.0.1'
 const INBOUND_CONNECTION_TIMEOUT = 30 * 1000
 
 export interface WebRTCTransportInit {
-  rtcConfiguration?: RTCConfiguration
+  rtcConfiguration?: RTCConfiguration | (() => RTCConfiguration | Promise<RTCConfiguration>)
   dataChannel?: DataChannelOptions
 
   /**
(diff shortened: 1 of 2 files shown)


achingbrain:
cc @justin0mcateer

PR #2598, opened 2024-06-20
feat: add url peer id

achingbrain:
To parse non-cryptographic peer ids from http servers, add the url peer id type.

## Change checklist


maschad (changes requested):
Looks good, but there are some type [incompatibilities](https://github.com/libp2p/js-libp2p/actions/runs/9599947204/job/26474961047?pr=2598#step:4:136) in [peer-id-factory](https://github.com/libp2p/js-libp2p/blob/main/packages/peer-id-factory/src/index.ts#L101-L112)

maschad (approved):

PR #2599, opened 2024-06-20
fix: send raw plaintext public key during handshake

oott123:
## Title

fix: make plaintext public key follow to the specs

## Description

In js, all "public keys" are marshaled using protobuf PublicKey message.

However, according to [specs](https://github.com/libp2p/specs/blob/master/plaintext/README.md) and other implementations, the `Data` field should contains the raw key, instead of the protobuf marshaled key, and then put in to the exchange.

Otherwise, js-libp2p will not connect to golang or rust libp2p, raising the "Public key did not match id" error, if only plaintext selected for encryption.

## Notes & open questions

Should we involve some cross-implementation tests to check if any further issue like that?

## Change checklist


packages/connection-encrypter-plaintext/src/index.ts
```diff
@@ -91,7 +91,7 @@ class Plaintext implements ConnectionEncrypter {
         id: localId.toBytes(),
         pubkey: {
           Type: type,
-          Data: localId.publicKey ?? new Uint8Array(0)
+          Data: localId.publicKey === null ? new Uint8Array(0) : (PublicKey.decode(localId.publicKey).Data ?? new Uint8Array(0))
```
achingbrain, line 94: To fix the build errors, I think this needs to be something like:

```suggestion
          Data: localId.publicKey == null ? new Uint8Array(0) : (PublicKey.decode(localId.publicKey).Data ?? new Uint8Array(0))
```

`x == null` is true if `x` is `null` or `undefined`, `x === null` is only true if `x` is `null`.

oott123:
So weird that tests are still failed. I think I have finished all the changes required... have no clue now

Issue #2623, opened 2024-07-18
private ipv6 address identified as public


marcus-pousette:
- **Version**:
1.8.1

- **Subsystem**:
dialer, defaultAddressSort

#### Severity:
Low

#### Description:
Address 
```ts
const address = '/ip6/::ffff:7f00:1/tcp/52498/p2p/12D3KooWFp4Uq323R3rF1ukZouXhmai37TejRJwMmsRg9fMM2jjY/p2p-circuit/p2p/12D3KooWPv7ef6tF1PjK1vFzjxW11rpcxp2rHjowtSLU4Xt2CNWP'
```
evaluates 
```ts
import { isPrivate } from './multiaddr/is-private.js';
```

```ts
isPrivate(address) 
```

is `false`,  but expected `true`?

This makes the default address sorter to prefer this address over other addresses

This one 
```
/ip4/127.0.0.1/tcp/52493/ws/p2p/12D3KooWFp4Uq323R3rF1ukZouXhmai37TejRJwMmsRg9fMM2jjY/p2p-circuit/p2p/12D3KooWPv7ef6tF1PjK1vFzjxW11rpcxp2rHjowtSLU4Xt2CNWP
```

evaluates to as private correctly

achingbrain:
Reading [RFC4291](https://datatracker.ietf.org/doc/html/rfc4291), the address `::ffff:7f00:1` is an [IPv4-Mapped IPv6 Address](https://datatracker.ietf.org/doc/html/rfc4291#section-2.5.5.2).

So the `::ffff` part marks it as this type of address and the final 32 bits are an IPv4 address.

Add the omitted zeroes and split the octets, and `7f00:1` becomes `7f.00.00.01`, which in base 10 is `127.0.0.1`.

I think we need to add an extra check in [isPrivateIp](https://github.com/libp2p/js-libp2p/blob/main/packages/utils/src/private-ip.ts#L58-L59) for this type of address, then convert the final 32 bits to an IPv4 address and test that like the others.

Would you like to open a PR?

marcus-pousette:
already did :)

PR #2624, opened 2024-07-18
fix: add ipv6Check regex for private address

marcus-pousette:
## Title

fix: add ipv6Check regex for private address

## Description

Solves #2623 

## Change checklist


achingbrain:
Will this handle other types of IPv4-Mapped IPv6 addresses made from private ipv4 addresses?

E.g.

`10.0.0.1` -> `::FFFF:0a00:1` - should be detected as private
`142.250.74.238` -> `::FFFF:8efa:4aee` - should be detected as public

marcus-pousette:
> Will this handle other types of IPv4-Mapped IPv6 addresses made from ipv4 addresses?
> 
> E.g.
> 
> `10.0.0.1` -> `::FFFF:0a00:1` - should be detected as private `142.250.74.238` -> `::FFFF:8efa:4aee` - should be detected as public

aaah yes, I realised that the pattern is not perhaps right given your explanation in the github issue

marcus-pousette:
Hmm this was a bit tricky, it seems like the multiaddr function removes the leading zeros, so for the 

example 
::ffff:0a00:1

we actually match against
::ffff:a00:1

the same can be for 1 or many leading zeros, 

It feels like this regex problem should have been solved many times other by someone else already

achingbrain:
Agreed, it doesn't sound like this should be a new problem..  Our version [comes from](https://github.com/libp2p/js-libp2p/pull/2362) the [private-ip](https://www.npmjs.com/package/private-ip) module, which hasn't seen an update to these patterns for years.

marcus-pousette:
I did a little bit of a not-like looking regex now where I explicitly try to match against some local ones. It works but might not be ideal

marcus-pousette:
> Agreed, it doesn't sound like this should be a new problem.. That said, our version [comes from](https://github.com/libp2p/js-libp2p/pull/2362) the [private-ip](https://www.npmjs.com/package/private-ip) module, which hasn't seen an update to these patterns for years.

this one might be right, but the problem is that toString() on a multiaddrs yield a format where leading 0s are removed which I guess does not work well with this pattern

achingbrain (changes requested):
This is nearly there - I've suggested a few more test cases for public addresses - it's failing on the first address above the class B range - `172.32.0.1`, the others seem fine.

marcus-pousette:
thanks for making the changes! did not have time to continue this week

achingbrain:
No problem - I wanted to get it in the next release, thanks for finding the issue and opening the PR!

PR #2639, opened 2024-07-29
fix!: remove autodialer

### LINKED ISSUE #2418: AutoDial retries creeps up CPU usage over time
- **Version**: 1.2.3

- **Platform**: `Linux 7e698928f50f 5.15.49-linuxkit-pr #1 SMP PREEMPT Thu May 25 07:27:39 UTC 2023 aarch64 Linux`

- **Subsystem**: Connection manager

### LINKED ISSUE #2621: Remove the autodialer?
The perceived wisdom is that we need active connections to ensure connectivity etc, so we have an autodialer that ensures we always have some connections.

This is quite bad for mobile apps, since the internet connection can be intermittent, the connection churn is higher so your phone starts to get warm and your app is more likely to be killed as a resource hog.

Some protocols like gossipsub will require active connections but others such as kad-dht generally don't since they are message based (in that we connect, send a message and read the response).

If the peer has an up to date peer book with some entry points to the network it may not be necessary to maintain more connections than just to a relay to ensure dialability, for example.

This needs some experimentation but it might be enough for the bootstrap module to dial the bootstrap peers instead of "discovering" them.

Once identify has run the bootstrappers will be in the DHT routing tables so there will be a starting point for subsequent queries and there should be no need to ensure we have active connections.

achingbrain:
The autodialer is a feature from an older time when the DHT was less reliable and we didn't have things like the random walk component.

There's not a lot of benefit in opening connections to any old peer, instead protocols now have better ways of targetting the kind of peers they require.

Actively dialing peers harms platforms where connections are extremely expensive such as react-native so this feature has been removed.

Closes #2621
Fixes #2418

BREAKING CHANGE: the autodialer has been removed as well as the corresponding config keys

## Change checklist


maschad (changes requested):
First pass, but shouldn't we be basing this off a `release-v2.0` branch? Similar to what we did with [release-v1.0](https://github.com/libp2p/js-libp2p/tree/release-v1.0) ?

maschad (changes requested):
Nice work 🚀  there are just a few little things to be addressed.

- The Peer Discovery docs still have a [few references](https://github.com/libp2p/js-libp2p/blob/feat/remove-autodialler/doc/PEER_DISCOVERY.md?plain=1#L24) to `minConnections` 
- The [`autoDialInterval`](https://github.com/libp2p/js-libp2p/blob/feat/remove-autodialler/packages/libp2p/test/connection-manager/index.spec.ts#L22) is now irrelevant
- We should include a migration guide essentially stating which configurations i.e. the _autoDial_ params,  have been removed and why.

achingbrain:
Need to double check the reconnect on disconnect for `KEEP_ALIVE` tagged peers still works.

packages/libp2p/src/connection-manager/auto-dial.ts
```diff
@@ -1,285 +0,0 @@
...
-  private readonly autoDialDiscoveredPeersDebounce: number
-  private autoDialInterval?: ReturnType<typeof setInterval>
-  private started: boolean
-  private running: boolean
-  private readonly log: Logger
-
-  /**
-   * Proactively tries to connect to known peers stored in the PeerStore.
-   * It will keep the number of connections below the upper limit and sort
-   * the peers to connect based on whether we know their keys and protocols.
```
SgtPooki, line 54: do we have this peer sorting functionality elsewhere? anything we need to keep in here?

Also kind of related to @maschad 's original ask about potentially keeping this for some users: There may be a need to have an auto-dialer for swarm-like functionality, where you want to stay connected to a specific set of peers? Can users accomplish that without auto-dial today?

packages/libp2p/src/connection-manager/auto-dial.ts
```diff
@@ -1,285 +0,0 @@
...
-
-    if (this.queue.size > this.autoDialMaxQueueLength) {
-      this.log('not enough connections %d/%d but auto dial queue is full', numConnections, this.minConnections)
-      this.sheduleNextAutodial()
-      return
-    }
-
-    this.running = true
-
-    this.log('not enough connections %d/%d - will dial peers to increase the number of connections', numConnections, this.minConnections)
```
SgtPooki, line 152: IIRC, we already have logic somewhere that prevents the need to dial bootstrappers if we have highly rated "close" peers in the peer store.. if it's elsewhere other than the auto-dialer, then this can all probably go away.

Question: does the auto-dialer tag/modify peers in the peer store that could change behavior for other services/components? It doesn't look like it does, but I could be missing something.

packages/libp2p/test/connection-manager/index.spec.ts
```diff
@@ -19,7 +19,6 @@ import type { TransportManager } from '@libp2p/interface-internal'
 
 const defaultOptions = {
   maxConnections: 10,
-  minConnections: 1,
   autoDialInterval: Infinity,
```
SgtPooki, line 22:
```suggestion
```

packages/libp2p/src/connection-manager/index.ts
```diff
@@ -187,10 +133,9 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
 
   constructor (components: DefaultConnectionManagerComponents, init: ConnectionManagerInit = {}) {
     this.maxConnections = init.maxConnections ?? defaultOptions.maxConnections
-    const minConnections = init.minConnections ?? defaultOptions.minConnections
 
-    if (this.maxConnections < minConnections) {
-      throw new CodeError('Connection Manager maxConnections must be greater than minConnections', codes.ERR_INVALID_PARAMETERS)
+    if (this.maxConnections < 0) {
+      throw new CodeError('Connection Manager maxConnections must be greater than zero', codes.ERR_INVALID_PARAMETERS)
```
SgtPooki, line 138: either 
```suggestion
    if (this.maxConnections <= 0) {
      throw new CodeError('Connection Manager maxConnections must be greater than zero', codes.ERR_INVALID_PARAMETERS)
```

or 

```suggestion
    if (this.maxConnections < 0) {
      throw new CodeError('Connection Manager maxConnections must be greater than or equal to zero', codes.ERR_INVALID_PARAMETERS)
```

SgtPooki (changes requested):
one potential debug log UX, or functional bug, if maxConnections isn't supposed to be set to zero.

packages/libp2p/src/connection-manager/auto-dial.ts
SgtPooki: i think the peer-discovery-bootstrap component does this..?

packages/libp2p/src/connection-manager/auto-dial.ts
achingbrain: Yes, tag them with `KEEP_ALIVE`.

packages/libp2p/src/connection-manager/index.ts
achingbrain: Fixed during merge of target branch.

packages/libp2p/src/connection-manager/auto-dial.ts
achingbrain: > does the auto-dialer tag/modify peers in the peer store

No, it only reads peers from the peer store.

> IIRC, we already have logic somewhere that prevents the need to dial bootstrappers if we have highly rated "close" peers in the peer store

The notion of "close" is down to the protocol that defines it.

For example KAD-DHT already populates it's routing tables with peer store peers on startup, but it doesn't need to actively dial those peers, only during an actual query.

In the case where those peers have gone away between restarts, the KAD paper defines how to ping old peers to make sure they're still contactable during routing table maintenance.

ZababurinSergei:
After extending autoDial, the module stopped working for me js-libp2p-pubsub-peer-discovery

https://github.com/libp2p/js-libp2p-pubsub-peer-discovery

How can I make the search start working?

PR #2644, opened 2024-07-31
feat: add connection monitor

### LINKED ISSUE #2643: Ping recording connection latency/aliveness
If the user has configured the ping protocol, we could have it periodically check that the remote is still available.  It could store the ping time as a 'latency' or 'rtt' property on the `Connection` object.

If it fails, the connection should be closed.

This is similar to https://github.com/alanshaw/it-ws/issues/113 or https://github.com/ChainSafe/js-libp2p-yamux/issues/76 except it would work across all transports and muxers, including those transports with their own muxers.

This would be non-breaking so can go in independently of 2.0.0.

achingbrain:
Adds a connection monitor that periodically ensures remote peers are still online and contactable by trying to send a single byte via the ping protocol, and sets the `.rtt` property of the connection to how long it took.

If the ping protocol is not supported by the remote, it tries to infer the round trip time by how long it took to fail.

If the remote is unresponsive or opening the stream fails for any other reason, the connection is aborted with the throw error.

It's possible to configure the ping interval, how long we wait before considering a peer to be inactive and whether or not to close the connection on failure.

Closes #2643

## Change checklist

### DIFF
--- a/packages/libp2p/src/connection-monitor.ts
+++ b/packages/libp2p/src/connection-monitor.ts
@@ -1,0 +1,121 @@
+import { serviceCapabilities } from '@libp2p/interface'
+import { anySignal } from 'any-signal'
+import { byteStream } from 'it-byte-stream'
+import type { ComponentLogger, Logger, Startable } from '@libp2p/interface'
+import type { ConnectionManager } from '@libp2p/interface-internal'
+
+const DEFAULT_PING_INTERVAL_MS = 10000
+const DEFAULT_PING_TIMEOUT_MS = 2000
+
+export interface ConnectionMonitorInit {
+  /**
+   * How often to ping remote peers in ms
+   *
+   * @default 10000
+   */
+  pingInterval?: number
+
+  /**
+   * How long the ping is allowed to take before the connection will be judged
+   * inactive and aborted
+   *
+   * @default 2000
+   */
+  pingTimeout?: number
+
+  /**
+   * If true, any connection that fails the ping will be aborted
+   *
+   * @default true
+   */
+  abortConnectionOnPingFailure?: boolean
+}
+
+export interface ConnectionMonitorComponents {
+  logger: ComponentLogger
+  connectionManager: ConnectionManager
+}
+
+export class ConnectionMonitor implements Startable {
+  private readonly components: ConnectionMonitorComponents
+  private readonly log: Logger
+  private heartbeatInterval?: ReturnType<typeof setInterval>
+  private readonly pingIntervalMs: number
+  private readonly pingTimeoutMs: number
+  private abortController?: AbortController
+
+  constructor (components: ConnectionMonitorComponents, init: ConnectionMonitorInit = {}) {
+    this.components = components
+
+    this.log = components.logger.forComponent('libp2p:connection-monitor')
+    this.pingIntervalMs = init.pingInterval ?? DEFAULT_PING_INTERVAL_MS
+    this.pingTimeoutMs = init.pingTimeout ?? DEFAULT_PING_TIMEOUT_MS
+  }
+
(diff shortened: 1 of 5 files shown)


achingbrain:
This was discussed on the maintainers call and there was a suggestion to make this a service that the user would have to configure.

I'm having a hard time thinking of when you wouldn't want this so it's implemented as part of libp2p and on by default.  

Happy to talk about this further before merge.

justin0mcateer:
LGTM. This will eliminate the complexity of managing this at the application layer which is a major improvement.

packages/libp2p/src/connection-monitor.ts
```diff
@@ -0,0 +1,127 @@
...
+export interface ConnectionMonitorInit {
+  /**
+   * How often to ping remote peers in ms
+   *
+   * @default 10000
+   */
+  pingInterval?: number
+
+  /**
+   * Timeout settings for How long the ping is allowed to take before the
```
maschad, line 19:
```suggestion
   * Timeout settings for how long the ping is allowed to take before the
```

packages/libp2p/src/connection-monitor.ts
```diff
@@ -0,0 +1,127 @@
...
+              })
+            ])
+
+            conn.rtt = Date.now() - start
+
+            await bs.unwrap().close({
+              signal
+            })
+          } catch (err: any) {
+            if (err.code !== 'ERR_UNSUPPORTED_PROTOCOL') {
```
maschad, line 101: If we are will to monitor connections that don't support the protocol, should we also include the `ERR_CONNECTION_DENIED` error as well? In a scenario should the connection fail to upgrade?

maschad (commented):
Nice work on this, I left a suggestion and I also have a question.

packages/libp2p/src/connection-monitor.ts
achingbrain: `ERR_CONNECTION_DENIED` is thrown when the remote has exceeded a rate limit and results in the connection being closed so I don't think it's relevant here.

SgtPooki (approved):
lgtm.. only comment i have is about being a little more explicit with the assertions in the tests instead of `.gte(0)` after a delay of 100ms

packages/libp2p/src/connection-monitor.ts
```diff
@@ -0,0 +1,127 @@
...
+            await Promise.all([
+              bs.write(new Uint8Array(1), {
+                signal
+              }),
+              bs.read(1, {
+                signal
+              })
+            ])
+
+            conn.rtt = Date.now() - start
```
wemeetagain, line 95: why is this not divided by 2 too?

wemeetagain (commented):
I'd like to see this be disable-able. This is definitely something that most applications will want to use, but users should be able to turn this off and not pay for it.

As an example, in Lodestar, we use an application-specific ping protocol, with a response that means something quite different. We would like to disable this and handle pings / keep-alives in an application-specific manner.

packages/libp2p/src/connection-monitor.ts
achingbrain: This one times how long it takes to send and receive one byte, (e.g. one round trip) but after the stream has been opened (e.g. after multistream select has finished).

The `ERR_UNSUPPORTED_PROTOCOL` error is thrown by multistream select, which takes two round trips, so that needs dividing by two but the happy path here doesn't.

maschad (approved):
Nice work

achingbrain:
> I'd like to see this be disable-able

@wemeetagain a flag has been added:

```js
  connectionMonitor: {
    enabled: false
  }
})
```

SgtPooki (approved):

Issue #2622, opened 2024-07-18
circuit-relay-server limits and shades of `transient`


marcus-pousette:
- **Version**:
1.8.1

- **Subsystem**:
registrar
circuit relay server, registrar

#### Severity:
Medium

#### Description:
Since limits are applied to connection by default, all circuit relay connections will become transient. This means protocol handles that have set notifyOnTransient to `false` will not receive onConnect events for these connections.

But, are non-unlimited circuit relay connections really transient in the same way as the one we get from `identify`?
What is an intuitive way on listen to topology changes: 

A. notifyOnTransient set to true, and filter out all transient connections except the relayed one
or
b. notifyOnTransient set to false, but set `applyDefaultLimit: false` on the circuit relay server config?

It feeeels, like `transient` coulb be replace with `limits` property object

and then every connection would have this property and we can use that to filter out unwanted connection, for example one that will timeout to soon, or does not allow us to send that much data

achingbrain:
> all circuit relay connections will become transient

This was true up until about a month ago when we merged https://github.com/libp2p/js-libp2p/pull/2575 - now relayed connections are only marked transient if the relay server actually applies limits.

> It feeeels, like transient could be replace with limits property object

I think this might be a good idea.  I'm certainly keen to move away from the 'transient' language in favour of 'limited' to better align with the changes around naming in [go-libp2p@0.34.0](https://github.com/libp2p/go-libp2p/releases/tag/v0.34.0) and showing the actual limits instead of having an opaque boolean value would make things a bit easier to work with.

E.g. the `.limits` object could have `bytes` and `seconds` properties that decrease as the allowances are used up.

We can then remove `runOnTransientConnection`/`notifyOnTransient` from `libp2p.handle` and `libp2p.register` (though keep it for `libp2p.dialProtocol`?  Or is it fair to say you are on your own if you dial a protocol using a circuit relay address?) and all stream handlers and topology listeners would be invoked regardless of the limit, then they can make the decision to use the connection or not themselves.

These are all (very) breaking changes so would cause `libp2p@2.x.x` to be released.

> But, are non-unlimited circuit relay connections really transient in the same way as the one we get from `identify`?

I'm not sure I understand this question.  What connections do you get from `identify`?

marcus-pousette:
> This was true up until about a month ago when we merged https://github.com/libp2p/js-libp2p/pull/2575 - now relayed connections are only marked transient if the relay server actually applies limits. 

[...]

PR #2645, opened 2024-07-31
fix!: rename "transient" connections to "limited"

achingbrain:
To better align with [go-libp2p@0.34.0](https://github.com/libp2p/go-libp2p/releases/tag/v0.34.0) rename "transient" connections to "limited".

BREAKING CHANGE: There are three breaking API changes:
  * The `.transient` boolean property has been replaced with a `.limits` object that may be undefined or may have `.bytes` and/or `.seconds` properties - the value of these properties decreases as the limits are used up
  * The `runOnTransientConnection` property of `libp2p.handle` and `libp2p.dialProtocol` has been renamed to `runOnLimitedConnection`
  * The `notifyOnTransient` property of `libp2p.register` has been renamed `notifyOnLimitedConnection`

Refs: #2622

## Change checklist

### DIFF
--- a/packages/integration-tests/test/circuit-relay.node.ts
+++ b/packages/integration-tests/test/circuit-relay.node.ts
@@ -638,12 +638,12 @@ describe('circuit-relay', () => {
       expect(circuitListener[0].relayStore.listenerCount('relay:removed')).to.equal(1)
     })
 
-    it('should mark an outgoing relayed connection as transient', async () => {
+    it('should mark an outgoing relayed connection as limited', async () => {
       // discover relay and make reservation
       const connectionToRelay = await remote.dial(relay1.getMultiaddrs()[0])
 
-      // connection to relay should not be marked transient
-      expect(connectionToRelay).to.have.property('transient', false)
+      // connection to relay should not be limited
+      expect(connectionToRelay).to.have.property('limits').that.is.undefined()
 
       await usingAsRelay(remote, relay1)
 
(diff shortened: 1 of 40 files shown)


achingbrain:
I've backed away from removing `runLimitedConnection` and `notifyOnLimitedConnection` because there are cases when you cannot make the decision ahead of time, such as:

```js
const stream = await libp2p.dialProcotol(peerId, '/my/protocol/1.0.0')
```

In the above since you're dialing the protocol on the remote peer and not a multiaddr, you don't know ahead of time if you are connected or if the connection is relayed/direct/limited so the `runLimitedConnection` option provides an escape hatch.

achingbrain:
Draft as I've yet to make the data limit decrease as the stream transfers bytes.

maschad (commented):
Nice work.

1. A minor comment adjustment [here](https://github.com/libp2p/js-libp2p/blob/feat/remove-autodialler/packages/protocol-dcutr/README.md?plain=1#L62)

2. we don't really have any docs around this but I suppose we still should have a migration guide correct?

SgtPooki:
> I've backed away from removing `runLimitedConnection` and `notifyOnLimitedConnection` because there are cases when you cannot make the decision ahead of time, such as:
> 
> ```js
> const stream = await libp2p.dialProcotol(peerId, '/my/protocol/1.0.0')
> ```
> 
> In the above since you're dialing the protocol on the remote peer and not a multiaddr, you don't know ahead of time if you are connected or if the connection is relayed/direct/limited so the `runLimitedConnection` option provides an escape hatch.
> 
> Happy to discuss further if anyone has a better idea for the API.

I think your rationale is sound. Though I'm not sure where the "removing" work was happening or what that consisted of. I'm guessing you were thinking of removing https://github.com/libp2p/js-libp2p/blob/c5dba70a9a7df43741c59b50a78f1dfda659396c/packages/interface/src/connection/index.ts#L193 ?

Do we have a best practice or prior art for whether we default API decisions to a "try, analyze, retry" vs "config for goal success" ? 

I know we don't like to "config all the things" but it seems like js-libp2p likes to go the "config for goal success" route: i.e. when we make a function call, we want devs to pass whatever they need to ensure they can achieve their ideal outcome on first try. so.. for `dialProtocol`, having a `runOnLimitedConnection` option available that devs can configure would be good.

achingbrain:
> Though I'm not sure where the "removing" work was happening or what that consisted of.

See the [linked issue](https://github.com/libp2p/js-libp2p/issues/2622#issuecomment-2236785018) mentioned in the op.

> Do we have a best practice or prior art for whether we default API decisions to a "try, analyze, retry" vs "config for goal success" ?

Best practice is generally to avoid network operations wherever possible as I/O is the death of performance.  In this case to make two attempts at opening a stream (and potentially a connection) could be very expensive so should be avoided.

PR #2652, opened 2024-08-07
fix: remove CustomEvent polyfill

### LINKED ISSUE #2420: TypeScript compilation errors within @libp2p/interface/src/event-target.ts
- **Version**:
```
$ npm ls | grep libp2p
├── @chainsafe/libp2p-noise@15.0.0
├── @chainsafe/libp2p-yamux@6.0.2
├── @libp2p/circuit-relay-v2@1.0.15
├── @libp2p/identify@1.0.14
├── @libp2p/mplex@10.0.15
├── @libp2p/webrtc@4.0.19
├── @libp2p/websockets@8.0.15
├── libp2p@1.2.3
```

- **Platform**:
```
$ uname -a
Linux schnippi 5.15.0-92-generic #102~20.04.1-Ubuntu SMP Mon Jan 15 13:09:14 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

- **Subsystem**:
event-target.ts

#### Severity:
Low

#### Description / Steps to reproduce the error:
```
$ tsc
node_modules/@libp2p/interface/src/event-target.ts:94:71 - error TS2345: Argument of type 'CustomEventInit<Detail>' is not assignable to parameter of type 'CustomEventInit<Detail> & EventInit & { detail: Detail; }'.
  Type 'CustomEventInit<Detail>' is not assignable to type '{ detail: Detail; }'.
    Property 'detail' is optional in type 'CustomEventInit<Detail>' but required in type '{ detail: Detail; }'.

94     return this.dispatchEvent(new CustomEvent<Detail>(type as string, detail))
                                                                         ~~~~~~

node_modules/@libp2p/interface/src/event-target.ts:111:5 - error TS2578: Unused '@ts-expect-error' directive.

111     // @ts-expect-error could be undefined
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

EliasOenal:
## Description

The EventTarget implementation of libp2p contained workarounds for incomplete runtime support for the feature. One of the workarounds was specifically marked to be removed once CustomEvent was implemented in NodeJS and the [upstream ticket](https://github.com/nodejs/node/issues/40678) was closed. This has happened two years ago. The implementation of the standard event in Node18 and the subsequent removal of the experimental flag in Node19 broke the current workaround and causes tsc to error. Some of these were suppressed with @ts-ignore, others were not. This fix closes #2420.

The fix removes the workaround as instructed by the source code and restores compatibility with recent versions of Node:
```
/**
 * CustomEvent is a standard event but it's not supported by node.
 *
 * Remove this when https://github.com/nodejs/node/issues/40678 is closed.
 *
 * Ref: https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent
 */
class CustomEventPolyfill<T = any> extends Event {
  /** Returns any custom data event was created with. Typically used for synthetic events. */
  public detail: T

  constructor (message: string, data?: EventInit & { detail: T }) {
    super(message, data)
    // @ts-expect-error could be undefined
    this.detail = data?.detail
  }
}

export const CustomEvent = globalThis.CustomEvent ?? CustomEventPolyfill
```

## Change checklist

- [X] I have performed a self-review of my own code
- [X] I have made corresponding changes to the documentation if necessary (this includes comments as well)
  - I don't believe documentation is needed.
  - I have not been able to setup the test environment on my MacOS, but the tests in my personal libp2p application pass. Please test the fix against the libp2p tests as well.

### DIFF
--- a/packages/interface/src/event-target.ts
+++ b/packages/interface/src/event-target.ts
@@ -105,22 +105,4 @@ export class TypedEventEmitter<EventMap extends Record<string, any>> extends Eve
   }
 }
 
-/**
- * CustomEvent is a standard event but it's not supported by node.
- *
- * Remove this when https://github.com/nodejs/node/issues/40678 is closed.
- *
- * Ref: https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent
- */
-class CustomEventPolyfill<T = any> extends Event {
-  /** Returns any custom data event was created with. Typically used for synthetic events. */
-  public detail: T
-
-  constructor (message: string, data?: EventInit & { detail: T }) {
-    super(message, data)
-    // @ts-expect-error could be undefined
-    this.detail = data?.detail
-  }
-}
-
-export const CustomEvent = globalThis.CustomEvent ?? CustomEventPolyfill
+export const CustomEvent = globalThis.CustomEvent


achingbrain (approved):
`CustomEvent` was added to node.js in `v18.7.0`.

`electron@29.x.x` was the first version to include a node version later than that so an aegir update was needed to test with a more recent Electron, since we were previously using `electron@27.x.x`.

All good now.

PR #2656, opened 2024-08-09
fix!: remove CustomEvent export from `@libp2p/interface`

achingbrain:
`CustomEvent` is global in Node.js as of `v18.7.0`. We support LTS and Current which means `v20+`, and Electron has upgraded to node 20 so the polyfill isn't necessary any more.

BREAKING CHANGE: `@libp2p/interface` no longer exports a `CustomEvent` polyfill

## Change checklist

### DIFF
--- a/packages/interface-compliance-tests/src/transport/listen-test.ts
+++ b/packages/interface-compliance-tests/src/transport/listen-test.ts
@@ -1,5 +1,5 @@
 /* eslint max-nested-callbacks: ["error", 8] */
-import { CustomEvent, TypedEventEmitter } from '@libp2p/interface'
+import { TypedEventEmitter } from '@libp2p/interface'
 import { expect } from 'aegir/chai'
 import drain from 'it-drain'
 import { pipe } from 'it-pipe'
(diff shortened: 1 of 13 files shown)


maschad:
@achingbrain  Should we temporarily disable the interop tests in this PR and then re-enable once gossipsub upgrades it's `@libp2p/interface` version?

achingbrain:
@maschad it shouldn't be necessary - when https://github.com/ChainSafe/js-libp2p-gossipsub/pull/500 is merged gossipsub will use the global `CustomEvent` and the tests should pass.

marcus-pousette:
I am using NodeJS 18.20.4 and get

"CustomEvent is not a constructor" 

error. But works with NodeJS 20. Have some version of 18 been verified to work?

leblowl:
@marcus-pousette [--experimental-global-customevent](https://nodejs.org/docs/latest-v18.x/api/cli.html#--experimental-global-customevent) seems to work for me, but when I import 'node:events' I don't see it on 18.15.0

PR #2658, opened 2024-08-12
docs: add migration guide for v2.0

achingbrain:
Adds instructions for how to upgrade from v1.x to v2.x

## Change checklist


doc/migrations/v1.0.0-v2.0.0.md
```diff
@@ -0,0 +1,245 @@
...
+} catch (err: any) {
+  if (err.name === 'NotFoundError') {
+    // handle the error appropriately
+  }
+}
+```
+
+## pubsub, identify, circuit-relay v2 and record envelope sealing use `privateKey` component
+
+The Peer ID of the current node contains a public/private keypair that is used
```
wemeetagain, line 232: 2.0 could be an opportunity to remove the private key from the peer id entirely

doc/migrations/v1.0.0-v2.0.0.md
```diff
@@ -0,0 +1,245 @@
...
+
+The strategies for peer access vary considerably by protocol. For example
+gossipsub may wish to maintain connections to mesh peers for certain topics in
+order to send/receive messages promptly, but KAD-DHT may just wish to know the
+addresses of a range of peers in the ID space outside of the first KAD-bucket
+and not have a need to be constantly connected to them.
+
+To that end individual protocols can ensure that they have sufficient
+connections for their purposes and libp2p will not try to second-guess their
+rules to apply at a global level.
```
SgtPooki, line 38: Are we planning to encapsulate the auto-dialer, or something else, into a re-usable module that individual protocols can extend to get similar auto-dialer functionality?

doc/migrations/v1.0.0-v2.0.0.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+## "Transient" connections have been renamed "limited"
+
+In libp2p terms, a `transient` connection is one that has limits applied to it
+for how long it's expected to remain open and/or how much data can be
+transferred over it.
+
+The most common place this type of connection will be encountered of this is one
+that is via a [v2 Circuit Relay](https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md).
```
SgtPooki, line 71:
```suggestion
The most common place this type of connection will be encountered is via a [v2 Circuit Relay](https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md).
```

doc/migrations/v1.0.0-v2.0.0.md
```diff
@@ -0,0 +1,245 @@
...
+could be treated as a regular connection.
+
+The intention is for network nodes to run v2 Circuit Relay servers as a public
+good that allows nodes to create direct connections via, for example WebRTC.
+
+To prevent abuse, a relay will normally apply strict time/transfer limits, and
+only have a limited number of relay slots so as to not run as an open relay.
+
+`go-libp2p@0.34.0` included a breaking change of renaming `transient`
+connections as `limited`.
```
SgtPooki, line 84:
```suggestion
`go-libp2p@0.34.0` included a [breaking change of renaming `transient`
connections as `limited`](https://github.com/libp2p/go-libp2p/releases/tag/v0.34.0).
```

doc/migrations/v1.0.0-v2.0.0.md
```diff
@@ -0,0 +1,245 @@
...
+})
+
+// dial a peer
+const conn = await node.dial(multiaddr('/ip4/...'), {
+  signal: AbortSignal.timeout(1000)
+})
+
+// is the connection limited?
+console.info(conn.limits?.seconds) // number | undefined
+console.info(conn.limits?.bytes) // bigint | undefined
```
SgtPooki, line 140: will `conn.limits == null` if there are no limits? we should provide a snippet for devs wanting a similar `conn.transient` check

```suggestion
// is the connection limited?
console.info(conn.limits != null) // true = limited

// if it's limited, how is it limited?
console.info(conn.limits?.seconds) // number | undefined
console.info(conn.limits?.bytes) // bigint | undefined
```

doc/migrations/v1.0.0-v2.0.0.md
SgtPooki: @wemeetagain is there a tracking issue/discussion for removing the private key from the peer ID?

SgtPooki (commented):
a few comments and suggestions for improvement

doc/migrations/v1.0.0-v2.0.0.md
wemeetagain: https://github.com/libp2p/js-libp2p/issues/2659

doc/migrations/v1.0.0-v2.0.0.md
achingbrain: If individual protocols need to ensure a minimum number of connections to random or arbitrary peers then they should do that.  That way it doesn't burden every other protocol/configuration with potentially unnecessary dials.

That said, we can make something like the autodialer available as a service if we discover that it's necessary.

doc/migrations/v1.0.0-v2.0.0.md
achingbrain: #2660

SgtPooki (approved):
lgtm!

PR #2660, opened 2024-08-13
fix!: remove private key field from peer id

### LINKED ISSUE #2659: Remove private key from peer id
in libp2p 1.0, the peer id encapsulates a type, multihash, public key, and a private key.

```ts
export interface PeerId {
  type: PeerIdType
  multihash: MultihashDigest
  privateKey?: Uint8Array
  publicKey?: Uint8Array
  ...
}
```

While conceptually, the peer id is highly related and mostly synonymous with "public key", the same cannot be said for "private key".

With codepaths used 99% of the time, peer ids should never contain a private key.
There is only a single peer id that contains a private key, and that is a node's own peer id.

That said, my opinion is that it's a mistake to combine the private key in the peer id.
Rather, access to a node's own private key should be exposed explicitly (as is already being done in #2303).

2.0 release is an opportunity to remove the private key from peer id.

achingbrain:
The only time you should ever see a private key on a `PeerId` is for the id of the currently running node.  An exception to this is the keychain which used `PeerId`s to export private keys, but a nicer API there is to just deal with `PrivateKey` instances directly.

At runtime a service can obtain the unmarshaled private key by adding a `privateKey: PrivateKey` field to it's components map so there's no need to have the field on every `PeerId`.

The `publicKey` field of a `PeerId` was a `Uint8Array` which (if present) held the public key bytes in a protobuf wrapper along with a field saying what type of key it was.  This was because we wanted to avoid pulling `@libp2p/crypto` into front ends, since it had a dependency on `node-forge` which was a large blob of untreeshakable CSJ code.  This dependency has been removed so `@libp2p/crypto` is now comparatively lightweight so we can use the `PublicKey` type instead of `Uint8Array`, which also saves CPU time since we don't need to unmarshal the key before we can use it.

Fixes #2659

BREAKING CHANGE:
  - The `.privateKey` field of the `PeerId` interface has been removed
  - The `.publicKey` field of the `PeerId` interface is now a `PublicKey` instead of a `Uint8Array`
  - `createLibp2p` now accepts a `privateKey` instead of a `peerId`
  - The keychain operates on `PrivateKey` instances instead of `PeerId`s with `.privateKey` fields
  - `@libp2p/peer-id-factory` has been removed, use `generateKeyPair` and `peerIdFromPrivateKey` instead

## Change checklist


achingbrain:
Opened as draft so we can agree early on the nomenclature.

- Things like `peerIdFromString` return a `PeerId` which is a Ed public key or RSA key hash or what have you
- Functions like `createEd25519PeerId` return a `LocalPeerId` which is a `PeerId` with the `.privateKey` field and are mostly used in tests

I suppose we could change the creation of Peer Ids so you create a keypair, then turn the public key into a PeerId?

packages/interface/src/peer-id/index.ts
```diff
@@ -4,39 +4,73 @@ import type { MultihashDigest } from 'multiformats/hashes/interface'
...
+  toCID(): CID
+  toBytes(): Uint8Array
+  equals(other?: PeerId | Uint8Array | string): boolean
 }
 
-export interface PeerId {
-  type: PeerIdType
-  multihash: MultihashDigest
-  privateKey?: Uint8Array
-  publicKey?: Uint8Array
```
SgtPooki, line 30: Should we take this opportunity to do something that requires a little less maintenance for these very common interfaces?

```ts

export interface LocalPeerIdBase {
  readonly publicKey: Uint8Array
  readonly privateKey: Uint8Array
}

export interface PeerIdBase <pType extends PeerIdType, key extends Uint8Array = Uint8Array> {
  readonly type: pType
  readonly publicKey: key
  readonly multihash: MultihashDigest

  toString(): string
  toCID(): CID
  toBytes(): Uint8Array
  equals(other?: PeerId | Uint8Array | string): boolean
}

export type RSAPeerId = PeerIdBase<'RSA'>
export type LocalRSAPeerId = RSAPeerId & LocalPeerIdBase
export type PeerId = RSAPeerId | Ed25519PeerId | Secp256k1PeerId | URLPeerId
```

[basic demo on typescript playground](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgAoWgSQCYBUCeADigN4C+AUKJLIigLICuANmMABZwDO7AIsAHMIXMMggAPSCGxdkAVWoAOAIJQocfMnJVw0eEjQYoOLZWp66yAMKZeYyRGmyAymCigBpihQmEA9lCi5rQGADJ+CHDM6FjYAELcpBTIyFAQcNh+IMyahIwARszACADSEPgAXPJKqur4yanpmdm57gBucJBlldXgKmoaFJS+AUG6ISgxxvGJyAA8hATE9lIyhrFLEAA0yADW5SuOawp9tRrIALy9YP11AHxaDWkZWTnIYEQQVYufT02vuQKRVK5Sq+3qKWezTeAFsWGxODwqkxWBxuHxBMIwN4UmA-K53CABAAKACUVREhIEDTxNl4ZKqdJpfji+EgXAZ11ugxSEAAjowohy-GB2NAAPxVKYmAA+XLOmjllI85OQ+T8fmY6RAQ28I0C70+yAASs5lNLsJd1tMElwIHMAOSm5QOu4+cT+A0fZbhSLMZ0Wq0BowmABkyF9UQttog3gA9HHkAgsiJkDANVVg7ErSQKMmQKn034qpH-WbA1dc7jPlUnWaHVsGnlCsVulUQBAAO7ygb4YkAbQAuqTGylCO1OhA28gO92TjcFQPh6PkHDUYj2O2uz26kuR8yCR4yVpGmBGFAQMgHeywA7kGQV7TbMeSKfz5fr1i7w-maz2S+3wvGdt3nbk+1Je8V35QVmGFUVoAg180jPID4FglAf0ofMuE1CAADpmD8Eki1JIA)

SgtPooki: idk that we're planning on adding many more peerId types or anything, but it results in less code that seems easier to reason about for me at least. and will be less lines in the future if we need to extend PeerIdBase

SgtPooki:
> I suppose we could change the creation of Peer Ids so you create a keypair, then turn the public key into a PeerId?

I think that's what you and @wemeetagain agreed to on the maintainers' call. 

Is there prior art in go-libp2p or rust-libp2p on how they generate peer IDs? Do we have a need to accept a keypair fully, so we're validating the person providing that public key so we get out of any impersonation attempts, or noise, with duplicate peerIds on the network?

packages/interface/src/peer-id/index.ts
achingbrain: I somewhat did this on purpose.  As suggested the generated API docs will have you clicking through from `PeerId` ->  `Ed25519PeerId` -> `PeerIdBase` to see the docs for the fields on the type.

No-one will ever deal with a `PeerIdBase` in day to day code so I just copied the fields/methods to make the docs simpler.

achingbrain:
> Is there prior art in go-libp2p or rust-libp2p on how they generate peer IDs? Do we have a need to accept the full keypair, so we're validating the person providing that public key so we get out of any impersonation attempts, or noise, with duplicate peerIds on the network?

We don't need the keypair to create a `PeerId`, just the public key (for Ed and secp keys) or the hash of the public key (for RSA keys).

We validate the remote peer has the private key corresponding to the public key during the connection upgrade process.

Duplicate `PeerId` being used by different hosts would cause a problem but fixing that is outside the scope of this PR.

packages/connection-encrypter-tls/src/pb/index.proto
```diff
@@ -3,7 +3,7 @@ syntax = "proto3";
 enum KeyType {
   RSA = 0;
   Ed25519 = 1;
-  Secp256k1 = 2;
+  secp256k1 = 2;
```
wemeetagain, line 6: why?

packages/crypto/src/keys/keys.proto
```diff
@@ -3,7 +3,7 @@ syntax = "proto3";
 enum KeyType {
   RSA = 0;
   Ed25519 = 1;
-  Secp256k1 = 2;
+  secp256k1 = 2;
```
wemeetagain, line 6: why?

packages/connection-encrypter-tls/src/pb/index.proto
achingbrain: It's referred to as `secp256k1` in [it's paper](https://www.secg.org/sec2-v2.pdf), plus where we have it as [a string constant](https://github.com/libp2p/js-libp2p/blob/main/packages/interface/src/keys/index.ts#L37) we have it as "secp256k1" so it's just bringing this in line with that.

On the wire `2` is still `2` so there's no difference.

packages/crypto/src/keys/keys.proto
achingbrain: https://github.com/libp2p/js-libp2p/pull/2660#discussion_r1743241774

packages/connection-encrypter-tls/src/pb/index.proto
wemeetagain: should include it in the breaking changes list

wemeetagain (approved):
LGTM

PR #2671, opened 2024-08-26
fix: connection monitor compatible with other ping implementations

dozyio:
## Description

Update the number of bytes sent by the connection manager so that it is compatible with other `ping` implementations

https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/ping/ping.go#L23
https://github.com/libp2p/rust-libp2p/blob/master/protocols/ping/src/protocol.rs#L48
https://github.com/libp2p/js-libp2p/blob/main/packages/protocol-ping/src/constants.ts#L2

## Change checklist

### DIFF
--- a/packages/libp2p/src/connection-monitor.ts
+++ b/packages/libp2p/src/connection-monitor.ts
@@ -1,3 +1,4 @@
+import { randomBytes } from '@libp2p/crypto'
 import { serviceCapabilities } from '@libp2p/interface'
 import { AdaptiveTimeout } from '@libp2p/utils/adaptive-timeout'
 import { byteStream } from 'it-byte-stream'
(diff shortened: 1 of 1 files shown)


achingbrain:
We send one byte to measure the RTT, there’s no need to send more.

Are you seeing breakage in the wild with this?

achingbrain:
That said, the ping spec does say it should be 32 bytes.

dozyio:
> We send one byte to measure the RTT, there’s no need to send more.
> 
> Are you seeing breakage in the wild with this?

Yes, seeing connection drop to Go peers due to them expecting 32 bytes before replying

achingbrain:
Thanks!

PR #2710, opened 2024-09-20
fix: update P-521 bit length in Web Crypto

twiss:
ECDH over P-521 returns 66 bytes, i.e. 528 bits. Requesting 521 bits causes the last 7 bits of the derived value to be set to 0, which seems unintended (the Node.js version doesn't do this). So, request 66 * 8 bits instead.

Also, remove the superfluous `namedCurve` property in the algorithm identifier passed to `deriveBits` (it's only necessary when importing or generating keys).


achingbrain:
Thanks for opening this PR.  Is there a test you could add that shows how it's currently broken in main?  The tests are run in node, chrome, firefox, etc so this will help prevent regressions in future.

twiss:
Hmhm, done :)

Btw, the comment in the tests said `'P-521' fails in tests :( no clue why`, ironically enough, but they actually passed for me even without this patch. So I've expanded the tests a bit to compare the derived key with a fixture generated in Go as well. That test fails without this patch (in the browser) and succeeds with it.

achingbrain (approved):

achingbrain:
> the comment in the tests said...

Like they say, comments become out of date the moment you write them 😆  

Thanks for updating the tests 🚀

PR #2712, opened 2024-09-22
feat: add the ability to limit inbound/outbound connections

2color:
## What's in this PR

This PR introduces `maxInboundConnections` and `maxOutboundConnections` which replace the `maxConnections` variable which is deprecated by this PR.

While working on this PR, I couldn't find `maxConnections` being respected for outbound connections (neither in openConnection or in dial). So I've added the check in openConnection.

## Change checklist

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -607,7 +607,8 @@ const node = await createLibp2p({
     noise()
   ],
   connectionManager: {
-    maxConnections: Infinity
+    maxInboundConnections: Infinity,
+    maxOutboundConnections: Infinity
   }
 })
 ```
--- a/doc/LIMITS.md
+++ b/doc/LIMITS.md
@@ -35,7 +35,7 @@ const node = await createLibp2p({
     /**
      * The total number of connections allowed to be open at one time
      */
-    maxConnections: 100,
+    maxInboundConnections: 100,
 
     /**
      * How many connections can be open but not yet upgraded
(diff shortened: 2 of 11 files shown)


achingbrain:
Ah, I though we'd covered this briefly face to face while looking at the bootstrapper dashboard.

>  I couldn't find maxConnections being respected for outbound connections (neither in openConnection or in dial). 

This is by design.  The reason we limit connections is to prevent a node from consuming all available resources on the machine either by accident or by being attacked by malicious peers.

The node should be able to make new outgoing connections at all times, this aids recovery from a denial of service attack whereby a set of remote peers could cause a node to reach `maxConnections`.

Once `maxConnections` is exceeded, the connection pruner will [select connections to close](https://github.com/libp2p/js-libp2p/blob/c628c44c588ad7102ce9522594ac888e751f35ba/packages/libp2p/src/connection-manager/connection-pruner.ts#L129) based on connection age, direction, stream count and tag value, so, for example, we would retain connections to KAD-close peers but drop new incoming connections that don't have any active streams open.

This approach also means we can easily control the total number of connections through `maxConnections` and not have to worry about `maxInboundConnections + maxOutboundConnections` creating a much higher value than `maxConnections` would otherwise.

e.g. we want 200 connections total, do we do `maxInboundConnections=100, maxOutboundConnections=100` or `maxInboundConnections=200, maxOutboundConnections=200` - then we effectively have a max of 400, etc.  What if we're expecting to have mostly incoming connections (like a bootstrapper), do we do `maxInboundConnections=190, maxOutboundConnections=10` - what if it's hard to predict and we needed 11.  How do we explain this tradeoff to users, etc.

PR #2713, opened 2024-09-22
feat: add pending connections count to metrics

2color:
## Change checklist

### DIFF
--- a/packages/libp2p/src/connection-manager/index.ts
+++ b/packages/libp2p/src/connection-manager/index.ts
@@ -278,6 +278,13 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
       }
     })
 
+    // track total number of pending inbound connections
+    this.metrics?.registerMetric('libp2p_connection_manager_pending_inbound_connections', {
+      calculate: () => {
+        return this.incomingPendingConnections
+      }
+    })
+
     // track total number of streams per protocol
     this.metrics?.registerMetricGroup('libp2p_protocol_streams_total', {
       label: 'protocol',
--- a/packages/libp2p/src/connection-manager/index.ts
+++ b/packages/libp2p/src/connection-manager/index.ts
@@ -162,6 +162,7 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
   private readonly deny: Multiaddr[]
   private readonly maxIncomingPendingConnections: number
   private incomingPendingConnections: number
+  private outboundPendingConnections: number
   private readonly maxConnections: number
 
   public readonly dialQueue: DialQueue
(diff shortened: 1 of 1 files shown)


achingbrain:
Given that this value can change rapidly between scrapes, do you think this might be better off as a counter rather than a gauge?

It should probably also have the `_total` suffix - see [Metric and Label naming](https://prometheus.io/docs/practices/naming/)

2color:
Counters can't go down. So we can't use them here since pending requests can go both up and down. 

https://prometheus.io/docs/tutorials/understanding_metric_types/#counter

achingbrain:
> Counters can't go down

This is true - but when graphing them, we can use a [rate()](https://prometheus.io/docs/prometheus/latest/querying/functions/#rate) function to show their change over time.  This way we don't miss spikes between scrapes.

2color:
>To pick between counter and gauge, there is a simple rule of thumb: if the value can go down, it is a gauge.
> Counters can only go up (and reset, such as when a process restarts). They are useful for accumulating the number of events, or the amount of something at each event. For example, the total number of HTTP requests, or the total number of bytes sent in HTTP requests. Raw counters are rarely useful. Use the rate() function to get the per-second rate at which they are increasing. https://prometheus.io/docs/practices/instrumentation/#counter-vs-gauge-summary-vs-histogram

If at `t(0)` we have `100` pending connections and at `t(1)` we have `80` pending connections, how can capture that? 

> This way we don't miss spikes between scrapes.

I think catching spikes is a matter of the scrape interval. We could add a counter to track the total incoming connections (prior to upgrade unlike `libp2p_connection_manager_connections` which only counts upgraded connections), but that would still not necessarily help catch spikes between scrapes.

wemeetagain:
Are pending connections only inbound?

We could use a histogram or summary for the # of pending connections observed when a new inbound connection is triggered.

We can also have a histogram/summary to track time from pending to fully connected.

achingbrain:
How about adding the pending connections count to the existing [inbound/outbound connections metric](https://github.com/libp2p/js-libp2p/blob/main/packages/libp2p/src/connection-manager/index.ts#L262-L265) as a label?  

Then it'd graph with those automatically which might make things easier to observe?  This is similar to the "inbound unnegotiated" and "outbound unnegotiated" stream metrics.

wemeetagain:
> How about adding the pending connections count to the existing [inbound/outbound connections metric](https://github.com/libp2p/js-libp2p/blob/main/packages/libp2p/src/connection-manager/index.ts#L262-L265) as a label?

Sounds good to me from a consistency perspective.

2color (commented):
LGTM with the latest changes

PR #2719, opened 2024-09-24
fix: simplify connection upgrade

### LINKED ISSUE #2477: Memory leak issue

### LINKED ISSUE #2537: investigation: hanging TCPSocketWrap in js-libp2p-amino-dht-bootstrapper
related to https://github.com/libp2p/js-libp2p-amino-dht-bootstrapper/issues/18#issuecomment-2101025419

I will be adding various observations here while looking around at connection open and close code in js-libp2p

#### 1. timeout events not handled

https://github.com/libp2p/js-libp2p/blob/4ad63bb79c0c2e5c670b32aa534906b923dcf150/packages/transport-tcp/src/socket-to-conn.ts#L158-L178

When `maConn.close` is called, we try to explicitly close a socket, handling `close` and `error` events, but not `timeout` events which can definitely occur. 

From https://nodejs.org/docs/latest-v20.x/api/net.html#socketsettimeouttimeout-callback

> When an idle timeout is triggered the socket will receive a ['timeout'](https://nodejs.org/docs/latest-v20.x/api/net.html#event-timeout) event but the connection will not be severed. The user must manually call [socket.end()](https://nodejs.org/docs/latest-v20.x/api/net.html#socketenddata-encoding-callback) or [socket.destroy()](https://nodejs.org/docs/latest-v20.x/api/net.html#socketdestroyerror) to end the connection.

This seems like it could easily lead to TCPSocketWrap continuing to increase.

##### Questions

1. By calling `setTimeout` again, does it still allow the callback defined at https://github.com/libp2p/js-libp2p/blob/4ad63bb79c0c2e5c670b32aa534906b923dcf150/packages/transport-tcp/src/socket-to-conn.ts#L62C1-L75C5 to trigger? (it seems like that's the intention).
2. If the original timeout event handler is called, and then we call maConn.close, will that same handler be called again or is it removed?
 
#### 2. Manual `socket.destroySoon` logic
 
https://github.com/libp2p/js-libp2p/blob/4ad63bb79c0c2e5c670b32aa534906b923dcf150/packages/transport-tcp/src/socket-to-conn.ts#L183-L194

I'm not sure if we're intentionally not using `socket.destroySoon` but AFAIK, the only thing we would be missing by switching this code to `socket.destroySoon` is the `log('%s socket drained', lOptsStr)` log line, which we could easily add up above.

#### 3. Socket `'drop'` event

https://github.com/libp2p/js-libp2p/blob/4ad63bb79c0c2e5c670b32aa534906b923dcf150/packages/transport-tcp/test/max-connections.spec.ts#L64-L66

We have some tests that test our maxConnections, but we have no code that handles the `'drop'` event in https://github.com/libp2p/js-libp2p/blob/4ad63bb79c0c2e5c670b32aa534906b923dcf150/packages/transport-tcp/src/socket-to-conn.ts. 

##### Questions

1. Is it possible that when reaching maxConnections & experiencing socket aborts/errors/timeouts, some of those `drop`ped Sockets are not closed out properly?

#### 4. `Socket.unref`

I don't see that we're calling `socket.unref` anywhere, and adding it kind of feels like a hack, but I think there are a few locations `maConn.close` and others where we could call `socket.unref()` to make sure we're explicit about when the socket is no longer needed. 

e.g. once we call `maConn.close` and `'drain'` is emitted, `socket.unref()` could be safely called?

#### 1. Transport class vs toMaConn socket event cleanup

https://github.com/libp2p/js-libp2p/blob/4ad63bb79c0c2e5c670b32aa534906b923dcf150/packages/transport-tcp/src/index.ts#L242-L245

TCP Transport class cleans up events it adds to sockets, why doesn't `socket-to-conn.ts` or `listener.ts`?

##### Questions

1. Should we clean up `allEventListeners` in the TCP transport class? 
2. should socket-to-conn.ts be removing event handlers it adds, and if so, when?
3. listener.ts doesn't remove event handlers either, should it?

achingbrain:
- Unifies inbound/outbound upgrade
- Adds default timeouts for inbound/outbound upgrades and stream protocol negotation
- Ensure tcp listener closes sockets when upgrade times out
- Passes abort signal to connection encrypter
- Uses simple promise rejection to abort tcp connections

Before:

image

After:

image

Fixes #2537
Fixes #2477

## Change checklist

### DIFF
--- a/packages/transport-tcp/src/listener.ts
+++ b/packages/transport-tcp/src/listener.ts
@@ -8,20 +8,8 @@ import {
   type NetConfig
 } from './utils.js'
 import type { TCPCreateListenerOptions } from './index.js'
-import type { ComponentLogger, Logger, LoggerOptions, MultiaddrConnection, Connection, CounterGroup, MetricGroup, Metrics, Listener, ListenerEvents, Upgrader } from '@libp2p/interface'
+import type { ComponentLogger, Logger, MultiaddrConnection, Connection, CounterGroup, MetricGroup, Metrics, Listener, ListenerEvents, Upgrader } from '@libp2p/interface'
 import type { Multiaddr } from '@multiformats/multiaddr'
-
-/**
- * Attempts to close the given maConn. If a failure occurs, it will be logged
- */
-async function attemptClose (maConn: MultiaddrConnection, options: LoggerOptions): Promise<void> {
-  try {
-    await maConn.close()
-  } catch (err: any) {
-    options.log.error('an error occurred closing the connection', err)
-    maConn.abort(err)
-  }
-}
 
 export interface CloseServerOnMaxConnectionsOpts {
   /**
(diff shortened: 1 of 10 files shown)


packages/transport-tcp/src/listener.ts
```diff
@@ -205,69 +193,51 @@ export class TCPListener extends TypedEventEmitter<ListenerEvents> implements Li
...
             })
           }
-
-          this.safeDispatchEvent('connection', { detail: conn })
         })
-        .catch(async err => {
-          this.log.error('inbound connection failed', err)
-          this.metrics?.errors.increment({ [`${this.addr} inbound_upgrade`]: true })
 
-          await attemptClose(maConn, {
```
2color, line 253: Was the source of the problem that `attemptClose` would close without aborting?

achingbrain: Not exactly, it was calling `.close` which throw if closing gracefully fails, and then call `.abort`, but we were doing it in response to an error condition so we should just `.abort` and move on.

As it turns out this hasn't solved the problem completely but it has made it a little better so we're getting closer:


Before:

image

After:

image

I think the bug may be in `@libp2p/tls`, without it the "inbound pending" count stays near 0 the whole time:

image

Issue #2714, opened 2024-09-23
Prevent duplicate connections when dialing a multiaddr without a peer id


achingbrain:
If we dial a multiaddr without a peer id (e.g. `/dns/example.com`), and we exchange peer IDs with the remote, if we already have a non-circuit relay connection to that peer, the connection manager should close the new connection and return the existing connection.

At the moment we end up with multiple connections to the same peer which wastes resources.

tabcat:
Could this be handled by the transports that initiate connections via a relayed connection? They would cleanup the connection after the direct connection is established.

tabcat:
The issue of preventing duplicate connections to the same multiaddr looks like it was handled by https://github.com/libp2p/js-libp2p/commit/0c5957836d1416566f18233f58c92e7db6ab5525#diff-f67bf5a2d5ca417004f9cab5d20bb74d754a6a96cf4e4df6cf540d9a62388090R533-R540 two days after this issue was opened.

I'd be happy to work on a PR for a separate issue that addresses closing indirect connections used to open direct connections.

PR #2734, opened 2024-09-29
fix: prevent duplicate connections to multiaddr

tabcat:
## Title

fix: prevent duplicate connections to multiaddr

## Description

After opening a new connection, check if there is an existing connection
to the same peer. If the existing connection is not limited or the new
connection is also limit, return the existing connection.

This will prevent multiple connections to the same peer in cases where
the multiaddr being dialed does not contain a peer id.

## Notes & open questions

Closes https://github.com/libp2p/js-libp2p/issues/2714

There is a small change from what was requested in the issue. An existing limited connection will be returned if the new connection is also limited.

Should `conn.close()` be awaited before returning the existing connection?

Where should a test for this be added?

## Change checklist

### DIFF
--- a/packages/libp2p/src/connection-manager/dial-queue.ts
+++ b/packages/libp2p/src/connection-manager/dial-queue.ts
@@ -135,11 +135,12 @@ export class DialQueue {
    */
   async dial (peerIdOrMultiaddr: PeerId | Multiaddr | Multiaddr[], options: OpenConnectionOptions = {}): Promise<Connection> {
     const { peerId, multiaddrs } = getPeerAddress(peerIdOrMultiaddr)
+    const { force } = options
 
     // make sure we don't have an existing connection to any of the addresses we
     // are about to dial
     const existingConnection = Array.from(this.connections.values()).flat().find(conn => {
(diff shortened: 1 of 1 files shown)


tabcat:
I've added two tests to check that the existing or new connection is returned appropriately.

achingbrain:
> An existing limited connection will be returned if the new connection is also limited.

In practice this shouldn't happen since the only limited connections we have are circuit relay connections, and we can't dial one without the target peer being in the multiaddr (as the relay needs to know which onward peer we are trying to dial), so the connection should already have been de-duplicated based on the target peer id in the multiaddr and a second relayed connection not opened.

tabcat:
What should be the behaviour when already using an unlimited relayed connection? The original issue mentioned replacing relayed connections directly, not limited connections.

Also the changes here won't actually address the problem the original issue brings up, closing existing relayed-connections used to set up the direct connections. I suggest closing this and making changes where the indirect connection used to open the direction connection is done.

I think the only case in which returning an existing connection via dial is when on the peerId is supplied, which dial already handles. Thoughts?

PR #2763, opened 2024-10-11
fix(@libp2p/tcp): race condition in onSocket

### LINKED ISSUE #2760: race condition causes @libp2p/tcp close method to hang
- **Version**: latest

- **Subsystem**: @libp2p/tcp

#### Description

The hanging promise is to do with shutting down the tcp server

https://github.com/libp2p/js-libp2p/blob/1cbfd6cf392b0db80f744cfe4a0841ed80760c84/packages/transport-tcp/src/listener.ts#L364-L372

For `server.close` to resolve all socket connections must be closed.
The connections are closed inside of the close method.

https://github.com/libp2p/js-libp2p/blob/1cbfd6cf392b0db80f744cfe4a0841ed80760c84/packages/transport-tcp/src/listener.ts#L306-L308

The connections are tracked in the onSocket listener method with the property this.connections.

https://github.com/libp2p/js-libp2p/blob/1cbfd6cf392b0db80f744cfe4a0841ed80760c84/packages/transport-tcp/src/listener.ts#L197-L202

The race condition is triggered when close is called before the connection is upgraded. At that point the socket is not added to this.connections.

https://github.com/tabcat/delay5

tabcat:
## Title

fix(@libp2p/tcp): race condition in onSocket

## Description

This fixes a race condition in the onSocket listener method.
onSocket gets a net.Socket parameter that needs to be closed later
before the tcp server can close.

The net.Socket is used to create a MultiaddrConnection but the
connection is not tracked with this.connections until after it has been
upgraded.

If the tcp listener.close is called while a the MultiaddrConnection is waiting
to be upgraded, then the MultiaddrConnection socket cannot be closed as
it does not exist in this.connections.

Instead of adding the MultiaddrConnection to this.connections on
creation I decided to create a separate property named
this.maConnections. I decided to track the non-upgraded connections
separately because I was unsure how this.connections must be used with other things
like metrics.

## Notes & open questions

Closes #2760

## Change checklist


packages/transport-tcp/src/listener.ts
```diff
@@ -195,11 +196,13 @@ export class TCPListener extends TypedEventEmitter<ListenerEvents> implements Li
     }
 
     this.log('new inbound connection %s', maConn.remoteAddr)
+    this.maConnections.add(maConn)
 
     this.context.upgrader.upgradeInbound(maConn)
       .then((conn) => {
         this.log('inbound connection upgraded %s', maConn.remoteAddr)
         this.connections.add(maConn)
+        this.maConnections.delete(maConn)
```
achingbrain, line 205: The `maConn` is only deleted from the set after a successful upgrade, if the upgrade fails it stays in the set until the transport is closed - this could be many hours, days or months after the transport was started so it introduces a memory leak.

achingbrain (changes requested):
I'm not sure this is necessary.  When we close the transport and call `this.pause(true)` we close the socket server which should destroy all open sockets.  

This is the same as calling `maConn.abort(err)` - the interesting thing it does is call `socket.destroy()` - see `toMultiaddrConnection` for this.

It's possible there's something missing though - could you please add a test to show the problem you're seeing?

achingbrain:
Actually [server.close says](https://nodejs.org/docs/latest/api/net.html#class-netserver):

> Stops the server from accepting new connections and keeps existing connections.

So yes, we'd need to close un-upgraded connections separately, but we should do so in a way that doesn't keep them in the set of sockets even if they fail to upgrade.

tabcat:
They are aborted inside of the catch block of the upgrade call currently. The problem is that maConn is added to this.connections inside of the then block of the upgrade. If close is called before the then block is ran then the connection cannot be closed and this.server.close never resolves.

packages/transport-tcp/src/listener.ts
tabcat: Ah this is a problem; maConn should be removed from maConnections inside of the catch block.

tabcat:
> It's possible there's something missing though - could you please add a test to show the problem you're seeing?

I can add a test. The test that I have been using exists in https://github.com/tabcat/delay5 which should be easily replicated here.

packages/transport-tcp/src/listener.ts
```diff
@@ -304,20 +306,28 @@ export class TCPListener extends TypedEventEmitter<ListenerEvents> implements Li
...
-    this.maConnections.forEach(conn => {
-      conn.abort(err)
+    // synchronously close any open connections - should be done after closing
+    // the server socket in case new sockets are opened during the shutdown
+    this.sockets.forEach(socket => {
+      if (socket.readable) {
+        events.push(pEvent(socket, 'close'))
+        socket.destroy()
+      }
     })
```
tabcat: If the shutdownController is aborted, would this cause the catch block to be called on the upgrade in onSocket? That block removes the socket from this.sockets. Wondering if this could be an issue?

tabcat: Ah i see the maConn is aborted directly after.

tabcat (commented):
Looks much cleaner :)

achingbrain:
Sorry for jumping in, I thought I'd just drag this across the line.

This has shown up a slightly different problem in that we currently close the socket server but wait for the `"close"` event via the callback passed to `this.server.close` - the `"close"` event is only emitted once all connections have closed - if the connections don't close the promise will never resolve which could be what's causing https://github.com/ChainSafe/lodestar/issues/6053

The callback was added in https://github.com/libp2p/js-libp2p/pull/2058 it was not in the original PR that added the "pause" functionality - https://github.com/libp2p/js-libp2p-tcp/pull/218

achingbrain (approved):

PR #2768, opened 2024-10-12
fix: error logging in firefox

achingbrain:
Firefox has an error type with a `.stack` property that is an empty string, so guard on the various properties being empty before logging them.

## Change checklist


wemeetagain (approved):

SgtPooki (approved):
Which error has the empty stack property?

maschad (approved):

achingbrain:
> Which error has the empty stack property?

I've seen it with errors emitted from WebTransport streams.

PR #2769, opened 2024-10-13
fix(mock-connection): use muxer to close/abort streams

### LINKED ISSUE #2675: Mock connection doesn't close stream muxer
It looks like `MockConnection#close` should also close `this.muxer`

tabcat:
## Title

fix(mock-connection): use muxer to close/abort streams

## Description

Uses this.muxer.close and this.muxer.abort in MockConnection's close and abort methods.

## Notes & open questions

Fixes #2675

I looked at https://github.com/libp2p/js-libp2p/commit/a541748c32c5644aa8098c81e96e51940cb25eaa to see if there was a reason this was done but did not see one.

## Change checklist

### DIFF
--- a/packages/interface-compliance-tests/src/mocks/connection.ts
+++ b/packages/interface-compliance-tests/src/mocks/connection.ts
@@ -100,9 +100,7 @@ class MockConnection implements Connection {
 
   async close (options?: AbortOptions): Promise<void> {
     this.status = 'closing'
-    await Promise.all(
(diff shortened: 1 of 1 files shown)


achingbrain:
Thanks for opening this and sorry it's taken so long to get to.

Most of the mocks have been removed since they had their own weird quirks and foibles that complicated development and we can use stubbed interfaces from sinon/sinon-ts instead.

PR #2776, opened 2024-10-22
fix!: make circuit relay listen on addresses like other transports

achingbrain:
Refactors the circuit relay transport to behave like other transports

1. Specify the addresses to listen on in the `addresses.listen` config
2. Remove the `discoverRelays` option
3. Remove the transport manager listen call to just listen on configured addresses
4. Emit the `listen` event when listen addresses change like other transports

This fixes the instances where we miss relay address updates and the corresponding `self:peer:update` event never fires.

BREAKING CHANGE: The `discoverRelays` option has been removed, instead add one or more instances of `"/p2p-circuit"` to the libp2p config under the `addresses.listen` key

## Change checklist

### DIFF
--- a/packages/integration-tests/test/circuit-relay.node.ts
+++ b/packages/integration-tests/test/circuit-relay.node.ts
@@ -20,7 +20,7 @@ import pRetry from 'p-retry'
 import pWaitFor from 'p-wait-for'
 import sinon from 'sinon'
 import { Uint8ArrayList } from 'uint8arraylist'
-import { discoveredRelayConfig, doesNotHaveRelay, getRelayAddress, hasRelay, notUsingAsRelay, usingAsRelay, usingAsRelayCount } from './fixtures/utils.js'
+import { discoveredRelayConfig, doesNotHaveRelay, getRelayAddress, hasRelay, usingAsRelay } from './fixtures/utils.js'
 import type { Libp2p, Connection } from '@libp2p/interface'
 import type { Registrar } from '@libp2p/interface-internal'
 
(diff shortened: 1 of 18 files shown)


packages/transport-circuit-relay-v2/src/errors.ts
```diff
@@ -17,3 +17,27 @@ export class DurationLimitError extends Error {
...
+ * An attempt to open a relayed connection over a relayed connection was made
+ */
+export class DoubleRelayError extends Error {
+  static name: string = 'DoubleRelayError'
+  name: string = 'DoubleRelayError'
+}
+
+/**
+ * An attempt to open a relayed connection over a relayed connection was made
+ */
```
SgtPooki, line 39: It doesn't seem like this is used anywhere, but this is the same description as `DoubleRelayError`

```suggestion
/**
 * An attempt to open a relayed connection when we already have `reservationConcurrency` relay reservations
 */
```

packages/integration-tests/test/circuit-relay.node.ts
```diff
@@ -352,45 +326,32 @@ describe('circuit-relay', () => {
...
-            tcp(),
-            circuitRelayTransport({
-              discoverRelays: 1
-            })
-          ]
+          addresses: {
+            listen: [
+              '/p2p-circuit',
+              '/p2p-circuit',
+              '/p2p-circuit'
```
SgtPooki, line 333: This feels more odd than `discoverRelays: 3`

SgtPooki (commented):
just a quick peek

packages/integration-tests/test/circuit-relay.node.ts
achingbrain: Tbh there’s not a great reason to listen on more than one relay and a number of downsides to it - bigger identify messages, bigger provider records/peer-info objects.

Also given that the number of browser nodes can vastly outnumber publicly dialable server nodes it’s probably not something we should encourage anyway as a small number of relay clients could end up consuming all available relay slots on the network.

packages/transport-circuit-relay-v2/src/errors.ts
achingbrain: It's thrown when the relay reservation queue is full.

PR #2783, opened 2024-10-24
feat: Use CIDR format for connection-manager allow/deny lists

acul71:
@SgtPooki @wemeetagain @dhuseby
## Description

This PR updates the connection manager to treat multiaddrs in the allow/deny lists using the standard IP CIDR format (e.g. `/ip4/52.55.0.0/ipcidr/16`) rather than string prefixes (e.g. `/ip4/52.55`). This allows us to validate multiaddrs accurately and ensures better control over IP address matching.

### Changes:
- Converted the `.allow` and `.deny` properties in the connection manager to use `IpNet[]` for IP range validation.
- Updated tests to use the `IpNet` format for both allow and deny lists.
- This will improve validation, making it more aligned with IP network management best practices by using CIDR blocks.

### Relevant issues:
- Related to: https://github.com/libp2p/js-libp2p/issues/1510.

## Notes & open questions

- How developer will know that they have to use IpNet format in allow and deny list?
- This PR focuses on using the `IpNet` format to ensure the connection manager handles IPs in CIDR notation.
- Any future changes related to this feature would involve expanding IP management logic.

## Change checklist


packages/libp2p/test/connection-manager/index.spec.ts
```diff
@@ -208,7 +208,7 @@ describe('Connection Manager', () => {
         connectionManager: {
           maxConnections: max,
           allow: [
-            '/ip4/83.13.55.32'
+            '/ip4/83.13.55.32/ipcidr/32'
```
SgtPooki, line 211: If `/ip4/83.13.55.32` won't work, we need to release this as a breaking change. I think we should still accept specific IP addresses

packages/libp2p/src/connection-manager/index.ts
```diff
@@ -571,7 +573,7 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
   async acceptIncomingConnection (maConn: MultiaddrConnection): Promise<boolean> {
     // check deny list
     const denyConnection = this.deny.some(ma => {
-      return maConn.remoteAddr.toString().startsWith(ma.toString())
+      return ma.contains(maConn.remoteAddr.nodeAddress().address)
     })
```
SgtPooki, line 577: nit: instead of `ma =>` we should indicate that it's an ipNet

packages/libp2p/src/connection-manager/index.ts
```diff
@@ -581,7 +583,7 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
 
     // check allow list
     const allowConnection = this.allow.some(ma => {
```
SgtPooki, line 585: nit: instead of `ma =>` we should indicate that it's an ipNet

packages/libp2p/src/connection-manager/connection-pruner.ts
```diff
@@ -108,7 +108,7 @@ export class ConnectionPruner {
       this.log('too many connections open - closing a connection to %p', connection.remotePeer)
       // check allow list
       const connectionInAllowList = this.allow.some((ma) => {
```
SgtPooki, line 110: nit: instead of `(ma) =>` we should indicate that it's an ipNet

packages/libp2p/test/connection-manager/index.spec.ts
SgtPooki: @acul71 let me know if this is not clear enough :) we probably just need to check if the supplied `allow`/`deny` is an `ipcidr` multiaddr or not, and if not, append `/ipcidr/32` internally to create the IpNet

acul71: @SgtPooki I was tempted to to the first commented way then I choose to go with the second.
PR soon. Ciao

```typescript
// allow/deny lists
    /**
    this.allow = (init.allow ?? []).map(a => {
      if (a.includes('/ipcidr/')) {
        return convertToIpNet(multiaddr(a))
      } else {
        return convertToIpNet(multiaddr(a + '/ipcidr/32'))
      }
    })
    this.deny = (init.deny ?? []).map((a) => {
      if (a.includes('/ipcidr/')) {
        return convertToIpNet(multiaddr(a))
      } else {
        return convertToIpNet(multiaddr(a + '/ipcidr/32'))
      }
    })
    */
    // allow/deny lists
    this.allow = (init.allow ?? []).map((a) => {
      try {
        // Try to parse a as a multiaddr directly
        return convertToIpNet(multiaddr(a))
      } catch (error) {
        // If parsing fails, append '/ipcidr/32' and try again
        return convertToIpNet(multiaddr(a + '/ipcidr/32'))
      }
    })

    this.deny = (init.deny ?? []).map((a) => {
      try {
        // Try to parse a as a multiaddr directly
        return convertToIpNet(multiaddr(a))
      } catch (error) {
        // If parsing fails, append '/ipcidr/32' and try again
        return convertToIpNet(multiaddr(a + '/ipcidr/32'))
      }
    })
```

packages/libp2p/src/connection-manager/index.ts
```diff
@@ -216,8 +218,25 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
...
-    this.allow = (init.allow ?? []).map(ma => multiaddr(ma))
-    this.deny = (init.deny ?? []).map(ma => multiaddr(ma))
+    this.allow = (init.allow ?? []).map((a) => {
+      try {
+        // Try to parse a as a multiaddr directly
+        return convertToIpNet(multiaddr(a))
+      } catch (error) {
+        // If parsing fails, append '/ipcidr/32' and try again
+        return convertToIpNet(multiaddr(a + '/ipcidr/32'))
+      }
```
SgtPooki, line 228: The current could would try to parse the multiaddr twice for the new (recommended) ipcidr multiaddrs. 
I feel like we could do `a.contains('ipcidr')` to either append `ipcidr/32` or not. 

however, there's probably an `.append` or modify multiaddr method.. i'd need to look deeper

SgtPooki: `.encapsulate` is the method to add to a multiaddr instance

acul71: @SgtPooki  
Is this ok?

```typescript
// allow/deny lists
    this.allow = (init.allow ?? []).map((a) => {
      try {
        // Attempt to parse `a` with the required /ipcidr/32 if missing
        let ma
        if (a.includes('/ipcidr')) {
          ma = multiaddr(a) // Parse directly if it already includes /ipcidr
        } else {
          ma = multiaddr(a).encapsulate('/ipcidr/32') // Encapsulate with /ipcidr/32 if missing
        }

        return convertToIpNet(ma)
      } catch (error) {
        throw new Error(`Invalid multiaddr format in allow list: ${a}`)
      }
    })

    this.deny = (init.deny ?? []).map((a) => {
      try {
        // Attempt to parse `a` with the required /ipcidr/32 if missing
        let ma
        if (a.includes('/ipcidr')) {
          ma = multiaddr(a) // Parse directly if it already includes /ipcidr
        } else {
          ma = multiaddr(a).encapsulate('/ipcidr/32') // Encapsulate with /ipcidr/32 if missing
        }

        return convertToIpNet(ma)
      } catch (error) {
        throw new Error(`Invalid multiaddr format in deny list: ${a}`)
      }
    })
 ```
 
 
 ########
 
 
 Command failed with exit code 1: npm run dep-check


Missing production dependencies:


@chainsafe/netmask

  /home/runner/work/js-libp2p/js-libp2p/packages/libp2p/src/connection-manager/connection-pruner.ts

  /home/runner/work/js-libp2p/js-libp2p/packages/libp2p/src/connection-manager/index.ts

Do I have to add  "@chainsafe/netmask" ?

git diff

diff --git a/package.json b/package.json

index 886a8d17..17e1f070 100644

--- a/package.json

+++ b/package.json

@@ -52,5 +52,8 @@

   ],

   "overrides": {

     "protons-runtime": "^5.5.0"

+  },

+  "dependencies": {

+    "@chainsafe/netmask": "^2.0.0"

   }

 }



This test error seems unrelated to PR: 
Error: #11 2.514 test/fixtures/relay.ts(20,5): error TS2561: Object literal may only specify known properties, but 'connectionEncryption' does not exist in type 'Libp2pOptions<{ identify: Identify; relay: CircuitRelayService; }>'. Did you mean to write 'connectionEncrypters'?

________________________________
UNRELATED:
Could I work on libp2p http transport example https://github.com/libp2p/js-libp2p/issues/2312

Comment (SgtPooki) [reply packages/libp2p/src/connection-manager/index.ts]: > ```ts
> if (a.includes('/ipcidr')) {
> ```

I chatted with @achingbrain about this during Helia WG. Instead of a check on the string, we should construct the multiaddr, `ma = multiaddr(a)` and then check if that instance contains ipcidr with something like [this](https://codepen.io/SgtPooki/pen/LYwgvBe?editors=0011):

```
const ma = multiaddr('/ip4/83.13.55.32')
const maWithCidr = multiaddr('/ip4/83.13.55.32/ipcidr/32')

console.log('ma has ipcidr?', ma.protoNames().includes('ipcidr')) // false
console.log('maWithCidr has ipcidr?', maWithCidr.protoNames().includes('ipcidr')) // true
```

--- 

> Do I have to add  "@chainsafe/netmask" ?

if we directly depend on it, we need to add it to deps in this package.

---

> Could I work on libp2p http transport example https://github.com/libp2p/js-libp2p/issues/2312

I commented on that issue. go for it

Comment (acul71) [reply packages/libp2p/src/connection-manager/index.ts]:
```typescript
// allow/deny lists
this.allow = (init.allow ?? []).map((a) => {
  try {
    let ma = multiaddr(a) // Attempt to parse the multiaddr

    // Check if the multiaddr already includes '/ipcidr'
    if (!ma.protoNames().includes('ipcidr')) {
      ma = ma.encapsulate('/ipcidr/32') // Encapsulate with /ipcidr/32 if missing
    }

    return convertToIpNet(ma)
  } catch (error) {
    throw new Error(`Invalid multiaddr format in allow list: ${a}`)
  }
})

this.deny = (init.deny ?? []).map((a) => {
  try {
    let ma = multiaddr(a) // Attempt to parse the multiaddr

    // Check if the multiaddr already includes '/ipcidr'
    if (!ma.protoNames().includes('ipcidr')) {
      ma = ma.encapsulate('/ipcidr/32') // Encapsulate with /ipcidr/32 if missing
    }

    return convertToIpNet(ma)
  } catch (error) {
    throw new Error(`Invalid multiaddr format in deny list: ${a}`)
  }
})
```

SgtPooki: that looks right to me

acul71:
@SgtPooki @achingbrain 
# Commit History and Changes

This document summarizes the commits made to the repository, detailing the changes and the rationale behind each commit.

## 1. Commit: Add `multiaddrToIpNet` utility function

### Summary:
- Added a new utility function `multiaddrToIpNet` in the `packages/utils/src/multiaddrToIpNet.ts` file.

### Explanation:
- **Purpose**: This function converts a multiaddr string or object into an `IpNet` object. If the multiaddr doesn't already include the `/ipcidr/32` segment, it encapsulates it with `/ipcidr/32` to ensure the correct format for IP network parsing.
  
- **Impact**: This function is used to parse and encapsulate multiaddresses correctly for use in IP-based operations throughout the project.

## 2. Commit: Update `connection-manager` to use `multiaddrToIpNet`

### Summary:
- Refactored `packages/libp2p/src/connection-manager/connection-pruner.ts` to use the newly created `multiaddrToIpNet` function.

### Explanation:
- **Purpose**: The code was modified to replace the previous logic that directly handled multiaddr-to-IP network conversions with the new `multiaddrToIpNet` utility, improving modularity and maintainability.
  
- **Impact**: This change simplifies the conversion process and centralizes logic in the `multiaddrToIpNet` function, reducing duplication across the codebase.

## 3. Commit: Refactor `connection-manager` to handle allow/deny lists

### Summary:
- Refactored the parsing of `allow` and `deny` lists in `packages/libp2p/src/connection-manager/index.ts`.

### Explanation:
- **Purpose**: Simplified the parsing of the `allow` and `deny` lists, moving the logic into a new `parseIpNetList` function. This improves the clarity and maintainability of the code.
  
- **Impact**: The new function reduces duplication and clarifies how multiaddresses are converted to `IpNet` objects within the `DefaultConnectionManager`.

## 4. Commit: Add tests for `allow` and `deny` list parsing

### Summary:
- Added a new test in `packages/libp2p/test/connection-manager/index.spec.ts` to ensure that the `allow` and `deny` lists are correctly parsed and stored as `IpNet` objects in `ConnectionManager`.

### Explanation:
- **Purpose**: This test ensures that the `ConnectionManager` correctly processes multiaddresses from `allow` and `deny` lists and converts them to `IpNet` objects.
  
- **Impact**: Provides confidence that the code correctly handles and stores network information for allowed and denied connections, which is vital for connection management.

## 5. Commit: Fix test for pruning behavior in connection manager

### Summary:
- Fixed a test in `packages/libp2p/test/connection-manager/index.spec.ts` that ensures a connection on the allow list is not closed when pruning connections.

### Explanation:
- **Purpose**: The test now correctly verifies that a connection listed in the `allow` list is not closed when pruning, ensuring expected behavior for connections deemed "safe" to keep alive.
  
- **Impact**: This fix addresses issues with pruning connections and ensures the stability of critical connections within the connection manager.

- **Note**: It took many hours to figure out why the test `should not close connection that is on the allowlist when pruning` was going into a timeout. After extensive debugging, it was discovered that the `remoteAddr` property was missing in the connection object. This caused the `Promise` to never resolve, leading to the timeout. Once the `remoteAddr` was correctly assigned, the test passed as expected.

## 6. Commit: Add `@chainsafe/netmask` dependency

### Summary:
- Added `@chainsafe/netmask` dependency to `packages/libp2p/package.json`.

### Explanation:
- **Purpose**: The new package `@chainsafe/netmask` was added to handle IP network operations such as subnet matching and parsing. This is a key dependency for working with IP networks within the project.
  
- **Impact**: Ensures the availability of necessary tools for IP network parsing and operations, supporting the changes made in `connection-manager`.

## 7. Commit: Update `package.json` to reflect new package installation

### Summary:
- Committed the changes to `package.json` to include the newly added dependency `@chainsafe/netmask`.

### Explanation:
- **Purpose**: This commit finalizes the installation of the `@chainsafe/netmask` package by adding it to the `dependencies` section of `package.json`.

- **Impact**: Ensures that the dependency is correctly tracked for future installations and builds.

---

## Conclusion

These commits primarily focus on enhancing the handling of IP networks within the connection management system, adding utility functions, improving parsing mechanisms, and ensuring correct functionality through tests. These changes improve the overall maintainability, reliability, and flexibility of the codebase in dealing with network-related operations.

packages/utils/src/multiaddrToIpNet.ts
achingbrain, line 1: Please follow the existing conventions in this monorepo.  All file names are kebab-case, this one should be too.

packages/utils/src/multiaddrToIpNet.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Converts a multiaddr string or object to an IpNet object.
+ * If the multiaddr doesn't include /ipcidr/32, it will encapsulate with /ipcidr/32.
+ *
+ * @param {string | Multiaddr} ma - The multiaddr string or object to convert.
+ * @returns {IpNet} The converted IpNet object.
+ * @throws {Error} Throws an error if the multiaddr is not valid.
+ */
+export function multiaddrToIpNet (ma: string | Multiaddr): IpNet {
```
achingbrain, line 14: This function is only used in the `libp2p` module so please define it there in the [src/connection-manager/utils.ts](https://github.com/libp2p/js-libp2p/blob/main/packages/libp2p/src/connection-manager/utils.ts) file.  Please also move the tests to that module.

If we later find that we need it elsewhere we can move it to the utils package, but right now there's no need to expand the public API.

packages/utils/test/multiaddrToIpNet.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+    const ma = '/ip4/127.0.0.1'
+    const ipNet = multiaddrToIpNet(ma)
+    expect(ipNet.toString()).to.equal('127.0.0.1/32')
+  })
+
+  it('should convert a multiaddr with a ipcidr to an IpNet', () => {
+    const ma = '/ip4/127.0.0.1/ipcidr/32'
+    const ipNet = multiaddrToIpNet(ma)
+    expect(ipNet.toString()).to.equal('127.0.0.1/32')
+  })
```
achingbrain, line 17: Please can you add some tests for IPv6 addresses

packages/libp2p/src/connection-manager/index.ts
```diff
@@ -571,7 +583,7 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
   async acceptIncomingConnection (maConn: MultiaddrConnection): Promise<boolean> {
```
achingbrain, line 583: Please can you add some tests that show that incoming connections are denied for multiaddrs in the deny list and allowed for ones in the allow list.

They should show that individual addresses and ranges are handled correctly in both IPv4 and IPv6 formats.

It should be a case of adding more tests like [this one](https://github.com/libp2p/js-libp2p/blob/main/packages/libp2p/test/connection-manager/index.spec.ts#L144-L171) and [this one](https://github.com/libp2p/js-libp2p/blob/main/packages/libp2p/test/connection-manager/index.spec.ts#L75-L91).

packages/utils/src/multiaddrToIpNet.ts
acul71: ok renamed to *multiaddr-to-ipnet.ts*

packages/utils/src/multiaddrToIpNet.ts
acul71: I've moved the multiaddrToIpNet function to the connection-manager/utils.ts file as requested and relocated the corresponding tests. I initially placed it in the general utils package for broader use, but I understand your point about keeping the API focused for now. It's done!

packages/utils/test/multiaddrToIpNet.spec.ts
acul71: Done. 
But 
```typescript
it('should convert a simple IPv6 multiaddr to an IpNet', () => {
    const ma = '/ip6/::1/ipcidr/128'
    const ipNet = multiaddrToIpNet(ma)
    expect(ipNet.toString()).to.equal('0000:0000:0000:0000:0000:0000:0000:0001/128')
  })
```
Don't work it gives error so I removed it, **but it should be a basic test it shouldn't fail**

I'm not expert in ip6 but look at this examples: is there a bug in @multiformats/multiaddr ?

```javascript
import { multiaddr} from '@multiformats/multiaddr'

function test (a) {
  const ma = multiaddr(a)
  console.log(a, ma.toString())
}

test('/ip6/0:0:0:0:0:0:0:1') // /ip6/::1
test('/ip6/::1') // /ip6/::1
test('/ip6/2001:db8:3333:4444:5555:6666:7777:8888/ipcidr/99') // /ip6/2001:db8:3333:4444:5555:6666:7777:8888/ipcidr/99
test('/ip6/2001:db8::/ipcidr/99') // /ip6/2001:db8::/ipcidr/99
test('/ip6/2001:db8::/ipcidr/100')// throw new SyntaxError('Unexpected end of data')
test('/ip6/::1/ipcidr/128') // throw new SyntaxError('Unexpected end of data')
test('/ip6/0000:0000:0000:0000:0000:0000:0000:0001/ipcidr/128') // throw new SyntaxError('Unexpected end of data')
test('/ip6/0:0:0:0:0:0:0:1/ipcidr/128') // throw new SyntaxError('Unexpected end of data')
```

```bash
node test.mjs 
/ip6/0:0:0:0:0:0:0:1 /ip6/::1
/ip6/::1 /ip6/::1
/ip6/2001:db8:3333:4444:5555:6666:7777:8888/ipcidr/99 /ip6/2001:db8:3333:4444:5555:6666:7777:8888/ipcidr/99
/ip6/2001:db8::/ipcidr/99 /ip6/2001:db8::/ipcidr/99
file:///home/luca/Informatica/Learning/PNL_Launchpad_Curriculum/Libp2p/tmp/test-multiformats/node_modules/multiformats/dist/src/bases/base.js:156
        throw new SyntaxError('Unexpected end of data');
              ^

SyntaxError: Unexpected end of data
    at decode (file:///home/luca/Informatica/Learning/PNL_Launchpad_Curriculum/Libp2p/tmp/test-multiformats/node_modules/multiformats/dist/src/bases/base.js:156:15)
    at Decoder.decode [as baseDecode] (file:///home/luca/Informatica/Learning/PNL_Launchpad_Curriculum/Libp2p/tmp/test-multiformats/node_modules/multiformats/dist/src/bases/base.js:199:20)
    at Decoder.decode (file:///home/luca/Informatica/Learning/PNL_Launchpad_Curriculum/Libp2p/tmp/test-multiformats/node_modules/multiformats/dist/src/bases/base.js:52:25)
    at fromString (file:///home/luca/Informatica/Learning/PNL_Launchpad_Curriculum/Libp2p/tmp/test-multiformats/node_modules/uint8arrays/dist/src/from-string.node.js:20:25)
    at convertToBytes (file:///home/luca/Informatica/Learning/PNL_Launchpad_Curriculum/Libp2p/tmp/test-multiformats/node_modules/@multiformats/multiaddr/dist/src/convert.js:100:20)
    at stringToMultiaddrParts (file:///home/luca/Informatica/Learning/PNL_Launchpad_Curriculum/Libp2p/tmp/test-multiformats/node_modules/@multiformats/multiaddr/dist/src/codec.js:44:23)
    at new Multiaddr (file:///home/luca/Informatica/Learning/PNL_Launchpad_Curriculum/Libp2p/tmp/test-multiformats/node_modules/@multiformats/multiaddr/dist/src/multiaddr.js:58:21)
    at multiaddr (file:///home/luca/Informatica/Learning/PNL_Launchpad_Curriculum/Libp2p/tmp/test-multiformats/node_modules/@multiformats/multiaddr/dist/src/index.js:195:12)
    at test (file:///home/luca/Informatica/Learning/PNL_Launchpad_Curriculum/Libp2p/tmp/test-multiformats/test.mjs:4:14)
    at file:///home/luca/Informatica/Learning/PNL_Launchpad_Curriculum/Libp2p/tmp/test-multiformats/test.mjs:12:1

Node.js v22.10.0

```

acul71: In [@multifomats/multiaddr/src/convert.ts](https://github.com/multiformats/js-multiaddr/blob/d4164fbce730d3b43a10682b651bcc96db845e7c/src/convert.ts)

```typescript
export function convertToBytes (proto: string | number, str: string): Uint8Array {
  const protocol = getProtocol(proto)
  switch (protocol.code) {
    case 4: // ipv4
      return ip2bytes(str)
    case 41: // ipv6
      return ip2bytes(str)
    case 42: // ipv6zone
      return str2bytes(str)

    case 6: // tcp
    case 273: // udp
    case 33: // dccp
    case 132: // sctp
      return port2bytes(parseInt(str, 10))

    case 53: // dns
    case 54: // dns4
    case 55: // dns6
    case 56: // dnsaddr
    case 400: // unix
    case 449: // sni
    case 777: // memory
      return str2bytes(str)

    case 421: // ipfs
      return mh2bytes(str)
    case 444: // onion
      return onion2bytes(str)
    case 445: // onion3
      return onion32bytes(str)
    case 466: // certhash
      return mb2bytes(str)
    case 481: // http-path
      return str2bytes(globalThis.decodeURIComponent(str))
    default:
      return uint8ArrayFromString(str, 'base16') // no clue. convert from hex
  }
}```

There is no case 43 (ipcidr) so in the default case 

The issue arises because the string 100 (or >99) is being interpreted as hexadecimal when passed to uint8ArrayFromString with the base16 encoding. Hexadecimal (base16) encoding expects strings consisting of characters from 0-9 and a-f. In base16, each character represents 4 bits, so the total number of characters must be even to decode into full bytes.

If I add 

```typescript
case 43: // ipcidr
      console.log('convertToBytes case 43:', str)
      str = str.length % 2 !== 0 ? '0' + str : str
      console.log('convertToBytes uint8ArrayFromString', uint8ArrayFromString(str, 'base16'))
      return uint8ArrayFromString(str, 'base16')
```

MarcoPolo: I'm a bit confused by this thread, but this may help:

The CIDR mask is limited to 0-32 for ipv4 and 0-128 for ipv6. When you convert the string form of the CIDR mast to bytes, you only need a single byte to represent all options. So the string to byte logic should convert the given decimal string number to a byte. And vice versa for the bytes to string part. See the go-multiaddr implementation for an example:

https://github.com/multiformats/go-multiaddr/pull/177/files#diff-38293d4bb2018b8660c43e7007e16a293752d0205fdc660558c419d6d3d199a6R59-R72

acul71: @MarcoPolo 
Make sense. But the existing code is expecting a uint8Array so I was trying to change the less code possibile. Maybe I have to go take a deeper look...

acul71: @SgtPooki @achingbrain @MarcoPolo 

https://github.com/multiformats/js-multiaddr/pull/394

packages/libp2p/src/connection-manager/connection-pruner.ts
```diff
@@ -29,13 +31,22 @@ export class ConnectionPruner {
...
+    this.allow = (init.allow ?? []).map((ma) => {
+      try {
+        if (!ma.protoNames().includes('ipcidr')) {
+          ma = ma.encapsulate('/ipcidr/32') // Encapsulate with /ipcidr/32 if missing
+        }
+        return convertToIpNet(ma)
+      } catch (error) {
+        throw new Error(`Invalid multiaddr format in allow list: ${ma}`)
+      }
+    })
```
achingbrain, line 49: This is redundant because `multiaddrToIpNet` adds the `CIDR` tuple if it's missing, and handles IPv6 as well:

```suggestion
    this.allow = (init.allow ?? []).map(ma => multiaddrToIpNet(ma))
```

packages/libp2p/src/connection-manager/connection-pruner.ts
```diff
@@ -107,8 +118,8 @@ export class ConnectionPruner {
     for (const connection of sortedConnections) {
       this.log('too many connections open - closing a connection to %p', connection.remotePeer)
       // check allow list
-      const connectionInAllowList = this.allow.some((ma) => {
-        return connection.remoteAddr.toString().startsWith(ma.toString())
+      const connectionInAllowList = this.allow.some((ipNet) => {
+        return ipNet.contains(connection.remoteAddr.nodeAddress().address)
```
achingbrain, line 122: Prefer `toOptions` because `nodeAddress` will throw if the protocol isn't `tcp` or `udp`:

```suggestion
        return ipNet.contains(connection.remoteAddr.toOptions().host)
```

packages/libp2p/src/connection-manager/index.ts
```diff
@@ -237,7 +239,7 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
       logger: components.logger
     }, {
       maxConnections: this.maxConnections,
-      allow: this.allow
+      allow: init.allow != null ? init.allow.map(a => multiaddr(a)) : undefined
```
achingbrain, line 242: Slightly more concise:

```suggestion
      allow: init.allow?.map(a => multiaddr(a))
```

packages/libp2p/src/connection-manager/connection-pruner.ts
```diff
@@ -1,6 +1,8 @@
 import { PeerMap } from '@libp2p/peer-collections'
 import { safelyCloseConnectionIfUnused } from '@libp2p/utils/close'
+import { convertToIpNet } from '@multiformats/multiaddr/convert'
 import { MAX_CONNECTIONS } from './constants.js'
```
achingbrain, line 4: Use `multiaddrToIpNet` instead:

```suggestion
import { MAX_CONNECTIONS } from './constants.js'
import { multiaddrToIpNet } from './utils.js'
```

packages/libp2p/src/connection-manager/index.ts
```diff
@@ -216,8 +218,8 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
     this.onDisconnect = this.onDisconnect.bind(this)
 
     // allow/deny lists
-    this.allow = (init.allow ?? []).map(ma => multiaddr(ma))
-    this.deny = (init.deny ?? []).map(ma => multiaddr(ma))
+    this.allow = init.allow != null ? this.parseIpNetList(init.allow) : []
+    this.deny = init.deny != null ? this.parseIpNetList(init.deny) : []
```
achingbrain, line 222: The functionality from `this.parseIpNetList` is encapsulated in `multiaddrToIpNet` so let's just use that:

```suggestion
    this.allow = (init.allow ?? []).map(str => multiaddrToIpNet(str))
    this.deny = (init.allow ?? []).map(str => multiaddrToIpNet(str))
```

packages/libp2p/src/connection-manager/index.ts
```diff
@@ -265,6 +267,30 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
...
+        } else {
+          ma = multiaddr(a).encapsulate('/ipcidr/32') // Encapsulate with /ipcidr/32 if missing
+        }
+        return multiaddrToIpNet(ma)
+      } catch (error) {
+        throw new Error(`Invalid multiaddr format in list: ${a}`)
+      }
+    })
+  }
+
```
achingbrain, line 293: Not needed since `multiaddrToIpNet` does all this.

```suggestion

```

packages/libp2p/src/connection-manager/index.ts
```diff
@@ -575,7 +601,7 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
   async acceptIncomingConnection (maConn: MultiaddrConnection): Promise<boolean> {
     // check deny list
     const denyConnection = this.deny.some(ma => {
-      return maConn.remoteAddr.toString().startsWith(ma.toString())
+      return ma.contains(maConn.remoteAddr.nodeAddress().address)
```
achingbrain, line 604: Prefer `toOptions` because `nodeAddress` will throw if the protocol isn't `tcp` or `udp`:

```suggestion
      return ma.contains(maConn.remoteAddr.toOptions().host)
```

packages/libp2p/src/connection-manager/index.ts
```diff
@@ -584,8 +610,8 @@ export class DefaultConnectionManager implements ConnectionManager, Startable {
     }
 
     // check allow list
-    const allowConnection = this.allow.some(ma => {
-      return maConn.remoteAddr.toString().startsWith(ma.toString())
+    const allowConnection = this.allow.some(ipNet => {
+      return ipNet.contains(maConn.remoteAddr.nodeAddress().address)
```
achingbrain, line 614: Prefer `toOptions` because `nodeAddress` will throw if the protocol isn't `tcp` or `udp`:

```suggestion
      return ipNet.contains(maConn.remoteAddr.toOptions().host)
```

packages/libp2p/src/connection-manager/utils.ts
```diff
@@ -28,3 +29,35 @@ export async function resolveMultiaddrs (ma: Multiaddr, options: ResolveOptions
...
+    let parsedMa: Multiaddr
+    if (typeof ma === 'string') {
+      parsedMa = multiaddr(ma)
+    } else {
+      parsedMa = ma
+    }
+
+    // Check if /ipcidr is already present
+    if (!parsedMa.protoNames().includes('ipcidr')) {
+      const isIPv6 = parsedMa.protoNames().includes('ip6')
```
achingbrain, line 54: Can avoid creating the proto names array twice, also just check the first proto for `ip6` to handle future craziness like `/ip4/.../tcp/1234/tunnel/ip6/.../udp/123/p2p/Qmfoo`:

```suggestion
    const protoNames = parsedMa.protoNames()

    // Check if /ipcidr is already present
    if (!protoNames.includes('ipcidr')) {
      const isIPv6 = protoNames[0] === 'ip6'
```

packages/libp2p/test/connection-manager/connection-pruner.spec.ts
```diff
@@ -219,6 +219,39 @@ describe('connection-pruner', () => {
...
+      {
+        mask: new Uint8Array([255, 255, 255, 0]),
+        network: new Uint8Array([192, 168, 1, 0])
+      }
+    ]
+
+    // Verify that the allow list in the pruner matches the expected IpNet objects
+    // eslint-disable-next-line @typescript-eslint/dot-notation
+    expect(pruner['allow']).to.deep.equal(expectedAllowList)
+  })
```
achingbrain, line 253: Add an IPv6 value

```suggestion
  it('should correctly parse and store allow list as IpNet objects in ConnectionPruner', () => {
    const mockInit = {
      allow: [
        multiaddr('/ip4/83.13.55.32/ipcidr/32'),
        multiaddr('/ip4/83.13.55.32'),
        multiaddr('/ip4/192.168.1.1/ipcidr/24'),
        multiaddr('/ip6/2001::0/ipcidr/64')
      ]
    }

    // Create a ConnectionPruner instance
    const pruner = new ConnectionPruner(components, mockInit)

    // Expected IpNet objects for comparison
    const expectedAllowList = [
      {
        mask: new Uint8Array([255, 255, 255, 255]),
        network: new Uint8Array([83, 13, 55, 32])
      },
      {
        mask: new Uint8Array([255, 255, 255, 255]),
        network: new Uint8Array([83, 13, 55, 32])
      },
      {
        mask: new Uint8Array([255, 255, 255, 0]),
        network: new Uint8Array([192, 168, 1, 0])
      },
      {
        mask: new Uint8Array([255, 255, 255, 255, 255, 255, 255, 255, 0, 0, 0, 0, 0, 0, 0, 0]),
        network: new Uint8Array([32, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
      }
    ]

    // Verify that the allow list in the pruner matches the expected IpNet objects
    // eslint-disable-next-line @typescript-eslint/dot-notation
    expect(pruner['allow']).to.deep.equal(expectedAllowList)
  })
```

packages/libp2p/test/connection-manager/multiaddr-to-ipnet.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  /*
+  // TODO: Re-enable when check is implemented
+  it('should throw an error for invalid CIDR value in IPv4 and IPv6 multiaddr', () => {
+    const invalidIPv6 = '/ip6/2001:db8::1/ipcidr/256'
+    const invalidIPv4 = '/ip4/192.168.1.1/ipcidr/33'
+
+    expect(() => multiaddrToIpNet(invalidIPv6)).to.throw(Error, 'Invalid CIDR value')
+    expect(() => multiaddrToIpNet(invalidIPv4)).to.throw(Error, 'Invalid CIDR value')
+  })
+  */
```
achingbrain, line 51: Can be enabled now:

```suggestion
  it('should throw an error for invalid CIDR value in IPv4 and IPv6 multiaddr', () => {
    const invalidIPv6 = '/ip6/2001:db8::1/ipcidr/256'
    const invalidIPv4 = '/ip4/192.168.1.1/ipcidr/33'

    expect(() => multiaddrToIpNet(invalidIPv6)).to.throw(Error, 'Invalid multiaddr format')
    expect(() => multiaddrToIpNet(invalidIPv4)).to.throw(Error, 'Invalid multiaddr format')
  })
```

achingbrain (changes requested):
LGTM, comments inline

packages/libp2p/src/connection-manager/connection-pruner.ts
acul71: of course can't understand how it happened maybe because too many weeks has passed, and I remember having troubles with with git

packages/libp2p/src/connection-manager/index.ts
acul71: Shouldn't line 222 be  this.deny = (init.deny ?? []).map(str => multiaddrToIpNet(str)) 
not init.allow ?

packages/libp2p/src/connection-manager/utils.ts
acul71: Yep! 😄

packages/libp2p/src/connection-manager/index.ts
achingbrain: Yes, good spot.

achingbrain (approved):

achingbrain:
Thanks for your patience with this!

PR #2785, opened 2024-10-24
feat: add reprovide

achingbrain:
Adds a reprovide queue to kad-dht to republish provider records two hours before they would expire.

Adds a function to the content routing interface to cancel the reprovide.

## Change checklist


packages/kad-dht/test/rpc/handlers/get-value.spec.ts
```diff
@@ -79,7 +79,7 @@ describe('rpc - handlers - GetValue', () => {
     const value = uint8ArrayFromString('world')
     const record = new Libp2pRecord(key, value, new Date())
 
-    await datastore.put(utils.bufferToRecordKey(key), record.serialize().subarray())
+    await datastore.put(utils.bufferToRecordKey('//record', key), record.serialize().subarray())
```
SgtPooki, line 82: Is it expected that a prefix would look like this? Can we use something here that better reflects what this might be in the wild? Some people might look through tests to see how things work, and we should set them up for success.

packages/kad-dht/test/rpc/handlers/put-value.spec.ts
```diff
@@ -75,7 +75,7 @@ describe('rpc - handlers - PutValue', () => {
     const response = await handler.handle(sourcePeer, msg)
     expect(response).to.deep.equal(msg)
 
-    const key = utils.bufferToRecordKey(uint8ArrayFromString('hello'))
+    const key = utils.bufferToRecordKey('//record', uint8ArrayFromString('hello'))
```
SgtPooki, line 78: ditto to the above

packages/kad-dht/test/utils/create-peer-id.ts
```diff
@@ -2,14 +2,35 @@ import { generateKeyPair } from '@libp2p/crypto/keys'
...
+}
+
+/**
+ * Creates a PeerId
+ */
+export async function createPeerId (): Promise<PeerId> {
+  const ids = await createPeerIds(1)
+
+  return ids[0]
+}
```
SgtPooki, line 24:
```suggestion
/**
 * Creates multiple PeerIds
 */
export async function createPeerIds (length: number): Promise<PeerId[]> {
  return Promise.all(
    new Array(length).fill(0).map(async () => createPeerId())
  )
}

/**
 * Creates a PeerId
 */
export async function createPeerId (): Promise<PeerId> {
  const privateKey = await generateKeyPair('Ed25519')
  return peerIdFromPrivateKey(privateKey)
}
```

packages/kad-dht/test/utils/create-peer-id.ts
```diff
@@ -2,14 +2,35 @@ import { generateKeyPair } from '@libp2p/crypto/keys'
...
 export type PeerIdWithPrivateKey = PeerId & {
   privateKey: PrivateKey
 }
 
 /**
- * Creates multiple PeerIds
+ * Creates multiple PeerIds with private keys
  */
-export async function createPeerIds (length: number): Promise<PeerIdWithPrivateKey[]> {
+export async function createPeerIdsWithPrivateKey (length: number): Promise<PeerIdWithPrivateKey[]> {
```
SgtPooki, line 33: Ditto for encapsulating the logic and calling the singular function instead..

SgtPooki (commented):
a few comments

packages/kad-dht/test/providers.spec.ts
```diff
@@ -39,13 +36,16 @@ describe('Providers', () => {
 
     const provs = await providers.getProviders(cid)
     const ids = new Set(provs.map((peerId) => peerId.toString()))
-    expect(ids.has(peers[0].toString())).to.be.eql(true)
+    expect(ids.has(peers[0].toString())).to.equal(true)
   })
 
-  it('duplicate add of provider is deduped', async () => {
+  it('should duplicate multiple adds of same provider', async () => {
```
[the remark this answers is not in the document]
achingbrain: Fixex

packages/kad-dht/test/rpc/handlers/get-value.spec.ts
achingbrain: I've filled out the log prefix

packages/kad-dht/test/utils/create-peer-id.ts
achingbrain: Fixed

packages/kad-dht/src/reprovider.ts
```diff
@@ -0,0 +1,240 @@
...
+
+      // Commit the deletes to the datastore
+      if (deleted.size > 0) {
+        this.log('deleting %d / %d entries', deleteCount, count)
+        await batch.commit()
+      } else {
+        this.log('nothing to delete')
+      }
+
+      this.log('Cleanup successful (%dms)', Date.now() - start)
```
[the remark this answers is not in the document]
achingbrain: The actual re-providing is done in a queue which has metrics for it's length, in-flight tasks, etc. I think this should be good enough to start with, we can always add more in future?

packages/kad-dht/src/providers.ts
```diff
@@ -1,292 +1,104 @@
...
-   */
-  cleanupInterval?: number
-  /**
-   * How long is a provider valid for. (in seconds)
-   *
-   * @default 86400
-   */
-  provideValidity?: number
+  logPrefix: string
+  lock: Mortice
```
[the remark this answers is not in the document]
achingbrain: Good shout - I've added top-level config to do this that has no magic.

packages/kad-dht/src/constants.ts
```diff
@@ -15,15 +15,21 @@ export const MAX_RECORD_AGE = 36 * hour
 
 export const PROTOCOL = '/ipfs/kad/1.0.0'
 
-export const RECORD_KEY_PREFIX = '/dht/record'
+export const PROVIDERS_VALIDITY = 24 * hour
```
2color, line 18: A bit late to the party, but isn't this 48 hours 

See https://github.com/libp2p/go-libp2p-kad-dht/blob/18a758c0806da2525d7d4d7fecf8c2cf7cf12e0a/providers/providers_manager.go#L38

achingbrain: It would seem so, can haz PR?

packages/kad-dht/src/reprovider.ts
```diff
@@ -0,0 +1,226 @@
...
+      this.log.trace('re-providing %c', cid)
+
+      // use adaptive timeout
+      const signal = this.reprovideTimeout.getTimeoutSignal(options)
+
+      try {
+        // reprovide
+        await this.reprovide(options.cid, options)
+      } finally {
+        this.reprovideTimeout.cleanUp(signal)
```
2color, line 209: This runs irrespective of whether the reprovide for the CID succeeds. Is that a problem?

PR #2798, opened 2024-11-02
feat: add auto-tls service

achingbrain:
Adds an optional service that requests a Let's Encrypt-style TLS certificate when publicly dialable addresses are detected.

This will allow transports such as WebSockets to upgrade themselves to be the secure version.

## Change checklist


packages/auto-tls/README.md
```diff
@@ -0,0 +1,105 @@
...
+    webSockets()
+  ],
+  connectionEncrypters: [
+    noise()
+  ],
+  streamMuxers: [
+    yamux()
+  ],
+  services: {
+    autoTLS: autoTLS(),
```
achingbrain, line 66: Need to add identify as per instructions

packages/auto-tls/src/auto-tls.ts
```diff
@@ -0,0 +1,200 @@
...
+  private readonly fetchCertificates: DebouncedFunction
+  private renewTimeout?: ReturnType<typeof setTimeout>
+
+  constructor (components: AutoTLSComponents, init: AutoTLSInit = {}) {
+    this.log = components.logger.forComponent('libp2p:certificate-manager')
+    this.addressManager = components.addressManager
+    this.privateKey = components.privateKey
+    this.peerId = components.peerId
+    this.events = components.events
+    this.forgeEndpoint = init.forgeEndpoint ?? 'registration.libp2p.direct'
```
2color, line 47: Endpoint implies a URL with path, no? 

Also suggestion to move these defaults into a constants file.

achingbrain: The path is `/v1/_acme-challenge` - I omitted the path here to prevent people specifying `/v2/_acme-challenge` and expecting it to work.

We could make them add `https://` on the front if you think it's more intuitive, otherwise I guess we should add a note that the `/v1/..` bit will be appended?

packages/auto-tls/src/auto-tls.ts
```diff
@@ -0,0 +1,200 @@
...
+  private async fetchCertificate (mulitaddrs: Multiaddr[], options?: AbortOptions): Promise<void> {
+    this.log('fetching certificate')
+
+    // TODO: handle rate limit errors like "too many new registrations (10) from this IP address in the last 3h0m0s, retry after 2024-11-01 09:22:38 UTC: see https://letsencrypt.org/docs/rate-limits/#new-registrations-per-ip-address"
+
+    const base36EncodedPeer = base36.encode(this.peerId.toCID().bytes)
+    const domain = `${base36EncodedPeer}.${this.forgeDomain}`
+
+    // Create CSR
+    const [certificatePrivateKey, csr] = await acmeClient.forge.createCsr({
```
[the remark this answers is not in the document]
achingbrain: No reason, just stopped at the first thing that worked.  TBH I wanted to use [@peculiar/acme-client](https://www.npmjs.com/package/@peculiar/acme-client) as we have other `@peculiar/...` deps in the tree but couldn't get it to work in the time I was looking at it.

packages/auto-tls/src/auto-tls.ts
```diff
@@ -0,0 +1,200 @@
...
+      commonName: domain,
+      altNames: []
+    })
+
+    const accountPrivateKey = await acmeClient.forge.createPrivateKey()
```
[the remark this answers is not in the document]
achingbrain: I couldn't see a "renew"-type operation in the API so I guess not, unless I've missed something.

I do want to add storing the cert/key in the datastore so we can re-use it after a restart as they are quite long-lived.

packages/auto-tls/src/auto-tls.ts
```diff
@@ -0,0 +1,200 @@
...
+      })
+  }
+
+  private async fetchCertificate (mulitaddrs: Multiaddr[], options?: AbortOptions): Promise<void> {
+    this.log('fetching certificate')
+
+    // TODO: handle rate limit errors like "too many new registrations (10) from this IP address in the last 3h0m0s, retry after 2024-11-01 09:22:38 UTC: see https://letsencrypt.org/docs/rate-limits/#new-registrations-per-ip-address"
+
+    const base36EncodedPeer = base36.encode(this.peerId.toCID().bytes)
+    const domain = `${base36EncodedPeer}.${this.forgeDomain}`
```
achingbrain, line 122: Need to be a wildcard:

```suggestion
    const domain = `*.${base36EncodedPeer}.${this.forgeDomain}`
```

packages/auto-tls/src/auto-tls.ts
2color: Yeah note also works.

packages/auto-tls/src/auto-tls.ts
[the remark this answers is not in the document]
2color: Indeed, there's isn't a a renew operation. It's the same process, except that you can reuse the CSR and account key (IIUC).

packages/auto-tls/src/auto-tls.ts
```diff
@@ -0,0 +1,200 @@
...
+            'Content-Type': 'application/json'
+          },
+          body: JSON.stringify({
+            value: keyAuthorization,
+            addresses
+          }),
+          ...options
+        })
+
+        if (!response.ok) {
```
SgtPooki, line 157: Do we want to allow all 2xx responses?

packages/auto-tls/src/auto-tls.ts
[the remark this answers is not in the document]
SgtPooki: We should definitely load a cert on start if we've already got one.

packages/auto-tls/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+   * @default 5000
+   */
+  delay?: number
+
+  /**
+   * How long before the expiry of the certificate to renew it in ms
+   *
+   * @default 60000
+   */
+  renewThreshold?: number
```
SgtPooki, line 117: This situation is a little different than domain cert renewal, but cert renewal should allow for resolving any errors that pop up when a renewal is attempted. Should we set the renewal time to a full day or week before expiry instead of only one minute?

packages/interface/src/index.ts
```diff
@@ -267,6 +279,16 @@ export interface Libp2pEvents<T extends ServiceMap = ServiceMap> {
    */
   'connection:close': CustomEvent<Connection>
 
+  /**
+   * This event notifies listeners that a TLS certificate is available for use
```
SgtPooki, line 283:
```suggestion
   * This event notifies listeners that a TLS certificate is available for use for the first time
```

packages/interface/src/index.ts
```diff
@@ -267,6 +279,16 @@ export interface Libp2pEvents<T extends ServiceMap = ServiceMap> {
...
+   */
+  'certificate:provision': CustomEvent<TLSCertificate>
+
+  /**
+   * This event notifies listeners that a TLS certificate is available for use
```
SgtPooki, line 288:
```suggestion
   * This event notifies listeners that a new TLS certificate is available for use
```

packages/auto-tls/src/auto-tls.ts
```diff
@@ -0,0 +1,200 @@
...
+import type { AutoTLSComponents, AutoTLSInit, AutoTLS as AutoTLSInterface } from './index.js'
+import type { PeerId, PrivateKey, Logger, TypedEventTarget, Libp2pEvents, AbortOptions, TLSCertificate } from '@libp2p/interface'
+import type { AddressManager } from '@libp2p/interface-internal'
+import type { DebouncedFunction } from '@libp2p/utils/debounce'
+import type { Multiaddr } from '@multiformats/multiaddr'
+
+const crypto = new Crypto()
+x509.cryptoProvider.set(crypto)
+
+type CertificateEvent = 'certificate:provision' | 'certificate:renew'
```
SgtPooki, line 20:
```suggestion
type CertificateEvent = 'certificate:provision' | 'certificate:renew' | `certificate:load`
```

add event for loading a cert during startup?

packages/auto-tls/test/index.spec.ts
```diff
@@ -0,0 +1,5 @@
+describe('auto-tls', () => {
+  it('should fetch a TLS certificate', async () => {
+
```
SgtPooki, line 3: I'm sure you are aware, but a callout here for some tests. Some I can think of: 

* invalid forgeEndpoint expectations
* certifical renewal time
* certificate loading on startup (not implemented yet)
* event emitting
* triggering fetch of cert only on certain listening maddrs

SgtPooki (commented):
just looking through the initial impl

packages/auto-tls/src/auto-tls.ts
achingbrain: I thought about doing something like this but it wasn't clear to me what the actual difference between `certificate:provision` and `certificate:load` would be, they both just mean "a certificate is available and one wasn't before", whereas `certificate:renew` means "a certificate is available but you may need to undo what you did before to start using it".

packages/auto-tls/src/auto-tls.ts
achingbrain: Done - note & constants file added

packages/auto-tls/src/auto-tls.ts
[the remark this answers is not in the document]
achingbrain: The cert is now cached and loaded on startup.  We generate a new CSR every time, but that's to skip having to verify if an existing CSR loaded from the datastore matches the current certificate private key.  Can always revisit if this turns out to be a bottleneck.

packages/auto-tls/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "@peculiar/webcrypto": "^1.5.0",
+    "@peculiar/x509": "^1.12.3",
+    "acme-client": "^5.4.0",
+    "multiformats": "^13.3.1"
+  },
+  "devDependencies": {
+    "aegir": "^44.0.1"
+  },
+  "sideEffects": false
+}
```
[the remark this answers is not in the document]
achingbrain: Yes, this will throw now.

packages/auto-tls/src/auto-tls.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    const cert = new x509.X509Certificate(certString)
+    const renewAt = new Date(cert.notAfter.getTime() - this.renewThreshold)
+
+    this.log('certificate expiry %s - renewing at %s', cert.notAfter, renewAt)
+
+    // schedule renewing the certificate
+    this.renewTimeout = setTimeout(() => {
+      this.certificate = undefined
+      this._fetchCertificates()
```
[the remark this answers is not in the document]
achingbrain: Makes sense, in case there's a clash with the self update event.

packages/auto-tls/src/auto-tls.ts
achingbrain: Yes, I live in hope that maybe one day we'll return a 204 for requests with no content instead of always 200s.

packages/interface/src/index.ts
achingbrain: Reworded

packages/auto-tls/src/auto-tls.ts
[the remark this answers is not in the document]
achingbrain: This uses the non-deprecated API now

packages/interface/src/index.ts
achingbrain: Updated

packages/auto-tls/src/auto-tls.ts
```diff
@@ -0,0 +1,200 @@
...
+    this.events = components.events
+    this.forgeEndpoint = init.forgeEndpoint ?? 'registration.libp2p.direct'
+    this.forgeDomain = init.forgeDomain ?? 'libp2p.direct'
+    this.acmeDirectory = init.acmeDirectory ?? 'https://acme-v02.api.letsencrypt.org/directory'
+    this.timeout = init.timeout ?? 10000
+    this.renewThreshold = init.renewThreshold ?? 60000
+    this.clientAuth = new ClientAuth(this.privateKey)
+    this.started = false
+    this.fetching = false
+    this.fetchCertificates = debounce(this._fetchCertificates.bind(this), init.delay ?? 5000)
```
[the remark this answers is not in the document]
achingbrain: I've updated the method names.

packages/auto-tls/src/auto-tls.ts
```diff
@@ -0,0 +1,200 @@
...
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json'
+          },
+          body: JSON.stringify({
+            value: keyAuthorization,
+            addresses
+          }),
+          ...options
+        })
```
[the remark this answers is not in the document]
achingbrain: I've added validation of the forge and acme endpoints in the constructor to ensure they are valid URLs.

`@libp2p/http-fetch` will throw if the response code isn't a success code and and failures will be logged.  I'm not sure we can do more than that.

packages/auto-tls/test/index.spec.ts
achingbrain: There are tests now 😅

achingbrain:
Merging to make landing #2800 easier

SgtPooki (commented):
posting review that was in progress

packages/auto-tls/README.md
```diff
@@ -0,0 +1,109 @@
...
+$ npm i @libp2p/plaintext
+```
+
+## Browser `<script>` tag
+
+Loading this module through a script tag will make it's exports available as `Libp2pPlaintext` in the global namespace.
+
+```html
+<script src="https://unpkg.com/@libp2p/plaintext/dist/index.min.js"></script>
+```
```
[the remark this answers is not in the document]
achingbrain: Shouldn't be `Libp2pPlaintext` anyway..

PR #2800, opened 2024-11-03
feat: auto-tls for websockets

achingbrain:
Starts a tcp server on the listen port and hands connections off to internal http or https servers (depdending on connection type).

Upgrade requests from both servers are handled by a websocket server.

The https server is enabled when either a secure websocket address is listened to explicitly, or when a TLS certificate is provisioned by another libp2p component, likely `@libp2p/auto-tls`.

This means we don't need to add another port mapping for the https server since we run http and https over the same port.

## Change checklist


achingbrain:
Depends on #2798 and probably some others.

packages/transport-websockets/test/browser.ts
```diff
@@ -1,12 +1,41 @@
...
+    const valid = ws.dialFilter([ma1, ma2, ma3, ma4])
+
+    if (isBrowser || isWebWorker) {
+      expect(valid.length).to.equal(2)
+      expect(valid).to.deep.equal([ma2, ma4])
+    } else {
+      expect(valid.length).to.equal(4)
+    }
+  })
+
```
achingbrain, line 35:
```suggestion
```

achingbrain:
Merging to get a rc out for further testing

sfroment:
@achingbrain Hey 👋 , this as broke the usage of websockets with pubsubPeerDiscovery, when pubsub is calling the getAddresses() here: https://github.com/libp2p/js-libp2p-pubsub-peer-discovery/blob/main/src/index.ts#L244 it will receive `Listener is not ready yet` from https://github.com/libp2p/js-libp2p/blob/main/packages/transport-websockets/src/listener.ts#L346. I'll try to fix it on my own but if I don't let me know if you have a clue of what might cause this 🙏

sfroment:
@achingbrain So I digged a bit more and the problem is coming from 
https://github.com/libp2p/js-libp2p/blob/main/packages/libp2p/src/components.ts#L74-L80 
as the pubsub afterStart will be called at the same time as the ws after start. So they'll both be awaited at the same time and since the ws isn't started when pubsub afterstart is called it will return the above error. 
one quick fix would be to do:
```js
async _invokeStartableMethod(methodName) {
        for (const [key, value] of Object.entries(this.components)) {
            if (isStartable(value)) {
                await value[methodName]?.();
            }
        }
        }
```
        But i don't think this is what you want ?

sfroment:
We could also change the transport manager to call listen on start and not afterStart but I don't know the implication of such a change

PR #2801, opened 2024-11-04
fix: increase providers validity to 48 hours

2color:
## Description

- **chore: remove irrelevant comment**
- **fix: increase providers validty based on spec as discussed in https://github.com/libp2p/js-libp2p/pull/2785**

## Notes & open questions

[The kad-dht specs says](https://github.com/libp2p/specs/blob/master/kad-dht/README.md#content-provider-advertisement-and-discovery):

>Provider Record Republish Interval: The content provider needs to make sure that the nodes chosen to store the provider record are still online when clients ask for the record. In order to guarantee this, while taking into account the peer churn, content providers republish the records they want to provide. Choosing the particular value for the Republish interval is network-specific and depends on several parameters, such as peer reliability and churn.
> For the IPFS network it is currently set to 22 hours.

However in our implementation we take a different approach, whereby we run the `cleanUp` method every 2 hours and only if the record expires in 2 hours do we queue it for a reprovide. **This effectively results in a 44 hour Provider Record Republish Interval**

https://github.com/libp2p/js-libp2p/blob/7c085d10c4d9f491ab85b2467ab1202769ec7a5a/packages/kad-dht/src/reprovider.ts#L150-L154

**Should we increase the `REPROVIDE_THRESHOLD` from 2 hours to 24 hours?**

## Change checklist


achingbrain:
I think maybe we need to split this into two values.

The intention is to mark the provider records we have published as valid for 24 hours, run the cleanup every hour, and reprovide when the record is two hours from expiry, compliant with the spec.  If we were feeling clever, maybe we'd have a variable threshold that also takes into account the average time a record takes to republish.

The 48 hour provide validity setting in kubo I think is for records we've been asked to host, so they are served for 48 hours, which gives the publisher 26 hours grace to republish the record - this is what we are missing.

achingbrain (approved):

Issue #2770, opened 2024-10-15
Ping Service Error: UnexpectedEOFError - Unexpected End of Input


fastalks:
- **Version**:

"libp2p": "^2.1.9",

- **Platform**:

64-bit (Windows) , Mac mini (Apple M1)
- **Subsystem**:

"@libp2p/ping": "^2.0.8",
#### Description:

- What you did
 I set up node1 on Windows and node2 on Mac. They discovered each other via mDNS and established a connection. I then added ping services and set the environment variable DEBUG=libp2p:*
- What happened
I encountered an UnexpectedEOFError: unexpected end of input. Sometimes, the connection disconnects permanently.

#### Steps to reproduce the error:

```
const libp2p = await createLibp2p({
    start: false,
    addresses: {
      listen: [
        '/ip4/0.0.0.0/tcp/0',
        '/ip4/0.0.0.0/tcp/0/ws',
        '/webrtc']
    },
    transports: [
      tcp(),
      webSockets({
        // this allows non-secure WebSocket connections for purposes of the demo
        filter: filters.all
      }),
      // support dialing/listening on WebRTC addresses
      webRTC(),
      circuitRelayTransport({
        discoverRelays: 1
      }) //
    ],
    streamMuxers: [
      yamux()
    ],
    connectionEncrypters: [
      noise()
    ],
    connectionManager: {
      maxConnections: 100
    },
    peerDiscovery: [
      mdns({
        interval: 20e3
      }),
    ],
    connectionGater: {
      denyDialMultiaddr: () => {
        // by default we refuse to dial local addresses from browsers since they
        // are usually sent by remote peers broadcasting undialable multiaddrs and
        // cause errors to appear in the console but in this example we are
        // explicitly connecting to a local node so allow all addresses
        return false
      }
    },
    services:{
      identify: identify(),
      ping: ping(),
      pubsub: gossipsub(),
    },
  });

```
```
[1]   libp2p:ping incoming ping from 12D3KooWFztNbYJ3eLNHomLtMyFgVb4ZGSLaTYTsCCJ2yiZPFFa9 complete in 109ms +109ms
[1]   libp2p:connection:outbound:2jjjmb1728963020005 incoming stream opened on /ipfs/ping/1.0.0 +10s
[1]   libp2p:ping incoming ping from 12D3KooWFztNbYJ3eLNHomLtMyFgVb4ZGSLaTYTsCCJ2yiZPFFa9 +10s
[...]

PR #2819, opened 2024-11-13
fix(ping): gracefully handle remote stream closure.

privacyguard:
## Description

This PR fixes the [UnexpectedEOFError issue](https://github.com/libp2p/js-libp2p/issues/2770). This issue happens because the ping protocol does not handle the case where the remote peer closes the stream and with the introduction of `connectionMonitor` this became an issue. It opens a stream for the ping protocol, sends a ping, gets the response then closes the stream. The local peer throws an exception and triggers a yamux reset.

## Notes & open questions

We only tested these changes in the context described above. Not sure if there are other cases to handle.

## Change checklist


achingbrain:
Closing in favour of #2822

PR #2840, opened 2024-11-22
fix!: use ip/port mapping

achingbrain:
Switch to the new public address mapping api in the address manager to ensure all host/port pairs are advertised with external ip addresses

BREAKING CHANGE: the `autoConfirmAddress` option has been removed

## Change checklist

### DIFF
--- a/packages/upnp-nat/src/check-external-address.ts
+++ b/packages/upnp-nat/src/check-external-address.ts
@@ -1,6 +1,5 @@
 import { NotStartedError, start, stop } from '@libp2p/interface'
 import { repeatingTask } from '@libp2p/utils/repeating-task'
-import { multiaddr } from '@multiformats/multiaddr'
 import pDefer from 'p-defer'
 import { raceSignal } from 'race-signal'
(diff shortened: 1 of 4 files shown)


packages/upnp-nat/src/check-external-address.ts
```diff
@@ -36,13 +35,13 @@ class ExternalAddressChecker implements ExternalAddress, Startable {
...
+  private readonly onExternalAddressChange?: (newExternalAddress: string) => void
 
-  constructor (components: ExternalAddressCheckerComponents, init: ExternalAddressCheckerInit = {}) {
+  constructor (components: ExternalAddressCheckerComponents, init: ExternalAddressCheckerInit) {
     this.log = components.logger.forComponent('libp2p:upnp-nat:external-address-check')
     this.client = components.client
     this.addressManager = components.addressManager
-    this.autoConfirmAddress = init.autoConfirmAddress ?? false
+    this.onExternalAddressChange = init.onExternalAddressChange
     this.started = false
```
SgtPooki, line 45: isn't this a breaking change for this package?

PR #2866, opened 2024-11-30
fix: do not add peers to routing table during RPC handling

dozyio:
## Description

I was having a issue where 2 docker dht nodes failed to dht ping each other on startup. The streams would abort due to a read timeout. 

Also removed `key` from the ping tests as not required.

## Notes & open questions

I don't think there needs to be a call to `this.routingTable.add` in `onIncomingStream` as it should be handled by the topology register.  Although alternatively maybe it could be moved to after the `pipe`?

Couldn't replicate the issue on localhost between 2 non-dockerised instances although I can provide docker files / test repo / logs if needed

## Change checklist

### DIFF
--- a/packages/kad-dht/src/rpc/index.ts
+++ b/packages/kad-dht/src/rpc/index.ts
@@ -95,12 +95,6 @@ export class RPC {
       const { stream, connection } = data
       const peerId = connection.remotePeer
 
-      try {
-        await this.routingTable.add(peerId)
-      } catch (err: any) {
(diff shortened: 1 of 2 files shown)


achingbrain:
Makes sense, thanks for opening this.  I think we might also end up trying to add DHT clients to the routing table like this, which would later fail because they wouldn't respond to a ping so it's a waste of resources.

achingbrain (approved):

Issue #2902, opened 2024-12-25
Listener is not ready yet (Websocket and pubsub discovery)


sfroment:
Hello 👋 

- **Version**: @libp2p+websockets@9.1.0

- **Platform**: MacOS (but the bug will be present on every platform)

#### Description: So while working on Topology [ts-drp](https://github.com/topology-foundation/ts-drp). I encounter this 
```
/node_modules/.pnpm/@libp2p+websockets@9.1.0/node_modules/@libp2p/websockets/src/listener.ts:347
      throw new Error('Listener is not ready yet')
            ^


Error: Listener is not ready yet
    at WebSocketListener.getAddrs (/node_modules/.pnpm/@libp2p+websockets@9.1.0/node_modules/@libp2p/websockets/src/listener.ts:347:13)
    at DefaultTransportManager.getAddrs (/node_modules/.pnpm/libp2p@2.4.2/node_modules/libp2p/src/transport-manager.ts:136:40)
    at AddressManager.getAddressesWithMetadata (/node_modules/.pnpm/libp2p@2.4.2/node_modules/libp2p/src/address-manager/index.ts:358:40)
    at AddressManager.getAddresses (/node_modules/.pnpm/libp2p@2.4.2/node_modules/libp2p/src/address-manager/index.ts:304:29)
    at @libp2p/pubsub-peer-discovery._broadcast (/node_modules/.pnpm/@libp2p+pubsub-peer-discovery@11.0.1/node_modules/@libp2p/pubsub-peer-discovery/src/index.ts:247:44)
    at @libp2p/pubsub-peer-discovery.afterStart (/node_modules/.pnpm/@libp2p+pubsub-peer-discovery@11.0.1/node_modules/@libp2p/pubsub-peer-discovery/src/index.ts:203:5)
    at <anonymous> (/node_modules/.pnpm/libp2p@2.4.2/node_modules/libp2p/src/components.ts:80:6)
    at Array.map (<anonymous>)
    at Proxy._invokeStartableMethod (/node_modules/.pnpm/libp2p@2.4.2/node_modules/libp2p/src/components.ts:77:10)
    at Proxy.afterStart (/node_modules/.pnpm/libp2p@2.4.2/node_modules/libp2p/src/components.ts:96:9)
```
After some digging I found out it was due to this PR https://github.com/libp2p/js-libp2p/commit/8a9258a24168d13172eb139d32bc6889e71f81dc which nevertheless is amazing ❤️ . 
[...]

PR #2903, opened 2024-12-25
fix: #2902

sfroment:
## Title 
[Fix ](fix: #2902)

## Description 
change the way the we call listen so that pubsubdisco can call the getAddresses

## Notes & open questions
Detailed issue here https://github.com/libp2p/js-libp2p/issues/2902
(I don't think it's the good way of fixing this but I might be wrong)

## Change checklist

### DIFF
--- a/packages/transport-websockets/test/node.ts
+++ b/packages/transport-websockets/test/node.ts
@@ -9,13 +9,15 @@ import { multiaddr } from '@multiformats/multiaddr'
 import { WebSockets, WebSocketsSecure } from '@multiformats/multiaddr-matcher'
 import { expect } from 'aegir/chai'
 import { isLoopbackAddr } from 'is-loopback-addr'
+import { createLibp2p } from 'libp2p'
 import { pEvent } from 'p-event'
 import pWaitFor from 'p-wait-for'
 import Sinon from 'sinon'
 import { stubInterface } from 'sinon-ts'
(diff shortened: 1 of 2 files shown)


packages/transport-websockets/test/node.ts
```diff
@@ -9,13 +9,15 @@ import { multiaddr } from '@multiformats/multiaddr'
 import { WebSockets, WebSocketsSecure } from '@multiformats/multiaddr-matcher'
 import { expect } from 'aegir/chai'
 import { isLoopbackAddr } from 'is-loopback-addr'
+import { createLibp2p } from 'libp2p'
```
achingbrain, line 12: This module doesn't depend on `libp2p` but adding this test here would mean the dependency would need to be added.

This is intentional as it creates weird dependency loops that mean (among other things) a patch release of `@libp2p/websockets` would also cause a patch release of `libp2p` even if no changes were made there.

If you need to do integration-style testing, tests should be added to the [integration-tests](https://github.com/libp2p/js-libp2p/tree/main/packages/integration-tests) module, this won't cause any cascading releases.

achingbrain (commented):
Thanks for opening this PR - I'm going to merge https://github.com/libp2p/js-libp2p/pull/2904 instead but I wanted to add a quick note to the test here for future reference.

achingbrain:
Closing in favour of #2904

PR #2921, opened 2025-01-23
docs: wss as been replaced with tls/ws and autoNat is mandatory

sfroment:
Signed-off-by: Sacha Froment <sfroment42@gmail.com>

## Change checklist


achingbrain:
Thanks for opening this.  The READMEs are generated from the `@packageDocumentation` block in `src/index.js` so this changed needs to be made [here](https://github.com/libp2p/js-libp2p/blob/main/packages/auto-tls/src/index.ts#L17-L58) otherwise it'll get overwritten during the release.

achingbrain (approved):

PR #2926, opened 2025-01-27
fix: use publicKey.x instead of privateKey.x

d70-t:
## Description

Although [RFC7518 6.3.2](https://datatracker.ietf.org/doc/html/rfc7518#section-6.3.2) specifies that private key JWKs must include all the fields present in the public key, this is not the case for the implementation of `node:crypto.generateKeyPairSync` in the deno runtime.

While Node returns the (same) `x` property in both, the `privateKey` and the `publicKey`, Deno only returns the `x` property in `publicKey` (i.e. [no x here](https://github.com/denoland/deno/blob/88490d092751288f736855b2418a4da606a31ce7/ext/node/ops/crypto/keys.rs#L1475)).

This change should not affect any Node users, but would enable the use of libp2p on Deno.

## Change checklist

### DIFF
--- a/packages/crypto/src/keys/ed25519/index.ts
+++ b/packages/crypto/src/keys/ed25519/index.ts
@@ -45,7 +45,7 @@ export function generateKey (): Uint8ArrayKeyPair {
   // @ts-expect-error node types are missing jwk as a format
   const privateKeyRaw = uint8arrayFromString(key.privateKey.d, 'base64url')
   // @ts-expect-error node types are missing jwk as a format
-  const publicKeyRaw = uint8arrayFromString(key.privateKey.x, 'base64url')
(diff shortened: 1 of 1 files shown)


achingbrain:
That RFC link only refers to RSA keys, the code change here is for Ed25519 keys?

Seems fine though, thanks for opening this PR.

achingbrain (approved):

d70-t:
Sorry, my bad. Elliptic curves are in section 6.2 right above. But it's the same story.

achingbrain:
Very true.  Is there an issue on the Deno repo about this?  It should implement the `MUST` parts of the RFC completely.

d70-t:
I opened another issue at https://github.com/denoland/deno/issues/27972

PR #2932, opened 2025-02-03
fix: send user agent during auto tls

achingbrain:
Sends a js-libp2p specific user agent to libp2p.direct while configuring the ACME DNS-01 response.

The user agent can be overridden but the default is something like:

```
js-libp2p/2.5.1 node/20.1.5
```

Refs: https://github.com/ipshipyard/p2p-forge/issues/48

## Change checklist

### DIFF
--- a/packages/auto-tls/src/auto-tls.ts
+++ b/packages/auto-tls/src/auto-tls.ts
@@ -1,3 +1,4 @@
+import process from 'node:process'
 import { ClientAuth } from '@libp2p/http-fetch/auth'
 import { serviceCapabilities, serviceDependencies, setMaxListeners, start, stop } from '@libp2p/interface'
 import { debounce } from '@libp2p/utils/debounce'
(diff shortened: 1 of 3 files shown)


packages/auto-tls/src/auto-tls.ts
```diff
@@ -77,6 +79,8 @@ export class AutoTLS implements AutoTLSInterface {
     const base36EncodedPeer = base36.encode(this.components.peerId.toCID().bytes)
     this.domain = `${base36EncodedPeer}.${this.forgeDomain}`
     this.email = `${base36EncodedPeer}@${this.forgeDomain}`
+    this.userAgent = init.userAgent ?? `${this.components.nodeInfo.name}/${this.components.nodeInfo.version} ${process.release.name}/${process.version.replaceAll('v', '')}`
```
maschad, line 82: Could the user agent string could contain unexpected characters from `nodeInfo.name` or `nodeInfo.version`?

maschad (commented):
LGTM just left a comment

packages/auto-tls/src/auto-tls.ts
achingbrain: We pass the user agent string to axios or undici, if there are characters that trip them up then I think the bug is with those modules.

What sort of characters are you thinking of?

maschad (approved):

PR #2954, opened 2025-02-11
feat: Configure maxDataLength for it-length-prefixed in PeerStreams

acul71:
## Description

This PR introduces a new feature that allows users to configure the `maxDataLength` option used by `it-length-prefixed` through the `PeerStreams` options. Previously, the library defaulted to a maximum data length of 4MB, which could lead to issues when processing larger messages. With this change, both the encoder and decoder in `it-length-prefixed` can be configured to use a custom `maxDataLength` (e.g., 8MB), thus accommodating larger messages.

In addition, I have upgraded `it-length-prefixed` to version `10.0.0` in `packages/pubsub/package.json`. However, I'm not sure if this is the only place where the upgrade is required. A review of other packages that depend on `it-length-prefixed` might be necessary.

A new test has been added to validate the behavior, ensuring that a 6MB message is correctly transmitted and received without triggering an error, and that the content of the sent and received messages are identical.

## Notes & open questions

- The upgrade to `it-length-prefixed@10.0.0` in `packages/pubsub/package.json` has been performed, but please verify if additional changes are needed in other parts of the repository.
- The new configuration option is applied to both encoding and decoding. Feedback on the approach or any potential backward compatibility issues would be appreciated.
- Are there any additional tests or documentation updates required for this feature?

## Change checklist


dhuseby:
answer: modify it everywhere

acul71:
@achingbrain @SgtPooki let me know

SgtPooki:
Is the interop CI failure due to these changes or do we have a flaky test?

packages/pubsub/src/peer-streams.ts
```diff
@@ -102,7 +108,8 @@ export class PeerStreams extends TypedEventEmitter<PeerStreamEvents> {
     this._rawInboundStream = stream
     this.inboundStream = pipe(
       this._rawInboundStream,
-      (source) => lp.decode(source)
+      // (source) => lp.decode(source)
```
SgtPooki, line 111:
```suggestion
```

packages/pubsub/src/peer-streams.ts
```diff
@@ -16,6 +16,11 @@ export interface PeerStreamsComponents {
   logger: ComponentLogger
 }
 
+// Define the DecodeOptions type locally
+interface DecoderOptions {
+  maxDataLength?: number
+}
```
SgtPooki, line 22: why not import `DecoderOptions` from `it-length-prefixed`?
```suggestion
import { type DecoderOptions as LpDecoderOptions } from 'it-length-prefixed'

// ...

// Define the DecodeOptions type locally
interface DecoderOptions extends LpDecoderOptions {
  // other custom options we might want for `attachInboundStream`
}
```

packages/pubsub/test/peer-streams.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+import { ConnectionPair } from './utils/index.js'
+
+describe('PeerStreams large message handling', () => {
+  let otherPeerId: PeerId
+
+  beforeEach(async () => {
+    otherPeerId = peerIdFromPrivateKey(await generateKeyPair('Ed25519'))
+  })
+
+  it('should receive messages larger than MAX_DATA_LENGTH when maxDataLength is set', async () => {
```
SgtPooki, line 19:
```suggestion
  it('should receive messages larger than internal MAX_DATA_LENGTH when maxDataLength is set', async () => {
```

maybe i'm being pedantic here, but reading this was slightly confusing

packages/pubsub/test/peer-streams.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    const receivedMessages: Uint8ArrayList[] = []
+    for await (const msg of inbound) {
+      receivedMessages.push(msg)
+    }
+
+    // Check if received correctly
+    expect(receivedMessages).to.have.lengthOf(1)
+    expect(receivedMessages[0].byteLength).to.equal(messageSize)
+    // Check that the content of the sent and received messages are identical
+    // expect(receivedMessages[0].slice()).to.eql(largeMessage.slice())
```
SgtPooki, line 65:
```suggestion
```

packages/pubsub/src/peer-streams.ts
acul71: Of course, I forgot to get rid of the comment :)

packages/pubsub/src/peer-streams.ts
acul71: Yes.
This is because I first tried to modify peer-streams.ts and then I realized I had to change it-length-prefixed to make it work, but then I forgot to import the type, because the code was already written. 
Nice also the `interface DecoderOptions extends LpDecoderOptions ` it would never come to my mind, if you didn't showed me

packages/pubsub/test/peer-streams.spec.ts
acul71: Yes you're right, only who knows the source code of `it-length-prefixed` would have understood that.

packages/pubsub/test/peer-streams.spec.ts
acul71: Yes I leaved the comment, on purpose,  because I wanted to ask if it was better to do
```typescript
    expect(receivedMessages[0].slice()).to.eql(largeMessage.slice())```
or 
```
    const data = receivedMessages[0].slice()
    const input = largeMessage.slice()
    expect(data.length).to.equal(input.length)
    expect(data).to.deep.equal(input)
```

The first sentence is more succinct, the second is longer but it's easier to debug in case of errors.

**Comment by acul71** [comment]
@SgtPooki 

```javascript
Error: @libp2p/pubsub: src/peer-streams.ts(19,15): error TS2305: Module '"it-length-prefixed"' has no exported member 'DecoderOptions'.
```

```typescript
- // Define the DecodeOptions type locally
- interface DecoderOptions {
- maxDataLength?: number
- }

+ import { type DecoderOptions as LpDecoderOptions } from 'it-length-prefixed'
+ // ...
+ // Define the DecodeOptions type locally
+ interface DecoderOptions extends LpDecoderOptions {
+   // other custom options we might want for `attachInboundStream`
+ }
```

Do you want me to modify `it-lenght-prefixed` so that it exports that type to be able to apply this change?

packages/pubsub/src/peer-streams.ts
acul71: @SgtPooki 
Ops It doesn't wok:

```javascript
Error: @libp2p/pubsub: src/peer-streams.ts(19,15): error TS2305: Module '"it-length-prefixed"' has no exported member 'DecoderOptions'.
```

```typescript
- // Define the DecodeOptions type locally
- interface DecoderOptions {
- maxDataLength?: number
- }

+ import { type DecoderOptions as LpDecoderOptions } from 'it-length-prefixed'
+ // ...
+ // Define the DecodeOptions type locally
+ interface DecoderOptions extends LpDecoderOptions {
+   // other custom options we might want for `attachInboundStream`
+ }
```

Do you want me to modify `it-lenght-prefixed` so that it exports that type to be able to apply this change?

SgtPooki: I wouldn't worry about changing it again and blocking on it.. but it probably could use 
a tracking issue. You *could* do some typescript fanciness and do something like this:

```
import {decode} from 'it-length-prefixed'

type LpDecoderOptions = NonNullable<Parameters<typeof decode>[1]>
```

which gets the actual type. [See in typescript playground](https://www.typescriptlang.org/play/?ssl=3&ssc=66&pln=1&pc=1#code/JYWwDg9gTgLgBAbwCYFMDGFUF84DMoQhwDkwMAtADYoB2A5jABblhQq7AAeKSxAUHxgBPMCjgAZMABF0mFFADyYGMAg0AznAC8cAHJrdAV0qUAhgCNqAHgAKpqKZAoY89VeGiIuOKgyoAfADaAIwAuv5AA)

SgtPooki: i'm not 100% sure how @achingbrain feels about this though. He may prefer that type be exported (which is the correct long-term thing to do), but they both result in the same type.

ideally, if `it-length-prefixed` changes parameters, it could break the type here, but that should be a major version bump in `it-length-prefixed` so ultimately the difference is trivial IMO

acul71: @SgtPooki @achingbrain 
Ok I'm doing this for now thanks
```typescript
import { type decode } from 'it-length-prefixed'
// import { type DecoderOptions as LpDecoderOptions } from 'it-length-prefixed' // Enable this when DecoderOptions will be exported from it-length-prefixed
type LpDecoderOptions = NonNullable<Parameters<typeof decode>[1]>
```
and I will open an issue on `it-length-prefixed` about exporting type and do a PR

acul71:
@SgtPooki @achingbrain @dhuseby Update `it-length-prefixed` dependency version in all packages to `^10.0.0.`

acul71:
> @SgtPooki @achingbrain @dhuseby Update `it-length-prefixed` dependency version in all packages to `^10.0.0.`

packages/pubsub/src/peer-streams.ts
```diff
@@ -1,11 +1,14 @@
 import { TypedEventEmitter } from '@libp2p/interface'
 import { closeSource } from '@libp2p/utils/close-source'
 import * as lp from 'it-length-prefixed'
+import { type decode } from 'it-length-prefixed'
 import { pipe } from 'it-pipe'
 import { pushable } from 'it-pushable'
 import { Uint8ArrayList } from 'uint8arraylist'
 import type { ComponentLogger, Logger, Stream, PeerId, PeerStreamEvents } from '@libp2p/interface'
 import type { Pushable } from 'it-pushable'
+// import { type DecoderOptions as LpDecoderOptions } from 'it-length-prefixed' // Enable this when DecoderOptions will be exported from it-length-prefixed
```
achingbrain, line 10: These types are exported as of `it-length-prefixed@10.0.1`

packages/pubsub/test/peer-streams.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    // Simulate sending data from the outbound side
+    await pipe(
+      [largeMessage],
+      (source) => lp.encode(source, { maxDataLength: messageSize }),
+      async function * (source) {
+        for await (const chunk of source) {
+          yield chunk
+        }
+      },
```
achingbrain, line 48: This step in the pipeline seems redundant - is it necessary?

achingbrain (approved):
LGTM, small nit about the test inline.

acul71 (commented):
Ok if I understand this correctly the 
```typescript
import type { DecoderOptions as LpDecoderOptions } from 'it-length-prefixed'
```
only help in checking the type but it don't include it in the actual code...

achingbrain:
```ts
import type { DecoderOptions as LpDecoderOptions } from 'it-length-prefixed'
```

and

```ts
import { type DecoderOptions as LpDecoderOptions } from 'it-length-prefixed'
```

are the same thing, we've just settled on the former in the codebase (for the most part) though vscode defaults to the latter if no existing import is present 🤷

acul71 (commented):
About 
```typescript
if (this._rawOutboundStream != null)
```
yes this better
```typescript
this._rawOutboundStream?.closeWrite()
```

About 
```typescript 
    async function * (source) {
       for await (const chunk of source) {
          yield chunk
        }
      },
```
This is because when I started to write the test I thought the the receiving peer was not important because I wanted to test if it gave an error in encoding, but it was not working so I added the code with console.log (that I've removed before submitting) to test why it was not working then I realized that if I didn't add the receiving host it wasn't working but then I just removed the console.log and not the `yield stuff`

PR #2961, opened 2025-02-13
chore: fix double browser install

achingbrain:
`playwright-test` installs the target browser anyway.

## Change checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -151,7 +151,6 @@ jobs:
       - run: |
           sudo rm /etc/apt/sources.list.d/microsoft-prod.list
           sudo apt-get update
-      - run: npx playwright install-deps
       - run: npm run --if-present test:webkit
       - uses: codecov/codecov-action@13ce06bfc6bbe3ecf90edbbf1bc32fe5978ca1d3 # v5.3.1
         with:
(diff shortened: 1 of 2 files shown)


interop/Dockerfile
```diff
@@ -1,18 +1,17 @@
...
 
 WORKDIR /app
 
-COPY package.json ./
+COPY package.json package-lock.json ./
 COPY ./packages ./packages
 COPY ./interop ./interop
 
-# disable colored output and CLI animation from test runners
+# disable coloured output and CLI animation from test runners
```
maschad, line 10: lol

packages/integration-tests/test/connections.spec.ts
```diff
@@ -259,7 +259,7 @@ describe('connections', () => {
     ({ dialer, listener } = await createPeers({
       transports: [
         memory({
-          latency: 1000
+          latency: 5_000
```
maschad, line 262: Why did the latency need to be increased here? I can't recall seeing flakiness but may be my ignorance

maschad (commented):
Looks good, just one comment really

packages/integration-tests/test/connections.spec.ts
achingbrain: It wasn't reliably triggering timeout errors.  There was flakiness in Windows in CI.

PR #2967, opened 2025-02-17
fix(webrtc): remove vulnerable unmaintained stun package dependency

devlux76:
Note: This is my second attempt, I put it on a completely clean feature branch this time.

## Description

This PR removes the dependency on the abandoned `stun` package from the WebRTC transport implementation, which had critical security vulnerabilities through its dependencies:

- `ip`: SSRF vulnerability in isPublic categorization
- `parse-path`: Authorization bypass vulnerability
- `trim-newlines`: Uncontrolled resource consumption
- `yargs-parser`: Prototype pollution vulnerability

The change replaces the STUN message handling with the built-in implementation from `@ipshipyard/node-datachannel`, which we already use for most of our WebRTC functionality. This simplifies our dependency tree and removes these security issues without affecting functionality.

All tests are passing and npm audit now reports zero vulnerabilities.

## Notes & open questions

The change was straightforward as we already had a more secure alternative available through our existing dependency. No API changes were required.

## Change checklist

- [x] I have made corresponding changes to the documentation if necessary (implementation changes only, no API changes)
- [x] I have added tests that prove my fix is effective or that my feature works (existing tests pass and verify the functionality)

### DIFF
--- a/packages/transport-webrtc/package.json
+++ b/packages/transport-webrtc/package.json
@@ -71,7 +71,6 @@
     "it-stream-types": "^2.0.2",
     "multiformats": "^13.3.1",
     "p-defer": "^4.0.1",
-    "p-event": "^6.0.1",
     "p-timeout": "^6.1.3",
     "p-wait-for": "^5.0.2",
     "progress-events": "^1.0.1",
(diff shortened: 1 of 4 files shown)


achingbrain:
Thanks for opening this.  The `stun` dep was originally added because there was a question mark over whether libjuice would support multiple STUN listeners per process (necessary for things like testing where we run two or more libp2p nodes in the same process).

Support was added in paullouisageneau/libjuice#292 so this could simplify further and remove the `dgramListener` codepath entirely.

Issue #2989, opened 2025-02-22
Rotate WebRTC Direct certificates


achingbrain:
At the moment we configure one long-lived certificate for WebRTC Direct.  Instead we should make the duration configurable and generate a new certificate when we approach the expiry date.

When a new certificate is generated, the `"listening"` event should be re-emitted so the transport manager updates our peer record which will cause identify-push to notify connected peers that our addresses have changed.

Aryakoste:
Hello @achingbrain I would like to work on this issue.

achingbrain:
Wonderful, please open a PR with some tests!

achingbrain:
Fixed by https://github.com/libp2p/js-libp2p/pull/3073

PR #2997, opened 2025-02-23
feat: Rotate WebRTC Direct certificates

Aryakoste:
## Title

Rotating web RTC certificate when expiring 

## Description

Fixes https://github.com/libp2p/js-libp2p/issues/2989

## Notes & open questions

I know this PR is not be perfect. Please provide me suggestions or changes that needed to be done.

## Change checklist


packages/transport-webrtc/src/private-to-public/listener.ts
```diff
@@ -154,20 +164,19 @@ export class WebRTCDirectListener extends TypedEventEmitter<ListenerEvents> impl
       server: Promise.resolve()
         .then(async (): Promise<StunServer> => {
           // ensure we have a certificate
-          if (this.certificate == null) {
+          if (this.certificate == null || this.isCertificateExpiring()) {
```
achingbrain, line 167: This check will only run when the node is started.

If the certificate expires while the node is running it will continue using it and connections will start to fail.

Instead a timeout should be set that fires a short(ish) time before the certificate expiry that creates a new certificate, restarts the server with the new cert and emits the "listening" event.

packages/transport-webrtc/src/index.ts
```diff
@@ -280,6 +280,7 @@ export interface TransportCertificate {
    * The hash of the certificate
    */
   certhash: string
+  notAfter: string
```
achingbrain, line 283: This would be better as a `number` so we don't have to convert the string back to a date in order to see if the cert is about to expire, instead it can just be compared to `Date.now() - threshold `.

packages/transport-webrtc/src/private-to-public/listener.ts
```diff
@@ -83,6 +84,15 @@ export class WebRTCDirectListener extends TypedEventEmitter<ListenerEvents> impl
     }
   }
 
+  private isCertificateExpiring (): boolean {
+    if (this.certificate == null) return true
+    const expiryDate = new Date(this.certificate.notAfter)
+    const now = new Date()
+    const timeToExpiry = expiryDate.getTime() - now.getTime()
+    const threshold = 30 * 24 * 60 * 60 * 1000
```
achingbrain, line 92: `threshold` should be configurable.

packages/transport-webrtc/test/certificate.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  beforeEach(() => {
+    components = {
+      peerId: { toB58String: () => 'QmPeerId' } as any,
+      privateKey: {} as any,
+      logger: {
+        forComponent: () => ({
+          trace: () => {},
+          error: () => {}
+        })
+      } as any,
```
achingbrain, line 25: These fields can be filled using actual values, see other tests here for an example.

packages/transport-webrtc/test/certificate.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+      transportManager: {} as any
```
achingbrain, line 26: You can use `stubInterface<TransportManager>()` from `sinon-ts` here to avoid casting to `any`.

achingbrain (changes requested):
Thanks for opening this, comments inline.

Aryakoste:
Making this changes

packages/transport-webrtc/src/private-to-public/listener.ts
```diff
@@ -171,6 +171,37 @@ export class WebRTCDirectListener extends TypedEventEmitter<ListenerEvents> impl
...
+    if (this.certificate == null) {
+      return
+    }
+
+    const expiryDate = new Date(this.certificate.notAfter)
+    const now = new Date()
+    const timeToExpiry = expiryDate.getTime() - now.getTime()
+    const timeoutDuration = timeToExpiry - 7 * 86400000
+
+    setTimeout(async () => {
```
achingbrain, line 198: If `node.stop()` is called, this timeout will prevent the process from stopping until it fires when the certificate expires long into the future.

Instead you should store a ref to the returned value and pass it to `clearTimeout` when the listener is stopped.

packages/transport-webrtc/src/private-to-public/listener.ts
```diff
@@ -171,6 +171,37 @@ export class WebRTCDirectListener extends TypedEventEmitter<ListenerEvents> impl
...
+  }
+
+  private setCertificateExpiryTimeout (): void {
+    if (this.certificate == null) {
+      return
+    }
+
+    const expiryDate = new Date(this.certificate.notAfter)
+    const now = new Date()
+    const timeToExpiry = expiryDate.getTime() - now.getTime()
```
achingbrain, line 195: No need to allocate a new `Date` object:

```suggestion
    const timeToExpiry = expiryDate.getTime() - Date.now()
```

packages/transport-webrtc/src/private-to-public/listener.ts
```diff
@@ -171,6 +171,37 @@ export class WebRTCDirectListener extends TypedEventEmitter<ListenerEvents> impl
...
+    const timeoutDuration = timeToExpiry - 7 * 86400000
```
achingbrain, line 196: This should use the configurable threshold.

> 7 * 86400000

Please make/reuse this default value as a constant at the top of the file.

packages/transport-webrtc/test/certificate.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    const isCertificateExpiringSpy = sinon.spy(listener as any, 'isCertificateExpiring')
+    const generateCertificateSpy = sinon.spy(listener as any, 'createAndSetCertificate')
+
+    await (listener as any).startUDPMuxServer('127.0.0.1', 0)
+
+    expect(isCertificateExpiringSpy.returned(true)).to.be.true
+    expect(generateCertificateSpy.called).to.be.true
+  })
```
achingbrain, line 93: For these tests please do not depend on internal state (e.g. private fields and methods), it makes the tests fragile and which then makes them hard to maintain.

Instead, make the `WebRTCDirectListener.certificate` field public, configure short timeouts for certificate expiry/renewal thresholds (e.g. under 100ms), use [delay](https://www.npmjs.com/package/delay) and assert on the certificate status after time has passed.

packages/transport-webrtc/src/index.ts
```diff
@@ -283,6 +283,7 @@ export interface TransportCertificate {
    * The hash of the certificate
    */
   certhash: string
+  notAfter: number
```
achingbrain, line 286: Please add a jsdoc comment explaining what this field is for.

achingbrain:
Closing as abandoned, superseded by #3073

PR #3001, opened 2025-02-25
fix: ignore failures to listen on IPv6 addresses when IPv4 succeeds

### LINKED ISSUE #2977: Failure to create a circuit relay reservation when both ipv6 and ipv4 are present but the client only supports ipv4
## Description

In https://github.com/libp2p/universal-connectivity, we rely on a dedicated bootstrapper for circuit relay reservations. We forgo instantiating a DHT Client because it leads to many connections and noise that become hard to distinguish from chat/app-specific peers.

To create the reservation, we take the multiaddrs of the bootstrapper (resolved with the delegated routing client) and append `/p2p-circuit` and pass it to the `addresses.listen` in the `createLibp2p` option. 

https://github.com/libp2p/universal-connectivity/blob/4e5aa6ebd81fb46c68c30583b66c88822b83e1bd/js-peer/src/lib/libp2p.ts#L188-L190
https://github.com/libp2p/universal-connectivity/blob/4e5aa6ebd81fb46c68c30583b66c88822b83e1bd/js-peer/src/lib/libp2p.ts#L45

The problem with this approach is that the boostrapper has both ipv4 and ipv6 and the if browser is in a network without IPv6 support, it fails to connect which causes libp2p to throw an exception and fails to start:

> failed to start libp2p NoValidAddressesError: Transport (@libp2p/circuit-relay-v2-transport) could not listen on any available address

If I filter out ipv6 addresses, the problem is fixed, but I'm not sure this is a good idea. 

What's the recommended way to configure libp2p to deal with this scenario?

## Related problem

If you start libp2p with `'/p2p-circuit'` in the `addresses.listen`, but have no DHT client, js-libp2p will enter an infinite loop:

```
libp2p:random-walk walk iteration for .... and 1 walkers finished, found 0 peers +0ms
libp2p:random-walk walk iteration for .... and 1 walkers finished, found 0 peers +0ms
```

achingbrain:
If listening on all IPv4 addresses succeeds but listening on all IPv6 addresses fails, we may be on a network that doesn't support IPv6 so ignore the failure.

Fixes #2977

## Change checklist

### DIFF
--- a/packages/libp2p/src/errors.ts
+++ b/packages/libp2p/src/errors.ts
@@ -66,13 +66,6 @@ export class NoValidAddressesError extends Error {
   }
 }
 
-export class NoSupportedAddressesError extends Error {
-  constructor (message = 'No supported addresses') {
-    super(message)
-    this.name = 'NoSupportedAddressesError'
-  }
-}
-
 export class ConnectionInterceptedError extends Error {
   constructor (message = 'Connection intercepted') {
     super(message)
--- a/packages/libp2p/src/transport-manager.ts
+++ b/packages/libp2p/src/transport-manager.ts
@@ -278,15 +278,18 @@ export class DefaultTransportManager implements TransportManager, Startable {
               if (IP6.matches(addr)) {
                 listenStats.ipv6.success++
               }
+            }, (err) => {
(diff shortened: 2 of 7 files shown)


achingbrain:
Needs #3002

maschad (approved):
Nice work

PR #3013, opened 2025-02-26
fix: add DER decoder and allow passing protobuf digest

achingbrain:
Adds enough of a DER decoder/encoder to parse/store public keys from PKIX bytes and private keys in PKCS#1 bytes.

Also allows passing the digest of a public key in to `publicKeyFromProtobuf` so we can avoid having to recalculate it.

This becomes a big bottleneck when loading a routing table's worth of peers from the peer store where many of them are RSA peers.

## Change checklist


achingbrain:
Needs #3013

PR #3014, opened 2025-02-26
fix: pass digest to publicKeyFromProtobuf

achingbrain:
This with the changes from #3013 roughly doubles the throughput of loading RSA peers from the peer store which is a massive bottleneck when restarting the amino bootstrapper.

Before:

```
 % node benchmarks/load-peers.js
read all x 5.94 ops/sec ±1.37% (33 runs sampled)
```

After:

```
% node benchmarks/load-peers.js
read all x 12.43 ops/sec ±1.18% (63 runs sampled)
```

## Change checklist


achingbrain:
Needs #3013

PR #3022, opened 2025-03-03
feat: allow early muxer selection by connection encrypters

achingbrain:
Some encrypters allow sending application protocols along with the initial handshake.

We can use these to select muxers and save the round-trips of mss.

## Change checklist


2color:
Would this save a roundtrip only for transports that don't have their own stream multiplexer, like TCP and WebSockets?

PR #3023, opened 2025-03-03
fix: decrease default routing table size

achingbrain:
Reduces the default routing table size to 1280 (down from 5120).

This improves CPU/memory/network usage as we don't need to sort as many peers when finding the closest ones to a given key, nor do we need to contact as many to ensure only live peers are still in the table.

## Change checklist

### DIFF
--- a/packages/kad-dht/src/index.ts
+++ b/packages/kad-dht/src/index.ts
@@ -428,15 +428,36 @@ export interface KadDHTInit {
    * can be stored.
    *
    * Storing more peers means fewer lookups (and network operations) are needed
-   * to locate a certain peer, but also that more memory is consumed.
+   * to locate a certain peer, but also that more memory is consumed and more
+   * CPU while responding to queries (e.g. with more peers in the table sorting
+   * the closest peers becomes more expensive) and CPU/network during table
+   * maintenance (e.g. checking peers are still online).
    *
-   * @default 32
+   * The larger this value, the more prefix bits must be the same for a peer to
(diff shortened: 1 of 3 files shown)


packages/kad-dht/src/index.ts
```diff
@@ -428,16 +428,37 @@ export interface KadDHTInit {
...
+   * contain.
+   *
+   * The total number of peers in the table is a factor of `prefixLength` and
+   * `kBucketSize`:
+   *
+   * ```
+   * (2 ^ prefixLength) * kBucketSize
+   * ```
+   *
+   * @default 8
```
2color, line 447: Where is the actual change in the code?

achingbrain:
> How does this compare to the defaults in [libp2p/go-libp2p-kad-dht](https://github.com/libp2p/go-libp2p-kad-dht?rgh-link-date=2025-03-03T12%3A26%3A37Z)?

go-libp2p-kad-dht uses a different data structure for it's routing table, something more similar to #2747 so a direct comparison is difficult.

Instead js-libp2p has a binary trie with peers equally distributed across the address space - the idea is this results in faster lookups for arbitrary data since we know of more peers further from our own peer id.  To ensure we stay well connected it also maintains a separate list of the 20 peers closest to the node's peer ID so far encountered.

packages/kad-dht/src/index.ts
achingbrain: Oops, pushed.

PR #3031, opened 2025-03-05
feat(protocol-http): add HTTP protocol implementation for libp2p

devlux76:
This PR adds a new HTTP protocol implementation for js-libp2p, allowing services using the HTTP protocol. The implementation provides:

- HTTP server and client capabilities for libp2p nodes using transport agnostic libp2p streams
- Support for HTTP/1.1 request/response patterns including websocket upgrades.
- Integration with existing libp2p connection handling

This implementation enables js-libp2p nodes to serve and access HTTP-based services directly through the libp2p network, expanding interoperability with web-based systems and making development of services over libp2p more natural and intuitive.

Related: [https://github.com/libp2p/specs/tree/master/http]


devlux76:
Ugh, those linting errors are embarrassing. I'll be pushing a fix for them shortly.

packages/protocol-http/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "parserOptions": {
+      "project": true,
+      "sourceType": "module"
+    }
+  },
+  "scripts": {
+    "start": "node dist/src/main.js",
+    "build": "aegir build",
+    "test": "aegir test",
+    "clean": "aegir clean && rimraf examples/dist",
```
achingbrain, line 43: I think all the references to examples need to be removed here, they aren't in this PR any more so it's causing the build to fail in CI.

packages/protocol-http/development.md
```diff
@@ -0,0 +1,173 @@
...
+   - The Libp2pResponse interface (from our package's interfaces.ts) doesn't have an `ok` property that we're checking in client code
+   - Our Logger implementation doesn't match the Logger interface expected by the examples
+
+3. **Type Resolution Issues**: 
+   - The examples have @libp2p/interface and @libp2p/interface-internal as peerDependencies rather than direct dependencies
+   - Some imports are using version mismatches between the examples environment and main package
+
+### 4. Comprehensive Solution Plan
+
+1. **Fix the Logger Implementation**:
```
achingbrain, line 122: Instead you should [accept a `ComponentLogger` instance](https://github.com/libp2p/js-libp2p/blob/5fe2a18fc13e090efe23138f7a89aed59d15591f/packages/protocol-echo/src/index.ts#L60) as part of your service components

You can then [create a logger](https://github.com/libp2p/js-libp2p/blob/5fe2a18fc13e090efe23138f7a89aed59d15591f/packages/protocol-echo/src/echo.ts#L19)  for your service.

achingbrain:
Thanks for opening this.  Before I dig in too deeply - how does this compare to the existing libp2p over HTTP implementation [here](https://github.com/libp2p/js-libp2p-http-fetch)?

I notice they both use the same protocol name but I cannot see (for example) where you register a handler for `.well-known/libp2p/protocols`.

devlux76:
Apologies for the noise in this PR. I'm going to close it for now and perform another PR when it's more fully baked.

To address the question of the difference between this and fetch. The difference is night and day. Fetch is for file transfers. In fact I don't see where it's actually implementing HTTP spec at all. The fetch code seems to be implementing something akin to ftp. 

Hence the reason I made this. This is a full webserver implementation in the browser. It strictly complies with HTTP 1.1 including proxy https://httpwg.org/specs/rfc9112.html it also has websocket upgrades from HTTP 2

achingbrain:
Ok, then unless I'm missing something it seems you've implemented "HTTP over libp2p streams" rather than `@libp2p/http-fetch` which is "libp2p streams over HTTP".

> The fetch code seems to be implementing something akin to ftp

You may be confusing [@libp2p/fetch](https://www.npmjs.com/package/@libp2p/fetch) ([spec](https://github.com/libp2p/specs/tree/master/fetch)) with [@libp2p/http-fetch](https://github.com/libp2p/js-libp2p-http-fetch) ([spec](https://github.com/libp2p/specs/tree/master/http)) - it's very different despite tediously having the same name as the browser HTTP client.

> This is a full webserver implementation in the browser

I would maybe take a look at [go-libp2p-http](https://github.com/libp2p/go-libp2p-http) which has a similar goal of "HTTP over libp2p".

Instead of implementing a full webserver/proxy/websockets/etc stack it just provides the underlying streaming primitive (e.g. a libp2p stream instead of a TCP socket) and lets existing webserver implementations do the heavy lifting.

You may be able to use [it-to-stream](https://www.npmjs.com/package/it-to-stream) to turn a libp2p stream into a node stream, then use the `http` module in node and something like `stream-http` in browsers, `ws` for WebSocket support, etc.

You'd then also get compatibility with application frameworks like fastify, express, etc for free.

PR #3032, opened 2025-03-05
fix: optimize TCP connection establishment time

### LINKED ISSUE #3029: Connection establishment time increase between tcp v9 and v10
To connect to another js-libp2p peer and send one byte takes 0.2s with `@libp2p/tcp@9.x.x` but 0.4s with `@libp2p/tcp@10.x.x`.

An extra round-trip may have been added - it needs some investigation as to why.

Image

https://observablehq.com/@libp2p-workspace/performance-dashboard?branch=04d78b7e02f35038bed154f5772cbeade2b87912

The relevant deps are:

*Before*

```
"@chainsafe/libp2p-noise": "^14.0.0"
"@chainsafe/libp2p-yamux": "^6.0.1"
"@libp2p/tcp": "^9.0.4"
"libp2p": "^1.0.5"
```

*After*

```
"@chainsafe/libp2p-noise": "^16.0.3"
"@chainsafe/libp2p-yamux": "^7.0.1"
"@libp2p/tcp": "^10.1.6"
"libp2p": "^2.8.0"
```

panditdhamdhere:
# Fix: Optimize TCP Connection Establishment Time

## Problem
Connection establishment time increased from ~0.2s in TCP v9 to ~0.4s in TCP v10 (as reported in issue #3029).

## Changes Made
1. Reduced `CLOSE_TIMEOUT` from 500ms to 250ms in `constants.ts`
2. Added `noDelay: true` to socket configuration in `tcp.ts` to disable Nagle's algorithm
3. Added `socket.setNoDelay(true)` in `socket-to-conn.ts` to ensure consistent socket configuration

## Testing
Manual testing shows connection times of ~15-19ms, significantly faster than the previous 400ms.

## Related Issue
Fixes #3029

## Change checklist

- [X] I have performed a self-review of my own code
- [X] I have made corresponding changes to the documentation if necessary (this includes comments as well)
- [X] I have added tests that prove my fix is effective or that my feature works


achingbrain:
Thanks for opening this.

We already set `noDelay` to `true` by default for incoming sockets [here](https://github.com/libp2p/js-libp2p/blob/main/packages/transport-tcp/src/listener.ts#L82) and outgoing sockets [here](https://github.com/libp2p/js-libp2p/blob/main/packages/transport-tcp/src/tcp.ts#L78)

> Manual testing shows connection times of ~15-19ms, significantly faster than the previous 400ms.

I think you may only be testing locally here.  The 400ms time is between data centres in us-east-1 and us-west-1.

packages/transport-tcp/TEST_RESULTS.md
```diff
@@ -0,0 +1,54 @@
+# TCP Connection Time Test Results
```
achingbrain, line 1: There is no need to submit files like this.  Please add your findings to the PR and it'll be in the git history instead for interested parties.

packages/transport-tcp/src/constants.ts
```diff
@@ -4,7 +4,7 @@ export const CODE_CIRCUIT = 290
 export const CODE_UNIX = 400
 
 // Time to wait for a connection to close gracefully before destroying it manually
-export const CLOSE_TIMEOUT = 500
+export const CLOSE_TIMEOUT = 250
```
achingbrain, line 7: Why is this necessary?

packages/transport-tcp/test/connection-time.js
```diff
@@ -0,0 +1,74 @@
+/**
```
achingbrain, line 1: This is a TypeScript project, please only submit TypeScript files.

packages/transport-tcp/test/connection-time.spec.js
```diff
@@ -0,0 +1,87 @@
...
+      })
+      
+      const connectionTime = Date.now() - start
+      console.log(`TCP connection established in ${connectionTime}ms`)
+      
+      // Close the socket
+      socket.end()
+      
+      // Verify connection time is under threshold
+      expect(connectionTime).to.be.below(300, 'Connection time should be under 300ms')
```
achingbrain, line 41: This only runs locally so yes, the connection time will be significantly under 300ms.

packages/transport-tcp/test/connection-time.spec.js
```diff
@@ -0,0 +1,87 @@
...
+    } finally {
+      // Close the server
+      await new Promise(resolve => {
+        server.close(resolve)
+      })
+    }
+  })
+  
+  it('should establish a libp2p connection in under 300ms', async () => {
```
achingbrain, line 50: I'm not sure this test works as expected - the nodes don't have any listen addresses, stream muxers or connection encrypters configured.

Did you run this before submitting the PR?

packages/transport-tcp/test/connection-time.spec.js
```diff
@@ -0,0 +1,87 @@
+/**
```
achingbrain, line 1: No need to add `packageDocumentation` tags - test files are not part of the generated docs and besides, you should only have one `packageDocumentation` tag per module - the one for `@libp2p/tcp` is in the `/src/index.ts` file.

packages/transport-tcp/src/tcp.ts
```diff
@@ -121,6 +121,10 @@ export class TCP implements Transport<TCPDialEvents> {
         ...options
       }) as (IpcSocketConnectOpts & TcpSocketConnectOpts)
 
+      // Set TCP_NODELAY to true to disable Nagle's algorithm
+      // This reduces latency by sending data immediately without buffering
+      cOpts.noDelay = true
```
achingbrain, line 126: `noDelay` is already set [here](https://github.com/libp2p/js-libp2p/blob/main/packages/transport-tcp/src/tcp.ts#L78)

packages/transport-tcp/src/socket-to-conn.ts
```diff
@@ -80,6 +80,10 @@ export const toMultiaddrConnection = (socket: Socket, options: ToConnectionOptio
   // by default there is no timeout
   // https://nodejs.org/dist/latest-v16.x/docs/api/net.html#socketsettimeouttimeout-callback
   socket.setTimeout(inactivityTimeout)
+  
+  // Ensure TCP_NODELAY is set to true to disable Nagle's algorithm
+  // This reduces latency by sending data immediately without buffering
+  socket.setNoDelay(true)
```
achingbrain, line 86: `noDelay` is already set [here](https://github.com/libp2p/js-libp2p/blob/main/packages/transport-tcp/src/listener.ts#L82)

packages/transport-tcp/test/connection-time.spec.js
achingbrain, line 1: This is a TypeScript project, please only submit TypeScript files.

achingbrain (changes requested):

achingbrain:
I'm closing this PR because all it does is change the default close timeout.  The problem observed in #3029 occurs during connection establishment, not while closing the connection.

Please open a new PR if you find a solution to the problem.

PR #3041, opened 2025-03-07
fix: drain timeout is 30s no 301s

sfroment:
## Title

fix: drain timeout is 30s no 301s

## Description
Before it was 30s https://github.com/libp2p/js-libp2p/pull/3005/files#diff-622f25d2f456a8282d3c1a3fe645b8a63044e11f384fabdad6ac7fc78eafa090L16 so I think it is just a typo 

## Change checklist


achingbrain (approved):
Thanks!

PR #3042, opened 2025-03-11
feat: support parsing base32 and base36 libp2p-key CIDs in peerIdFromString

2color:
## Title

Given the prevalence of both CID encoded PeerIDs and base58mh encoded ones, I found myself defining this function in multiple places. 

This PR introduces this as a helper (as was supported in an earlier version of js-libp2p)

## Change checklist

### DIFF
--- a/packages/peer-id/src/index.ts
+++ b/packages/peer-id/src/index.ts
@@ -37,6 +37,9 @@ export function peerIdFromString (str: string, decoder?: MultibaseDecoder<any>):
     // identity hash ed25519/secp256k1 key or sha2-256 hash of
     // rsa public key - base58btc encoded either way
     multihash = Digest.decode(base58btc.decode(`z${str}`))
+  } else if (str.startsWith('k51qzi5uqu5') || str.startsWith('kzwfwjn5ji4') || str.startsWith('k2k4r8') || str.startsWith('bafz')) {
+    // base36 encoded CIDv1 with libp2p-key and identity hash (for ed25519/secp256k1/rsa) or base32 encoded CIDv1 with libp2p-key and identity hash (for ed25519/secp256k1/rsa)
+    return peerIdFromCID(CID.parse(str))
   } else {
     if (decoder == null) {
       throw new InvalidParametersError('Please pass a multibase decoder for strings that do not start with "1" or "Q"')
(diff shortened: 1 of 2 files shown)


achingbrain (changes requested):
Looking at the utility functions exported by this module, if I have a string and I want a PeerId it would not be obvious to me whether to call `peerIdFromString` or this new `getPeerId` function.

Given that the spec [mentions](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#string-representation) base32 (not base36 as in the comment [here](https://github.com/libp2p/js-libp2p/pull/3042/files#diff-03404c4c2b9d68fdd3cc5903cbdc0958a10b3a89c8070f1b498e60492fe6a20dR141)) encoded `libp2p-key` CIDs as a canonical string encoding, let's check form that prefix in `peerIdFromString` instead.

2color:
> Given that the spec [mentions](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#string-representation) base32 (not base36 as in the comment [here](https://github.com/libp2p/js-libp2p/pull/3042/files#diff-03404c4c2b9d68fdd3cc5903cbdc0958a10b3a89c8070f1b498e60492fe6a20dR141)) encoded `libp2p-key` CIDs as a canonical string encoding, let's check form that prefix in `peerIdFromString` instead.

Do you mean checking the [multibase prefix in the string for `b`](https://github.com/multiformats/multibase/blob/master/multibase.csv#L13) and decoding it as a CID if it is in the  `peerIdFromString` function? 

Regarding base36 cids, I realise it's not in the peerIDs spec, but it's the [standard string form for ipns names](https://specs.ipfs.tech/ipns/ipns-record/#string-representation). Do you feel strongly about not including it here?

achingbrain:
> Do you mean checking the [multibase prefix in the string for b](https://github.com/multiformats/multibase/blob/master/multibase.csv?rgh-link-date=2025-03-11T13%3A11%3A01Z#L13)

Yes, but with few more characters to include the `libp2p-key` codec in the check so we are more sure it's a `CID`.

> it's the [standard string form for ipns names](https://specs.ipfs.tech/ipns/ipns-record/#string-representation). Do you feel strongly about not including it here?

Groan. The usual reason not to include lots of different base encoder/decoders is because they bloat bundle sizes but I suppose here we have a dependency on the CID class so base36 is in there anyway.

2color:
> Yes, but with few more characters to include the `libp2p-key` codec in the check so we are more sure it's a `CID`.

This turned out to be rather confusing. I took a look at the [string encoding of cids](https://github.com/multiformats/cid?tab=readme-ov-file#variant---stringified-form) and I'm not sure I got the right prefix characters for base36/base32 CID encoded public keys.

achingbrain (approved):

Issue #3030, opened 2025-03-05
I built an http service plugin to allow you to run a webserver over libp2p streams


devlux76:
I was looking for a way to provide a fullsome and complete implementation of the http application layer protocol service in libp2p using libp2p streams as the transport.

I found an official spec at https://github.com/libp2p/specs/tree/master/http but I noticed that there had been no action in over 5 months.

I decided to create the plugin myself by forking libp2p and putting it in the packages under protocol-http

I've built this, it seems to run well and it even provides websocket upgrades over libp2p streams. It strictly complies with http 1.1 spec and the interface it provides to the developer is the same one provided by nodejs. 

This means it is event driven and porting a webservice to libp2p should be simple now. For simplicity sake the whole thing is just using protobuf and it's working fine with yamux.

There's a webserver "listener" that is peerId based and there is a fetch compatible component so your browser app can fetch('http://peerId/some/endpoint") and you can treat the whole thing like any other webservices connection, even though it is transport layer agnostic.

Basically, I wrote something to strictly comply with the official libp2p-http spec.

Honestly, I'm not sure where to go from here. I don't necessarily want to maintain this long term, but I'm willing to be the maintainer of it for the time being.  

Would it be uncouth to just open a PR from my feature branch? 
https://github.com/syntropiq/js-libp2p/tree/feature/protocol-http
It's an embarrassingly bad commit history but at least it's clean and doesn't touch any other parts of the stack.

Or is there some other process it needs to undergo?

Thanks in advance!

achingbrain:
Just to be clear (and for future readers/issue spelunkers), the [HTTP spec](https://github.com/libp2p/specs/tree/master/http) is "libp2p streams over HTTP" and there is an existing js implementation of that spec [here](https://github.com/libp2p/js-libp2p-http-fetch), whereas what this issue appears to be about is more like "HTTP over libp2p streams".

> I was looking for a way to provide a fullsome and complete implementation of the http application layer protocol service in libp2p using libp2p streams as the transport

I would take a look at [go-libp2p-http](https://github.com/libp2p/go-libp2p-http) which has a similar goal of "HTTP over libp2p streams".

Instead of implementing a full webserver/proxy/websockets/etc stack it just provides the underlying streaming primitive (e.g. a libp2p stream instead of a TCP socket) and lets existing webserver implementations do the heavy lifting.

[...]

PR #3046, opened 2025-03-14
feat(protocol): add HTTP protocol implementation for libp2p

devlux76:
## Title
feat(protocol): add HTTP protocol implementation for libp2p

## Description

This PR introduces a complete HTTP protocol implementation for libp2p, following the [official libp2p HTTP specification](https://github.com/libp2p/specs/blob/master/http/README.md). The implementation provides both client and server components allowing peers to exchange HTTP messages over libp2p streams.

Key features include:
- HTTP client and server with HTTP/1.1 semantics over libp2p streams
- WHATWG Fetch API compatibility layer for familiar web usage
- Path-based request routing with middleware support
- Protocol Buffers-based message encoding/decoding
- Protocol discovery via `.well-known/libp2p/protocols` as specified
- Comprehensive test coverage and documentation

This implementation enables HTTP-based communication in peer-to-peer networks using libp2p, opening up possibilities for distributed web applications, REST APIs over libp2p, and more interoperable P2P services.

## Notes & open questions

- The implementation deviates slightly from traditional HTTP due to Protocol Buffers v3 constraints, using a composition pattern instead of inheritance for message structures.
- Authentication mechanisms beyond libp2p's transport security are not fully implemented as the specification indicates these will be defined in future specs.
- Not all HTTP features (caching, range requests, etc.) are fully implemented yet, as noted in the DESIGN document. These could be added in future iterations.
- There may be opportunities to optimize performance for high-throughput scenarios in follow-up PRs.

## Change checklist


devlux76:
Seems the build is failing because interfaces is relying on an outdated version of a package I've just set to the latest version? I'm not really sure what the best path to take here is. I don't feel right trying to update someone else's dependencies, but it build and runs cleanly here.

achingbrain:
Thanks for submitting this.  Did you see this response in regard to your approach here? - https://github.com/libp2p/js-libp2p/issues/3030#issuecomment-2713084584

devlux76:
> Thanks for submitting this. Did you see this response in regard to your approach here? - [#3030 (comment)](https://github.com/libp2p/js-libp2p/issues/3030#issuecomment-2713084584)

I didn't see it before. I see it now. I respectfully disagree though.  Existing web-server stacks make a lot of presumptions about the transport layer. We aren't going to be able to interface with them at the transport layer like that.

It's far better IMHO to create the web stack and expose familiar semantics, specifically HTTP semantics over libp2p. 
This way people just need to import this lib instead of express for instance.
As far as maintainability, this relies solely on existing libp2p code with no external deps. 
The HTTP protocol stack isn't likely to go anywhere, anytime soon.  I'd say it's pretty solid in that regard.

packages/protocol-http/DESIGN.md
```diff
@@ -0,0 +1,203 @@
...
+
+This document outlines the design and implementation of the HTTP protocol over libp2p based on the [official libp2p HTTP specification](https://github.com/libp2p/specs/blob/master/http/README.md). It describes the current implementation architecture, design decisions, and deviations from the specification due to technical constraints.
+
+## Specification Compliance
+
+The implementation follows the libp2p HTTP specification with some adaptations:
+
+1. **Transport Layer**: Uses libp2p streams as the underlying transport mechanism
+2. **Protocol Identification**: Uses `/https/1.1` as the protocol identifier for stream negotiation
+3. **Message Format**: Implements HTTP semantics using Protocol Buffers v3, which required several structural adaptations
```
lidel, line 13: :warning: I am afraid this is not compatible with current shape of https://github.com/libp2p/specs/tree/master/http:

- spec defines `/http/1.1` and this PR uses `/https/1.1`
  - `/http/1.1` is _cleartext_ version of HTTP, as defined in [RFC 2616](https://datatracker.ietf.org/doc/html/rfc2616), no TLS, no extra transformations/serializations
    - `/http/1.1` over libp2p does not have hard dependency on Protocol Buffer by design
  
This makes this PR effectively introduce a new, protobuf-based custom protocol with HTTP semantics, which does not have any agreed specs.

I see a lot of work and effort went into this. Some potential paths forward:
- Change this PR to use cleartext HTTP 1.1, remove Protobuf serialization, follow `/http/1.1` wire format and specs from https://github.com/libp2p/specs/tree/master/http 
  - this would make this work useful for entire ecosystem, allowing interop with spec-compliant Golang implementations (e.g. Kubo already supports [`/http/1.1/ipfs/gateway` as opt-in experiment](https://github.com/ipfs/kubo/blob/master/docs/experimental-features.md#http-gateway-over-libp2p))
- OR start discussion in https://github.com/libp2p/specs about limitations of `/http/1.1`, and propose ways to fix it
  - this would mean parking this PR until spec work is resolved

lidel (changes requested):

achingbrain:
This PR uses a different protocol id to the HTTP spec and defines it's own protobuf format for sending HTTP requests and reading responses.

Thanks for taking the time to open this PR, it's clear a lot of work has gone into it but as implemented it doesn't implement an existing specification.

There may be a place for it but this code can be published as its own module external to the js-libp2p monorepo.

If you want to take this idea further I'd suggest following @lidel's advice and open a discussion on the specs repo.

Issue #2988, opened 2025-02-22
Store WebRTC Direct certificate in the keychain


achingbrain:
We generate a new TLS certificate for the WebRTC Direct transport on startup.  This means our multiaddr changes between reboots even though the port/peer id might remain the same.

Instead we should store it in the keychain and re-use it if it exists and the expiry is within some configurable threshold.

This way previously shared multiaddrs would still be valid.

Crosstons:
Hey @achingbrain  I would like to work on this

achingbrain:
Wonderful, please open a PR with some tests!

Crosstons:
@achingbrain Can you give me some advice on how to approach this issue?

achingbrain:
Fixed by https://github.com/libp2p/js-libp2p/pull/3071

PR #3048, opened 2025-03-16
feat: implement WebRTC certificate storage in the keychain

Crosstons:
## Title

feat: Implement WebRTC certificate persistence

## Description

This PR implements the WebRTC certificate persistence functionality. The following changes have been made:

- Added `certificate-store.ts` to handle storing and retrieving certificates from the keychain.
- Created tests for certificate persistence in `certificate-persistence.spec.ts`.

This addresses the issue of ensuring certificates can be stored and reused across node restarts, improving the reliability and efficiency of WebRTC connections.

Related https://github.com/libp2p/js-libp2p/issues/2988.

## Notes & open questions

- The test file `certificate-persistence.spec.ts` contains two tests. One test passes successfully, while the second test related to persisting certificates between restarts still needs changes. Further debugging and adjustments are required to ensure all tests pass.

## Change checklist

### DIFF
--- a/packages/transport-webrtc/src/private-to-public/utils/certificate-store.ts
+++ b/packages/transport-webrtc/src/private-to-public/utils/certificate-store.ts
@@ -1,0 +1,151 @@
+import { Crypto } from '@peculiar/webcrypto'
+import * as x509 from '@peculiar/x509'
+import type { Keychain } from '@libp2p/keychain'
+import type { TransportCertificate } from '../../index.js'
+import { importPrivateKey } from './certificate-parser.js'
+import { generateTransportCertificate } from './generate-certificates.js'
+import type { Logger } from '@libp2p/logger'
+
+const crypto = new Crypto()
+x509.cryptoProvider.set(crypto)
+
+// Certificate storage constants
+export const CERTIFICATE_KEY_PREFIX = 'webrtc-certificate'
+export const DEFAULT_CERTIFICATE_VALIDITY_DAYS = 365 // 1 year by default
+export const DEFAULT_MIN_REMAINING_VALIDITY_DAYS = 30 // Regenerate if less than 30 days validity remaining
+
+export interface CertificateStoreOptions {
+  validityDays?: number
+  minRemainingValidityDays?: number
+}
+
+/**
+ * Check if a stored certificate exists and is valid
+ */
+export async function getStoredCertificate (keychain: Keychain, 
+                                           options: CertificateStoreOptions = {}, 
+                                           log?: Logger): Promise<TransportCertificate | null> {
+  const minRemainingDays = options.minRemainingValidityDays ?? DEFAULT_MIN_REMAINING_VALIDITY_DAYS
+  
+  try {
+    // Try to find an existing certificate
+    const keyName = CERTIFICATE_KEY_PREFIX
+    
+    // Export the certificate key material
+    const privateKey = await keychain.exportKey(keyName)
+    
+    // Parse the stored certificate format to extract metadata
+    const parsedCert = await importPrivateKey(privateKey)
+    
+    if (parsedCert == null) {
+      log?.trace('stored certificate could not be parsed')
+      return null
+    }
+    
+    // Check if the certificate is still valid with sufficient remaining time
+    const now = new Date()
+    const minValidUntil = new Date(now)
+    minValidUntil.setDate(minValidUntil.getDate() + minRemainingDays)
+    
+    if (parsedCert.expiryDate > minValidUntil) {
+      log?.trace(`found valid certificate (expires ${parsedCert.expiryDate.toISOString()})`)
+      return parsedCert.certificate
+    }
+    
+    log?.trace(`certificate expires too soon (${parsedCert.expiryDate.toISOString()})`)
+    return null
+  } catch (err) {
+    // If certificate doesn't exist or has an issue, return null
+    log?.trace('failed to retrieve stored certificate', err)
+    return null
+  }
+}
+
+/**
+ * Store a certificate in the keychain
+ */
+export async function storeCertificate (keychain: Keychain, certificate: TransportCertificate, log?: Logger): Promise<void> {
+  try {
+    // Create a suitable key for the keychain from our certificate
+    const certKey = await createPrivateKeyFromCertificate(certificate)
+    
+    try {
+      // Try to remove any existing certificate before storing the new one
+      await keychain.removeKey(CERTIFICATE_KEY_PREFIX)
+    } catch (err) {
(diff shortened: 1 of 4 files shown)


packages/transport-webrtc/test/certificate-persistent.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+          }
+        }
+      })
+    }
+    
+    it('should use certificate hash in multiaddr', async () => {
+      // Create a node with certificate persistence enabled
+      const node = await createNode(null, {
+        certificates: [], // Empty array to not use any pre-defined certificates
+        useLibjuice: false, // Use WebRTC's built-in STUN/TURN
```
achingbrain, line 83: This option is ignored, it will only use libjuice now.

```suggestion
```

packages/transport-webrtc/test/certificate-persistent.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+        rtcConfiguration: {
+          iceTransportPolicy: 'all',
+        }
```
achingbrain, line 86: I don't think this config is relevant to the test.

```suggestion
```

packages/transport-webrtc/test/certificate-persistent.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+          webRTCDirect({
+            ...options,
+            dataChannel: {
+              maxMessageSize: 1 << 16
+            }
+          })
+        ],
+        services: {
+          keychain: async (components) => {
+            const keychainModule = await import('@libp2p/keychain')
```
achingbrain, line 64: Why is the dynamic import necessary?

packages/transport-webrtc/test/certificate-persistent.spec.ts
```diff
@@ -0,0 +1,176 @@
+/**
```
achingbrain, line 1: Because this test imports `libp2p` and other modules, it should be moved to `/packages/integration-tests` in this monorepo.

Only unit tests should be added here.  Obviously the definition is hazy but if you're starting a libp2p node or having to jump through hoops to import other modules from the monorepo, it's probably an integration test.

packages/transport-webrtc/src/private-to-public/utils/certificate-store.ts
```diff
@@ -0,0 +1,151 @@
...
+    publicKey: certificate.pem,
+    // Store additional metadata
+    _certificate: certificate,
+    // Add required methods
+    export: async () => certificate.privateKey,
+    sign: async () => { throw new Error('Not implemented') },
+    // Add required properties
+    id: certificate.certhash,
+    expiryDate: getExpiryDateFromPem(certificate.pem)
+  }
```
achingbrain, line 141: This needs to be changed to use actual `PrivateKey` instances.

packages/transport-webrtc/src/private-to-public/utils/certificate-store.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+export async function generateAndStoreCertificate (
+  keychain?: Keychain,
+  options: CertificateStoreOptions = {},
+  log?: Logger
+): Promise<TransportCertificate> {
+  const validityDays = options.validityDays ?? DEFAULT_CERTIFICATE_VALIDITY_DAYS
+  
+  log?.trace('generating new ECDSA P-256 key pair')
+  const keyPair = await crypto.subtle.generateKey({
```
achingbrain, line 99: You can now import `generateKeyPair` from `@libp2p/crypto/keys` and do `await generateKeyPair('ECDSA')`.

Then the keypair can be stored in the keychain and you can use `privateKeyToCryptoKeyPair` (also imported from `@libp2p/crypto/keys`) to get a WebCrypto-friendly keypair to generate the cert.

packages/transport-webrtc/test/certificate-persistent.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+      return await createLibp2p({
+        addresses: {
+          listen: ['/ip4/127.0.0.1/udp/0/webrtc-direct']
+        },
+        datastore: nodeDatastore,
+        connectionGater: {
+          denyDialMultiaddr: async () => false
+        },
+        streamMuxers: [mplex()],
+        connectionEncrypters: [noise()],
```
achingbrain, line 53: Why do you need to supply a muxer/encrypter?

packages/transport-webrtc/test/certificate-persistent.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+        transports: [
+          webRTCDirect({
+            ...options,
+            dataChannel: {
+              maxMessageSize: 1 << 16
+            }
+          })
```
achingbrain, line 60: Why do you need to set the message size?

packages/transport-webrtc/test/certificate-persistent.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+        rtcConfiguration: {
+          iceTransportPolicy: 'all',
+        }
+      })
+      
+      try {
+        await node.start()
+
+        // Wait for addresses to be available
+        await new Promise(resolve => setTimeout(resolve, 1000))
```
achingbrain, line 93: Please don't add arbitrary delays, they make the tests take a long time to run and can be flakey.

Instead, use something like `p-wait-for` and test for the thing you are waiting for.

packages/transport-webrtc/test/certificate-persistent.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+      } finally {
+        await node.stop()
+      }
+    })
+    
+    /**
+     * This test checks that the WebRTC transport is at least attempting to persist and retrieve certificates.
+     * After fixing the createPrivateKeyFromCertificate implementation, it should now work properly.
+     */
+    it('should attempt certificate persistence between restarts', async function() {
```
achingbrain, line 115: This test should just start a node, record any WebRTC-Direct multiaddrs it's listening on, then stop the node and start it again and assert that it's re-used the same cert.  No need for deep stubbing of module internals.

achingbrain (changes requested):
Thanks for opening this.  As implemented it won't actually persist the certificate as it sets it as a `_certificate` field on a private key.

As of #3059 we can now generate and store ECDSA keypairs as `PrivateKey`s and as of  #3062 we can store PEM-encoded x509 certificates in the keychain which you may have been missing.

Please can you update the implementation to use these new methods.

PR #3053, opened 2025-03-18
fix: do not expire self multiaddrs

### LINKED ISSUE #3051: PeerStore expires and evicts own self's peer data, breaking identify protocol
- **Version**: libp2p 2.8.0; @libp2p/identify 3.0.25; @libp2p/peer-store 11.1.0

- **Platform**: Darwin arm64

- **Subsystem**: PeerStore, Identify

#### Description:

#3019 adds expirations to peer data to @libp2p/peer-store, and automatically evicts peers with expired data, throwing `NotFoundError` if the peer data is present but not updated within the expiration time.

This also applies to evicting a peer's own peer data. If you run a peer with a minimal configuration with just the identify protocol, and provide an array of `listen` and `announce` multiaddrs, it will become **undialable** after the default expiration time.

```ts
// Peer A
	addresses: {
		listen: ["/ip4/127.0.01/tcp/8080/ws"],
		announce: ["/ip4/127.0.01/tcp/8080/ws"],
	},
	transports: [webSockets({})],

	peerStore: {
		maxAddressAge: 5000,
		maxPeerAge: 5000,
	},

	streamMuxers: [yamux()],
	connectionEncrypters: [noise({})],

	services: {
		identify: identify({}),
	},
})
```

If we wait 5 seconds and then try dialing from another peer B, the connection is opened, but the identify protocol fails. The logs on the original peer A show:

```
  libp2p:identify:error could not respond to identify request NotFoundError: Not found
    at PersistentStore.load (file:///Users/joelgustafson/Projects/canvas/node_modules/@libp2p/peer-store/dist/src/store.js:77:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async PersistentPeerStore.get (file:///Users/joelgustafson/Projects/canvas/node_modules/@libp2p/peer-store/dist/src/index.js:80:20)
    at async Identify.handleProtocol (file:///Users/joelgustafson/Projects/canvas/node_modules/@libp2p/identify/dist/src/identify.js:121:30) +0ms
```

The referenced line in `Identify.handleProtocol` is [this one](https://github.com/libp2p/js-libp2p/blob/907518badde245a241f5b6cbfc7579d9c78639af/packages/protocol-identify/src/identify.ts#L163) that loads the data for the peer's own PeerId:

```ts
const peerData = await this.peerStore.get(this.peerId)
```

I'm not sure what the ideal behavior here is. Maybe we should only remove the expired multiaddrs when loading peers from the peer store, and not throw NotFoundError? Maybe we shouldn't evict data for a peer's own PeerId ever?

If not, some documentation (in the changelog/readme for peer-store?) on how to re-register self data? This was a patch version in peer-store but *completely* breaks many basic peer configurations.

achingbrain:
Do not apply address/record expiry to the peer store entry for the self peer.

Fixes #3051

## Change checklist

### DIFF
--- a/packages/peer-store/src/store.ts
+++ b/packages/peer-store/src/store.ts
@@ -100,12 +100,12 @@ export class PersistentStore {
     const buf = await this.datastore.get(key)
     const peer = PeerPB.decode(buf)
 
-    if (this.#peerIsExpired(peer)) {
+    if (this.#peerIsExpired(peerId, peer)) {
       await this.datastore.delete(key)
       throw new NotFoundError()
     }
 
-    return pbToPeer(peerId, peer, this.maxAddressAge)
+    return pbToPeer(peerId, peer, this.peerId.equals(peerId) ? Infinity : this.maxAddressAge)
   }
 
   async save (peerId: PeerId, data: PeerData): Promise<PeerUpdate> {
@@ -152,12 +152,12 @@ export class PersistentStore {
       const peer = PeerPB.decode(value)
 
       // remove expired peer
-      if (this.#peerIsExpired(peer)) {
+      if (this.#peerIsExpired(peerId, peer)) {
         await this.datastore.delete(key)
         continue
       }
 
-      yield pbToPeer(peerId, peer, this.maxAddressAge)
+      yield pbToPeer(peerId, peer, this.peerId.equals(peerId) ? Infinity : this.maxAddressAge)
     }
   }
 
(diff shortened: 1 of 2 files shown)


SgtPooki (approved):
to be clear, we don't expire self addresses, but updates do to network changes will update the addresses in the store, which will remove old invalid addresses, correct?

achingbrain:
Yes, the address manager updates the peer store record when listening addresses change.

PR #3059, opened 2025-03-24
feat: support ECDSA private keys

achingbrain:
Adds support for creating, using and storing ECDSA private keys.

This is a prerequisite for storing x509 certs and associated keys in the keychain.

Keys are serialized to DER PKI messages (see https://datatracker.ietf.org/doc/html/rfc4210)

## Change checklist


packages/crypto/src/keys/ecdsa/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  }, false, ['sign'])
+
+  const signature = await crypto.subtle.sign({
+    name: 'ECDSA',
+    hash: {
+      name: 'SHA-256'
+    }
+  }, privateKey, msg.subarray())
+
+  return new Uint8Array(signature, 0, signature.byteLength)
```
maschad, line 31: why can't you simply return `new Uint8Array(signature)` here?

packages/crypto/src/keys/ecdsa/utils.ts
```diff
@@ -0,0 +1,211 @@
...
+      y
+    }, {
+      ...P_521_KEY_JWK,
+      key_ops: ['verify'],
+      x,
+      y
+    })
+  }
+
+  throw new InvalidParametersError(`Private key length was wrong length, got ${privateKey.byteLength}, expected 32`)
```
maschad, line 98: This key error says "expected 32" but aren't  48, and 66 byte keys also supported?

packages/crypto/src/keys/ecdsa/utils.ts
```diff
@@ -0,0 +1,211 @@
...
+}
+
+export function unmarshalECDSAPublicKey (bytes: Uint8Array): ECDSAPublicKey {
+  const message = decodeDer(bytes)
+  const coordinates = message[1][1][0]
+  let x: string
+  let y: string
+
+  if (coordinates.byteLength === 65) {
+    x = uint8ArrayToString(coordinates.subarray(1, 33), 'base64url')
```
maschad, line 108: Maybe we should move these magic numbers to named constants, it's not immediately obvious why 1,33 should be used.

maschad (commented):
Nice work here, left some comments

packages/crypto/src/keys/ecdsa/utils.ts
achingbrain: Fixed

packages/crypto/src/keys/ecdsa/utils.ts
achingbrain: Fixed

PR #3076, opened 2025-04-10
fix: close handshake datachannel after use

### LINKED ISSUE #2625: `.stop()` sometimes hangs indefinitely
- **Version**:
1.8.1

- **Subsystem**:
@libp2p/websocket , connection manager, components

#### Severity:
High

I have a test scenario with a few nodes that I connect to each other and send some data

Sometimes when terminating a node it endlessly waits for the Websocket transport to shutdown. This is because there are connections that does close (?) and in the `it-ws` library there are no calls to `server.closeAllConnections()` before `close`. 

I have managed to make the shutdown work better if  modify https://github.com/libp2p/js-libp2p/blob/73f2b6b6d767492f8f1d740fac382a2d22c3fca1/packages/libp2p/src/components.ts#L72 to terminate the components in the reverse order in sequentially... so I suspect there is some kind of race condition going on or leak of some sort

Also no issues if I use Tcp transport instead.

This is not a good description on how to reproduce the issue, because it is a bit random when it occurs. But wanted to create this issue if others have the same problem and will update the description if I have a isolated a good test

achingbrain:
Some times the handshake datachannel can remain open which causes a memory leak and stops the process from exiting so explicitly close the channel after the noise handshake has completed.

Also removes a redundant `raceSignal` - it's not necessary as we pass the signal in to the `connect` method.

Fixes #2625

## Change checklist

### DIFF
--- a/packages/transport-webrtc/src/private-to-public/transport.ts
+++ b/packages/transport-webrtc/src/private-to-public/transport.ts
@@ -6,7 +6,6 @@ import { BasicConstraintsExtension, X509Certificate, X509CertificateGenerator } 
 import { Key } from 'interface-datastore'
 import { base64url } from 'multiformats/bases/base64'
 import { sha256 } from 'multiformats/hashes/sha2'
-import { raceSignal } from 'race-signal'
 import { equals as uint8ArrayEquals } from 'uint8arrays/equals'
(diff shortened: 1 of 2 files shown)


achingbrain:

PR #3078, opened 2025-04-14
feat(types): Support custom events on PubSub and PeerStreams

tysonclugg:
## Title
feat(types): Support custom events on PubSub and PeerStreams

## Description

Allow implementations of PubSub and PeerStreams to maintain type safety when using custom events.

## Notes & open questions

Given these changes don't affect functional code paths and that there's no clear direction on how typescript only changes should be tested in js-libp2p, I've not added any tests.

## Change checklist

### DIFF
--- a/packages/interface/src/pubsub.ts
+++ b/packages/interface/src/pubsub.ts
@@ -66,7 +66,7 @@ export interface PubSubRPC {
   messages: PubSubRPCMessage[]
 }
 
-export interface PeerStreams extends TypedEventTarget<PeerStreamEvents> {
+export interface PeerStreams<PeerEvents extends PeerStreamEvents = PeerStreamEvents> extends TypedEventTarget<PeerEvents> {
   id: PeerId
   protocol: string
   outboundStream?: Pushable<Uint8ArrayList>
--- a/packages/pubsub/src/index.ts
+++ b/packages/pubsub/src/index.ts
@@ -40,7 +40,7 @@ import {
   verifySignature
 } from './sign.js'
 import { toMessage, ensureArray, noSignMsgId, msgId, toRpcMessage, randomSeqno } from './utils.js'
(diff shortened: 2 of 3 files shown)


p-shahi:
@tysonclugg We're thinking of removing pubsub from the js-libp2p monorepo (in the future perhaps as a part of v3) and so, this PR is better served to be made against https://github.com/ChainSafe/js-libp2p-gossipsub
@achingbrain will create an issue in that repo describing the current state of wonky interfaces between pubsub and gossipsub in that repo as well

tysonclugg:
I'm of the opinion that js-libp2p-gossipsub should be merging into js-libp2p instead of the other way around.

In the interim (until things are moved one way or the other), this is still the correct place for this PR. There is no other place to submit this code change which has clear benefits for the typing system. I would very much appreciate this being merged now, and not put off because of a _potential_ future change in policy (which hasn't been announced anywhere at this stage that I can see).

Q1: @achingbrain What assurances do we have that anything is being done in the ChainSafe repo? Are PRs such as [#523](https://github.com/ChainSafe/js-libp2p-gossipsub/pull/523) being looked at, and are ChainSafe still actively developing js-libp2p-gossipsub for use with their own _active_ projects?

Q2: @p-shahi What assurances do we have that PubSub will remain as a [level 3A](https://github.com/libp2p/specs/blob/master/pubsub/README.md) (recommended active) part of libp2p, and isn't destined for deprecation?

My concerns about moving PubSub out of libp2p is that it will rapidly degenerate into disrepair, thereby forcing a negative response to Q2 above.  Moving gossipsub into libp2p helps to keep things from going stale and overall should improve the situation (eg: PR 523 linked above).

achingbrain:
> What assurances do we have that anything is being done in the ChainSafe repo?

There are [recent commits](https://github.com/ChainSafe/js-libp2p-gossipsub/commits/master/) so things are happening.

> are ChainSafe still actively developing js-libp2p-gossipsub for use with their own active projects?

Yes. It is used extensively by [lodestar](https://github.com/ChainSafe/lodestar).

> What assurances do we have that PubSub .. isn't destined for deprecation?

Nothing is destined for deprecation.  These are protocols/services that you can use with js-libp2p but js-libp2p doesn't need to know about them in order to ship (i.e. not everything needs to live in this repo).

I appreciate that you have several PRs open against GossipSub that have not been reviewed.  I would suggest trying to engage with the ChainSafe dev team to get them moved along.

PR #3082, opened 2025-04-15
fix: allow empty error events

achingbrain:
Some `ErrorEvent` implementations don't have a `.detail` field so null-guard on it before accessing properties.

## Change checklist


achingbrain:

PR #3083, opened 2025-04-16
fix: partial revert of #3076 - do not close handshake channel

achingbrain:
Removes the closing of the handshake datachannel after the connection has been opened because it breaks compatibility with rust-libp2p 53.

This can be revisited after https://github.com/libp2p/rust-libp2p/issues/5986 is resolved.

## Change checklist


achingbrain:

PR #3084, opened 2025-04-16
fix: skip muxer negotiation

achingbrain:
Skips early muxer negotiation during noise handshake since WebRTC incorporates it's own muxer.

Fixes an edge case where a node is only configured with the WebRTC-Direct transport and no additional muxers.

Refs: https://github.com/ChainSafe/js-libp2p-noise/pull/519

## Change checklist


maschad (commented):
The https://github.com/ipshipyard/roadmaps/issues/22 is broken but based on https://github.com/ChainSafe/js-libp2p-noise/pull/519 this makes sense.

~~Question wouldn't this also apply to webtransport as well since it uses TLS?~~

Nevermind I just saw https://github.com/libp2p/js-libp2p/pull/3081 was merged 😅

maschad (approved):

PR #3093, opened 2025-05-02
fix: allow partial routing implementations

achingbrain:
So routing implementations don't have to have empty methods for features they don't support, guard on the routing method being present before invoking it.

## Change checklist


maschad (commented):
Looks good, but why not change to the [ContentRouting](https://github.com/libp2p/js-libp2p/blob/main/packages/interface/src/content-routing.ts#L34-L98) interface to make the said methods optional?

PR #3094, opened 2025-05-06
test: add test to catch unhandled abort signal

2color:
## Description

This adds a test which reproduces the bug observed in https://github.com/ipfs/helia/issues/780.

It seems that the abort signal passed to the KadDHT `provide` method doesn't cause the returned generator to terminate under certain circumstances. I haven't been able to isolate why the abortion gets swallowed sometimes, but this test is as close as I got to reliably reproducing it.

I suspect this that during the delay in the test I added, the iterator from `parallel` is still consuming from its source and doesn't realise the abort signal has fired. The abort event fires from a separate context and doesn't automatically interrupt the generator's execution:

https://github.com/libp2p/js-libp2p/blob/a7d77ddd1ea39a2f9ac43078e2fdb36844ca3c02/packages/kad-dht/src/content-routing/index.ts#L133-L145

I initially observed that the abort signal doesn't get handled from this call: 
```ts
for await (const event of libp2p.services.dht.provide(cid, {
  signal: AbortSignal.timeout(5_000)
})) {
  console.log('Event:', event)
}
```

## Change checklist


2color:
Noting that this has been fixed by https://github.com/achingbrain/it/pull/159/

PR #3095, opened 2025-05-06
fix: update WebRTC types

achingbrain:
Ensure we are using the global types and not node-datachannel polyfills to be compatible with browser implementations.

## Change checklist


maschad:
Do you want to remove [these exports](https://github.com/libp2p/js-libp2p/blob/main/packages/transport-webrtc/src/webrtc/index.ts#L1-L2) as well since we are not longer using the polyfills?

PR #3100, opened 2025-05-12
perf(Ed25519-Key): Improve validation speeds in browsers

Chandelier-02:
## Title
perf(Ed25519-Key): Improve validation speeds in browsers

## Description
In this PR, I've replaced most Ed25519 key operations with Web Crypto alternatives. With many browsers soon rolling out full support for the Ed25519 key type, this is an opportune time to integrate the new API into Libp2p. The primary benefit is the significant performance improvement that WebCrypto offers.

Simple benchmarks show substantially higher throughput compared to the @noble/curves library implementation of the Ed25519 key.
Screenshot from 2025-05-12 15-12-38

I've also tested the performance difference in one of my projects that receives numerous messages over gossipsub (verified via Ed25519). The results show a dramatic reduction in main thread usage during high message rates, especially with large messages.

## Notes & open questions
I could not figure out how to derive a Ed25519 public key from a private key (don't think it's supported in the WebCrypto API), so I am utilizing the Noble Curves library to handle that. Not certain of the security implications of this. 

Since Ed25519 key support is currently behind experimental flags in Chromium browsers, I've implemented fallback detection that reverts to the previous Noble curves implementation when the API is unavailable.

There are two tests in the  [random walk test file](https://github.com/Chandelier-02/js-libp2p/blob/main/packages/libp2p/test/core/random-walk.spec.ts) that I haven't been able to resolve due to timing conditions: 'should continue random walk until all consumers satisfied' and 'should not block walk on slow consumers'. All other tests are passing.

## Change checklist

- [ x] I have performed a self-review of my own code

### DIFF
--- a/packages/crypto/src/keys/ed25519/ed25519.ts
+++ b/packages/crypto/src/keys/ed25519/ed25519.ts
@@ -37,7 +37,7 @@ export class Ed25519PublicKey implements Ed25519PublicKeyInterface {
     return uint8ArrayEquals(this.raw, key.raw)
   }
 
-  verify (data: Uint8Array | Uint8ArrayList, sig: Uint8Array): boolean {
+  verify (data: Uint8Array | Uint8ArrayList, sig: Uint8Array): boolean | Promise<boolean> {
     return crypto.hashAndVerify(this.raw, sig, data)
   }
 }
@@ -62,7 +62,7 @@ export class Ed25519PrivateKey implements Ed25519PrivateKeyInterface {
     return uint8ArrayEquals(this.raw, key.raw)
   }
 
-  sign (message: Uint8Array | Uint8ArrayList): Uint8Array {
+  sign (message: Uint8Array | Uint8ArrayList): Uint8Array | Promise<Uint8Array> {
     return crypto.hashAndSign(this.raw, message)
   }
 }
--- a/packages/crypto/src/keys/ed25519/index.browser.ts
+++ b/packages/crypto/src/keys/ed25519/index.browser.ts
@@ -1,4 +1,7 @@
 import { ed25519 as ed } from '@noble/curves/ed25519'
+import { fromString as uint8arrayFromString } from 'uint8arrays/from-string'
+import { toString as uint8arrayToString } from 'uint8arrays/to-string'
+import crypto from '../../webcrypto/index.js'
 import type { Uint8ArrayKeyPair } from '../interface.js'
 import type { Uint8ArrayList } from 'uint8arraylist'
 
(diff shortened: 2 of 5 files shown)


Chandelier-02:
I have spent some time looking into the reason for why those two specific tests are failing in the 'random-walk.spec.ts' file. The specific test that's failing is this one:
```
it('should continue random walk until all consumers satisfied', async () => {
    let yielded = 0

    peerRouting.getClosestPeers
      .onFirstCall().callsFake(async function * (key, options?: AbortOptions) {
        for (let i = 0; i < 100; i++) {
          options?.signal?.throwIfAborted()
          yielded++
          yield await createRandomPeerInfo()
        }
      })
      .onSecondCall().returns(slowIterator())

    await Promise.all([
      drain(take(randomWalk.walk(), 1)),
      drain(take(randomWalk.walk(), 2))
    ])

    expect(yielded).to.equal(3)
  })
```

After adding a bunch of logging in the random-walk file, I do see that each consumer is getting satisfied, but the yield is only occurring twice. 

First, PeerInfo-A gets generated, and both consumers see it. The first consumer is then finished as it saw its one peer it wanted. Then, PeerInfo-B gets generated. The second consumer is then finished. 

When comparing the old PeerId implementation vs the new one in this PR, the only difference I see that would change test behavior is that the new PR's PeerIds get generated much faster. So, I am thinking there is a race condition in the test. Maybe the test is trying to gauge the wrong metric (number of peer infos yielded vs satisfied consumers). 

Maybe I am misunderstanding what this test is trying to do.

p-shahi:
@Chandelier-02 can you take a look at the test failures?

Chandelier-02:
Okay so the problem had to do with the typing of the Uint8Array's buffer in one of the functions I had changed. I think this is due to the swap over to es2024 in the aegir config. However, fixing that created a new problem. Now, the tests that use this new function, which now has to check if the buffer is a SAB instance, fail because "SharedArrayBuffer is not defined."

achingbrain:
> I could not figure out how to derive a Ed25519 public key from a private key

Me neither.  I put a little [demo together here](https://codepen.io/achingbrain/pen/MYweBLb?editors=0012) but it fails to export a private key previously imported from PKCS#8 data, even though the same code works with a newly generated keypair 🤷 

> Not certain of the security implications of this.

Given that the current release uses `@noble/curves` to derive the key from the seed, it's no less secure that the latest version.

> I think this is due to the swap over to es2024 in the aegir config

It was caused by the typescript upgrade from 5.4.x to 5.8.x, `Uint8Array`s became generic which seems to cause [all manner of weirdness](https://github.com/microsoft/TypeScript/issues/61793).

Thanks for putting this together!

achingbrain:
> I put a little [demo together here](https://codepen.io/achingbrain/pen/MYweBLb?editors=0012) but it fails to export a private key previously imported from PKCS#8 data

It turns out this is a bug in Firefox as the same demo works fine in Safari and Chrome (with Ed25519 support enabled).

I've filed a bug here: https://bugzilla.mozilla.org/show_bug.cgi?id=1970321 (possibly a dupe of https://bugzilla.mozilla.org/show_bug.cgi?id=1932396)

Issue #3034, opened 2025-03-06
Evaluate @roamhq/wrtc


achingbrain:
The old [wrtc](https://www.npmjs.com/package/wrtc) module has been rebooted as [@roamhq/wrtc](https://www.npmjs.com/package/@roamhq/wrtc). We should evaluate it's performance vs [node-datachannel](https://www.npmjs.com/package/node-datachannel).

The [@libp2p/webrtc](https://github.com/libp2p/js-libp2p/tree/main/packages/transport-webrtc) should be updated to use `@roamhq/wrtc` - if a PR can be opened with all tests passing we can test the throughput using the perf tests in the [libp2p/test-plans](https://github.com/libp2p/test-plans) repo.

Khwahish29:
Hey I'd like to work on this :)

achingbrain:
Great, please submit a PR!

AkMo3:
I am sorry, I actually didn't track this and tried to solve this without prior communication. In any case, I have pointed a few consideration [here](https://github.com/libp2p/js-libp2p/pull/3145#issue-3084174300) which are worth taking a look. It would be help if you can run the CI to check the compatibility of the package across different platforms

luu-alex:
While looking into the issue I realized their current build of webrtc does not support ice-lite as [commented from the author](https://github.com/WonderInventions/node-webrtc/issues/29#issuecomment-2988643381) and is incompatible as a replacement for node-datachannel atm.

Rushhaabhhh:
@achingbrain  , @p-shahi  ~
### Developer Note : Addressing WebRTC Performance and @roamhq/wrtc Evaluation

Hello team, 
I am a contributor from PLDG and as assigned by @seetadev I've analyzed issues #3033  (WebRTC streaming performance) and #3034 (evaluation of @roamhq/wrtc), I want to share a detailed assessment and propose a pragmatic, multi-stage plan to address the performance challenges and architectural limitations.

#### Summary of Core Challenges

- **Issue #3033 :** The current node-datachannel-based WebRTC transport exhibits streaming throughput that is an order of magnitude slower than TCP in Node.js environments. The bottleneck lies primarily in the underlying libdatachannel and libjuice components, particularly in SCTP buffer and ICE handling performance.
[...]

PR #3145, opened 2025-05-22
fix: replace node-datachannel with @roamhq/wrtc

AkMo3:
## Title

Attempt to replace node-datachannel with @roamhq/wrtc, feel free to close if someone else is working!

## Description

Related: https://github.com/libp2p/js-libp2p/issues/3034

I have attempted to replace the `node-datachannel` with `@roamhq/wrtc`, a few observations are:
- The `@roamhq/wrtc` package is not a direct replacement for `node-datachannel` due to different spec and opinions they follow
- The `wrtc` package is not very stable and might be unsuitable for production, reason being when Node.js tries to destroy RTCPeerConnection and RTCDataChannel objects during process exit, it encounters a NAPI error that causes a fatal crash.
- No `IceUdpMuxListener` in`@roamhq/wrtc` that can be used to create a stun listener

## Change checklist

### DIFF
--- a/packages/transport-webrtc/package.json
+++ b/packages/transport-webrtc/package.json
@@ -46,7 +46,6 @@
   "dependencies": {
     "@chainsafe/is-ip": "^2.1.0",
     "@chainsafe/libp2p-noise": "^16.1.3",
-    "@ipshipyard/node-datachannel": "^0.26.6",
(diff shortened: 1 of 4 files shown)


AkMo3:
I have some new idea, will try them and comment in the issue if I make any progress

PR #3173, opened 2025-06-03
feat: add stream middleware

achingbrain:
Adds middleware handlers for protocol streams.  They are invoked for incoming and outgoing streams and allow access to the stream and connection before the handler (incoming) or caller (outgoing) receive them.

This way middleware can wrap streams in transforms, or deny access, or something else.

```ts
libp2p.use('/my/protocol/1.0.0', (stream, connection, next) => {
  const originalSource = stream.source

  // increment all byte values in the stream by one
  stream.source = (async function * () {
    for await (const buf of originalSource) {
      buf = buf.map(val => val + 1)

      yield buf
    }
  })()

  // pass the stream on to the next middleware
  next(stream, connection)
})
```

## Change checklist

### DIFF
--- a/doc/GETTING_STARTED.md
+++ b/doc/GETTING_STARTED.md
@@ -250,7 +250,13 @@ If you want to know more about libp2p peer discovery, you should read the follow
 
 ## Debugging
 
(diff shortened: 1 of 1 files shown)


achingbrain:
cc @dozyio

dozyio:
Amazing @achingbrain! thank you - Will have a play over the weekend

PR #3182, opened 2025-06-13
fix: return closest known peers, even if they are not closer

achingbrain:
Aligns with Kad-DHT spec PR, when responding to a `FIND_NODE` query, return the `k` known closest peers, excluding the server and the requester, even if they are not closer than the requester.

## Change checklist


achingbrain:

PR #3196, opened 2025-06-24
feat: implement AutoNATv2

achingbrain:
Implements the `AutoNATv2` [spec](https://github.com/libp2p/specs/blob/master/autonat/autonat-v2.md) for greater flexibility when verifying which addresses are publicly dialable.

## Change checklist


achingbrain:

PR #3203, opened 2025-06-25
fix: check for signal abort in tls during cert generation

achingbrain:
Ensure we don't miss abort events. Also removes abort listener from signal which can cause leaks if the signal is long-lived.

## Change checklist


achingbrain:

PR #3211, opened 2025-07-08
feat: add isLibp2p function for type guarding

achingbrain:
Adds a simple function that detects libp2p-like objects.

## Change checklist


achingbrain:

PR #3215, opened 2025-07-09
fix: scope logging to connection and stream

achingbrain:
Updates log scope for streams to use the muxer scope, which uses the connection scope so we can trace log lines per stream/connection instead of having component scope.

## Change checklist


achingbrain:

PR #3224, opened 2025-07-13
feat(dht): enhance performance and reliability with caching and adapt…

pranavkonde:
### Description
Improves DHT performance and reliability with caching, adaptive timeouts, and better resource management.

### Changes
- Add record caching (LRU, 1000 entries, 5min TTL)
- Implement adaptive timeouts based on peer response times
- Add a peer scoring system for better peer selection
- Add batched cleanup for stale provider records
- Add comprehensive metrics for monitoring

### New Metrics
- Query timing and error metrics
- Provider operation metrics
- Cleanup performance metrics
- Cache hit/miss rates

### Performance Impact
- Reduced query latency through caching
- Better query success rates with adaptive timeouts
- Improved resource utilization
- Faster peer selection

### Testing
- Verified cache behavior and expiration
- Tested adaptive timeout functionality
- Confirmed peer scoring system
- Checked metrics accuracy

### DIFF
--- a/packages/kad-dht/src/content-fetching/index.ts
+++ b/packages/kad-dht/src/content-fetching/index.ts
@@ -43,6 +43,13 @@ export class ContentFetching {
   private readonly queryManager: QueryManager
   private readonly network: Network
   private readonly datastorePrefix: string
+  private readonly recordCache: Map<string, { record: Libp2pRecord, expires: number }>
+  private readonly peerResponseTimes: Map<string, number[]>
+  private readonly MAX_CACHE_SIZE = 1000
+  private readonly MAX_CACHE_AGE = 300000 // 5 minutes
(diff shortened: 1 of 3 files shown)


achingbrain:
Thank you for opening this PR.

Is this in response to an issue you’ve seen in a deployed application or are you just trying to make general improvements to the codebase?

achingbrain:
Thanks for opening this PR.

I'm going to close it due to the lack of response to feedback from the contributor, but also because adding a record cache to the DHT adds complexity to the codebase and increases memory usage with little obvious benefit.

This PR also:

- Adds several timers that are not cleared when the node is stopped which would prevent the process from exiting
- Invents a new adaptive timeout when one already exists in the utils package
- Has no tests

PR #3225, opened 2025-07-22
fix: enable optimistic select by default

achingbrain:
If only one protocol is specified, send it with the first data block to skip the mss round-trip tax.

## Change checklist


achingbrain:
Test to see if it still breaks rust compatibility, not ready for merging.

PR #3238, opened 2025-08-06
fix: avoid wasteful reprovides outside threshold

2color:
# What

This PR fixes a couple of DHT bugs related to how provider records are handled and processed:

- fix: bug in the logic in `packages/kad-dht/src/reprovider.ts` that causes wasteful reproviding of records that long before reaching the threshold. This would cause valid records to be provided every `REPROVIDE_INTERVAL`
- fix: don't delete expired records if they're ours to ensure we persist the user intent to provide CIDs even if the node goes down beyond expiration. These will be reprovided in the next processRecords run
- fix: in RPC handler return records within `PROVIDERS_VALIDITY` instead of `MAX_RECORD_AGE` (36 hours) which seems to be from before the increase of DHT record validity to 48 hours, and is in essence an outdated duplicate. This fix ensures that provider records are returned in RPC requests as long as they are within the 48 hour PROVIDERS_VALIDITY.

Other small changes:
- Rename `cleanUp` (which is misleading given that it triggers the reprovides) to a more descriptive name `processRecords`
- Add a few more logs and reduce the level of logs from trace to normal logs

## Notes & open questions

- For self provider records, should we check the block store to ensure they haven't been garbage collected or removed? 

## Change checklist

### DIFF
--- a/packages/kad-dht/src/reprovider.ts
+++ b/packages/kad-dht/src/reprovider.ts
@@ -121,7 +121,7 @@ export class Reprovider extends TypedEventEmitter<ReprovideEvents> {
   private async processRecords (options?: AbortOptions): Promise<void> {
     try {
       this.safeDispatchEvent('reprovide:start')
-      this.log('Starting reprovide/cleanup')
+      this.log('starting reprovide/cleanup')
       // Get all provider entries from the datastore
       for await (const entry of this.datastore.query({
         prefix: this.datastorePrefix


packages/kad-dht/src/reprovider.ts
```diff
@@ -118,10 +118,10 @@ export class Reprovider extends TypedEventEmitter<ReprovideEvents> {
    * Check all provider records. Delete them if they have expired, reprovide
    * them if the provider is us and the expiry is within the reprovide window.
    */
-  private async cleanUp (options?: AbortOptions): Promise<void> {
+  private async processRecords (options?: AbortOptions): Promise<void> {
     try {
       this.safeDispatchEvent('reprovide:start')
-
+      this.log('Starting reprovide/cleanup')
```
achingbrain, line 124:
```suggestion
      this.log('starting reprovide/cleanup')
```

PR #3279, opened 2025-09-09
chore: firefox webrtc

achingbrain:
Fixes integration and transport interop tests.


packages/interface-compliance-tests/src/transport/index.ts
```diff
@@ -213,7 +213,7 @@ export default (common: TestSetup<TransportTestFixtures>): void => {
         .with.property('name', 'AbortError')
     })
 
-    it('should close all streams when the connection closes', async () => {
+    it.only('should close all streams when the connection closes', async () => {
```
achingbrain, line 216:
```suggestion
    it('should close all streams when the connection closes', async () => {
```

PR #3280, opened 2025-09-10
docs: fix broken links

MozirDmitriy:
## Description

some links of licence were broken
https://github.com/libp2p/js-libp2p/blob/main/interop/LICENSE-APACHE
https://github.com/libp2p/js-libp2p/blob/main/interop/LICENSE-MIT
but now correct are:
https://github.com/libp2p/js-libp2p/blob/main/packages/interop/LICENSE-APACHE
https://github.com/libp2p/js-libp2p/blob/main/packages/interop/LICENSE-MIT

## Change checklist

### DIFF
--- a/interop/README.md
+++ b/interop/README.md
@@ -116,8 +116,8 @@ $ docker run -e transport=tcp -e muxer=yamux -e security=noise -e is_dialer=fals
 
 Licensed under either of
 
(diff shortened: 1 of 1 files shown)


achingbrain:
Thanks for opening this PR.  Each package in this mono repo is self-contained so linking to the license of another package isn't the way to do it, instead the licence files needed to be added.

Superseded by #3282

Issue #3273, opened 2025-09-06
@libp2p/utils/merge-options is blowing the stack in a wasm environment


mjsir911:
- **Version**:
6.7.2

- **Platform**:
fastly compute wasm environment

- **Subsystem**:
`merge-options`

#### Severity:
  Medium - A non-essential functionality does not work, performance issues, etc.

#### Description:

I'm trying to use @helia/http-2.2.1 in a fastly edge compute environment. this is rather experimental / I'm just messing around, but I don't understand why `merge-options`/`getEnumerableOwnPropertyKeys`/`clone` is being *so* recursive:

```
Error while running request handler: too much recursion
Stack:
  getEnumerableOwnPropertyKeys@input.js:22828:26
  cloneOptionObject@input.js:22860:33
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
  cloneOptionObject@input.js:22860:42
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
  cloneOptionObject@input.js:22860:42
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
  cloneOptionObject@input.js:22860:42
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
  cloneOptionObject@input.js:22860:42
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
  cloneOptionObject@input.js:22860:42
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
  cloneOptionObject@input.js:22860:42
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
  cloneOptionObject@input.js:22860:42
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
  cloneOptionObject@input.js:22860:42
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
  cloneOptionObject@input.js:22860:42
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
  cloneOptionObject@input.js:22860:42
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
  cloneOptionObject@input.js:22860:42
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
  cloneOptionObject@input.js:22860:42
  clone@input.js:22847:14
  cloneOptionObject/<@input.js:22861:35
[...]

PR #3285, opened 2025-09-11
fix: [3273] @libp2p/utils/merge-options is blowing the stack

SuchitraSwain:
…ironment

## Change checklist


package.json
```diff
@@ -46,5 +46,10 @@
     "doc",
     "interop",
     "packages/*"
-  ]
+  ],
+  "dependencies": {
+    "@chainsafe/libp2p-noise": "^16.1.4",
+    "@libp2p/tcp": "^10.1.14",
+    "libp2p": "^2.8.9"
+  }
```
dozyio, line 54: why dependencies are being added?

achingbrain (changes requested):
Thanks for opening this.  In addition to @dozyio's comment, in #3273 I suggest removing the `mergeOptions` function entirely, it's only used in the keychain and to create the libp2p config so is a bit superfluous.

Each component that receives an init object should be in charge of inspecting it and/or setting default values, ideally through shallow (e.g. one level deep) property access, then we don't need the deep-merge functionality of `mergeOptions`.

achingbrain:
Closing in favour of #3294

PR #3291, opened 2025-09-19
fix!: remove pubsub

achingbrain:
Removes the @libp2p/pubsub module to let floodsub/gossipsub/etc defined what "pubsub" is and not have a "canonical" representation that may limit API design in the future.

BREAKING CHANGE: the `@libp2p/pubsub` module has been removed

## Change checklist


achingbrain:
Temporarily adds gossipsub to the codebase to ensure the tests still pass.  Will be removed after gossipsub has upgraded to the v3 api.

Issue #2647, opened 2024-08-02
WebRTC SDP handshake connection transient?


marcus-pousette:
- **Version**:
@libp2p/webrtc 4.1.3

- **Subsystem**:
WebRTC

#### Severity:
Medium

#### Description:
I continue my investigation from #2646 and found that there is a bunch of connection and disconnection events that goes through the protocol handler due to the temporary WebRTC SDP handshake connection. 

https://github.com/libp2p/js-libp2p/blob/944935f8dbcc1083e4cb4a02b49a0aab3083d3d9/packages/transport-webrtc/src/private-to-private/initiate-connection.ts#L57

When `shouldCloseConnection` is true, I would want to handle this connection as transient since it might close very soon after open. 

Doing a 

```
connection.transient = true

```

line after the dial does not however fix the problem completely since we still have the problem that all transient and non-transient connection go through the protocol handler, which is another issue #2369

marcus-pousette:
Going through some of the testing for transient connections this line

https://github.com/libp2p/js-libp2p/blob/fd4aea3efea861aa11b4b25b69c5ac8842208ab5/packages/libp2p/test/registrar/registrar.spec.ts#L224

does not seem to have an effect of the outcome of the test. 


If we would want to know whether a closing connection is transient or not the "peer:disconnect" info also needs to emit this 

https://github.com/libp2p/js-libp2p/blob/fd4aea3efea861aa11b4b25b69c5ac8842208ab5/packages/libp2p/src/connection-manager/index.ts#L474

paschal533:
Hi @marcus-pousette, I'd like to work on this. Planning to mark the SDP connection as transient and fix the topology notification inconsistency in the registrar.

Issue #2369, opened 2024-01-21
`notifyOnTransient` also ignore disconnection events?


marcus-pousette:
- **Version**:
1.1.0 

- **Subsystem**:
registrar

#### Severity:
Medium

#### Description:
It seems like onDisconnect events are still emitted  from transient connections shutting down even if `notifyOnTransient` is false. 

I am receiving onDisconnect events when failing to establish a WebRTC connection 

#### Steps to reproduce the error:

The seems to be intended behaviour (judging by the code). 

https://github.com/libp2p/js-libp2p/blob/f27138ca1f552c4ad3e5d325fef626ba6783f0fd/packages/libp2p/src/registrar.ts#L167 

Is it possible to fetch the closing connection here and skip if transient and notifyOnTransient is false ? 

Feels more intuitive that connect and disconnect events are handled in the same way.

marcus-pousette:
Going through some of the testing for transient connections this line

https://github.com/libp2p/js-libp2p/blob/fd4aea3efea861aa11b4b25b69c5ac8842208ab5/packages/libp2p/test/registrar/registrar.spec.ts#L224

does not seem to have an effect of the outcome of the test. 


If we would want to know whether a closing connection is transient or not the "peer:disconnect" info also needs to emit this 

https://github.com/libp2p/js-libp2p/blob/fd4aea3efea861aa11b4b25b69c5ac8842208ab5/packages/libp2p/src/connection-manager/index.ts#L474

PR #3376, opened 2026-02-05
fix(registrar): ensure onDisconnect respects notifyOnLimitedConnection

paschal533:
## Title

fix(registrar): ensure onDisconnect respects notifyOnLimitedConnection

## Description

Fixed an inconsistency in topology notification where `onDisconnect` could fire for connections that never triggered `onConnect`.

When a topology has `notifyOnLimitedConnection` unset (default `false`), connections with limits (like circuit relay connections used for WebRTC SDP signaling) correctly skip `onConnect`. However, when these connections closed, `onDisconnect` was still being called due to a logic error in `_onDisconnect`:
```typescript
// Before: when topology.filter is undefined, this evaluates to
// undefined === false → false, so the early return never triggers
if (topology.filter?.has(remotePeer) === false) { return }

// After: explicitly checks filter exists before checking membership
if (topology.filter != null && topology.filter.has(remotePeer) !== true) { return }
```

This caused spurious disconnect events in protocol handlers like GossipSub when WebRTC browser-to-browser connections use temporary relay connections for SDP exchange.

Fixes https://github.com/libp2p/js-libp2p/issues/2647
Fixes https://github.com/libp2p/js-libp2p/issues/2369

## Notes & open questions

- No changes were needed in `transport-webrtc` because circuit relay v2 connections already have `connection.limits` set. The bug was purely in the registrar's disconnect notification logic.
- The optional chaining behavior (`undefined === false → false`) is a subtle JS gotcha that allowed this to go undetected.

## Change checklist

### DIFF
--- a/packages/integration-tests/.aegir.js
+++ b/packages/integration-tests/.aegir.js
@@ -169,11 +169,11 @@ async function createGoLibp2pRelay () {
   const { defaultLogger } = await import('@libp2p/logger')
 
   const log = defaultLogger().forComponent('go-libp2p')
-  const controlPort = Math.floor(Math.random() * (50000 - 10000 + 1)) + 10000
-  const apiAddr = multiaddr(`/ip4/127.0.0.1/tcp/${controlPort}`)
+  const controlAddrUndefinedPort = multiaddr('/ip4/127.0.0.1/tcp/0')
   const deferred = pDefer()
+
   const proc = execa(p2pd(), [
-    `-listen=${apiAddr.toString()}`,
+    `-listen=${controlAddrUndefinedPort.toString()}`,
     // listen on TCP, WebSockets and WebTransport
     '-hostAddrs=/ip4/127.0.0.1/tcp/0,/ip4/127.0.0.1/tcp/0/ws,/ip4/127.0.0.1/udp/0/quic-v1/webtransport',
     '-noise=true',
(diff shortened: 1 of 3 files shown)


Faolain:
gentle bump @dozyio for review

packages/libp2p/src/registrar.ts
```diff
@@ -199,7 +199,11 @@ export class Registrar implements RegistrarInterface {
 
         await Promise.all(
           [...topologies.values()].map(async topology => {
-            if (topology.filter?.has(remotePeer) === false) {
+            // If the topology has a filter, only call onDisconnect if the peer
+            // was previously added to the filter (which happens on onConnect).
+            // This ensures limited connections that were never notified via
+            // onConnect don't trigger onDisconnect.
+            if (topology.filter != null && topology.filter.has(remotePeer) !== true) {
```
tabcat:
```suggestion
            if (topology.filter?.has(remotePeer) !== true) {
```

follows style from other handler methods

paschal533:
Hi @tabcat , Thanks for the review. I've applied the fix to line 244 in `_onPeerUpdate` and added a test for that    
  case.

  However, I need to keep the explicit null check (`topology.filter != null &&`) rather than simplifying  
  to optional chaining style. Here's the reason...

  The optional chaining pattern doesn't work correctly for topologies **without filters**

  ```typescript
  // This breaks topologies without filters:
  if (topology.filter?.has(peer) !== true) { return }
  // When filter is undefined: undefined !== true → true → always returns early!
```

  This causes the test "should call topology onConnect handler for limited connection when explicitly     
  requested" to fail (times out waiting for onConnect), because that topology has no filter.

  The explicit check handles all three cases correctly:
 ```typescript
  if (topology.filter != null && topology.filter.has(peer) !== true) { return }
  // When filter is undefined: false && ... → false → calls handler ✓
  // When filter exists but peer not in it: true && true → true → skips ✓
  // When filter exists and peer is in it: true && false → false → calls handler ✓
```

  The style difference exists because:
  - _onPeerIdentify checks === true (works with optional chaining)
  - _onDisconnect and _onPeerUpdate check !== true (requires explicit null check for topologies without   
  filters)

  All tests now pass including the limited connection tests. Let me know if you'd prefer a different      
  approach!

PR #3379, opened 2026-02-23
feat: implement partial messages gossipsub spec

matthewkeil:
## Summary

Implements the **Partial Messages Extension** for GossipSub as specified in [libp2p/specs#685](https://github.com/libp2p/specs/pull/685).

This extension allows nodes to exchange parts of large messages instead of full messages. It is designed for Ethereum's Fusaka Data Availability protocol, where nodes typically miss only 1-2 cells (~2 KiB each) of a 64 KiB message — transmitting only the missing parts saves ~500 KiB/slot/node compared to retransmitting the full message.

### Wire format (backward compatible — all new fields are `optional`)

- `SubOpts` extended with `requestsPartial` and `supportsSendingPartial`
- `ControlMessage` extended with `extensions` containing a `ControlExtensions` message
- `RPC` extended with `partial` containing a `PartialMessagesExtension` message

### New public API on `GossipSub`

- `subscribePartial(topic, opts)` — subscribe with partial message support flags
- `unsubscribePartial(topic)` — remove partial support for a topic
- `publishPartial(partialMsg)` — send partial data to requesting peers, metadata-only to supporting peers
- `gossipsub:partial-message` event — fired when a partial message is received

### New module `src/partial/`

- `BitwiseOrMerger` — default `PartsMetadataMerger` using bitwise OR to combine bitmasks
- `PartialMessageState` — per-topic group tracking with LRU eviction and TTL-based pruning

### Integration into `gossipsub.ts`

- **Subscription signaling**: partial flags included in `SubOpts` sent to peers
- **Extension handshake**: `control.extensions.partialMessages = true` sent on first RPC to each peer
- **Heartbeat**: gossips parts metadata to non-mesh partial peers; prunes expired groups
- **Peer cleanup**: `peerPartialOpts`, `sentExtensions`, and per-topic `PartialMessageState` entries cleaned on disconnect

### Configuration options (added to `GossipsubOpts`)

- `partsMetadataMerger` — custom merger (default: `BitwiseOrMerger`)
- `partialMessagesMaxGroups` — max groups per topic
- `partialMessagesGroupTTLMs` — group TTL

### Bug fix: `MockMuxer._createStream()` protocol override (`@libp2p/utils`)

`MockMuxer._createStream()` had `protocol: ''` at the **end** of its property spread, silently overriding whatever protocol was passed in `options`. This caused pre-existing failures in gossipsub tests (`2-nodes`, `signature-policy`, `gossip`, `mesh`) because:

1. **Inbound streams were never created** — the `'stream'` event handler in `connectPubsubNodes` matched `evt.detail.protocol` against registered gossipsub protocols, but the protocol was always `''`, so no handler was ever invoked.
2. **Heartbeat mesh selection was broken** — `this.protocols.includes(peerStreams.protocol)` always failed with `protocol: ''`, so no peers were ever eligible for the mesh.
3. **`streamPair()` protocol option was silently ignored** — callers passing `protocol` to `createStream()` never saw it propagated.

**Fix**: moved `protocol: ''` to the beginning of the object literal so it acts as a default that `...options` can override. Verified no regressions in `packages/libp2p` (20 tests) and `packages/integration-tests` mock-muxer compliance (46 tests).

### Test infrastructure fixes (`packages/gossipsub/test/`)

**`test/utils/create-pubsub.ts` — `connectPubsubNodes`**:
- Added `break` after first matching handler in `'stream'` event listeners to prevent duplicate handler invocation
- Added `pWaitFor` that waits for both outbound **and** inbound streams to be established before returning, fixing race conditions where tests proceeded before async stream setup completed

**`test/signature-policy.spec.ts`**:
- Moved `pEvent('subscription-change')` listener registration **before** `subscribe()` + `connectPubsubNodes()` calls. Subscriptions are exchanged during the connection handshake, so registering the listener after connecting missed the event.

### Tests

70 new tests across 10 files under `test/partial-messages/`:

| File | Tests | Coverage |
|------|-------|----------|
| `subscription-signaling.spec.ts` | 9 | SubOpts partial flags, peer tracking, send/remove |
| `protobuf-round-trip.spec.ts` | 7 | Encode/decode of all new protobuf fields, backward compat |
| `handle-received-partial.spec.ts` | 13 | Inbound partial message handling, state updates, events |
| `publish-partial.spec.ts` | 6 | Outbound routing by peer flags, topic filtering |
| `cleanup.spec.ts` | 5 | Peer disconnect cleanup, topic unsubscribe |
| `extension-handshake.spec.ts` | 4 | ControlExtensions negotiation, dedup |
| `gossip-heartbeat.spec.ts` | 4 | Parts metadata gossip, TTL pruning |
| `mixed-config-eager.spec.ts` | 6 | Mixed partial/non-partial networks, config options, eager push |
| `bitwise-or-merger.spec.ts` | 5 | BitwiseOrMerger unit tests |
| `partial-message-state.spec.ts` | 11 | PartialMessageState unit tests (LRU, TTL, capacity) |

Full suite: **151 passing**, 20 pending (pre-existing skips), 0 failing.

## Test plan

- [x] All 151 gossipsub tests pass (`npx aegir test --target node`)
- [x] `packages/libp2p` connection tests pass (20 tests — MockMuxer consumer)
- [x] `packages/integration-tests` mock-muxer compliance tests pass (46 tests — MockMuxer consumer)
- [ ] CI passes on all targets

## Notes

- **No scoring for partial messages.** By design in this initial version, partial messages bypass the message validation/scoring pipeline. Scoring can be added in a follow-up.
- **Application must call both `publish()` and `publishPartial()`.** Partial messages are a parallel channel — calling `publishPartial` does not replace regular `publish` for full message propagation.

## Change checklist


dozyio:
Hi Matthew, thanks for the PR! Just wondered if you want this reviewing yet as its still in draft?

matthewkeil:
> Hi Matthew, thanks for the PR! Just wondered if you want this reviewing yet as its still in draft?

Was just going through some extra testing in other repos first to make sure everything was as it should be.  It's in a pretty good state after adding a bunch more tests so should be good to go for a review.  I can mark as ready but in full disclosure there are still some things that need to be handled elsewhere.  I don't anticipate it changing the code here much though.

PR #3389, opened 2026-03-02
fix: setMaxConnections uses func arg in comparison 

dozyio:
## Description

Compare against the func arg before throwing

## Change checklist


tabcat:
Let's add a test.

tabcat (approved):
good catch

PR #3394, opened 2026-03-02
fix(metrics): avoid consuming protocol frames before app readers

### LINKED ISSUE #3393: metrics-prometheus trackProtocolStream can consume protocol frames before app readers attach
## Summary

`@libp2p/prometheus-metrics` can interfere with protocol readers by attaching a `message` listener too early in `trackProtocolStream()`.

This causes buffered stream data to be dispatched before app-level readers attach, which can make higher-level protocols (notably identify) miss their first frame and fail. Downstream, this surfaced as increased `Unknown` peer classification.

## Root cause

In `metrics-prometheus/src/index.ts`:

```ts
_track (stream, name) {
  stream.addEventListener('message', (evt) => {
    this._incrementValue(`${name} received`, evt.data.byteLength)
  })
  ...
}
```

In `@libp2p/utils` `AbstractMessageStream`, buffered inbound data is only dispatched when `listenerCount('message') > 0`.

So if metrics registers a `message` listener before protocol code starts reading, inbound bytes can be consumed/dispatched early and not available later when protocol handlers attach.

## Minimal repro

Using `streamPair()`:

1. call `metrics.trackProtocolStream(outbound)`
2. send data from `inbound`
3. wait briefly
4. start reading from `outbound[Symbol.asyncIterator]()`

Observed:
- without metrics tracking: first frame is read
- with metrics tracking (current): read times out (first frame was already consumed)

## Downstream impact (Lodestar)

On a local mainnet beacon run with libp2p v3 stack:

- baseline (`metrics-prometheus` current):
  - 11 samples / ~5 min
  - avg peers: 40.9
  - avg unknown peers: 1.36
  - max unknown peers: 2
- patched metrics behavior (no `message` listener side effect):
  - 10 samples / ~5 min
  - avg peers: 40.0
  - avg unknown peers: 0
  - max unknown peers: 0

This aligns with identify occasionally failing and peer client staying unknown.

## Proposed fix

In `metrics-prometheus` `_track`:

- **do not** add a `message` listener
- instead wrap `stream.dispatchEvent` and count bytes when `evt.type === 'message'`
- keep existing `send` monkey-patch for sent-byte accounting

This preserves metrics collection while avoiding changes to `listenerCount('message')` and read-buffer dispatch behavior.

I opened a PR with this fix + regression test.

lodekeeper:
## Summary

Fixes a protocol-stream observability race in `@libp2p/prometheus-metrics` where tracking could consume inbound frames before app/protocol readers attach.

Fixes #3393

## Root cause

`_track()` currently calls `stream.addEventListener('message', ...)`.

For `AbstractMessageStream`, any `message` listener can trigger dispatch of buffered inbound bytes. If metrics attaches first, protocol handlers that attach later may miss the first frame.

### 1) Avoid adding `message` listener in `_track`

Instead of adding an event listener, wrap `stream.dispatchEvent`:

- when `evt.type === 'message'`, increment received bytes metric
- then forward to original `dispatchEvent`

This preserves received-byte accounting but does **not** increase `listenerCount('message')`, so metrics no longer affects stream buffering/dispatch semantics.

### 2) Regression test

Added test in `packages/metrics-prometheus/test/streams.spec.ts`:

- tracks a stream before app listener is attached
- sends data first
- attaches app iterator later
- asserts first frame is still readable
- asserts metrics still count received bytes

## Validation done downstream

In Lodestar local beacon runs, this change eliminated observed short-window `Unknown` peer spikes linked to identify misses (while keeping metrics enabled).

### DIFF
--- a/packages/metrics-opentelemetry/package.json
+++ b/packages/metrics-opentelemetry/package.json
@@ -46,7 +46,6 @@
   },
   "devDependencies": {
     "@libp2p/logger": "^6.2.2",
-    "@libp2p/utils": "^7.0.12",
     "aegir": "^47.0.22"
   },
   "browser": {
--- a/packages/metrics-opentelemetry/test/index.spec.ts
+++ b/packages/metrics-opentelemetry/test/index.spec.ts
@@ -1,8 +1,6 @@
 import { start, stop } from '@libp2p/interface'
 import { defaultLogger } from '@libp2p/logger'
-import { multiaddrConnectionPair, streamPair } from '@libp2p/utils'
 import { expect } from 'aegir/chai'
-import { pEvent } from 'p-event'
 import { openTelemetryMetrics } from '../src/index.js'
 
 describe('opentelemetry-metrics', () => {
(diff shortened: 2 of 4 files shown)


nflaig:
I've deployed this branch to one of our node groups and it does resolve our issue and we no longer are seeing that many unknown peers

image

lodekeeper:
Thanks for validating this in production 🙌 Great signal that the fix addresses the unknown-peer spike in real deployments. I’ll keep this branch rebased and ready while we wait for maintainer review.

wemeetagain:
@lodekeeper please fix up the branch so CI can pass

lodekeeper:
Thanks for the ping — fixed and pushed now: bfb080b52 (lint fixes in streams.spec.ts for the new early-data test). CI has re-triggered on this commit; I’ll keep an eye on it and follow up if anything else is needed.

lodekeeper:
@tabcat thanks — addressed your review note. I applied the same dispatchEvent-based receive tracking change to:
- packages/metrics-opentelemetry/src/index.ts
- packages/metrics-simple/src/index.ts

Pushed in 4dbf55d82 (branch head now includes this + the earlier type-assertion cleanup).

dozyio:
@lodekeeper Could you update the tests for `metrics-simple` and `metrics-opentelemetry` so that they keep parity with prometheus please?

tabcat:
@dozyio there arent similar test files for those modules so fine merging without those.

lodekeeper:
@dozyio done 👍 I added parity tests for both backends so they mirror the prometheus behavior updates:
- packages/metrics-simple/test/index.spec.ts
- packages/metrics-opentelemetry/test/index.spec.ts

Pushed in commit 7c6a553ac.

lodekeeper:
Thanks @tabcat — acknowledged. I already added parity tests for both backends in 7c6a553ac, so we can merge with those in place (or I can drop them if you prefer minimal scope).

dozyio:
> @dozyio done 👍 I added parity tests for both backends so they mirror the prometheus behavior updates:
> 
> * packages/metrics-simple/test/index.spec.ts
> * packages/metrics-opentelemetry/test/index.spec.ts
> 
> Pushed in commit [7c6a553](https://github.com/libp2p/js-libp2p/commit/7c6a553ac3399576512ad05df57be3b0768034ab).

You will need to add `@libp2p/utils` to the `devDependencies` for `metrics-simple` and `metrics-opentelemetry`

lodekeeper:
Good catch — done. I added @libp2p/utils to devDependencies for both packages:
- packages/metrics-simple/package.json
- packages/metrics-opentelemetry/package.json

Pushed in commit cbaadfee4.

tabcat:
@lodekeeper just remove that last commit `cbaadfe`. no need to add tests for the other metrics packages, just keep the fixes there and looks good.

lodekeeper:
Done — I dropped the parity-test additions and corresponding devDependency change.\n\nI reverted commits 7c6a553ac and cbaadfee4 in bcb1710a2, keeping only the core fixes on the branch.

dozyio:
Thanks @lodekeeper for the PR - merged

PR #3395, opened 2026-03-03
feat(kad): namespace selector

### LINKED ISSUE #3377: PUT_VALUE handler overwrites stored records without version comparison
- **Version**: 15.1.11

- **Platform**: 6.1.0-42-amd64 Debian 6.1.159-1 (2025-12-30) x86_64 GNU/Linux

- **Subsystem**: kad-dht

#### Description
PutValueHandler.handle() validates the incoming record's signature. After that, it unconditionally overwrites whatever is stored locally. The selector is never called during PUT. This means a valid but older signed record can overwrite a newer one on any DHT node. A malicious peer can exploit this by aggressively re-publishing stale records.

Expected behavior: before overwriting, check the existing record using the selector and reject the incoming record if it's older.

Is this intentional? If so, what's the recommended mitigation for applications that store mutable versioned records? 
For example, I mitigated the issue like this: I forked the kad-dht PUT_VALUE handler to check if a record already exists locally before overwriting — if the incoming record has a lower version than the stored one, the PUT is rejected.

*Why I put it as Severity High - Low*: I did not know about this. After talking with some other people who use this library, I found out they do not know either. I even asked Sonnet 4.5 and after a 45 second research, it said "_Yes, the PUT handler does use selectors_". If this is common knowledge, I apologize. 

#### Steps to reproduce the error: N/A

dozyio:
## Description

`PUT_VALUE` now applies selector-based conflict checks before storing for namespaced keys with selectors. Stale records may be ignored earlier than before.

Closes https://github.com/libp2p/js-libp2p/issues/3377

## Change checklist

### DIFF
--- a/packages/kad-dht/src/content-fetching/index.ts
+++ b/packages/kad-dht/src/content-fetching/index.ts
@@ -4,6 +4,7 @@ import map from 'it-map'
 import parallel from 'it-parallel'
 import { pipe } from 'it-pipe'
 import { equals as uint8ArrayEquals } from 'uint8arrays/equals'
+import { toString as uint8ArrayToString } from 'uint8arrays/to-string'
 import {
   ALPHA
 } from '../constants.js'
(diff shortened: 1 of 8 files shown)


Realman78:
Can you also update the documentation? It would be very beneficial

PR #3401, opened 2026-03-10
fix(utils): clear repeatingTask anySignal per run

lodekeeper:
## Summary

Fixes a listener-retention path in `repeatingTask` timeout handling.

`repeatingTask` creates a composed abort signal on each run when `options.timeout` is set:

```ts
anySignal([shutdownController.signal, AbortSignal.timeout(options.timeout)])
```

But this composed signal was not being cleared during steady-state loop execution. That means abort listeners can accumulate while the task keeps running.

This patch:
- stores the per-run composed signal in a local variable
- calls `signal.clear()` in `.finally()` for each run

## Why

We observed long-running memory growth signatures consistent with composed-signal listener retention (WeakRef/WeakCell/Listener + timeout/abort churn) in downstream production-like workloads. Clearing per-run composed signals removes that accumulation path.

## Scope

- `packages/utils/src/repeating-task.ts`

No behavior change for task scheduling semantics; this is cleanup-only.

### DIFF
--- a/packages/utils/src/repeating-task.ts
+++ b/packages/utils/src/repeating-task.ts
@@ -93,7 +93,16 @@ export function repeatingTask (fn: (options?: AbortOptions) => void | Promise<vo
     })
       .catch(() => {})
       .finally(() => {
-        signal?.clear()
+        if (signal != null) {
+          if (signal.aborted) {
+            signal.clear()
+          } else {
+            // Clear listeners once this per-run signal eventually aborts
+            signal.addEventListener('abort', () => {
+              signal.clear()
+            }, { once: true })
+          }
+        }
 
         running = false
 


lodekeeper:
Quick CI triage note:

The current `Build`/`build` failure appears unrelated to this diff and reproducible on another PR branch around the same time.

Failure signature in both runs:
- during `npm install` in `go-libp2p` postinstall
- `ERR_MODULE_NOT_FOUND` for `@libp2p/logger/dist/src/index.js` imported by `blockstore-core/dist/src/tiered.js`

Example runs:
- this PR: `22926672927` / `22926672913`
- unrelated branch: `22925624715` / `22925624738`

Happy to recheck once CI is rerun.

lodekeeper:
Given CI triage points to an upstream/transient install failure, could a maintainer please re-run the failed checks when convenient? Happy to rebase/retest immediately if needed.

lodekeeper:
Pushed follow-up fix for the CI regressions (commit `85fa70700`).

Root cause from failed matrix logs: the previous change called `signal.clear()` in `finally`, which removed `AbortSignal.timeout` listeners before timeout fired, so `repeating-task.spec` (`should time out a task`) hung until the global 60s test timeout across browser/electron/node jobs.

Fix in this commit:
- if per-run `signal` is already aborted in `finally` => clear immediately
- otherwise attach a one-shot `abort` listener that calls `signal.clear()` after abort actually fires

This preserves timeout-abort semantics while still cleaning listeners to avoid long-lived accumulation.

New CI is running on this commit now.

dozyio:
@lodekeeper the PR is still in draft - do you want it reviewed?

lodekeeper:
Yep — ready now, thanks for the ping 🙏

I’ve marked it ready for review. CI is green on the latest commit (85fa70700), including the previously failing browser/electron/node jobs.

dozyio:
@lodekeeper thanks - merged

nflaig:
@dozyio thanks for taking care of these PRs so fast, just posting this here for posterity, the memory leak seems fixed on our end now, at least there hasn't been a noticeable increase and the node has been running with this patch for ~2 days

image

Issue #3015, opened 2025-02-27
Failed webRTC upgrade leads to dropped relay p2p-circuit connection


julienmalard:
**Version**:
* "@libp2p/circuit-relay-v2": "^3.2.4",
* "@libp2p/webrtc": "^5.2.5",
* "libp2p": "^2.7.4",

**Platform**:
* macOS 12.5.1
* Firefox browser (135.0.1 (64 bits))
* Chrome browser (Version 133.0.6943.127 (Build officiel) (x86_64))

- **Subsystem**:
webRTC and Circuit Relay

#### Severity:
High - The main functionality of the application does not work, API breakage, repo format breakage, etc.

#### Description:
When webRTC connections fail (e.g., because of a corporate firewall), the connection between the two peers is dropped instead of falling back to the relay server. This leads to an infinite loop, where both peers connect to the relay, connect to each other over their respective p2p-circuit addresses, attempt to upgrade to webRTC, and then drop the connection and start all over again.

I would expect that, if the upgrade fails, the peers would remain on the relayed connected (and potentially try to upgrade again eventually, but without dropping the relayed connection if webRTC fails).

I wonder if this is perhaps related to the fix for https://github.com/libp2p/js-libp2p/issues/2646.

#### Steps to reproduce the error:
Start two browser nodes (same machine, in my test) with webRTC and circuit relay enabled and a known circuit relay node in the bootstrap list over a network with a firewall that will prevent ICE from working.

NiKrause:
Did you find out why WebRTC upgrade does fail? I have had the experience that dcutr fails by reports missing multiaddresses

julienmalard:
In my case, webrtc failed due to an office firewall. However, I would expect the connection to fall back to the initial relay in such a case.

NiKrause:
Your expectation is understandable and I would agree, on the other hand I'd like to add that in the past, in my opinion, a browser could operate over a relay connection for ever without upgrading to direct peer-to-peer via WebRTC. 

It looks like to me now, and here I might be completely wrong, that for some reason with latest libp2p, the relay connection itself lives only for 120 seconds by default and if not upgraded, it will die.

julienmalard:
@NiKrause That was also my experience in the past. I believe that this might be linked to the following fix for a different issue:  https://github.com/libp2p/js-libp2p/issues/2646.

NiKrause:
[...]

PR #3406, opened 2026-03-13
fix(transport-webrtc): handle terminal peer connection states and sig…

### LINKED ISSUE #2646: When failing to establish WebRTC connection yields to "half open" connection
- **Version**:
1.8.1

- **Platform**:
23.0.0 Darwin Kernel Version 23.0.0: Fri Sep 15 14:41:43 PDT 2023; root:xnu-10002.1.13~1/RELEASE_ARM64_T6000 arm64. [Version 1.64.109 Chromium: 123.0.6312.58 (Official Build) (arm64)](https://brave.com/latest/)

- **Subsystem**:
Circuit Relay, WebRTC

#### Severity:
Medium

#### Description:
I am investing issues I am having when using the WebRTC transport between nodes in offline mode, and I seem to end up in a unexpected state sometimes when a WebRTC connection is failing to establish, but only one side realizes this. 

My main concern with this issue is that WebRTC connections could fail out in the wild (for other reasons than the reproduction method below) and if that leads to the user experience will not be good. 

When using this example 

https://github.com/libp2p/js-libp2p-examples/tree/main/examples/js-libp2p-example-browser-pubsub

Turn WIFI off completely on your computer (this seems to disable WebRTC connections to be establish at all even if you connect to another node on the same machine). 

Now run the example with two nodes and try to connect them to each other (one connects to the relay, the other one connects to the webrtc address of the first node). 

When the second node connects to the first node WebRTC address I get following error on the first node (expected maybe): 

```sh
chunk-KI5O2YSF.js?v=fd2899f7:17 Uncaught (in promise) AbortError: The operation was aborted
    at raceSignal (chunk-KI5O2YSF.js?v=fd2899f7:17:27)
    at YamuxStream.closeWrite (chunk-RSSFLS4Q.js?v=fd2899f7:261:13)
    at YamuxStream.close (chunk-RSSFLS4Q.js?v=fd2899f7:222:12)
    at stream.close (@libp2p_circuit-rela…?v=aaea501d:1019:11)
    at ConnectionImpl.close [as _close] (libp2p.js?v=094dbc14:6580:22)
    at async ConnectionImpl.close (libp2p.js?v=094dbc14:6157:7)
    at async initiateConnection (@libp2p_webrtc.js?v=2e74bfdf:1278:9)
    at async WebRTCTransport.dial (@libp2p_webrtc.js?v=2e74bfdf:1460:61)
    at async DefaultTransportManager.dial (libp2p.js?v=094dbc14:5605:14)
    at async queue.add.peerId.peerId [as fn] (libp2p.js?v=094dbc14:4496:26)
libp2p.js?v=094dbc14:4516 Uncaught (in promise) CodeError: signal is aborted without reason
    at queue.add.peerId.peerId [as fn] (libp2p.js?v=094dbc14:4516:21)
    at async raceSignal (chunk-KI5O2YSF.js?v=fd2899f7:22:12)
    at async Job.run (chunk-VZAQ7WEY.js?v=fd2899f7:1218:22)
   
```

Connected peers on the first node only shows the relay node (1 connection in total), expected because it failed to establish a WebRTC connection.
    
BUT the node that is dialing never realizes that something wrong has happened (!) and still shows that it is connected to both the relay and the WebRTC connection  (2 connections in total).

The expected outcome is that both nodes realizes that the WebRTC connection was never establish.

paschal533:
# Fix: WebRTC terminal peer connection states, signaling race condition, and stream muxer test reliability

 ## Problems fixed

  1. Half-open connection when ICE fails during SDP exchange

  When a WebRTC private-to-private connection fails to establish (WiFi disabled, NAT traversal fails, ICE times out),  
  the dialing side could remain in a "connected" state while the listening side correctly detected the failure. This   
  left a leaked, half-open connection the connection manager never cleaned up.

  The root cause: `RTCPeerConnectionMultiaddrConnection` only checked `connectionState` changes via `onconnectionstatechange` which does not fire for state transitions that already occurred before the handler was     
  registered. Added an explicit check of the initial `connectionState` at construction time.

  Closes #2646

  2. `UnexpectedEOFError` in `readCandidatesUntilConnected`

  A race condition caused the WebRTC "simple" transport compliance test to fail intermittently: the listener closes its
   signaling stream just before the initiator's `connectionstatechange: connected` event fires. The initiator then       
  receives an EOF on the signaling stream and throws an `UnexpectedEOFError`, discarding a perfectly good peer
  connection.

  Fixed by moving the `connectedPromise` outside the try block in `readCandidatesUntilConnected`. If the signaling stream  
  throws EOF and the peer connection isn't yet 'connected', we now await the `connectedPromise`, if the PC connects     
  anyway (just after the EOF), the error is swallowed; if it fails, the error is re-thrown.

  3. 'disconnected' incorrectly treated as a terminal peer connection state

  The `onconnectionstatechange` handler treated 'disconnected' the same as 'failed' and 'closed', calling
  `onTransportClosed()` and `peerConnection.close()` immediately. WebRTC's 'disconnected' state is transient and the ICE     
  agent may recover to 'connected' without intervention

  In Firefox, this caused a spurious DataChannel close during large data transfers, which raced against
  `sendCloseWrite`'s `pEvent(channel, 'close')` listener registration. If the channel closed before the listener was       
  registered, the FIN_ACK was missed, the `finAckTimeout` fired, and all branches of Promise.any rejected with an        
  `AggregateError`.

  Fixed by only treating 'failed' and 'closed' as terminal states.

  4. Flaky stream muxer compliance test (calling abort aborts streams)

  The test raced a 70ms timeout against pipes using `await delay(10)` per iteration which is leaving only a 20ms window after `dialer.abort()` at t=50ms for the abort to propagate. On loaded CI runners, NodeJS timer imprecision routinely exceeds
   this margin.

  Increased the timeout to 2000ms. On success, the test exits well under 100ms; the 2000ms only applies when abort     
  genuinely fails to propagate.

  Test plan

  - Existing WebRTC transport tests pass (Chrome, Firefox)
  - @libp2p/integration-tests WebRTC "simple" compliance test passes
  - @libp2p/mplex calling abort aborts streams passes reliably
  - No regression on happy-path connected peers


Faolain:
minor nit but looks like pre-existing TODO: were stripped?

paschal533:
> minor nit but looks like pre-existing TODO: were stripped?

Thanks for catching that... those slipped in during reformatting.  I fogot to restored them in the latest commit. I will do that right away. Thank you sm

Faolain:
I've been experiencing this exact issue I believe 

My code attempts WebRTC upgrades after relay connections (`useShareLink.ts:612`). If the upgrade enters a transient `'disconnected'` state (common behind NAT), `@libp2p/webrtc@6.0.10` immediately kills the peer connection. Per issue #3015, this can cascade and drop the underlying relay connection too. 

@dozyio and @tabcat would you guys please be able to give this one a lookover?

dozyio:
Thanks @paschal533  for this - quite a lot going on here... could you split the PR into the 4 separate fixes please and I'll get them merged

dozyio:
Also to say that I've being running the branch locally - seems good

paschal533:
> Thanks @paschal533 for this - quite a lot going on here... could you split the PR into the 4 separate fixes please and I'll get them merged

Sure, I will do that right away! Thanks

paschal533:
closing this PR now

PR #3416, opened 2026-03-21
fix: guard processSendQueue against non-writable stream status

### LINKED ISSUE #3415: bug: uncaught StreamStateError when drain event fires on closing stream
## Description

A race condition in `abstract-message-stream.ts` causes an **uncaught exception** when a `drain` event fires on a stream whose underlying transport is already closing. The `StreamStateError` is thrown inside an event listener callback with no surrounding try/catch, so it propagates as an uncaught exception to the Node.js process level.

## Environment

- `@libp2p/utils@7.0.13` (latest)
- `@chainsafe/libp2p-noise@17.0.0`
- `@libp2p/mplex@12.0.14`
- Lodestar v1.41.0 (Ethereum consensus client)
- Node.js (TCP transport)

## Stack Trace

```
[network] error: uncaughtException: Cannot write to a stream that is closing
StreamStateError: Cannot write to a stream that is closing
    at TCPSocketMultiaddrConnection.send (abstract-message-stream.js:112)
    at stream.send (@libp2p/prometheus-metrics/index.js:183)
    at EncryptedMessageStream.sendData (@chainsafe/libp2p-noise/utils.js:169)
    at EncryptedMessageStream.processSendQueue (abstract-message-stream.js:355)
    at EncryptedMessageStream.continueSendingOnDrain (abstract-message-stream.js:56)
```

## Root Cause Analysis

**The race:**

1. A TCP connection begins closing → the underlying stream's `writeStatus` transitions to `'closing'`
2. Before the close completes, a `drain` event fires on the underlying TCP stream (the OS/Node.js TCP socket signals buffer space is available)
3. The noise layer's `noiseOnDrain` handler ([libp2p-noise/utils.ts](https://github.com/ChainSafe/js-libp2p-noise/blob/main/src/utils.ts)) relays this as `safeDispatchEvent('drain')` on the encrypted stream
4. The `continueSendingOnDrain` listener in `abstract-message-stream.ts` (line 97-104) calls `processSendQueue()`
5. `processSendQueue()` (line 428) has guards for `writableNeedsDrain`, empty buffer, and `sendingData` — but **no guard for `writeStatus`**
6. It calls `sendData()` → noise's `sendData()` calls `this.stream.send()` on the underlying (now-closing) TCP stream
7. `send()` (line 169-171) throws `StreamStateError` because `writeStatus === 'closing'`
8. The throw propagates through the event listener call stack with no try/catch → **uncaught exception**

**Key observation:** The `continueSendingOnDrain` callback at line 97 only checks `this.writableNeedsDrain` before calling `processSendQueue()`. Neither it nor `processSendQueue()` verifies that the stream is still writable.

## Impact

- On Lodestar mainnet: **6 uncaught exceptions in 7 days** (3 distinct timestamps, occurring in bursts of 2-3 when a connection tears down)
- The error is logged as `[network] error: uncaughtException` — it reaches the Node.js `process.on('uncaughtException')` handler
- Currently non-fatal (Lodestar catches and logs it) but any `uncaughtException` is a process stability risk
- Affects any js-libp2p consumer using noise over TCP when connections tear down while the send queue is draining

## Proposed Fix

Add a `writeStatus` guard in `processSendQueue()`, which is the single chokepoint for all send paths:

```typescript
// In abstract-message-stream.ts, processSendQueue() — add after existing bail-out checks:

protected processSendQueue (): boolean {
    // bail if the underlying transport is full
    if (this.writableNeedsDrain) { /* ... existing ... */ }

    // bail if there is no data to send
    if (this.writeBuffer.byteLength === 0) { /* ... existing ... */ }

    // bail if we are already sending data
    if (this.sendingData) { /* ... existing ... */ }

    // bail if the stream is no longer writable — a drain event can
    // arrive after writeStatus transitions to 'closing'/'closed'
    if (this.writeStatus !== 'writable') {
      this.log.trace('not processing send queue as stream is %s', this.writeStatus)
      return false
    }

    // ... rest of method unchanged
}
```

This is safe because:
- It's a single guard at the beginning of the method, consistent with the existing bail-out pattern
- Data remaining in `writeBuffer` when the stream closes will be discarded as part of the normal close/abort cleanup
- The `send()` method already enforces this invariant by throwing — this just converts the throw into a graceful no-op at the right level

**Alternative (less preferred):** Add a try/catch around the `sendData()` call in `processSendQueue()` to swallow `StreamStateError` during close. This would be more defensive but masks the root cause.

## Reproduction

This is a timing-dependent race that occurs during TCP connection teardown. It's most likely to manifest under load when many connections are being established and torn down simultaneously. In our case, a mainnet Ethereum beacon node with ~200 peers.

The simplest conceptual reproduction:
1. Establish a noise-over-TCP connection with mplex
2. Send enough data to fill the write buffer (trigger `writableNeedsDrain`)
3. Close/abort the connection while the buffer is still draining
4. The OS delivers a drain event after the close has started → boom

---

*Reported from [Lodestar](https://github.com/ChainSafe/lodestar) — an Ethereum consensus client built on js-libp2p.*

lodekeeper:
## What

Add a `writeStatus` guard at the top of `processSendQueue()` in `abstract-message-stream.ts` to bail out when the stream is no longer writable.

## Why

A race condition causes an uncaught `StreamStateError` when a `drain` event fires on a stream whose underlying transport is already closing:

1. TCP connection starts closing → `writeStatus` transitions to `'closing'`
2. A `drain` event fires on the underlying TCP stream
3. The noise layer relays this as `safeDispatchEvent('drain')` on the encrypted stream
4. `continueSendingOnDrain` → `processSendQueue()` → `sendData()` → `send()` on the now-closing stream
5. `send()` throws `StreamStateError` — inside an event listener with no try/catch → **uncaught exception**

Observed on Lodestar mainnet (v1.41.0, `@libp2p/utils@7.0.13`): 6 uncaught exceptions in 7 days.

## Fix

Single guard at the top of `processSendQueue()`, consistent with the existing bail-out pattern:

```typescript
if (this.writeStatus !== 'writable') {
  this.log.trace('not processing send queue as stream write status is %s', this.writeStatus)
  return false
}
```

## Test

Added a regression test that:
1. Creates a stream pair with `capacity: 1` to trigger backpressure
2. Fills the send buffer until `send()` returns `false`
3. Sets `writeStatus` to `'closing'`
4. Dispatches a `drain` event
5. Asserts `sendData` is NOT called (the guard catches it) and no error is thrown

Closes #3415


dozyio:
@lodekeeper - can you review this commit 17d9da8abf14cd23941699e1c12fb51a8f589b05 please

dozyio (approved):

lodekeeper:
@dozyio Reviewed commit 17d9da8 — looks good overall.

**`abstract-message-stream.ts`**: The `willSend` defensive copy + comment makes the intent clear — guard against `sendData` mutating the slice before throwing. The rename is cleaner than the original `toSend` requeue. No issues here.

**Tests**: Both cases are solid:
- *preserve on error*: verifies buffer is intact after a StreamStateError on drain ✓
- *no duplication on transient error*: the `throwOnce` pattern correctly proves data isn't counted twice across two drain events ✓

One minor note: the second test sets `writeStatus = 'closing'` before the first drain but the `continueSendingOnDrain` guard only bails on statuses that are neither `'writable'` nor `'closing'`. So `'closing'` still reaches the try/catch — which is the right flow for the test to exercise. Just worth a comment in the test body if you want it to be self-documenting, but it's not required.

Happy to address any remaining feedback before merge.

dozyio:
@lodekeeper it seems test are failing after merging latest main branch - could you investigate?

lodekeeper:
@dozyio Investigated — pushed `cbc1f2e2d`.

**Root cause:** the merge of `main` brought in #3435 (commit `5917715d0c`), which wraps `processSendQueue()` in `queueMicrotask(...)` inside the drain handler to prevent a `Maximum call stack size exceeded` error. After that change, the two drain-race tests added in `17d9da8ab` (`should preserve queued bytes when sendData throws StreamStateError during drain` and `should not duplicate queued bytes after transient StreamStateError during drain`) dispatch a synchronous `drain` event and then immediately assert on `stub.calledOnce` / `writeBufferLength`, but `sendData` is now deferred to a microtask — so the stub has been called 0 times when the assertion fires (`expected false to be true` at line 495).

**Fix:** await a microtask flush (`await new Promise<void>((resolve) => { queueMicrotask(resolve) })`) between `dispatchEvent('drain')` and the post-drain assertions. Preserves test intent (sendData is still invoked, bytes are still preserved / requeued); just adapts to the deferred scheduling. No source change needed.

packages/utils/src/abstract-message-stream.ts
```diff
@@ -479,7 +488,24 @@ export abstract class AbstractMessageStream<Timeline extends MessageStreamTimeli
...
+          sendResult = this.sendData(toSend)
+        } catch (err: any) {
+          // the underlying transport may have closed between the drain event
+          // and this send attempt - treat as a failed send rather than letting
+          // the error propagate as an uncaught exception
+          if (err.name === 'StreamStateError') {
+            this.log('send failed during queue processing, stream is %s - %e', this.writeStatus, err)
+            // Requeue the defensive copy in case sendData mutated `toSend`
+            // before throwing.
+            this.writeBuffer.prepend(willSend)
```
tabcat: i dont like this. imo the data should be dropped. a lot could happen inside of this.sendData before it throws

packages/utils/src/abstract-message-stream.ts
```diff
@@ -479,7 +488,24 @@ export abstract class AbstractMessageStream<Timeline extends MessageStreamTimeli
...
+            return false
```
tabcat: send should error if we were not able to send the data

packages/utils/src/abstract-message-stream.ts
tabcat: wrapping this with try/catch and logging the error will fix the original uncaught runtime error symptom on its own.

tabcat, line 455: the write guard should go here (line 460), log, return false.

packages/utils/src/abstract-message-stream.ts
```diff
@@ -96,6 +96,15 @@ export abstract class AbstractMessageStream<Timeline extends MessageStreamTimeli
...
       if (this.writableNeedsDrain) {
+        // a drain event can arrive after writeStatus transitions to
+        // 'closing'/'closed' during connection teardown - bail out to
+        // avoid calling send() on a non-writable stream which would
+        // throw an uncaught StreamStateError
+        if (this.writeStatus !== 'writable' && this.writeStatus !== 'closing') {
+          this.log.trace('not processing send queue on drain as stream write status is %s', this.writeStatus)
+          return
+        }
+
```
tabcat: should live in processSendQueue

packages/utils/src/abstract-message-stream.ts
```diff
@@ -479,7 +488,24 @@ export abstract class AbstractMessageStream<Timeline extends MessageStreamTimeli
...
+            this.log('send failed during queue processing, stream is %s - %e', this.writeStatus, err)
+            // Requeue the defensive copy in case sendData mutated `toSend`
+            // before throwing.
+            this.writeBuffer.prepend(willSend)
+            return false
+          }
+
+          throw err
+        }
+
```
tabcat, line 492: should not be wrapped and return false on failure from sendData. can catch the uncaught runtime error by wrapping the call in continueSendingOnDrain

tabcat (commented):
Thanks for the work @lodekeeper 
I'm going to get this over the finish line for you. Feel free to leave comments but edits are not requested from you at this time.

lodekeeper (commented):
Understood — thanks @tabcat, appreciate you taking it over. Happy to answer any questions about the original implementation if context would help. Will keep an eye on the thread.

tabcat:
@lodekeeper care to review the current diff with main?

lodekeeper:
Reviewed the current diff against main. The implementation matches the design notes from your inline comments:

- **L453–457** — writeStatus guard moved into `processSendQueue` (returns `false` and logs), addresses the `should live in processSendQueue` / write-guard-on-line-460 notes.
- **L100–110** — drain-handler microtask wrapped in try/catch with `log.error`, addresses `wrapping this with try/catch and logging the error will fix the original uncaught runtime error symptom`.
- **Inner `sendData` try/catch removed** — `send()` now propagates `StreamStateError` consistently with the existing closed/closing check (matches `send should error if we were not able to send the data` + `should not be wrapped`).

Tests cover the three resulting paths:
1. drain on closed stream → no uncaught error
2. drain mid-closing while `sendData` throws `StreamStateError` → caught by microtask wrapper
3. direct `send()` when `sendData` throws → error propagates to caller

`build` and most `Examples` checks are green; `test-node` / browser / interop matrix still running, no failures so far. `mergeable: MERGEABLE`, status `UNSTABLE` only because CI is in progress. LGTM from my side — happy to address anything else if reviewers flag something.

PR #3425, opened 2026-03-31
fix(kad-dht): getClosestPeers timeout

dozyio:
## Description

- Move routing-table updates off the `getClosestPeers` hot path so query progress/event streaming is not blocked.
- Centralise update scheduling in `RoutingTable` and use a dedicated `PeerQueue` with per-peer dedupe (`find`/`join`).
- Apply the same queued update path for topology-triggered updates (`onPeerConnect`).
- Add configurable queue concurrency via `routingTableUpdateQueueConcurrency`.
- Expose queue depth metrics via queue metric wiring for Grafana.
- Update `getClosestPeers` to yield query results directly (no peer filtering/mutation).
- Wire missing `networkDialTimeout` (thanks @paschal533).

## Change checklist


dozyio:
@achingbrain any chance you could take a look at this one please?

packages/kad-dht/src/peer-routing/index.ts
```diff
@@ -275,9 +275,15 @@ export class PeerRouting {
...
+        if (!(await this.components.connectionManager.isDialable(peer.multiaddrs, {
+          signal: options.signal
+        }))) {
+          continue
+        }
+
         yield finalPeerEvent({
           from: this.components.peerId,
-          peer: await self.components.peerStore.getInfo(peer.id, options),
+          peer,
```
achingbrain: 👍

achingbrain (changes requested):
A few notes:

- Maybe make the routing table `.add` method synchronous and encapsulate the update queuing logic inside the routing table class instead.  The query manager seems like the wrong place to me.
- If you use the `PeerQueue` class from `@libp2p/utils` instead of the generic `Queue` it will deduplicate jobs based on the passed PeerId for you (call `.find(peerId)` on the queue and then `.join(opts)` on the job if present).
- Pass a `metrics` instance and `metricName` option to the `Queue`/`PeerQueue` constructor to have it show up in graphana (essential to tune parameters)
- You can continue to use a queue after `.abort()` is called on it, no need to have the queue field be `Queue | undefined` in the containing class - just create it in the constructor as normal
- Most of the routing table stats would be available to graphana via the built in metrics, it seems redundant to hold them in memory
- The routing table stats seem to be most useful for tests - consider using sinon spys instead
- There seems to be some duplicate logic around when to perform routing table updates - this is defined by the spec and the routing table already has an interval during which a peer will not be re-pinged
- "Routing" is an overloaded term in libp2p so the new config options should be fully qualified (e.g. routing*Table*UpdateQueueConcurrency)

dozyio:
Thanks @achingbrain for the feedback - Hopefully addressed all the issues although I didn't `make the routing table .add method synchronous` - would require a fair amount of change as many of the internals are async.

Also reduced the scope a bit by removing the peer TTL stuff and custom queue metrics.

packages/kad-dht/src/routing-table/index.ts
```diff
@@ -253,11 +262,48 @@ export class RoutingTable extends TypedEventEmitter<RoutingTableEvents> implemen
...
+      } finally {
+        if ('clear' in signal) {
+          (signal as any).clear()
+        }
+      }
+    }, {
+      peerId,
+      signal: options.signal
+    }).catch(err => {
+      if (this.shutdownController.signal.aborted || (err as any)?.name === 'AbortError') {
```
achingbrain: I think `err` is already of type `any` so the cast is unnecessary.

```suggestion
      if (this.shutdownController.signal.aborted || err?.name === 'AbortError') {
```

achingbrain (approved):
LGTM, tested locally & is much faster.

Couple of small comments inline.

packages/kad-dht/src/routing-table/index.ts
```diff
@@ -253,11 +262,48 @@ export class RoutingTable extends TypedEventEmitter<RoutingTableEvents> implemen
...
+      void existingJob.join(options).catch(() => {})
+      return
+    }
+
+    void this.routingTableUpdateQueue.add(async (jobOptions) => {
+      const signal = jobOptions.signal == null
+        ? this.shutdownController.signal
+        : anySignal([this.shutdownController.signal, jobOptions.signal])
+
+      setMaxListeners(Infinity, signal)
```
achingbrain: You can pass `null | undefined` signals into `anySignal` so it can be simplified and the later cast removed:

```suggestion
      const signal = anySignal([this.shutdownController.signal, jobOptions.signal])
      setMaxListeners(Infinity, signal)
```

packages/kad-dht/src/routing-table/index.ts
```diff
@@ -253,11 +262,48 @@ export class RoutingTable extends TypedEventEmitter<RoutingTableEvents> implemen
...
+
+      try {
+        await this.add(peerId, {
+          signal
+        })
+      } finally {
+        if ('clear' in signal) {
+          (signal as any).clear()
+        }
```
achingbrain: The cast can be removed if [this](https://github.com/libp2p/js-libp2p/pull/3425/changes#r3084545356) is applied.

```suggestion
        signal.clear()
```

packages/kad-dht/src/routing-table/index.ts
```diff
@@ -253,11 +262,48 @@ export class RoutingTable extends TypedEventEmitter<RoutingTableEvents> implemen
...
+      }
+    }, {
+      peerId,
+      signal: options.signal
```
achingbrain: The composite signal needs to be passed to the job.

```suggestion
      signal
```

packages/kad-dht/src/routing-table/index.ts
```diff
@@ -119,6 +122,12 @@ export class RoutingTable extends TypedEventEmitter<RoutingTableEvents> implemen
     this.shutdownController = new AbortController()
     setMaxListeners(Infinity, this.shutdownController.signal)
 
+    this.routingTableUpdateQueue = new PeerQueue<void>({
+      concurrency: init.routingTableUpdateQueueConcurrency ?? ROUTING_TABLE_UPDATE_QUEUE_CONCURRENCY,
+      metricName: `${init.metricsPrefix}_routing_table_update_queue`,
+      metrics: this.components.metrics
+    })
```
achingbrain, line 137: Should also allow configuring the [maxSize](https://libp2p.github.io/js-libp2p/interfaces/_libp2p_utils.QueueInit.html#maxsize) parameter otherwise this can grow in an unbounded fashion.

packages/kad-dht/src/routing-table/index.ts
```diff
@@ -277,19 +284,26 @@ export class RoutingTable extends TypedEventEmitter<RoutingTableEvents> implemen
     }
 
     void this.routingTableUpdateQueue.add(async (jobOptions) => {
-      const signal = anySignal([this.shutdownController.signal, jobOptions.signal])
-      setMaxListeners(Infinity, signal)
+      let addOptions = jobOptions
+      let signal: ReturnType<typeof anySignal> | undefined
```
achingbrain: You can `import type { ClearableSignal } from 'any-signal'` instead of deriving the return type.

packages/kad-dht/src/routing-table/index.ts
dozyio: Had an issue with `onPeerConnect` timing out when there is a backlog of jobs to process - thought maybe better to pass the `onPeerConnect` value and set the timeout when actually connecting

achingbrain:
Is this ready for release?  It might explain why the service worker gateway has such a strong preference for HTTP providers over libp2p/bitswap.

dozyio:
@achingbrain - should be ok, merging. Thanks for the reviews

PR #3426, opened 2026-03-31
fix(kad-dht): reprovide CIDs in Kademlia key order to reduce new dials

paschal533:
## Summary

During a reprovide run, the reprovider iterates all stored CIDs and calls `provide()` for each one. Each `provide()` opens connections to the K closest peers for that CID. Because CIDs are processed in datastore iteration order (essentially random), each CID's K closest peers are likely different, so every CID requires a fresh set of dials.

This ports the [SweepingProvider optimisation from go-libp2p](https://github.com/libp2p/go-libp2p/pull/2774): collect all CIDs that need reproviding, sort them by their Kademlia key, then queue them in that order.

XOR-adjacent CIDs share the same K closest peers. By processing them consecutively, the connections opened for one CID are still live when the next CID is queued, so they get reused instead of requiring new dials. Over a full reprovide run with many CIDs this significantly reduces the total number of new connections opened.

## Changes

- `src/reprovider.ts`: collect CIDs into an array before queueing, sort by Kademlia key, then queue in sorted order
- `test/reprovider.spec.ts`: inserts 5 CIDs in reverse Kademlia key order and verifies `contentRouting.provide` is called in ascending Kademlia key order

## Test plan

- `should reprovide in Kademlia key order`  inserts CIDs in reverse order, verifies provide calls arrive in correct ascending Kademlia key order
- All 150 existing tests continue to pass
- Lint and TypeScript build clean

### DIFF
--- a/packages/kad-dht/src/reprovider.ts
+++ b/packages/kad-dht/src/reprovider.ts
@@ -1,7 +1,7 @@
 import { AdaptiveTimeout, Queue } from '@libp2p/utils'
 import drain from 'it-drain'
 import { TypedEventEmitter, setMaxListeners } from 'main-event'
-import { PROVIDERS_VALIDITY, REPROVIDE_CONCURRENCY, REPROVIDE_INTERVAL, REPROVIDE_MAX_QUEUE_SIZE, REPROVIDE_THRESHOLD, REPROVIDE_TIMEOUT } from './constants.ts'
+import { PROVIDERS_VALIDITY, REPROVIDE_CONCURRENCY, REPROVIDE_INTERVAL, REPROVIDE_MAX_QUEUE_SIZE, REPROVIDE_SORT_BATCH_SIZE, REPROVIDE_THRESHOLD, REPROVIDE_TIMEOUT } from './constants.ts'
 import { convertBuffer, parseProviderKey, readProviderTime, timeOperationMethod } from './utils.ts'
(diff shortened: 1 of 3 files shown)


packages/kad-dht/src/reprovider.ts
```diff
@@ -116,11 +116,20 @@ export class Reprovider extends TypedEventEmitter<ReprovideEvents> {
...
+   * same K closest peers, connections opened for one CID are likely to be
+   * reused for the next, reducing the number of new dials per reprovide run.
    */
   private async processRecords (options?: AbortOptions): Promise<void> {
     try {
       this.safeDispatchEvent('reprovide:start')
       this.log('starting reprovide/cleanup')
+
+      // collect CIDs that need reproviding so we can sort them before queueing
+      const toReprovide: CID[] = []
```
achingbrain: Need to do this in a way that does not let this array grow in an unbounded way otherwise it will OOM with large CID collections

paschal533: Fixed. I replaced the unbounded array with a fixed-size batch approach (sortBatchSize, default 512). CIDs are sorted in 512-entry windows; each batch is flushed before the next begins, bounding peak memory to O(sortBatchSize). Added a test that verifies within-batch Kademlia ordering with sortBatchSize: 2

PR #3428, opened 2026-04-01
feat: add DialTarget type

achingbrain:
libp2p can dial `PeerId | Multiaddr | Multiaddr[]` and this is repeated in several places.

Add a `DialTarget` convenience type that is a union of the above.

This makes the code more concise and removes the need for some modules to have `@multiformats/multiaddr` as a production dep.

## Change checklist


dozyio:
Thanks @achingbrain

PR #3429, opened 2026-04-01
feat: add high-level connection and stream progress events

achingbrain:
Where an operation might involve multiple libp2p operations - opening connections and streams, it is useful to be able to track which connections and streams are being opened, to which peers and how long it takes using the same `onProgress` callback.

Add 4x new progress events with enough context in the event object to track these activities during calls to `libp2p.dial`, `libp2p.dialProtocol` and `connection.newStream`:

- `connection:open`
- `connection:opened`
- `connection:open-stream`
- `connection:opened-stream`

Uses `connection:xxx` event names to not tie the event names to internal libp2p architecture (e.g. `connection:open` instead of `connection-manager:open-connection`).

Updates libp2p interface and relevant protocols to use progress event types that contain the relevant events.

## Change checklist


achingbrain:
Depends on:

- [x] https://github.com/libp2p/js-libp2p/pull/3428

dozyio:
Thanks @achingbrain

PR #3435, opened 2026-04-04
test: fix firefox muxer stress test

dozyio:
## Description

Replaced generic drain-event waiting with the stream-native backpressure API in `packages/interface-compliance-tests/src/stream-muxer/stress-test.ts`:
- Changed `await pEvent(outboundStream, 'drain', { rejectionEvents: ['close'] })` to `await outboundStream.onDrain()`.
- This preserves behaviour (wait until writable; reject if stream closes/resets) while avoiding high per-iteration listener/promise overhead in the stress loop.
- Wrapped `this.processSendQueue` in `queueMicrotask` to avoid blowing up the stack.

Observed impact:
- Firefox: ~27–30s down to 3-4s.
- Chromium: 2-2.5s down to 1.5-2s

## Change checklist


dozyio:
Without patch Firefox taking ~30mins to run two tests in CI
 
@libp2p/mplex:         ✅ 1000 stream(s) with 100 msg(s) (685417ms) ~ 11 minutes
@libp2p/integration-tests:         ✅ 1000 stream(s) with 100 msg(s) (1105832ms) ~ 18 minutes

dozyio:
@achingbrain any thoughts on this one?

packages/interface-compliance-tests/src/stream-muxer/stress-test.ts
```diff
@@ -51,9 +51,7 @@ async function spawn (createMuxer: (maConn: MultiaddrConnection) => Promise<Stre
       const sendMore = outboundStream.send(buf)
 
       if (sendMore === false) {
-        await pEvent(outboundStream, 'drain', {
-          rejectionEvents: ['close']
-        })
+        await outboundStream.onDrain()
```
achingbrain: 👍 though it should probably be done everywhere `pEvent` is used (including the v2 -> v3 upgrade guide and a note should probably be added stating the pros/cons of each way).

`stream.onDrain()` also accepts abort options so this should be passed in where available.

packages/utils/src/abstract-message-stream.ts
```diff
@@ -98,7 +98,10 @@ export abstract class AbstractMessageStream<Timeline extends MessageStreamTimeli
       if (this.writableNeedsDrain) {
         this.log.trace('drain event received, continue sending data')
         this.writableNeedsDrain = false
-        this.processSendQueue()
+
+        queueMicrotask(() => {
```
achingbrain, line 102: This will add latency to streaming.  Is it necessary?  Have you seen the stack overflowing?

achingbrain (changes requested):

achingbrain:
Good catch, notes inline.

packages/utils/src/abstract-message-stream.ts
dozyio: Yeah stack blows up on chrome webworkers

achingbrain (approved):

PR #3436, opened 2026-04-06
test: change RSA to Ed25519 in "can rotate keychain passphrase" test

dozyio:
## Description

Attempt to fix flakey `can rotate keychain passphrase` test by changing slow RSA key to Ed25519

## Change checklist


tabcat:
Can it really take ~10 seconds to generate an RSA keypair? I'd assume this timeout is somewhere else.

dozyio:
Could be the `iterationCount`? We should swap to Ed25519 anyway

PR #3437, opened 2026-04-06
test: fix flaky peer store tests

dozyio:
## Description

Attempt to fix https://github.com/libp2p/js-libp2p/discussions/3387#discussioncomment-16465406

## Change checklist


tabcat (approved):
looks like these expiry only necessary for other tests.

PR #3445, opened 2026-04-09
fix: propagate progress events and events before async work

achingbrain:
Fixes two bugs with the Queue class:

1. When joining an existing job, progress events were not received by the second or subsequent context(s)
2. If a progress event was emitted before any async work was done in a queue job under the concurrency threshold, it was missed because the job result recipient had not been registered yet

## Change checklist


packages/utils/src/queue/job.ts
```diff
@@ -77,7 +78,12 @@ export class Job <JobOptions extends AbortOptions = AbortOptions, JobReturnType
 
       const result = await raceSignal(this.fn({
         ...(this.options ?? {}),
-        signal: this.controller.signal
+        signal: this.controller.signal,
+        onProgress: (evt: any): void => {
+          this.recipients.forEach(recipient => {
+            recipient.onProgress?.(evt)
```
tabcat, line 84: what should happen if one of the recipients' onProgress throws?

achingbrain: It should crash, callbacks are not allowed to throw

tabcat (approved):

PR #3449, opened 2026-04-10
fix: support require

achingbrain:
Add the `module-sync` field to the exports map for each module to support synchronous loading from CJS environments using `require`.

This allows use by older projects that cannot or do not want to upgrade to ESM, or in TypeScript projects that compile to CJS knowingly or unknowingly.

See:

 - https://nodejs.org/en/blog/release/v22.10.0#2024-10-16-version-22100-current-aduh95
 - https://nodejs.org/api/packages.html#conditional-exports

## Change checklist

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -26,7 +26,8 @@
   "exports": {
     ".": {
       "types": "./dist/src/index.d.ts",
-      "import": "./dist/src/index.js"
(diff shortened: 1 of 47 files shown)


dozyio (approved):

dozyio:
thanks @achingbrain

PR #3460, opened 2026-04-17
fix(transport-webrtc): fix EOF race in readCandidatesUntilConnected

paschal533:
## Problem

A race condition caused the WebRTC \"simple\" transport compliance test to fail intermittently: the listener closes its signaling stream just before the initiator's `connectionstatechange: connected` event fires. The initiator then received EOF on the signaling stream and threw an `UnexpectedEOFError`, discarding a perfectly good peer connection.

**Root cause:** `connectedPromise` was declared inside the `try` block, making it inaccessible in the `catch` handler for stream read errors.

## Solution

Move `connectedPromise` outside the `try` block. In the `catch` handler, if the peer connection is not yet `'connected'`, await `connectedPromise`:
- If it resolves → the connection succeeded, ignore the stream error
- If it rejects → the connection failed, re-throw the original stream error

Also corrects the log message from \"error parsing ICE candidate\" to \"error reading ICE candidates\" for accuracy.

## Test plan

- [ ] Added `readCandidatesUntilConnected` describe block with 4 tests
- [ ] Test: throws `ConnectionFailedError` when the peer connection enters `'failed'` state
- [ ] Test: does not fail when peer connection is temporarily disconnected then recovers
- [ ] Test: throws `ConnectionFailedError` when peer connection goes disconnected then failed
- [ ] Test: returns without error when peer connection reaches `'connected'` state

## Related PRs

- **This PR** — `UnexpectedEOFError` race in `readCandidatesUntilConnected`
- fix(transport-webrtc): do not treat `'disconnected'` as a terminal peer connection state
- fix(interface-compliance-tests): increase abort propagation timeout in stream muxer close test

### DIFF
--- a/packages/transport-webrtc/src/private-to-private/util.ts
+++ b/packages/transport-webrtc/src/private-to-private/util.ts
@@ -16,10 +16,10 @@ export interface ReadCandidatesOptions extends AbortOptions, LoggerOptions, Prog
 }
 
 export const readCandidatesUntilConnected = async (pc: RTCPeerConnection, stream: MessageStream<Message, Stream>, options: ReadCandidatesOptions): Promise<void> => {
-  try {
-    const connectedPromise = Promise.withResolvers<void>()
-    resolveOnConnected(pc, connectedPromise)
+  const connectedPromise = Promise.withResolvers<void>()
+  resolveOnConnected(pc, connectedPromise)
 
+  try {
     // read candidates until we are connected or we reach the end of the stream
     while (true) {
       // if we connect, stop trying to read from the stream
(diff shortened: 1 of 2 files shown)


peerbit-org:
Found a cancellation gap in `4aded5fa`: after a read error, the recovery wait can outlive the dial signal if no further connection-state event arrives.

Prepared a focused follow-up: [9cef2fe9](https://github.com/peerbit-org/js-libp2p/commit/9cef2fe995db1dfdbab94a3c214ea23c9b288d76). It uses the existing `raceSignal` dependency and preserves the caller's abort reason.

The two regressions—abort during read and abort after signaling EOF—remain pending on the current proposal. The EOF-then-connected control passes both versions. With the follow-up, all three pass in Node and Firefox using isolated source builds; this is not a full monorepo validation claim. No new dependency, retry, or deadline increase.

PR #3461, opened 2026-04-17
fix(transport-webrtc): disconnected is not a terminal state

paschal533:
## Problem

The `onconnectionstatechange` handler and `resolveOnConnected` both treated `'disconnected'` the same as `'failed'` and `'closed'`, immediately tearing down the connection and rejecting the connected promise.

WebRTC's `'disconnected'` state is **transient** — the ICE agent may recover to `'connected'` without any intervention. Treating it as terminal caused two issues:

1. Connections that could have recovered were torn down prematurely.
2. In Firefox, the spurious `DataChannel` close during large data transfers raced against `sendCloseWrite`'s `pEvent(channel, 'close')` listener registration. If the channel closed before the listener was registered, the FIN_ACK was missed, the `finAckTimeout` fired, and all branches of `Promise.any` rejected with an `AggregateError`.

## Solution

Only treat `'failed'` and `'closed'` as terminal states in both:
- `RTCPeerConnectionMultiaddrConnection.onconnectionstatechange`
- `resolveOnConnected` in `util.ts`

## Test plan

- [ ] Existing WebRTC transport tests pass (Chrome, Firefox)
- [ ] No regression in connection teardown on genuine `'failed'` / `'closed'` states

## Related PRs

- fix(transport-webrtc): fix `UnexpectedEOFError` race in `readCandidatesUntilConnected`
- **This PR** — `'disconnected'` not terminal
- fix(interface-compliance-tests): increase abort propagation timeout in stream muxer close test

### DIFF
--- a/packages/transport-webrtc/src/private-to-private/util.ts
+++ b/packages/transport-webrtc/src/private-to-private/util.ts
@@ -86,7 +86,6 @@ function resolveOnConnected (pc: RTCPeerConnection, promise: DeferredPromise<voi
         promise.resolve()
         break
       case 'failed':
-      case 'disconnected':
       case 'closed':
         promise.reject(new ConnectionFailedError(`RTCPeerConnection connection state became "${pc.connectionState}"`))
         break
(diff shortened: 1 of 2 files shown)


dozyio (approved):

dozyio:
Thanks @paschal533

PR #3463, opened 2026-04-18
fix(libp2p): default abortConnectionOnPingFailure to false

marshalleq:
## Motivation

`@libp2p/connection-monitor` defaults to `abortConnectionOnPingFailure: true`. On any probe failure this calls `conn.abort(err)`, which cascades through the muxer (GoAway) and every multiplexed stream, and terminates the underlying MA-connection with a TCP RST on `@libp2p/tcp`.

Under normal WAN conditions — trans-continental RTT, intermittent event-loop contention on either peer, transient network jitter — a single 10s probe exchange failing is routine. The current default treats it as evidence the connection is dead and destroys it.

## Evidence

A five-host deployment across three continents (Germany, Australia, New Zealand) on libp2p 3.x with yamux keepalive enabled (`enableKeepAlive: true`, `keepAliveInterval: 10_000`). Baseline measurements per host over a 30-minute window:

| Host | Role | Disconnect rate |
|---|---|---|
| bootstrap (IONOS, DE) | VPS bootstrap | ~36/hr |
| bootstrap-de (Contabo, DE) | VPS bootstrap | ~42/hr |
| station3 (Contabo, AU) | VPS node | ~25/hr |
| station .11 (TrueNAS, NZ) | LAN guardian | ~48/hr |
| station .13 (LAN, NZ) | LAN guardian | ~62/hr |

Packet capture on the AU→NZ link confirmed the Node process itself sends the TCP RST mid-traffic, in the same event-loop tick as a preceding yamux GoAway burst. Instrumenting every abort call site in production produced stack traces rooted overwhelmingly at \`connection-monitor.ts\` → \`conn.abort(err)\`.

Setting \`abortConnectionOnPingFailure: false\` on all five hosts dropped the measured disconnect rate to **0/hr across all hosts** over multi-hour windows — no other change was required. Yamux keepalive handled liveness without the RST cascade. I do have a full writeup available of the hypotheses tested and their results to get to this point, but it's currently an offline copy in StreamResetInvestigation.md.  Please advise if required.

## Change

Flips \`DEFAULT_ABORT_CONNECTION_ON_PING_FAILURE\` from \`true\` to \`false\`. The option is still honoured when set explicitly. Applications that want the aggressive behaviour opt in rather than opt out.

## Tests

- Two existing tests that relied on the old default (\`should abort a connection that times out\`, \`should abort a connection that fails\`) now pass \`abortConnectionOnPingFailure: true\` explicitly — their intent was always to test the abort mechanism, not the default.
- New regression test \`should not abort a connection that fails by default\` guards against accidental regression of the new default.
- All 9 connection-monitor tests pass.

## Alternative considered

A more conservative middle-ground would be to require N consecutive probe failures before aborting (with N defaulting to e.g. 3), preserving the dead-connection-detection intent while eliminating the false-positive cascade. Happy to follow up with that if maintainers prefer — but changing the default to \`false\` is the simpler, strictly-safer change and matches what every deployment we're aware of ends up doing manually after hitting this problem.

### DIFF
--- a/packages/libp2p/src/connection-monitor.ts
+++ b/packages/libp2p/src/connection-monitor.ts
@@ -11,7 +11,14 @@ const PROTOCOL_VERSION = '1.0.0'
 const PROTOCOL_NAME = 'ping'
 const PROTOCOL_PREFIX = 'ipfs'
 const PING_LENGTH = 32
-const DEFAULT_ABORT_CONNECTION_ON_PING_FAILURE = true
+// Probe failures are not reliable evidence a connection is dead under real
+// WAN conditions — trans-continental RTT, event-loop contention, and transient
+// network jitter routinely cause a single 10s ping exchange to fail. Yamux
+// keepalive detects genuinely dead connections at the muxer layer without
+// tearing down the TCP socket, so aborting on a single probe failure is
+// measurable self-harm at non-trivial scale. Applications that want the
+// aggressive behaviour can still opt in by setting this to true.
+const DEFAULT_ABORT_CONNECTION_ON_PING_FAILURE = false
 
 export interface ConnectionMonitorInit {
   /**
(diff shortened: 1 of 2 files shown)


Faolain:
@achingbrain thoughts on this?

tabcat:
@Faolain Myself and @dozyio are primarly maintaining js-libp2p now. We're looking at this, thanks.

tabcat:
@marshalleq I've opened a new PR that changes how the connection manager closes stale connections. it handles network interruption and line blocking of the ping messages. Feedback would be appreciated. Thanks.

At this moment changing a default option is not on the table, but I want to make sure the UX is better here.

PR #3475, opened 2026-04-24
feat: add quic transport

achingbrain:
Adds a QUIC transport based on the upcoming Node.js PR [here](https://github.com/nodejs/node/pull/62876).

Currently you need to build a custom Node.js binary for the tests to pass using the instructions on the linked PR.

There are lots of `@ts-expect-error` where types do not align, and TODOs to be resolved before this is ready.

The code is based on the WebTransport and TCP transports, so expect some copy/paster and some options that are unused at the moment.  Also a lot of the protections in the TCP transport around stopping accepting new connections are not present.  These should really be pushed higher up into libp2p anyway so all transports can benefit from the protections.

Depends on:

- [x] https://github.com/libp2p/js-libp2p/pull/3476
- [ ] https://github.com/nodejs/node/pull/62876

## Change checklist


achingbrain:
This has been on my hard drive in various forms since last summer so now that the Node.js PR has all the required features it's time to get it somewhere more visible.

packages/transport-quic/src/stream.ts
```diff
@@ -0,0 +1,122 @@
...
+    while (data.byteLength > 0) {
+      const toWrite = Math.min(data.byteLength, this.writer.desiredSize ?? 0)
+
+      if (toWrite === 0) {
+        break
+      }
+
+      // TODO: have to copy before write otherwise error is thrown:
+      // TypeError: Provided key doesn't match [[ArrayBufferDetachKey]]
+      const bytes = data.subarray(0, toWrite).slice()
```
achingbrain, line 48: Ref: https://github.com/nodejs/node/pull/62876#issuecomment-4314102877

dozyio:
Running on macOS

```
brew install ninja
git clone https://github.com/nodejs/node.git
cd node
./configure --experimental-quic --ninja
./out/Release/node --experimental-quic -e "console.log(require('node:quic'))"
export PATH=$(pwd)/out/Release:$PATH
```

dozyio:
Added a small ts declarations file - should probably be removed in the future `packages/transport-quic/src/node-quic.d.ts`

PR #3477, opened 2026-04-24
chore: use random bytes for echo

achingbrain:
Ensures that where bufs are split for sending, they are reassembled in the correct order.

## Change checklist

### DIFF
--- a/packages/interface-compliance-tests/src/transport/index.ts
+++ b/packages/interface-compliance-tests/src/transport/index.ts
@@ -351,7 +351,7 @@ export default (common: TestSetup<TransportTestFixtures>): void => {
 
       const connection = await dialer.dial(dialAddrs[0])
 
-      const input = uint8ArrayConcat(new Array(160).fill(0).map(() => randomBytes(65535)))
+      const input = uint8ArrayConcat(new Array(160).fill(0).map(() => randomBytes(65536)))
       const output = await dialer.services.echo.echo(connection.remotePeer, input, {
         signal: AbortSignal.timeout(timeout)
       })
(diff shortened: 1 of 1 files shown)


tabcat:
was thinking of adding this before https://github.com/libp2p/js-libp2p/pull/3467

are you seeing issues? i was assuming libp2p ordered streams would catch it. looks like its passing :)

achingbrain:
More like a backfill.  It would have caught https://bugzilla.mozilla.org/show_bug.cgi?id=1983831 unfortunately that was observed in the wild.

achingbrain:
Also I wasn't sure if this was covered elsewhere in the test suite - feel free to close if you don't think it's worth merging.

tabcat:
>  It would have caught https://bugzilla.mozilla.org/show_bug.cgi?id=1983831

lets keep it. 

might as well do the same for many small writes as well.

packages/interface-compliance-tests/src/transport/index.ts
```diff
@@ -349,7 +351,7 @@ export default (common: TestSetup<TransportTestFixtures>): void => {
 
       const connection = await dialer.dial(dialAddrs[0])
 
-      const input = new Uint8Array(1024 * 1024 * 10).fill(5)
+      const input = uint8ArrayConcat(new Array(160).fill(0).map(() => randomBytes(65535)))
```
tabcat: if we want the same size should be 65536

packages/interface-compliance-tests/src/transport/index.ts
tabcat: might as well make this random too.

tabcat (approved):

PR #3479, opened 2026-04-24
chore: filter empty multiaddrs in identify protocol

### LINKED ISSUE #3478: Multiaddr "/" coming from peerStore
- **Subsystem**:

@multiformats/multiaddr
PeerStore

#### Severity:

Medium

#### Description:

https://github.com/libp2p/js-libp2p/pull/3400 
`InvalidParametersError: Multiaddr / was not an IPv4, IPv6, DNS, DNS4, DNS6 or DNSADDR address`
@dozyio noticed the / multiaddr in the error message. I believe this is an invalid Multiaddr and should not exist/be parsable.

Not sure yet. Possible suspects:

- https://github.com/libp2p/js-libp2p/blob/61d594a6c9af2b6f42b7d50d88891efbc8d76b27/packages/peer-store/src/utils/dedupe-addresses.ts#L50

dozyio:
## Description

Add a check for empty multiaddrs from identify to stop them being added to the peer store.

Fixes #3478

## Change checklist


tabcat:
**Issues:**
- `result.listenAddrs` at `utils.ts:172` still used raw `message.listenAddrs.map(buf => multiaddr(buf))`, so `peer:identify` consumers and the return value of `identify.identify()` still surfaced bare `/` to anything that read them.
- Malformed bytes would throw `InvalidMultiaddrError` at that same line *after* `peerStore.patch` had already run — partial-state failure, same crash class as #3400.

**Changes:**
- Hoisted the cleaned `listenAddrs` once and reused it for both `peer.addresses` and `result.listenAddrs`.
- Folded the empty check into `getCleanMultiaddr` so it returns `undefined` for zero-byte multiaddrs too (also helps `observedAddr` in `identify.ts:111`).
- Added a junk-byte (`Uint8Array.from([0xff, 0xff, 0xff])`) test case — the existing `new Uint8Array(0)` short-circuits in `getCleanMultiaddr` before reaching the try/catch.

tabcat (approved):

PR #3483, opened 2026-04-27
fix: emit self:peer:update on observed address removal

tabcat:
## Description

`AddressManager.removeObservedAddr` had inverted polarity vs.
`confirmObservedAddr`: `startingConfidence` was initialised to `false`,
the inner guards used the wrong sign, and the final check was negated.
`_updatePeerStoreAddresses()` was never invoked from the removal path.

Effect: callers that mark an address unreachable —
`packages/protocol-autonat-v2/src/client.ts`,
`packages/protocol-autonat/src/autonat.ts`,
`packages/transport-circuit-relay-v2/src/transport/listener.ts` —
removed the address from internal state but never patched the peer
store, so `self:peer:update` did not fire and downstream consumers
(DHT mode switching, identify-push, application subscribers) could not
react.

Fix mirrors the polarity used by `confirmObservedAddr`.

## Change checklist

### DIFF
--- a/packages/libp2p/test/addresses/address-manager.spec.ts
+++ b/packages/libp2p/test/addresses/address-manager.spec.ts
@@ -226,6 +226,33 @@ describe('Address Manager', () => {
     expect(peerStore.patch).to.have.property('callCount', 1)
   })
 
+  it('should update the peer store when a verified observed address is removed', async () => {
+    const ma = '/ip4/123.123.123.123/tcp/39201'
+    const am = new AddressManager({
+      peerId,
+      transportManager: stubInterface<TransportManager>({
+        getAddrs: Sinon.stub().returns([]),
+        getListeners: Sinon.stub().returns([])
+      }),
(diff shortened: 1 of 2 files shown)


packages/libp2p/test/addresses/address-manager.spec.ts
```diff
@@ -226,6 +226,33 @@ describe('Address Manager', () => {
...
+      }),
+      peerStore,
+      events,
+      logger: defaultLogger()
+    })
+
+    am.addObservedAddr(multiaddr(ma))
+    am.confirmObservedAddr(multiaddr(ma))
+
+    await delay(1500)
```
dozyio: maybe `pWaitFor` instead of `delay`?

tabcat: we should do this a few other places delay is called. could save some time.

dozyio (approved):

PR #3487, opened 2026-05-04
fix(gossipsub): use monotonic seqnos

tabcat:
## Summary

- Replace `randomBytes(8)` in the `Signing` branch of `buildRawMessage` with a module-scope counter seeded from `Date.now()` in nanoseconds, written big-endian.
- Mirrors [go-libp2p-pubsub](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go) and [rust-libp2p](https://github.com/libp2p/rust-libp2p/blob/master/protocols/gossipsub/src/behaviour.rs), which both seed from a high-res timestamp and atomically increment per message.
- Adds a unit test asserting strictly increasing big-endian uint64 seqnos.

## Why

The [pubsub spec](https://github.com/libp2p/specs/blob/master/pubsub/README.md#the-message) requires `seqno` to be a "linearly increasing 64-bit big-endian uint". `@libp2p/gossipsub` was the last major impl still emitting random seqnos, which made it non-compliant.

This started causing real interop pain: kubo 0.40+ enables `BasicSeqnoValidator` by default, and that validator drops any message whose seqno isn't strictly greater than the previously-seen max from that peer. With random 64-bit seqnos, ~50% of consecutive messages from a JS publisher are dropped at any given kubo peer; over a mesh, the reproducer below sees ~80% end-to-end loss.

- Spec violation analysis (lidel): chainsafe/js-libp2p-gossipsub#545
- Kubo-side investigation: ipfs/kubo#11304
- Reproducer (helia↔kubo loses ~80%, helia↔helia clean): https://github.com/Rinse12/kubo_helia_libp2p_gossipsub_bug_reproductions

## Notes on the seed

`Date.now() * 1_000_000n` gives sub-microsecond headroom seeded from wall clock. Across a process restart this still produces seqnos larger than the previous batch (assuming wall clock doesn't move backwards), matching the go/rust behavior. `performance.now()` was rejected for the seed because it starts near 0 each process — restarts would reset seqnos and trip the validator on the receiver.

The receive/validation path (`validateToRawMessage`) is intentionally untouched — JS already accepts any 8-byte seqno, which is fine. Adding a `BasicSeqnoValidator`-equivalent on the JS side is a separate change.

## Test plan

- [x] New unit test `buildRawMessage > Signing seqno > produces strictly increasing big-endian uint64 seqnos` (publishes 100 messages, asserts strict monotonicity).
- [x] `npm run test:node` in `packages/gossipsub` — 94 passing, 21 pending, no regressions.
- [x] `npm run lint` clean.
- [x] `npm run dep-check` clean.

### DIFF
--- a/packages/gossipsub/src/utils/buildRawMessage.ts
+++ b/packages/gossipsub/src/utils/buildRawMessage.ts
@@ -1,4 +1,3 @@
-import { randomBytes } from '@libp2p/crypto'
 import { publicKeyFromProtobuf } from '@libp2p/crypto/keys'
 import { peerIdFromMultihash } from '@libp2p/peer-id'
 import * as Digest from 'multiformats/hashes/digest'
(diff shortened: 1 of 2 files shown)


tabcat:
Closing — opened prematurely, will reopen after I'm ready.

tabcat:
Think we should wait on adding Seqno tracking per peer and get this working with go impl.

dozyio (approved):

PR #3488, opened 2026-05-05
feat(utils): add multiaddr sort comparators

tabcat:
## Description

Move the four multiaddr-only sort comparators (`reliableTransportsFirst`, `loopbackAddressLast`, `publicAddressesFirst`, `circuitRelayAddressesLast`) from `packages/libp2p/src/connection-manager/address-sorter.ts` into `@libp2p/utils`, retyped to take `Multiaddr` directly. Also adds a `defaultMultiaddrSorter(multiaddrs: Multiaddr[])` convenience that chains the four in the standard priority order (loopback last, public first, relay last, reliable transports as innermost tiebreaker).

The `Address`-aware composition (`certifiedAddressesFirst`, `defaultAddressSorter`) stays in libp2p and now feeds the utils comparators through a small `asMultiaddr` adapter. Sort order is preserved verbatim — same five-key priority chain in `defaultAddressSorter`. `dial-queue.ts` import path is unchanged. The `AddressSorter` interface in `@libp2p/interface` is unchanged.

`@libp2p/utils` gains a new dependency on `@multiformats/multiaddr-matcher`.

This unblocks #3423, which puts the addresses we send in identify messages in priority order.

## Notes & open questions

- I couldn't think of a solution that would keep the logic simple and allow `defaultAddressSorter` to consume `defaultMultiaddrSorter`. `certifiedAddressesFirst` needs to slot at priority 4 inside the chain, which doesn't fit cleanly into a no-options `defaultMultiaddrSorter`.

## Change checklist


packages/utils/src/multiaddr/sorters.ts
```diff
@@ -0,0 +1,147 @@
+import { Circuit, WebSockets, WebSocketsSecure, WebRTC, WebRTCDirect, WebTransport, TCP } from '@multiformats/multiaddr-matcher'
+import { isLoopback } from './is-loopback.ts'
+import { isPrivate } from './is-private.ts'
+import type { Multiaddr } from '@multiformats/multiaddr'
+
+/**
+ * Sorts addresses by order of reliability, where they have presented the fewest
+ * problems:
+ *
+ * TCP -> WebSockets/Secure -> WebRTC -> WebRTCDirect -> WebTransport
```
tabcat, line 10: wondering if this needs to be updated.

packages/utils/src/multiaddr/sorters.ts
```diff
@@ -0,0 +1,147 @@
...
+ * Sort multiaddrs by the default multiaddr-only ordering: loopback addresses
+ * last, public addresses first, circuit-relay addresses last, with reliable
+ * transports as the innermost tiebreaker.
+ */
+export function defaultMultiaddrSorter (multiaddrs: Multiaddr[]): Multiaddr[] {
+  return multiaddrs.sort((a, b) =>
+    loopbackAddressLast(a, b) ||
+    publicAddressesFirst(a, b) ||
+    circuitRelayAddressesLast(a, b) ||
+    reliableTransportsFirst(a, b)
```
tabcat, line 145: equivalent to .sort chain (although primary key order changes). could make this change to defaultAddressSorter.

packages/utils/src/multiaddr/sorters.ts
dozyio: think this is still the best order

dozyio (approved):

PR #3489, opened 2026-05-05
fix(transport-webtransport): abort connection when session dies

### LINKED ISSUE #1982: WebTransport throws NetworkError when creating a new stream on a (very) old connection
I opened a WebTransport connection to a peer, closed my laptop and forgot about it.

The next day I opened my laptop, the node still thinks it has a connection, `connection.newStream()` calls `.createBidirectionalStream()` on the `WebTransport` object which throws `DOMException: Failed to execute 'createBidirectionalStream' on 'WebTransport': No connection.`.

We should handle this error.  Probably abort the connection with the error?

paschal533:
## Problem

When a WebTransport session becomes invalid after a network interruption or laptop sleep/resume, `connection.newStream()` calls `createBidirectionalStream()` on the dead `WebTransport` object which throws:

```
DOMException: Failed to execute 'createBidirectionalStream' on 'WebTransport': No connection.
```

Three independent gaps caused the stale connection to remain in the connection manager indefinitely:

1. **`muxer.ts` `onCreateStream()`** — the DOMException from `createBidirectionalStream()` propagated unhandled; neither the muxer nor the connection were cleaned up.
2. **`muxer.ts` incoming stream reader** — errors from the `incomingBidirectionalStreams` reader were only logged; no cleanup triggered.
3. **`index.ts` `wt.closed` rejection** — `cleanUpWTSession()` only set `maConn.timeline.close = Date.now()`. This never dispatches a `close` event on `maConn`, so the connection manager was never notified even when `wt.closed` eventually fired.

### `src/muxer.ts`

**`onCreateStream()`** — wrap `createBidirectionalStream()` in try/catch. On error, call `this.abort(err)` (all open streams) and `this.maConn.abort(err)` before rethrowing. `maConn.abort()` triggers: `sendReset()` → `cleanUpWTSession()` → `wt.close()`, then dispatches a `StreamAbortEvent` ('close' event) on `maConn`. This propagates through `Connection` → upgrader → `connection:close` event bus → connection manager removes the entry.

**Incoming stream reader** — add `this.abort(err)` and `this.maConn.abort(err)` to the `.catch()` handler (same cleanup for the inbound path).

### `src/index.ts`

Split `.catch().finally()` into `.then()/.catch()`. In the `.catch()`, call `cleanUpWTSession('remote_close')` first (sets `closed = true` to prevent a double `wt.close()` call via `sendReset()`), then `maConn?.abort(err)` to fire the full close event chain.

### `test/muxer.spec.ts` (new)

Two unit tests using sinon-ts stubs:
- `onCreateStream` catches DOMException and aborts `maConn`
- Incoming stream reader failure aborts `maConn`

Closes #1982

## Test plan

- [ ] `cd packages/transport-webtransport && npx aegir test -t node` — both new muxer tests pass
- [ ] `npx aegir build` — no TypeScript errors
- [ ] Browser integration test (requires go-libp2p server): intentionally idle a connection and confirm that `newStream()` rejects cleanly and the connection is removed from the connection manager after resume


paschal533:
Closing in favour of a structural refactor per feedback from @tabcat — the new PR (#3498) aligns the WebTransport session lifecycle with the WebRTC transport pattern, where `WebTransportSessionMultiaddrConnection` directly owns the `WebTransport` session rather than delegating through a closure.

PR #3490, opened 2026-05-05
feat(autonat-v2): emit address reachability events

### LINKED ISSUE #3472: AutoNAT v2 does not expose reachability events
This issue distills findings from a systematic cross-implementation evaluation of AutoNAT v2, conducted by [ProbeLab](https://probelab.io/). The final report of the investigation is available in the following [link](https://github.com/probe-lab/autonat-perf-audit/blob/main/docs/final-report.md).

#### Severity:

Low — this is an API completeness / developer ergonomics gap. No runtime crash or incorrect reachability verdict; the information is tracked internally but not surfaced to applications.

#### Description:

**What I did.** Registered the `autoNATv2()` service on a js-libp2p Node.js node and attempted to subscribe to a reachability event from the service so the application could react to per-address REACHABLE / UNREACHABLE verdicts.

**What happened.** No event fires. `@libp2p/autonat-v2` tracks per-address outcomes internally in a `dialResults` Map but does not emit any event. No `dispatchEvent` / `safeDispatchEvent` calls exist anywhere in `packages/protocol-autonat-v2/`.

**What I expected.** Per-address reachability changes surfaced as events on the service, consistent with the rest of js-libp2p, where `TypedEventEmitter` is used extensively:

- `Libp2p` emits `peer:connect`, `peer:disconnect`, `peer:discovery`, `connection:open`/`close`.
- PeerStore emits `self:peer:update` and `peer:update`.
- Identify emits `peer:identify`.
- Circuit Relay v2 emits `relay:created-reservation`, `relay:removed`, `relay:found-enough-relays`.

AutoNAT v2 is the outlier with no event surface. Concrete consequences: applications cannot directly observe reachability changes (they can only *infer* from downstream PeerStore updates, and the per-address REACHABLE/UNREACHABLE granularity is lost); for nodes behind symmetric NAT running AutoNAT v2 over QUIC, every dial-back fails, the address is silently removed via `removeObservedAddr()` after 8 failures, and the application sees the address disappear without ever receiving an explicit UNREACHABLE signal.

AutoNAT v2 already influences downstream behavior indirectly: observed addresses are added by Identify as unverified, and only elevated to the PeerStore entry after AutoNAT v2 calls `confirmObservedAddr()` on the AddressManager ([`client.ts:592`](https://github.com/libp2p/js-libp2p/blob/main/packages/protocol-autonat-v2/src/client.ts#L592)). DHT mode switching (via `self:peer:update`) therefore already consumes an AutoNAT-verified address list. What is missing is only the API-level event surface for applications.

#### Proposal

Expose v2 probe results via `TypedEventEmitter`, consistent with the rest of the library:

- Per-address event: `reachability:changed { addr, state, confidence }` where state is `REACHABLE | UNREACHABLE | UNKNOWN`.
- Optional aggregated host-level event equivalent to go-libp2p's `EvtLocalReachabilityChanged`, derived from a per-address reducer.

tabcat:
## Description

Closes https://github.com/libp2p/js-libp2p/issues/3472.

Adds an event surface to `@libp2p/protocol-autonat-v2` so applications can
observe per-address reachability as AutoNAT probes it. The service emits three
events on its own event target, each carrying an `AddressReachabilityChange`
`{ addr, type, success, failure }` (the address, its type, and the running
success/failure dial tally for the current round):

- `address:verifying`: emitted as each peer contributes a dial result for an
  address, so a series of these tracks verification progress. A verdict is not
  guaranteed to follow.
- `address:reachable`: the address was confirmed to be externally dialable.
- `address:unreachable`: the address was confirmed not to be externally
  dialable.

Subscribe with:

`libp2p.services.autoNAT.addEventListener('address:reachable', (evt) => { ... })`

## Notes & open questions

- Events are dispatched on the service's own event target (consistent with
  Circuit Relay v2), not the main libp2p event bus.
- The event surface is a stateless projection of the probe results: no
  reachability state is retained. `address:verifying` fires per contributing
  peer (deduplicated per network segment); the verdict events carry the tally
  for the round.
- Near the connection limit a previously verified address is re-affirmed
  without a fresh probe, so `address:reachable` can carry a zero tally.
- An address disappearing is part of the AddressManager lifecycle and is
  observable via `self:peer:update` / `getMultiaddrs()`, so there is no
  separate "removed" event.

## Change checklist


tabcat:
@srene any feedback here is greatly appreciated.

dozyio (approved):

tabcat:
I've refactored this to just emit events. There is no saved state exposed.

PR #3495, opened 2026-05-10
deps: upgrade to aegir v48

tabcat:
## Description

aegir 48 ships TypeScript 6.x with a generic `Uint8Array` and runs Node tests directly against `.ts` sources without a build step.

Changes in this PR:

- bump aegir to `^48.0.4` across all `package.json` files (51)
- rewrite ~700 relative `.js` imports to `.ts` in `.ts` sources — Node strip-types resolves paths literally, so `.js` on a TS source breaks tests when no build step has run
- merge `@libp2p/crypto`'s node and browser webcrypto shims into one file using `globalThis.crypto` — the Electron 18 workaround that motivated splitting them is obsolete now that aegir ships Electron 41/42
- cast `Uint8Array` at WebCrypto / WebSocket / RTCDataChannel / X509 / WebTransport boundaries — TS 6's default `Uint8Array<ArrayBufferLike>` is no longer assignable to `BufferSource`
- import `Debugger` from `weald` in `@libp2p/logger` — the ambient `@types/debug` namespace is no longer auto-included now that aegir restricts `types: ["@types/mocha", "@types/node"]`
- implement `[Symbol.dispose]` / `[Symbol.asyncDispose]` on the OpenTelemetry generator wrappers — TS 6's `Generator` / `AsyncGenerator` interfaces require them

The bulk of the diff (~700 lines) is the mechanical `.js` → `.ts` import rewrite. Code changes are concentrated in `@libp2p/crypto`, `@libp2p/keychain`, `@libp2p/logger`, `@libp2p/metrics-opentelemetry`, `@libp2p/connection-encrypter-tls`, `@libp2p/transport-{websockets,webrtc,webtransport}`.

## Notes & open questions

There are **28 `as Uint8Array<ArrayBuffer>` casts** at WebCrypto / native-API call sites. Eliminating them would require either (a) tightening the function/iterator input type (e.g. `msg: Uint8Array<ArrayBuffer> | Uint8ArrayList` in `hashAndSign`, or `Uint8ArrayList` itself yielding `Uint8Array<ArrayBuffer>` from upstream) — a breaking change for callers that currently pass a generic `Uint8Array<ArrayBufferLike>` — or (b) copying via `new Uint8Array(buf)`, which adds a memcpy in hot paths (websocket/datachannel send, signing). Casts are the pragmatic answer; runtime is correct because libp2p never uses `SharedArrayBuffer`.

A follow-up PR will tighten `@libp2p/crypto`'s output types (e.g. `privateKeyToProtobuf`) to return `Uint8Array<ArrayBuffer>` — non-breaking (subtype) and lets downstream consumers drop `@ts-expect-error` workarounds when feeding crypto outputs into native APIs like `new Blob([...])`.

## Change checklist


tabcat:
@dozyio could split this up into separate prs, lmk

dozyio (approved):

achingbrain:
> There are 28 as Uint8Array<ArrayBuffer> casts at WebCrypto / native-API call sites. Eliminating them would require either (a) tightening the function/iterator input type (e.g. msg: Uint8Array<ArrayBuffer> | Uint8ArrayList in hashAndSign, or Uint8ArrayList itself yielding Uint8Array<ArrayBuffer> from upstream) — a breaking change for callers that currently pass a generic Uint8Array<ArrayBufferLike> — or (b) copying via new Uint8Array(buf), which adds a memcpy in hot paths (websocket/datachannel send, signing). Casts are the pragmatic answer; runtime is correct because libp2p never uses SharedArrayBuffer.

This is the wrong approach.  In general you want to accept `Uint8Array` as an arg but return `Uint8Array<ArrayBuffer>` wherever possible.

Internally use [withArrayBuffer](https://github.com/achingbrain/uint8arrays#witharraybufferbuf) from `uint8arrays` to convert the backing buffer if necessary and maintain type safety if not.

If the `Uint8Array` passed to `withArrayBuffer` is already backed by an `ArrayBuffer` it's returned unchanged so it's a no-copy operation almost all of the time since `SharedArrayBuffer` is seldom used.

achingbrain:
A heads up as well, `extract-zip` (used by aegir to unzip electron binaries during testing) is [broken in Node 26](https://github.com/max-mapper/extract-zip/issues/154) (and also [Node 24/lts](https://github.com/max-mapper/extract-zip/issues/154#issuecomment-4531953884) as of a few days ago).  There is an upstream fix in `yauzl@3.x.x` but `extract-zip` still depends on  `yauzl@2.x.x`.

Unfortunately `extract-zip` is [unmaintained](https://github.com/max-mapper/extract-zip/issues/71) so I don't expect a fix to land any time soon - there's a workaround in `aegir@48.0.11` so if you're seeing problems with electron testing in CI upgrading will fix it.

PR #3497, opened 2026-05-11
handle WebSocket implementations missing bufferedAmount

ernest-okot:
## Description

React Native's WebSocket implementation does not expose bufferedAmount (it's undefined). The current code in WebSocketMultiaddrConnection.sendData compares it directly with a number, which evaluates undefined < 4MB to false, marking every write as needing drain. The eventual 'close' event then makes pEvent reject with undefined, surfacing in Upgrader._encryptOutbound as the cryptic TypeError: Cannot read property 'message' of undefined.

Same guard applied to checkBufferedAmount, which otherwise starts a polling task that never terminates.

Reproducer: a React Native Expo project using @libp2p/websockets to dial a Node.js relay. Before the fix, every dial failed at `mss.select` with an undefined message error. After the fix, dial completes and noise handshake proceeds normally.
 
Refs
- React Native WebSocket source: https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/WebSocket/WebSocket.js — bufferedAmount is not implemented.
- MDN: bufferedAmount is required for browsers, but not all WebSocket implementations conform.

## Change checklist

### DIFF
--- a/packages/transport-websockets/src/websocket-to-conn.ts
+++ b/packages/transport-websockets/src/websocket-to-conn.ts
@@ -62,7 +62,14 @@ class WebSocketMultiaddrConnection extends AbstractMultiaddrConnection {
       this.websocket.send(buf)
     }
 
-    const canSendMore = this.websocket.bufferedAmount < this.maxBufferedAmount
+    // Some WebSocket implementations (notably React Native) don't expose
+    // `bufferedAmount`. When unavailable, treat it as 0; we can't track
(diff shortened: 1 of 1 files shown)


tabcat:
This would effectively remove all backpressure guards from the transport inside of React Native and any other environment where bufferedAmount is undefined. You're very likely to hit errors further down the road which are harder to debug.

Have you tried other transports for your usecase? I'm not able to find a good way to handle this atm.

PR #3504, opened 2026-05-13
feat(libp2p): tolerate transient ping failures

tabcat:
## Description

Connection-monitor currently aborts on the first failed ping. With the `AdaptiveTimeout` 5s floor, transient loss or backpressure on a healthy connection routinely trips the abort even though TCP would have recovered.

Supersedes #3463, which fixes the same problem by flipping the default to `false` (see it for the trans-Pacific disconnect evidence); this PR keeps the default but aborts only on genuine transport silence.

The default `abortConnectionOnPingFailure: true` is preserved but the abort is now gated on actual transport silence:

- **`@libp2p/interface`**: optional `lastReadAt?: number` added to `MessageStreamTimeline`; new `ConnectionStaleError` export.
- **`@libp2p/utils`**: `AbstractMessageStream.onData()` sets `timeline.lastReadAt = Date.now()` whenever bytes arrive — covers every transport in one line since they all route through this method.
- **`libp2p`**: connection-monitor adds `connectionStaleTimeout` (default 60s). On ping failure it aborts only if `Date.now() - (conn.timeline.lastReadAt ?? conn.timeline.open) > connectionStaleTimeout`. Aborts emit `ConnectionStaleError` so consumers can distinguish gate-driven aborts from transport-level failures.

Net effect: ping failures from heavy traffic on other streams, GC pauses, TCP retransmit bursts, or path-level jitter no longer kill healthy connections — only sustained silence does.

## Behavior change

`abortConnectionOnPingFailure` keeps its `true` default, but its meaning changes: previously a single failed ping aborted the connection; now a failed ping aborts only after the peer has sent no data for `connectionStaleTimeout` (default 60s). Deployments relying on `true` for immediate teardown will keep connections open longer when pings fail while data is still flowing. The option and its default are unchanged, so no config action is required.

## Notes & open questions

Some structural cleanups (drop `AdaptiveTimeout` from this file, simplify `pingTimeout`, drop `abortConnectionOnPingFailure`) become natural once the abort decision lives elsewhere — they're breaking and noted for v4.

## Change checklist


tabcat:
@marshalleq would you mind testing these changes and see if they work for you. it tolerates multiple ping failures and handles cases where the ping messages havent been decoded yet but data has been read from the stream. default stale connection timeout is 1 minute.

PR #3511, opened 2026-05-16
chore: fix gossipsub benchmarks

dozyio:
## Description

Fix gossipsub benchmarks so that they can run

## Change checklist


dozyio:
Might want to migrate off benchmark.js as it hasn't been updates in a long while

tabcat:
i think tinybench is new standard https://github.com/tinylibs/tinybench

dozyio:
Thinking to get this one in as it matches the rest of the repo and then migrate all off benchmark.js

tabcat:
- **Heartbeat was timing the publish, not `heartbeat()`.** The migration folded
    the old untimed `beforeEach` publish into the timed fn, so it measured
    ~275 ms/op instead of ~3.66 ms/op. Moved it to a one-time warm-up so it times
    `heartbeat()` alone.
  - **Node teardown** — the 50 nodes were never stopped, so their timers kept the
    process alive and the run hung after finishing. Added `stop()`.
  - **Deduped `runBenchmark`** → `test/utils/benchmark.ts` (was copied in all 4 files).
  - **`@types/benchmark`** added, dropped the `@ts-expect-error`.
  - **`mocha` → `aegir test -t node`** (mocha wasn't a direct dep), kept `--exit`
    as a safety net.

tabcat (approved):

PR #3513, opened 2026-05-17
chore: pull_request_target -> pull_request

dozyio:
## Description

Move from `pull_request_target` to `pull_request`

## Change checklist

### DIFF
--- a/.github/workflows/semantic-pull-request.yml
+++ b/.github/workflows/semantic-pull-request.yml
@@ -1,7 +1,7 @@
 name: Semantic PR
 
 on:
-  pull_request_target:
+  pull_request:
     types:
       - opened
       - edited


achingbrain:
From the docs:

> This event runs in the context of the default branch of the base repository, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

`pull_request_target` seems like the correct event for this action - why the change?

tabcat:
`Avoid using this event if you need to build or run code from the pull request.` but this workflow doesn't run code from the pr so pull_request_target was fine. The events are only unsafe if you checkout pr code with pull_request_target or add secrets that can be accessed by forks to a workflow using pull_request.

The only concrete difference i see is pull_request_target allows the Semantic PR check to run before maintainer approval.

Either is fine (workflow does not need write permission [pull_request] and does not checkout pr code [pull_request_target]). Kind of prefer pull_request_target for the semantic check without approval; it's also what release-please docs recommend.

PR #3523, opened 2026-05-29
fix: allow dialing WebTransport without PeerId

achingbrain:
PeerIds can be exchanged during the noise handshake so allow dialing a remote peer over WebTransport without a PeerId in the multiaddr.

Refs: https://github.com/ipfs/service-worker-gateway/pull/1097

## Change checklist


achingbrain:
CI failures need https://github.com/libp2p/js-libp2p/pull/3495

tabcat (approved):

Issue #3512, opened 2026-05-17
@libp2p/utils: AbstractMessageStream drain-microtask leaks StreamStateError as uncaughtException (regression in 7.1.0)


BlobMaster41:
**Version**

* `libp2p@3.3.0`
* `@libp2p/utils@7.2.0` (also reproduces on `7.1.0`)
* `@libp2p/interface@3.2.2`
* `@chainsafe/libp2p-yamux@8.0.1`
* `@chainsafe/libp2p-noise@17.0.0`

**Platform**: `Linux 6.8.0 x86_64`, Node.js `v25.8.0`

**Subsystem**: `@libp2p/utils`, `AbstractMessageStream`. Affects every yamux substream over noise (i.e. the default stack).

#### Severity

Critical. The process exits via `uncaughtException`. The trigger condition is backpressure followed by a close on the same connection, which is reachable by any peer, so this is also a remotely triggerable DoS against any node on the default libp2p stack on `libp2p >= 3.2.3`.

#### What happens

`AbstractMessageStream`'s `'drain'` listener defers `processSendQueue()` into a bare `queueMicrotask` with no error handler. If the underlying transport finishes closing between the `'drain'` dispatch and the queued microtask actually running, `processSendQueue` walks through yamux (`YamuxStream.sendData`, `YamuxMuxer.sendFrame`, `YamuxMuxer.send`) into `maConn.send` and throws `StreamStateError("Cannot write to a stream that is closed")`. Because the throw originates inside a detached microtask, no `try/catch` or `await` chain in user code can intercept it. It goes straight to Node's `uncaughtException`.

Two consequences for any node on this stack:

1. Hard crash. Reproduced by the first test group below.
2. Silent byte loss. Inside `processSendQueue`, `writeBuffer.consume(toSend.byteLength)` runs before `sendData(toSend)`. The `prepend` recovery only runs if `sendData` returns a partial result. On throw, control jumps to the surrounding `try { ... } finally { sendingData = false }` and the recovery never executes. Bytes are dropped from the queue, and the remote peer waits forever for the missing tail of a framed message. Reproduced by the second test group below.

If a downstream user installs an `uncaughtException` handler that logs and continues instead of exiting, there are further effects on long running processes: `writeBuffer` accumulates on the half broken stream until the muxer's close event finally tears it down, bounded by `maxWriteBufferLength` which defaults to `Infinity`. That's a downstream choice though, not something this bug causes directly.

#### Crash trace

```
[...]

PR #3527, opened 2026-05-30
fix(utils): abort message stream when sendData throws during drain flush

BlobMaster41:
## What

`AbstractMessageStream.processSendQueue` consumes bytes from the write buffer before it calls `sendData`:

```ts
this.writeBuffer.consume(toSend.byteLength)
const sendResult = this.sendData(toSend)
```

`sendData` can throw synchronously. The common trigger is the underlying transport closing between the `drain` event that schedules the flush and the microtask that actually runs it, at which point the write walks into a closed muxer or connection and throws `StreamStateError`.

## Background

This is the second half of #3512. That issue reported two consequences of deferring `processSendQueue` into a bare microtask:

1. The throw escaped as an `uncaughtException` and crashed the process.
2. The chunk consumed just before `sendData` was silently dropped, and the stream was left writable with no way to make progress.

The microtask guard shipped in `@libp2p/utils@7.2.1` (commit 8230eb5) fixed consequence 1: the throw is caught and logged. Consequence 2 is still present. The guard swallows the error, the consumed bytes are gone, and the stream stays writable.

## Change

Wrap the `sendData` call inside the send loop. On a throw, restore the consumed chunk to the write buffer and `abort` the stream, then rethrow so the existing microtask guard still logs it. A send failure now becomes an observable abort instead of a swallowed error on a stuck stream.

## Testing

Adds `packages/utils/test/abstract-message-stream.spec.ts`. It drives the drain microtask into a throwing `sendData` and asserts the stream ends up `aborted` rather than silently writable. I was not able to run the full monorepo suite locally, so CI is the source of truth for the wider package.

Refs #3512


packages/utils/src/abstract-message-stream.ts
```diff
@@ -488,7 +488,22 @@ export abstract class AbstractMessageStream<Timeline extends MessageStreamTimeli
...
+          sendResult = this.sendData(toSend)
+        } catch (err: any) {
+          // sendData can throw synchronously, for example when the underlying
+          // transport finishes closing between the 'drain' event that scheduled
+          // this microtask and the microtask running. The bytes for this chunk
+          // were already consumed from the write buffer above, so without this
+          // they would be silently dropped and the stream would be left in a
+          // writable state that can never make progress. Restore the chunk and
+          // abort the stream so the failure is observable instead of swallowed.
+          this.writeBuffer.prepend(willSend)
```
tabcat, line 498: correct for this reasoning: abort will only emit 'idle' if there are still bytes in the write buffer. 'idle' is not emitted from processSendQueue if sendData throws. and we need to emit 'idle' if the write buffer is emptied.

case that breaks without the prepend call: toSend consumes the entire writeBuffer -> sendData throws -> abort checks writeBuffer has bytes (it doesnt) -> skips emitting 'idle'

tabcat (approved):

tabcat:
Thanks!

Issue #2771, opened 2024-10-15
Dialling a peer and subscribing to a pubsub topic but "subscription-change" event does not always fire


haydenyoung:
- **Version**:
2.1.9

- **Platform**:
Linux 6.9.3-76060903-generic #202405300957~1721174657~22.04~abb7c06 SMP PREEMPT_DYNAMIC Wed J x86_64 x86_64 x86_64 GNU/Linux

- **Subsystem**:
Pubsub (GossipSub)

#### Severity:
Medium

#### Description:
OrbitDB has two peers connecting to one another and sharing messages using Libp2p and custom protocols. To initiate the message exchange, OrbitDB relies on the event subscription-change by attaching a listener to it. When fired, the listener will dial a protocol from peerA from peerB and initiate an exchange of data via streams.

When bootstrapping peerA's address in peerB, subscription-change fires reliably. However, when a discovery mechanism is used (we have been using mdns for local network discovery) and the peerB dials peerA manually, subscription-change becomes unreliable, sometimes firing, sometimes not.

Watching the [pipePeerReadStream function in GossipSub](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/d9e4374b568be6eb490cadc924f42576ed6cfea4/src/index.ts#L1036), it appears that  the [stream doesn't contain anything and the data handling is completely skipped](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/d9e4374b568be6eb490cadc924f42576ed6cfea4/src/index.ts#L1039), giving the appearance that the event doesn't fire when, in fact, it does but it skips over much of the function because there is no data to read.

#### Steps to reproduce the error:

Prerequisites:

```
git clone https://github.com/orbitdb/orbit-db-pinner.git
cd orbit-db-pinner
git checkout pinner/new-refactor-2-helia-upgrade
npm i
```

To see a successful test run:

In terminal 1 start the relay:
```
npm run start:relay
```

In terminal 2 start the first pinner server:
```
npm run start:orbiter1
```
In terminal 3 start the second pinner server:
```
npm run start:orbiter2
```

In terminal 4, run the unit tests:
```
npm run test
```
or
```
npm run test:ci
```

All tests should run successfully.

To see subscription-change fail:

Kill Orbiter1 and Orbiter2

[...]

PR #3531, opened 2026-06-04
fix(gossipsub): remove closed outbound streams from the registry

tabcat:
## Description

When an outbound stream closed (reset, transport drop, remote close), the closed stream was left in `streamsOutbound`. The `createOutboundStream` guard then short-circuited on the stale entry, so when the peer reconnected no new outbound stream was created and every further outbound message to that peer failed.

Remove an outbound stream from the registry when it closes, so a new one is created the next time the peer (re)connects.

Related https://github.com/libp2p/js-libp2p/issues/2771.
References https://github.com/ChainSafe/js-libp2p-gossipsub/pull/534, which fixes the same issue in the pre-migration repository.

## Notes & open questions

Scope: this restores the stream when the peer reconnects, which is the reported case. Re-establishing a stream that closes while its connection stays open, where the remote never re-opens, is left as a follow-up.

## Change checklist


tabcat:
Going to remove the heartbeat reconnect. Feeling its too aggressive. Could be good to have in the future.

dozyio (approved):

larrype8:
I am curious if there will be any ongoing work related to the scope that was not performed in this PR, "Scope: this restores the stream when the peer reconnects, which is the reported case. Re-establishing a stream that closes while its connection stays open, where the remote never re-opens, is left as a follow-up." I was working on a patch for gossipsub not handling the stream close event and having a dead stream in the list and found your fix for that which i have incorporated. I was also working toward not long handling the error but actively re-establishing the stream or disconnecting the stream. I am not sure waiting for he peer to reconnect should be the only strategy but I am not sure about that. Could you elaborate on any future plans to work on active re-establishment of the failed stream?

tabcat:
Hey @larrype8, I'm not convinced active re-dialing in gossipsub is necessary yet. Most of the problem this PR is solving is a case where a device goes offline, comes back sometime later, and then tries to dial the same gossipsub peer again (so we know the gossipsub peer is already being dialed). But if you experience a problem in implementation, please do create an issue. Thanks

larrype8:
Thanks for the reply. The main problem we experienced was the one you fixed in this PR. I will monitor it and log an issue if we experience a problem getting the peer reconnected with the exact condition because I am also not sure how it will behave with this fix in place. Thanks again.

PR #3534, opened 2026-06-06
fix: handle terminal middleware without mutating chains

dozyio:
## Description

Fixes two stream middleware issues:
- Copies registered inbound middleware before appending the protocol handler wrapper, preventing handlers from accumulating across streams.
- Allows middleware to return false to stop the chain without aborting the stream.

Adds regression coverage for repeated inbound streams, terminal sync/async middleware, outbound terminal middleware, and throw/reject abort behaviour.

## Change checklist


packages/interface/src/stream-handler.ts
```diff
@@ -9,9 +9,12 @@ export interface StreamHandler {
 
 /**
  * Stream middleware allows accessing stream data outside of the stream handler
+ *
+ * Return false to stop the middleware chain without aborting the stream.
+ * Throw or reject to abort the stream.
  */
 export interface StreamMiddleware {
-  (stream: Stream, connection: Connection, next: (stream: Stream, connection: Connection) => void): void | Promise<void>
+  (stream: Stream, connection: Connection, next: (stream: Stream, connection: Connection) => void): void | false | Promise<void | false>
```
tabcat, line 18: should this be `void | boolean | Promise<void | boolean>`?

packages/libp2p/src/connection.ts
```diff
@@ -353,6 +349,40 @@ function findOutgoingStreamLimit (protocol: string, registrar: Registrar, option
...
+
+      if (result === false) {
+        stopChain()
+      } else if (result != null) {
+        result.then(result => {
+          if (result === false) {
+            stopChain()
+          }
+        }).catch(reject)
+      }
```
tabcat: why not just await `mw` call?

packages/interface/src/stream-handler.ts
```diff
@@ -9,9 +9,12 @@ export interface StreamHandler {
 
 /**
  * Stream middleware allows accessing stream data outside of the stream handler
+ *
+ * Return false to stop the middleware chain without aborting the stream.
```
tabcat, line 13: could said returning true or void lets the middleware continue.

packages/interface/src/stream-handler.ts
dozyio: should be false - kind of want false to be an explicit option to cancel the middleware... or could make it true & void do the same thing which feels a bit off

tabcat: thinking about cases where the function returns some boolean logic. would have to coerce to void or false to stop or not stop the middleware while also fitting the type constraint.

dozyio: The semantics get a bit awkward because you need to choose what the boolean means - e.g. `isAuthorized()` vs `isBlocked()`

PR #3535, opened 2026-06-10
feat: allow creating nodes synchronously

achingbrain:
Adds a `createLibp2pSync` function that returns the libp2p instance without wrapping it in a promise.

Gives the user full control over the lifecycle of the node, e.g. they must then call `.start()` themselves.

When no private key was supplied to `createLibp2p()` the `.peerId` component was created using async - this has been moved to the `.start()` function, and the `.peerId` accessor will now throw a `NotStartedError` if accessed before it is set.

Monorepo modules that accessed the `.peerId` or `.privateKey` component before the node is started have been updated to not do that.

## Change checklist

### DIFF
--- a/packages/connection-encrypter-plaintext/src/index.ts
+++ b/packages/connection-encrypter-plaintext/src/index.ts
@@ -38,11 +38,11 @@ export interface PlaintextComponents {
 
 class Plaintext implements ConnectionEncrypter {
   public protocol: string = PROTOCOL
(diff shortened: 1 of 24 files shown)


achingbrain:
Could call it `createLibp2pOffline` instead to make it more obvious the node is not started?

achingbrain:
Actually, this isn't required, feel free to re-open if it's useful.

Issue #3536, opened 2026-06-11
Tracking: kad-dht query convergence & routing-table reliability


tabcat:
Tracking issue for a set of internal `@libp2p/kad-dht` reliability fixes found while hardening the IPNS-over-DHT write path. All are internal (no public API changes); each lands as its own PR, linked here as it opens. Ordered roughly by impact.

- [x] **`getClosestPeers` should only add peers that responded**
  - Every contacted peer is added to the closest set even if it never answered, so dead or stale entries fill the closest-K and receive wasted `PUT_VALUE`s. Mirrors go-libp2p, which excludes `PeerUnreachable` peers from the converged set.
  - #3537
- [x] **Terminate a query path once the closest-K set is full**
  - Lets a path stop instead of draining every seeded peer. Only honest once the live-peer fix lands, otherwise the set can fill with dead peers.
  - #3539
- [x] **Run the periodic routing-table refresh**
  - Only the startup refresh fires today; the interval refresh is throttled out because the per-CPL refresh timestamp is never recorded.
  - #3538
- [ ] ~~**Evict unresponsive peers from the routing table on query failure**~~
  - ~~A peer that fails a lookup is never removed (eviction only happens via bucket-full ping-replace), so dead entries are re-queried on every lookup. Ping-confirm before evicting. _(needs design + test)_~~
  - Scratched: a maintained routing table stays healthy on its own, so proactive eviction adds little. A 45-min live run on the #3538 refresh branch (passive node, no manual queries) held a stable table at ~5-8% unreachable with no upward drift (~46 peers, ~20 connections kept open to the KEEP_ALIVE closest-peer set).
- [ ] ~~**Unbiased Fisher-Yates shuffle for the disjoint-path seed**~~
  - Scratched: shuffling is not required for S/Kademlia disjoint-path lookups (the spec only says to distribute the k closest into d buckets); the bias is harmless to the lookup result and go-libp2p doesn't shuffle at all.
- [x] **Fix `alpha` option doc**
  - The JSDoc says `@default 3` but the code uses `ALPHA = 10`.
  - #3541
- [x] **Defer the dialability check until after the distance checks**
  - So it only runs for peers that pass the distance filter, not every candidate.
  - #3528

[...]

PR #3537, opened 2026-06-11
fix(kad-dht): only keep peers that responded in getClosestPeers

tabcat:
## Description

`getClosestPeers` adds every queried peer to the closest-K set even when it never
answered the `FIND_NODE`, so the set is padded with dead or stale routing-table
entries. `PUT_VALUE` writes to that set, so each dead peer is a wasted write and
can crowd out reachable peers slightly further out.

This adds a peer to the closest set only if it sent a `PEER_RESPONSE` during the
lookup, matching go-libp2p, which excludes `PeerUnreachable` peers from the set a
lookup converges on.

Writing IPNS records to the public Amino DHT, of the 20 closest peers a lookup returns:

| | before | after |
| --- | --- | --- |
| reachable on a fresh dial | ~3-16 | ~18-20 |
| stored the record (write reach) | ~4-18, varies per run | 18-20, consistent |

Part of #3536.

## Change checklist

### DIFF
--- a/packages/kad-dht/src/peer-routing/index.ts
+++ b/packages/kad-dht/src/peer-routing/index.ts
@@ -263,12 +263,8 @@ export class PeerRouting {
         yield event
       }
 
-      // only add the peer to the closest set if it actually responded - a peer
-      // we could not contact is a dead or stale routing-table entry, and
-      // including it both wastes a PUT_VALUE and lets dead peers fill the
-      // closest-K so the lookup stops before finding live peers slightly
-      // further out
       if (responded) {
+        // add the peer to the list if we've managed to contact it successfully
         peers.addWithKadId(peer, peerKadId, path)
       }
     }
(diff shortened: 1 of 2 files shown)


packages/kad-dht/src/peer-routing/index.ts
```diff
@@ -249,14 +249,24 @@ export class PeerRouting {
...
         signal,
         path
-      })
+      })) {
+        if (event.name === 'PEER_RESPONSE') {
+          responded = true
+        }
 
-      // add the peer to the list if we've managed to contact it successfully
-      peers.addWithKadId(peer, peerKadId, path)
```
tabcat, line 259: can see this was the intended result before. but there are other events that can be yielded that don't imply we've had a successful response from the peer.

dozyio (approved):

PR #3480, opened 2026-04-25
feat: webrtc-direct-v2

dozyio:
## Description

Initial implementation based on proposed spec https://github.com/libp2p/specs/pull/715

Requires node-datachannel v0.32.3+

## Change checklist

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation if necessary (this includes comments as well)
- [x] I have added tests that prove my fix is effective or that my feature works

PR #3356, opened 2025-11-20
feat: evaluate @roamhq/wrtc as alternative to node-datachannel

NiKrause:
This PR evaluates @roamhq/wrtc as a replacement for node-datachannel as requested in #3034.

## Implementation

Replaced node-datachannel with two libraries:
- **@roamhq/wrtc** for RTCPeerConnection (WebRTC peer connections)
- **stun** package for STUN server (WebRTC Direct listener)

## Changes

- Replace node-datachannel RTCPeerConnection with @roamhq/wrtc
- Implement STUN server using stun package for WebRTC Direct
- Remove DirectRTCPeerConnection wrapper class
- Add extractRemoteFingerprint helper for SDP parsing
- Remove node-datachannel-specific GC workarounds
- Add TypeScript definitions for stun package

## Test Results

- WebRTC transport functional
- WebRTC Direct functional
- Certificate management working
- SDP manipulation validated

## Known Issue

**Segmentation fault during test cleanup** - Native module crash in @roamhq/wrtc destructor. All tests pass before crash occurs. Does not affect functionality, only test harness cleanup.

Stack trace shows crash in `RTCDataChannel::~RefPtr()` and `RTCPeerConnection::~RTCPeerConnection()`.

PR #3544, opened 2026-06-16
deps(transport-webrtc): replace node-datachannel

mertushka:
## Description

Replaces `node-datachannel` with [@mertushka/webrtc-node](https://github.com/mertushka/webrtc-node) in `@libp2p/webrtc`.

`@mertushka/webrtc-node` is a Node.js WebRTC binding that exposes a W3C-compatible WebRTC API surface instead of a package-specific polyfill surface. Version `0.2.0` passes 620 strict Web Platform Tests for WebRTC API behavior and provides the standard objects `@libp2p/webrtc` already uses in browsers, including `RTCPeerConnection`, `RTCDataChannel`, `RTCSessionDescription`, `RTCIceCandidate`, `RTCCertificate`, and standard `EventTarget` behavior.

This is useful here because the existing Node path had to wrap `node-datachannel` APIs to inject ICE credentials, map ICE servers, import certificates, work around stale data channel state, and force native peer connection cleanup. With `@mertushka/webrtc-node`, the Node path can use the standard WebRTC-shaped APIs directly, with explicit `nonstandard` hooks only where WebRTC Direct needs Node-specific control over ICE credentials, certificate import, UDP muxing, fingerprint verification, and remote fingerprint access.

This removes the node-datachannel-specific compatibility layer and related runtime workarounds now that the replacement exposes the required standard WebRTC APIs plus the nonstandard APIs used by the WebRTC Direct path.

Changes include:

- replace the `node-datachannel` dependency with `@mertushka/webrtc-node`
- wire the new WebRTC implementation through the Node WebRTC entrypoints
- keep WebRTC Direct support using `nonstandard` APIs:
  - `configurePeerConnection`
  - `importCertificate`
  - `setLocalIceCredentials`
  - `getRemoteFingerprint`
  - `IceUdpMuxListener`
- remove node-datachannel-specific peer connection wrappers
- remove stale readyState/send error workarounds
- remove node-datachannel GC/closed-state workarounds

## Related open PRs

Potentially Fixes/Closes:

- https://github.com/libp2p/js-libp2p/pull/3356
- https://github.com/libp2p/js-libp2p/pull/3459
- https://github.com/libp2p/js-libp2p/pull/3460

Likely requires a rebase:

- https://github.com/libp2p/js-libp2p/pull/3480

## Verification

- `npm run dep-check -w @libp2p/webrtc`
- `npm run build -w @libp2p/webrtc`
- `npm run test:node -w @libp2p/webrtc -- --exit`
- `npm run test:chrome -w @libp2p/webrtc`


dozyio:
Thanks @mertushka for putting this together. I’m hesitant to merge this right now because the replacement dependency is still very new and hasn’t had much time to prove itself in production or across the ecosystem.

For a core dependency like this, I’d prefer we wait until it has more adoption, a clearer maintenance track record, and some time for early issues to shake out. This isn’t a rejection of the approach, but I don’t think we should take on that risk in js-libp2p yet. Happy to revisit once the dependency has matured a bit.

mertushka:
@dozyio Thanks for the feedback, that makes sense and I agree with the concern.

I opened this PR mostly to share the idea and start a discussion rather than to push for an immediate merge. The main reasons I explored replacing `node-datachannel` were performance, stability, and having a more standards-aligned WebRTC API surface. `@mertushka/webrtc-node` currently passes 620 strict WPT tests, which gave me more confidence in the behavior.

I originally built the module because one of my own projects, haxball.js, had several issues with `node-datachannel`, and I found `node-datachannel` difficult to work on or adapt when those issues came up.

That said, I completely understand that js-libp2p needs to be conservative with core dependencies. 

I’m happy for this PR to stay as an experiment/reference.

dozyio:
Moving to draft as likely won't be merged in the near term

PR #3550, opened 2026-07-03
fix: treat empty dnsaddr answer as no addresses

tabcat:
## Description

`@multiformats/dns` 1.0.15 changed `query()` to throw an empty-answer error (so it can fall through to other resolvers) instead of returning an empty result. As a side effect, dialing a `/dnsaddr/` that resolves to no records started rejecting with `DNSQueryFailedError` instead of `NoValidAddressesError`.

This catches the empty-answer case in the dnsaddr resolver so those dials fail with `NoValidAddressesError` again, while genuine DNS failures still surface as `DNSQueryFailedError`.

## Notes & open questions

Root cause is multiformats/js-dns#34 (released in `@multiformats/dns` 1.0.15). No new test needed: the existing `connections.spec.ts` "should fail to dial if resolve fails and there are no addresses to dial" test exercises this path (it was failing on 1.0.15, passes with this change).

## Change checklist

### DIFF
--- a/packages/libp2p/src/connection-manager/resolvers/dnsaddr.ts
+++ b/packages/libp2p/src/connection-manager/resolvers/dnsaddr.ts
@@ -32,10 +32,6 @@ class DNSAddrResolver implements MultiaddrResolver {
         ]
       })
     } catch (err: any) {
-      // @multiformats/dns >= 1.0.15 throws when a query returns no answers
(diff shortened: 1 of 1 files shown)


tabcat:
It's probably more helpful to throw the DNS error but this gets everything working with the same behavior as before. We can decide on what error to throw later.

PR #3552, opened 2026-07-06
fix(pnet): forward drain events from the underlying connection

### LINKED ISSUE #3551: @libp2p/pnet: PSK-protected connections permanently stall on first write backpressure ('drain' never forwarded)
- **Version**: @libp2p/pnet 3.0.23 (current `main`), with libp2p 3.x / @libp2p/utils 7.2.3

- **Platform**: All platforms. Verified on Darwin 25.4.0 arm64, Node.js v24.14.1.

- **Subsystem**: pnet (connection protection)

#### Severity:

High — any private network (PSK) deployment that transfers more than a few MB over a single connection stops working, with no error surfaced.

#### Description:

- **What you did**: Ran a bulk transfer between two nodes using `tcp` + `yamux` + `preSharedKey` connection protection (a plain stream writing data as fast as possible).
- **What happened**: The transfer stalls permanently at around 4 MB — the point where the underlying TCP socket first reports write backpressure. The outbound direction of the protected connection never sends again, while the connection still appears healthy (it is not closed or errored, and the inbound direction keeps working). Removing the connection protector, or applying the fix below, lets the same transfer complete (verified at 4 MB and 128 MB).
- **What you expected to happen**: The transfer completes; backpressure should only pause sending temporarily.

Root cause: `BoxMessageStream` in `packages/pnet/src/crypto.ts` wraps the transport `MultiaddrConnection` and relays its `'message'` and `'close'` events, but not `'drain'`. When `sendData()` returns `canSendMore: false`, `AbstractMessageStream` (in `@libp2p/utils`) sets `writableNeedsDrain` and buffers all subsequent writes until a `'drain'` event is dispatched on the stream — and the only code path that dispatches it is `onMuxerDrain()` (`abstract-message-stream.ts`, `safeDispatchEvent('drain')`). Since pnet never calls it when the inner connection drains, the first backpressure event wedges the outbound direction forever. `AbstractStreamMuxer` handles the same situation by relaying the inner connection's `'drain'` event (`abstract-stream-muxer.ts`); pnet is missing the equivalent relay.

1. Create two libp2p nodes with `transports: [tcp()]`, `streamMuxers: [yamux()]`, a standard connection encrypter, and `connectionProtector: preSharedKey({ psk })` using a shared swarm key.
2. Dial one node from the other, open a stream on any echo/sink protocol, and write ~8 MB of data.
3. The transfer stalls at roughly 4 MB (once the socket send buffer fills) and never completes; without the connection protector it completes.

A minimal unit reproduction (used as the regression test in the linked PR): protect both halves of `multiaddrConnectionPair()` from `@libp2p/utils`, call `send()` on the protected connection until it returns `false`, then wait for a `'drain'` event on it — the event never fires.

Fix: relay the inner connection's `'drain'` event to `onMuxerDrain()` in the `BoxMessageStream` constructor — see the linked PR.

peerbit-org:
## Description

`BoxMessageStream` relays `'message'` and `'close'` events from the wrapped `MultiaddrConnection`, but not `'drain'`. When the underlying connection reports write backpressure (`send()` returns `false`), `AbstractMessageStream` pauses sending and buffers writes until a `'drain'` event is dispatched on the stream — which only happens via `onMuxerDrain()`. Since pnet never invoked it, the first backpressure event permanently wedged the outbound direction of every PSK-protected connection while the connection still appeared healthy: a `tcp` + `yamux` + PSK bulk transfer stalls at ~4 MB (the point the socket send buffer first fills) and never recovers, while the inbound direction keeps working.

This PR relays the underlying connection's `'drain'` event to `this.onMuxerDrain()` in the `BoxMessageStream` constructor — the same pattern `AbstractStreamMuxer` uses for its wrapped connection. With the change, the same transfers complete (verified at 4 MB and 128 MB).

Includes a regression test that protects both halves of a `multiaddrConnectionPair()`, sends until the underlying connection reports backpressure, then asserts the protected connection emits `'drain'` and that all queued bytes arrive at the other end. The test times out without the fix and passes with it.

Fixes https://github.com/libp2p/js-libp2p/issues/3551.

## Notes & open questions

None — the change is intentionally minimal (a single event relay plus the regression test). `sendPause()`/`sendResume()` already handle the read side; this addresses the write side only.

## Change checklist

### DIFF
--- a/packages/pnet/src/crypto.ts
+++ b/packages/pnet/src/crypto.ts
@@ -43,6 +43,11 @@ export class BoxMessageStream extends AbstractMultiaddrConnection {
       }
     })
 
+    // resume sending when the underlying connection can accept more data
+    this.maConn.addEventListener('drain', () => {
+      this.onMuxerDrain()
+    })
+
     this.maConn.addEventListener('close', (evt) => {
       if (evt.error != null) {
         if (evt.local) {
--- a/packages/pnet/test/index.spec.ts
+++ b/packages/pnet/test/index.spec.ts
@@ -52,6 +52,50 @@ describe('private network', () => {
     expect(output).to.deep.equal([uint8ArrayFromString('hello world'), uint8ArrayFromString('doo dah')])
   })
 
+  it('should forward drain events from the underlying connection', async () => {
+    const [outboundConnection, inboundConnection] = multiaddrConnectionPair({
+      delay: 10
+    })
+
+    const protector = preSharedKey({
+      psk: swarmKeyBuffer
+    })()
+
+    const [outbound, inbound] = await Promise.all([
+      protector.protect(outboundConnection),
+      protector.protect(inboundConnection)
+    ])
+
+    let received = 0
+
+    inbound.addEventListener('message', (evt) => {
+      received += evt.data.byteLength
+    })
+
+    // send data until the underlying connection reports backpressure
+    const chunk = new Uint8Array(1024 * 64)
+    let sent = 0
+    let canSendMore = true
+
+    while (canSendMore) {
+      canSendMore = outbound.send(chunk)
+      sent += chunk.byteLength
+    }
+
+    // the protected connection should emit 'drain' when the underlying
+    // connection drains, otherwise sending never resumes
+    await pEvent(outbound, 'drain', {
+      timeout: 5_000
+    })
+
+    await Promise.all([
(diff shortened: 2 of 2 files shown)


Faolain:
@dozyio @tabcat would you guys be able to review? 🙏

tabcat:
Thanks @peerbit-org

PR #3553, opened 2026-07-07
chore: pin macos CI runner to macos-15

tabcat:
## Description

`macos-latest` migrated from macos-15 to macos-26 in June 2026 (actions/runner-images#14167). On macos-26 the `go-libp2p` postinstall that downloads the p2pd binary stalls (the fetch returns `200 OK`, then the response body read never completes and times out after 60s, retrying 11 times), so `test-node (macos-latest)` fails at the install step before any tests run. It only surfaces on a node_modules cache miss, which re-triggers the p2pd download, so it hits any deps-changing PR.

Pin the macOS runner to `macos-15` until the download works on macos-26.

## Notes & open questions

Stopgap. The underlying issue is go-libp2p's fetch-based p2pd download not completing on macos-26 (libp2p/npm-go-libp2p); unpin once that is fixed.

## Change checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,7 +48,10 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        # macos-latest migrated from macos-15 to macos-26 in June 2026
+        # (actions/runner-images#14167), which stalls the go-libp2p p2pd binary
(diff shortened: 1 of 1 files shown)


achingbrain:
Longer term you'll probably want to remove the postinstall script as it's going to be disabled by default in npm 12.

Instead you can download the binary on first run - see https://github.com/ipfs/npm-kubo/pull/83 for an example.

PR #3583, opened 2026-07-21
feat(gossipsub): graft peers into the mesh on subscribe

tabcat:
## Description

gossipsub only forwards a message to the peers in its per-topic mesh, and mesh links are grafted lazily on the heartbeat (default 1s). A message published just after subscribing could be forwarded to a still-empty mesh and silently dropped: the relay had recorded the peer's subscription but had not yet grafted it.

This grafts a peer into the mesh as soon as its SUBSCRIBE is received, provided we are meshed on the topic and the mesh is below `Dlo`, using the same eligibility the heartbeat applies (outbound stream on one of our protocols, not a direct peer, non-negative score, not backed off). Grafts arising from a single SUBSCRIBE RPC are coalesced into one GRAFT.

The graft bookkeeping (peer score, mesh membership, metrics, `gossipsub:graft` event) is shared with the heartbeat and inbound-GRAFT paths via a new `addToMesh` helper, so an eager graft is indistinguishable from any other.

## Notes & open questions

Mirrors rust-libp2p, which also grafts on subscribe while the mesh is below its low-water mark; go-libp2p forms the mesh only on `Join` and the heartbeat, so this is a deliberate divergence from go.

A possible follow-up: hand a newly grafted peer an IHAVE of very recent messages at graft time, which would also close the window for peers grafted by the heartbeat rather than on subscribe. Left out here as a broader change.

## Change checklist

### DIFF
--- a/packages/gossipsub/src/gossipsub.ts
+++ b/packages/gossipsub/src/gossipsub.ts
@@ -972,6 +972,8 @@ export class GossipSub extends TypedEventEmitter<GossipSubEvents> implements Typ
       // update peer subscriptions
 
       const subscriptions: Array<{ topic: TopicStr, subscribe: boolean }> = []
+      const fromStr = from.toString()
+      const graftTopics: TopicStr[] = []
 
       rpc.subscriptions.forEach((subOpt) => {
         const topic = subOpt.topic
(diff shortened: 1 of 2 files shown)


tabcat:
This fixes an issue with the gossipsub example ci job. Also matches a rust-libp2p gossipsub improvement which makes messages less likely to be lost when sent in the same heartbeat as subscribing.

PR #3585, opened 2026-07-26
fix(pnet): bound xsalsa20 wasm memory per message and per connection

saul-jb:
Sending large messages through connections protected by pnet throws:
```
WebAssembly.Memory.grow(): Maximum memory size exceeded
```

This PR windows the xsalsa xor so it never grows and exceeds its memory limits. There is also a test added that demonstrates this fix.


tabcat:
Added a few things to this branch.

pnet has never called `finalize()` on its ciphers. Each cipher holds a 64 byte slot in a pool that xsalsa20 shares across the whole process, and only `finalize()` hands it back, so every connection used up two for good. After a few thousand connections a node can't open any more, and the ones it has start failing. It's also the call that wipes the key from memory.

Opening and closing 500 connections now leaves the pool where it started. Before, it grew every time and never came back down.

This is on your PR rather than its own because it doesn't work without your chunking. Growing the wasm memory breaks `finalize()` for every other cipher, so one big message would stop all of them releasing.

I also reworked the tests. The large message test only checked how many bytes arrived, which passes even with encryption turned off, so it now checks the bytes themselves.

I renamed `xorWindowed` to `xorInChunks` and moved it out of the class, since the tests said chunk and the source said window.

PR #3591, opened 2026-08-10
fix(tcp): drain pending dials during shutdown

wemeetagain:
## Description

Make the TCP transport participate in component shutdown and drain outbound dial operations before shutdown completes.

The root cause is a one-time snapshot race. A dial can pass its admission check, invoke caller-controlled code through `onProgress` or DNS lookup, and re-enter shutdown before `net.connect()` returns. If shutdown snapshots only returned sockets, the new socket is registered after the snapshot and is left open.

This change:

- registers each dial operation before any caller-controlled boundary
- closes dial admission before taking a stable shutdown snapshot
- records shutdown on the operation even before its socket exists
- destroys sockets owned by pending operations and awaits their close
- releases operations that fail before socket creation
- retains operation ownership until socket close or successful connection handoff
- rejects new dials after shutdown begins

Related investigation: https://github.com/matthewkeil/js-libp2p/pull/1

## Validation

- `npm run lint --workspace @libp2p/tcp`
- `npm run build --workspace @libp2p/tcp`
- `npm run spell-check`
- shutdown regressions on Node.js 22.23.2 and 24.13.0
- 23 additional TCP tests outside the locally blocked `listen-dial.spec.js`

## Change checklist

- [x] Documentation changes are not necessary for this internal lifecycle fix
- [x] I have added tests that prove the fix is effective

### DIFF
--- a/packages/transport-tcp/src/tcp.ts
+++ b/packages/transport-tcp/src/tcp.ts
@@ -28,27 +28,38 @@
  */
 
 import net from 'net'
-import { AbortError, TimeoutError, serviceCapabilities, transportSymbol } from '@libp2p/interface'
+import { AbortError, NotStartedError, TimeoutError, serviceCapabilities, transportSymbol } from '@libp2p/interface'
 import { TCP as TCPMatcher } from '@multiformats/multiaddr-matcher'
 import { CustomProgressEvent } from 'progress-events'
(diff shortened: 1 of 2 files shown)


wemeetagain:
closing in favor of https://github.com/matthewkeil/js-libp2p/pull/2

PR #3593, opened 2026-08-10
fix(gossipsub)!: bound subscribed topic bytes per peer

tabcat:
## Description

Adds a per-peer budget for the topic subscriptions gossipsub tracks in its
`topics` map, which was previously unbounded per peer.

Each tracked topic costs its string length plus a small fixed per-entry
overhead, capped per peer by `maxTopicBytesPerPeer` (default 1 MiB, ~1000
topics). Subscriptions over the budget are ignored (the peer keeps what it
had), refunded on unsubscribe, and cleared when the peer is removed.

Behaviour change from the previous unbounded default: a peer that legitimately
subscribes to more than ~1000 topics will have the excess ignored; raise
`maxTopicBytesPerPeer` to restore the old behaviour.

## Change checklist

- [ ] I have made corresponding changes to the documentation if necessary

### DIFF
--- a/packages/gossipsub/src/constants.ts
+++ b/packages/gossipsub/src/constants.ts
@@ -264,3 +264,18 @@ export const BACKOFF_SLACK = 1
 
 export const GossipsubIdontwantMinDataSize = 512
 export const GossipsubIdontwantMaxMessages = 512
+
+/**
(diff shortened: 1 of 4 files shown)


wemeetagain (commented):
Sorry, late official review, it looks good to force sane defaults.

PR #3595, opened 2026-08-11
feat: enforce idontwant

matthewkeil:
Implements the gossipsub v1.2 MUST (gossipsub-v1.2.md line 62): peers found in the per-peer dont_send_message_ids set are now skipped when relaying. 

Previously received IDONTWANTs were tracked, pruned, and metered but never consulted on any send path, so duplicates were still sent to peers that refused them. Enforcement covers both forwardMessage and publish (content-derived message ids make pre-publish IDONTWANT possible), with a new gossipsub_idontwant_skipped_sends_total metric.

Also fixes a latent version gate: `sendIDontWants` matched `/meshsub/1.2.0` by exact equality, which would silently stop IDONTWANT to peers on any future protocol version. Version-dependent behavior is now gated by a feature-test predicate over an ordered version ladder (the go-libp2p-pubsub pattern), so future versions inherit lower-version features automatically.


packages/gossipsub/src/gossipsub.ts
```diff
@@ -2183,6 +2192,15 @@ export class GossipSub extends TypedEventEmitter<GossipSubEvents> implements Typ
     // If the message is anonymous or has a random author add it to the published message ids cache.
     this.publishedMessageIds.put(msgIdStr)
 
+    // skip peers that told us they already have the message - with content-derived
+    // message ids a peer may have sent IDONTWANT before we publish the same content
```
tabcat: this is the origin path and I can't find another implementation that filters here. this also brings in some inconsistency when it removes all remaining peers, recipients returns [] without an error. think it would be best to remove this. doesnt seem possible for this to trigger in a StrictSign setting, is there a specific instance you are thinking of in a content-derived id setting?

packages/gossipsub/src/gossipsub.ts
```diff
@@ -1476,6 +1479,13 @@ export class GossipSub extends TypedEventEmitter<GossipSubEvents> implements Typ
         processed++
 
         const msgIdStr = this.msgIdToStrFn(msgId)
+
+        // the peer told us it already has this message, so don't serve it back
+        if (this.idontwants.get(id)?.has(msgIdStr) === true) {
+          this.metrics?.onIdontwantSkippedSend(IDontWantSkipPath.iwant)
```
tabcat: this records the metric before we know if we have the messageId or not. let me know if you'd think it be better to record after or remove this filter entirely.

tabcat:
I've added a filter inside of handleIwant. Go and Rust both do this so its not irregular and saves some bandwidth.
Also I cleaned up the types a bit so there are fewer strings floating around even if they were typed before.
Replaced the Proxy metrics stub with a real register passed through init.

Left a comment for you on the handleIWant filter i added.

packages/gossipsub/src/gossipsub.ts
matthewkeil: i think you are right after speaking with @wemeetagain. will remove this

packages/gossipsub/src/gossipsub.ts
matthewkeil: moved it below the `.mcache` lookup

tabcat (approved):

tabcat:
@matthewkeil Thanks!

matthewkeil:
@tabcat good catch on the cleanup.  did it manually, should have pulled out claude

PR #3596, opened 2026-08-11
fix: drain outbound connections during shutdown

GrapeBaBa:
## Description

Prevent shutdown from finishing while outbound dial jobs or TCP sockets still own referenced native handles.

- mark the connection manager stopped before service teardown
- abort and await active dial jobs
- track outbound TCP sockets and wait for their close events
- avoid resetAndDestroy after graceful socket shutdown has started, since libuv does not allow mixing uv_shutdown with uv_tcp_close_reset

This was observed while investigating a Lodestar shutdown hang: https://gist.github.com/nflaig/5f41cfc50f38baf5046a034162943dc3

## Tested version

This draft intentionally points to commit c20c70314f5328ef240f6b061242a313eb38d860, the exact js-libp2p source used to build Lodestar image sha256:2c2bcfbe1a9adf72cb9f73dfc605fb3a149e283d8aeae6b0325bbefd5e1b13f9.

The affected compiled libp2p and TCP files in the running image were compared byte-for-byte with the artifacts built from this commit. All hashes match.

The image has completed 44 consecutive effective mainnet shutdown cycles without a hang so far. The 50-cycle soak is still running.

This branch is intentionally not merged with newer main commits yet, so the PR head remains identical to the tested version.

## Verification

- Added real TCP regression coverage for outbound socket close and the reset/graceful-close race.
- At this exact commit, the affected package lint, doc check, build, and 39 Node.js tests passed.
- No tests were rerun while opening this draft PR.

## Change checklist

### DIFF
--- a/packages/libp2p/src/connection-manager/dial-queue.ts
+++ b/packages/libp2p/src/connection-manager/dial-queue.ts
@@ -81,6 +81,7 @@ export class DialQueue {
   private readonly connections: PeerMap<Connection[]>
   private readonly log: Logger
   private readonly resolvers: Record<string, MultiaddrResolver>
+  private readonly activeDials: Set<Promise<Connection>>
 
   constructor (components: DialQueueComponents, init: DialerInit = {}) {
     this.addressSorter = init.addressSorter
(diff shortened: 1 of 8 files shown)


GrapeBaBa:
Superseded by #3597. A controlled 50-cycle mainnet shutdown soak showed that only the minimal socket reset guard was required: unmodified Lodestar, 50 clean exits, 0 hangs.

PR #3598, opened 2026-08-13
fix(gossipsub): graft known subscribers when their outbound stream opens

tabcat:
Grafts a peer we already know is subscribed as soon as our outbound stream to them is established, mirroring the existing graft-on-subscribe path (#3583).

`graftOnSubscribe` can only add a peer to the mesh once we have an outbound stream to them (the same guard the heartbeat applies). If a peer's SUBSCRIBE is handled before that stream exists, the graft is refused and the peer waits for the next heartbeat before joining the mesh. This adds the mirror trigger: when the outbound stream opens, graft any peer we already know is subscribed. Mesh formation no longer depends on the subscription and the stream arriving in a particular order, nor on the peer grafting us back.

rust-libp2p handles this earlier and more simply, by queuing the GRAFT into a per-peer send queue that the connection handler drains once a substream exists rather than requiring the outbound stream up front; there is a note to that effect next to the code.

### DIFF
--- a/packages/gossipsub/src/gossipsub.ts
+++ b/packages/gossipsub/src/gossipsub.ts
@@ -704,6 +704,11 @@ export class GossipSub extends TypedEventEmitter<GossipSubEvents> implements Typ
         this.log('send subscriptions to', id)
         this.sendSubscriptions(id, Array.from(this.subscriptions), true)
       }
+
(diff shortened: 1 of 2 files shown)


tabcat:
still trying to fix the pubsub example in ci.

PR #3610, opened 2026-08-27
feat(gossipsub): add floodPublish option to PublishOpts

markolazic01:
## Description

`floodPublish` is currently a constructor-only option, so the only way to
flood publish one topic is to enable it globally for all traffic. Added it
to `PublishOpts` so a single message can be flood published without turning
on global flood publish. Option can be also used in reverse, disabling flood publish for a single call.

## Notes & open questions

Additive and non-breaking.

## Change checklist

### DIFF
--- a/packages/gossipsub/src/gossipsub.ts
+++ b/packages/gossipsub/src/gossipsub.ts
@@ -1993,7 +1993,7 @@ export class GossipSub extends TypedEventEmitter<GossipSubEvents> implements Typ
     return tosend
   }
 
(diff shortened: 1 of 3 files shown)


tabcat:
it doesnt seem like anyone can actually reach this feature by importing the interface. the class is imported directly in the tests so the type error doesnt show up there.

tabcat:
Looks like lodestar overrides the types manually.

tabcat (approved):

PR #3625, opened 2026-09-09
fix: clean up aborted websocket connections

peerbit-org:
Fix two WebSocket cleanup paths:

- Close the socket when a dial fails or is aborted, before rethrowing the existing error.
- On reset, stop the buffered-amount poller and call `close()` without the invalid reserved code 1006.

Regression tests cover an aborted HTTP upgrade, established native/`ws` sockets, and poller cancellation/backpressure. Against the original source, five regressions fail and one control passes.

Local validation: 35 Node tests; 4 Chromium tests each in main and worker; package-local typecheck and lint. Full monorepo CI remains required.

No public API or timeout changes. Includes the dependency compatibility prerequisite #3628.

### DIFF
--- a/packages/connection-encrypter-noise/package.json
+++ b/packages/connection-encrypter-noise/package.json
@@ -84,7 +84,7 @@
     "@noble/ciphers": "^2.0.1",
     "@noble/curves": "^2.0.1",
     "@noble/hashes": "^2.0.1",
-    "protons-runtime": "^7.0.0",
+    "protons-runtime": "7.0.0",
     "uint8arraylist": "^3.0.2",
     "uint8arrays": "^6.1.1",
     "wherearewe": "^2.0.1"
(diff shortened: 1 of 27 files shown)


peerbit-org:

peerbit-org:
The fresh pubsub-example job still fails its **published-dependency baseline**, before testing this PR's replacement packages: node 2 receives every valid fruit, while node 3 receives none ([log](https://github.com/libp2p/js-libp2p/actions/runs/34688676022/job/103540477906)).

I reproduced a concrete scheduling defect in the external example: subscription announcements can arrive before the intermediate node has a forwarding mesh. The minimal fix is separate at https://github.com/libp2p/js-libp2p-example-pubsub/pull/3. Original source fails under a delayed first-heartbeat schedule; corrected source passes under that same schedule and normal timing, with every valid fruit checked. No timeout increase, retry, dependency change, or transport workaround.

The example fix needs to land in that repository before the example job can pick it up. I have not changed this PR's WebSocket code or waived the example failure. Both fresh build jobs are now passing with #3628; full platform CI is still running.

PR #3626, opened 2026-09-10
fix: preserve canonical floodsub sequence bytes

peerbit-org:
Floodsub loses leading zero bytes when converting signed sequence numbers through bigint. This changes the signing bytes and causes verification failures.

Encode signed sequences as exactly eight big-endian bytes using `DataView`; reject incoming non-eight-byte sequences before conversion and outgoing values outside uint64.

**Compatibility:** legacy short-width signed messages are rejected. Old receivers still cannot verify canonical leading-zero sequences; both peers need the fix for those values. No fallback verifier is added. Unsigned messages, message IDs, and random-sequence generation are unchanged.

Tests cover literal byte vectors, uint64 bounds, independently signed leading-zero messages, exact round trips, and valid legacy signatures. Local focused suites: 22 passing in Node and Firefox; original source fails five assertions. Package-local typecheck and lint pass.

Draft pending compatibility review and full CI. Includes prerequisite #3628; the separate lifecycle-test failure is addressed in #3627.

### DIFF
--- a/packages/connection-encrypter-noise/package.json
+++ b/packages/connection-encrypter-noise/package.json
@@ -84,7 +84,7 @@
     "@noble/ciphers": "^2.0.1",
     "@noble/curves": "^2.0.1",
     "@noble/hashes": "^2.0.1",
-    "protons-runtime": "^7.0.0",
+    "protons-runtime": "7.0.0",
     "uint8arraylist": "^3.0.2",
     "uint8arrays": "^6.1.1",
     "wherearewe": "^2.0.1"
@@ -109,7 +109,7 @@
     "mkdirp": "^3.0.1",
     "multiformats": "^14.0.0",
     "p-defer": "^4.0.1",
-    "protons": "^9.0.1",
+    "protons": "9.0.2",
     "sinon": "^22.0.0",
     "sinon-ts": "^2.0.0"
   },
--- a/packages/connection-encrypter-plaintext/package.json
+++ b/packages/connection-encrypter-plaintext/package.json
@@ -50,7 +50,7 @@
     "@libp2p/interface": "^3.3.0",
     "@libp2p/peer-id": "^6.0.15",
     "@libp2p/utils": "^7.4.1",
-    "protons-runtime": "^7.0.0",
+    "protons-runtime": "7.0.0",
     "uint8arraylist": "^3.0.2",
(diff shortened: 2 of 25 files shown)


peerbit-org:
First-head CI has now executed all four new sequence-width cases successfully on Ubuntu, macOS, Windows, and Firefox main/worker. The checkout candidate has the submitted head and expected base as parents, with the same tree as the submitted head.

Two caveats remain visible in the raw logs despite aggregate success:

- Firefox main and worker report the pre-existing lifecycle test's unhandled `new stream error`. The direct test callback drops a rejected promise; the production registrar already awaits and catches it. The independent, test-only correction is #3627 (net one line removed, no runtime change). On main it changes strict Node from 19 passing / 1 failing to 61 passing; Firefox main/worker each pass all 61 without the logged exception.
- The [pubsub example job](https://github.com/libp2p/js-libp2p/actions/runs/34448505126/job/102779651184) failed before dependency replacement, waiting for node3's orange message. Existing `continue-on-error` makes the Examples workflow green; this is not evidence that the replacement packages passed that example.

This PR's source/head remains unchanged. Keeping it draft pending the remaining validation and canonical-only compatibility review; neither failure is waived or claimed fixed by the sequence patch. No workflow reruns or timeout changes.

peerbit-org:
Replaced the outgoing hex-string roundtrip with a fresh eight-byte `Uint8Array` and `DataView.setBigUint64(..., false)`, matching the direct big-endian encoding already used by Gossipsub. The explicit uint64 bounds guard remains, since DataView alone would wrap out-of-range values. No helper, dependency, generic bigint conversion, message-ID, or test change.

At head `1bdb8e9df6943882adbfea049a6ed9d206ccff0c`, unchanged focused suites pass 22 tests each in strict Node, Firefox main, and Firefox worker; browser logs have no uncaught exception. Package-local strict type build and lint pass. Rebuilt real-codec interop passes all five boundary/leading-zero vectors, with Gossipsub acceptance and identical signed-byte roundtrips. No performance claim is made.

Draft status and the existing full-CI/compatibility caveats remain. The independent lifecycle correction in #3627 has not received remote test acceptance: its first build stopped in unchanged `@libp2p/record` with TS2365 before Floodsub tests ran.

peerbit-org:

PR #3627, opened 2026-09-10
test(floodsub): await expected topology connection rejection

peerbit-org:
Floodsub's `onConnect` error test discards a rejected topology-callback promise, causing an unhandled rejection.

Await and assert the exact injected error, verify that no peer was admitted, and remove unrelated incoming-stream setup. The registrar's production error handling is unchanged.

Local validation: the original strict Node suite reproduces the failure; the corrected Floodsub suite passes 61 tests each in strict Node, Firefox main, and Firefox worker. Package-local typecheck and lint pass.

The Floodsub change is test-only, net one line removed. Includes the dependency compatibility prerequisite #3628.


peerbit-org:

PR #3628, opened 2026-09-12
fix: pin protons to the generated codec API

peerbit-org:
Protons runtime 7.1.0 swapped the `StreamFunction` argument order. Our checked-in codecs still use the previous API, causing TypeScript failures and incorrect streaming results.

Pin `protons-runtime` to 7.0.0 across 21 packages and generator `protons` to 9.0.2 across 20 packages. No generated or production TypeScript changes.

Add a Record streaming regression using the existing Go fixture: runtime 7.1.0 yields no events; 7.0.0 returns the expected fields.

Local validation: all 8 Record tests pass in Node and Firefox, plus package-local typecheck and lint. A coordinated codec/API upgrade can follow separately.

### DIFF
--- a/packages/connection-encrypter-noise/package.json
+++ b/packages/connection-encrypter-noise/package.json
@@ -84,7 +84,7 @@
     "@noble/ciphers": "^2.0.1",
     "@noble/curves": "^2.0.1",
     "@noble/hashes": "^2.0.1",
-    "protons-runtime": "^7.0.0",
+    "protons-runtime": "7.0.0",
     "uint8arraylist": "^3.0.2",
     "uint8arrays": "^6.1.1",
     "wherearewe": "^2.0.1"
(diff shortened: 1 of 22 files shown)


achingbrain:
The regression has been fixed in the latest protons/protons-runtime release so this is no longer necessary.

peerbit-org:
Thanks — verified that runtime 7.1.1 restores the argument order and the Record typecheck passes. One compatibility failure remains with the unchanged checked-in codec:

```text
Array.from(Record.stream(fixture.serialized), event => event.field)
7.0.0: ["$.key", "$.value"]   — 8 tests pass
7.1.1: ["..key", "..value"]   — 7 pass, streaming regression fails
```

Runtime 7.1.1 starts the prefix at `.`; the existing codec emits `${prefix}.key`. This is a runtime-only comparison using the same emitted tests. Ordinary encode/decode tests pass.

Holding off on closing/removing these pins until the remaining streaming compatibility is resolved; no codec regeneration or relaxed assertion included.

PR #3630, opened 2026-09-12
fix(webrtc): allocate UDP ports and clean up failed listeners

peerbit-org:
Allocate WebRTC Direct's ephemeral ports using UDP instead of a TCP probe. Reselect only on native ephemeral-bind failure, bounded to five attempts; explicit-port failures still propagate. Remove the `get-port` dependency.

Unregister failed startups, record the actual allocated port, and allow `close()` to finish when startup fails.

Local validation: 17 listener/transport tests pass and exit naturally; all six new regressions fail against the original code. Firefox: 47 passing, 24 existing skips. Package-local typecheck and lint pass.

The full Node suite completes its assertions but retains the process on both baseline and fix; this is not counted as a clean lifecycle pass. Full CI remains required.

Includes #3629 and its #3628 build prerequisite; the UDP fix is a separate commit. #3627 tracks the pre-existing Floodsub browser exception.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,6 +62,11 @@ jobs:
       - uses: ipfs/aegir/actions/cache-node-modules@main
         with:
           directories: ${{ env.CACHE_DIRS }}
+      - name: Report WebRTC native versions
+        working-directory: packages/transport-webrtc
+        run: |
+          npm ls node-datachannel
+          node --input-type=module -e "import { getLibraryVersion } from 'node-datachannel'; console.log('libdatachannel:', getLibraryVersion())"
       - run: npm run --if-present test:node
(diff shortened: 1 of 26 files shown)


tabcat:
i'd like to open some pulls upstream so ephemeral ports could be reserved there, as that would be the correct fix, but this is a good solution until then. will review soon

peerbit-org:
Fixed the Windows collision fixture in f25543da: bind the blocker to `127.0.0.1`, matching the native listener. Windows can permit the previous wildcard/specific-address pair on one port ([Winsock behavior](https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse)). Also replaced the rejected spelling in the test name.

No production change, platform skip, timeout increase or test retry. All 17 focused tests pass locally with natural process exit; the failed-start regression still fails against original production. Package-local typecheck, lint and spellcheck pass. Fresh CI must confirm Windows.

